PR #236, opened 2020-06-30
Decompiled ovl_Shot_Sun

Sirius902:
pog


src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +55,170 @@ const ActorInit Shot_Sun_InitVars = {
...
+    s32 temp_v0;
+
+    // Translation: Ocarina secret occurrence
+    osSyncPrintf("%d ---- オカリナの秘密発生!!!!!!!!!!!!!\n", this->actor.params);
+    temp_v0 = this->actor.params & 0xFF;
+    if (temp_v0 == 0x40 || temp_v0 == 0x41) {
+        this->unk_1A4[0] = 0;
+        this->actor.flags |= 0x10;
+        this->actor.flags |= 0x2000000;
+        this->actionFunc = &func_80BADF0C;
```
fig02, line 76: for function pointers and arrays in general you dont need the &

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +55,170 @@ const ActorInit Shot_Sun_InitVars = {
...
+        Collider_DestroyCylinder(globalCtx, &this->collider);
+    }
+}
+
+void ShotSun_FairyCountdown(ShotSun* this, GlobalContext* globalCtx) {
+    s32 temp_v0 = this->actor.params & 0xFF;
+    s32 sp38;
+
+    if (this->spawnTimer > 0) {
+        this->spawnTimer -= 1;
```
fig02, line 101: spawnTimer--; if it works

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +55,170 @@ const ActorInit Shot_Sun_InitVars = {
...
+        return;
+    }
+
+    switch (temp_v0) {
+        case 0x40:
+            sp38 = 7;
+            break;
+        case 0x41:
+            sp38 = 7;
```
fig02, line 110: these are the params for the fairy spawn, so naming it params will do

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -14,7 +33,17 @@ void ShotSun_Init(Actor* thisx, GlobalContext* globalCtx);
...
 
-/*
+void ShotSun_FairyCountdown(ShotSun* this, GlobalContext* globalCtx);
+void ShotSun_StartFairyCountdown(ShotSun* this, GlobalContext* globalCtx);
+void func_80BADF0C(ShotSun* this, GlobalContext* globalCtx);
+void ShotSun_HyliaSunUpdate(ShotSun* this, GlobalContext* globalCtx);
+
+extern s32 func_8005B198();
+extern void func_80062718(ColliderCylinder* collider, Vec3s* pos);
+extern void func_80078884(u16 sfxId);
```
Random06457, line 43: `extern` isn't required for function prototypes.
```suggestion
s32 func_8005B198();
void func_80062718(ColliderCylinder* collider, Vec3s* pos);
void func_80078884(u16 sfxId);
```

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
Random06457, line 76: `&` is not required for function pointers. The same applies below.
```suggestion
        this->actionFunc = func_80BADF0C;
```

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +55,170 @@ const ActorInit Shot_Sun_InitVars = {
...
+    Vec3f spawnPos;
+
+    if ((this->collider.base.acFlags & 2) != 0) {
+        func_80078884(NA_SE_SY_CORRECT_CHIME);
+        osSyncPrintf(VT_FGCOL(CYAN) "SHOT_SUN HIT!!!!!!!\n" VT_RST);
+        if (INV_CONTENT(SLOT_ARROW_FIRE) == ITEM_NONE) {
+            Actor_Spawn(&globalCtx->actorCtx, globalCtx, ACTOR_ITEM_ETCETERA, 700.0f, -800.0f, 7261.0f, 0, 0, 0, 7);
+            globalCtx->csCtx.segment = SEGMENTED_TO_VIRTUAL(&D_02007020);
+
+            // Maybe something was optimized away?
```
fig02, line 191: you dont need to comment this, it happens all the time. you can also remove the whitespace around it

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +55,170 @@ const ActorInit Shot_Sun_InitVars = {
...
+                collectible->unk_15A = 6000;
+                collectible->actor.speedXZ = 0.0f;
+            }
+        }
+        Actor_Kill(&this->actor);
+    } else {
+        if (!(120.0f < this->actor.xzDistFromLink) && gSaveContext.dayTime >= 0x4555 && gSaveContext.dayTime < 0x5000) {
+            cylinderPos.x = *(f32*)(&player->unk_908[0x54]) + globalCtx->envCtx.unk_04.x * 0.16666667f;
+            cylinderPos.y = *(f32*)(&player->unk_908[0x58]) - 30.0f + globalCtx->envCtx.unk_04.y * 0.16666667f;
+            cylinderPos.z = *(f32*)(&player->unk_908[0x5C]) + globalCtx->envCtx.unk_04.z * 0.16666667f;
```
fig02, line 211: might be able to get away with 0.167 or even 0.17. try it out

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.h
```diff
@@ -10,9 +10,11 @@ typedef void (*ShotSunActionFunc)(struct ShotSun*, GlobalContext*);
 
 typedef struct ShotSun {
     /* 0x0000 */ Actor actor;
-    /* 0x014C */ char unk_14C[0x4C];
+    /* 0x014C */ ColliderCylinder collider;
     /* 0x0198 */ ShotSunActionFunc actionFunc;
-    /* 0x019C */ char unk_19C[0xC];
+    /* 0x019C */ Vec3s unk_19C;
+    /* 0x01A2 */ s16 spawnTimer; // Frames until fairy spawns
```
fig02, line 16: generally when an actor only has one timer we just call it "timer" cause it can be implied what it does from its usage. when theres more than one though feel free to name it

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +55,170 @@ const ActorInit Shot_Sun_InitVars = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Shot_Sun/func_80BADF0C.s")
+            this->unk_19C = cylinderPos;
```
fig02, line 213: can name this instance variable hitboxPos or something. its the cylinder that the player has to shoot to spawn fire arrows

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +55,170 @@ const ActorInit Shot_Sun_InitVars = {
...
+        if (this->unk_1A4[0] == 2 && globalCtx->msgCtx.unk_E3EE == 4) {
+            if (((temp_t0 == 0x40) && (*(u16*)(&globalCtx->msgCtx.unk_E3E8[4]) == 9))
+                || ((temp_t0 == 0x41) && (*(u16*)(&globalCtx->msgCtx.unk_E3E8[4]) == 0xB))) {
+                    this->actionFunc = &ShotSun_StartFairyCountdown;
+                    func_80080480(globalCtx, this);
+                    this->spawnTimer = 0;
+            } else {
+                this->unk_1A4[0] = 0;
+            }
+
```
fig02, line 171: super tiny but id remove this white space. clang might even do this for you which you should run if you havent already.
youll need to install clang-tidy and clang-format and run ./format.sh from root

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +55,170 @@ const ActorInit Shot_Sun_InitVars = {
...
+void func_80BADF0C(ShotSun* this, GlobalContext* globalCtx) {
+    Player* player = PLAYER;
+    s32 pad;
+    s32 temp_t0 = this->actor.params & 0xFF;
+
+    if (22500.0f < func_800CB650(&this->actor.posRot.pos, &player->actor.posRot.pos)) {
+        this->unk_1A4[0] = 0;
+    } else {
+        if (this->unk_1A4[0] == 0) {
+            if (((s32)player->stateFlags2 << 7) >= 0) {
```
fig02, line 149: this cast shouldnt be needed

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +55,170 @@ const ActorInit Shot_Sun_InitVars = {
...
+        func_80078914(&this->actor.projectedPos, NA_SE_EV_TRE_BOX_APPEAR);
+    }
+}
+
+void func_80BADF0C(ShotSun* this, GlobalContext* globalCtx) {
+    Player* player = PLAYER;
+    s32 pad;
+    s32 temp_t0 = this->actor.params & 0xFF;
+
+    if (22500.0f < func_800CB650(&this->actor.posRot.pos, &player->actor.posRot.pos)) {
```
fig02, line 145: you might be able to get away with doing func_800CB650(&this->actor.posRot.pos, &player->actor.posRot.pos > 22500.0f 
if not though then oh well

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +55,170 @@ const ActorInit Shot_Sun_InitVars = {
...
+        case 0x41:
+            sp38 = 7;
+            break;
+    }
+
+    Actor_Spawn(
+        &globalCtx->actorCtx, globalCtx, ACTOR_EN_ELF,
+        this->actor.initPosRot.pos.x, this->actor.initPosRot.pos.y, this->actor.initPosRot.pos.z,
+        0, 0, 0,
+        sp38
```
Random06457, line 118: Is this something the formatter script did? If not you can run `./format.sh`

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +55,170 @@ const ActorInit Shot_Sun_InitVars = {
...
+        this->actionFunc = &ShotSun_HyliaSunUpdate;
+        this->actor.flags &= ~1;
+    }
+}
+
+void ShotSun_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+    ShotSun* this = THIS;
+    s32 temp_v0;
+
+    temp_v0 = this->actor.params & 0xFF;
```
fig02, line 90: in general throughout this file, id probably name the temp "params". it doesnt seem the params is used for more than one purpose so i dont think you need to specify what kind of param it is.

not too important if you dont feel comfortable naming it right now though, just a suggestion

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -14,7 +33,17 @@ void ShotSun_Init(Actor* thisx, GlobalContext* globalCtx);
...
+void ShotSun_StartFairyCountdown(ShotSun* this, GlobalContext* globalCtx);
+void func_80BADF0C(ShotSun* this, GlobalContext* globalCtx);
+void ShotSun_HyliaSunUpdate(ShotSun* this, GlobalContext* globalCtx);
+
+extern s32 func_8005B198();
+extern void func_80062718(ColliderCylinder* collider, Vec3s* pos);
+extern void func_80078884(u16 sfxId);
+
+extern UNK_TYPE D_02007020;
+
```
fig02, line 46: this is a cutscene, which would be typed as an array of CutsceneData[]

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
Sirius902: I tried both of those and neither work.

fig02: rip

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
Sirius902: It doesn't work without it.

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
Sirius902: Okay so I should name them params even though it's only the bottom byte? Also I named the fairy params fairyParams instead because that same function uses its own params as well.

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
fig02: oh, doing a shift and then checking for greater than or equal to zero is a compiler optimization. pretty sure this is player->stateFlags2 & 0x1000000 (i could be way off, i sucks at bit shifting stuff lol)
should be able to remove case after its written properly.

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
fig02: yea, generally params are bit packed and do a bunch of shifting to access different parts of the params. but in this case it doesnt, its only used for one purpose.
I guess ideally we could name it and call it a "type" or something-- 0x40 and 0x41 only make one small difference when calling the fairies. but i think "params" works at least for now

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -14,7 +33,17 @@ void ShotSun_Init(Actor* thisx, GlobalContext* globalCtx);
...
 
-/*
+void ShotSun_FairyCountdown(ShotSun* this, GlobalContext* globalCtx);
+void ShotSun_StartFairyCountdown(ShotSun* this, GlobalContext* globalCtx);
+void func_80BADF0C(ShotSun* this, GlobalContext* globalCtx);
+void ShotSun_HyliaSunUpdate(ShotSun* this, GlobalContext* globalCtx);
+
+s32 func_8005B198();
+void func_80062718(ColliderCylinder* collider, Vec3s* pos);
+void func_80078884(u16 sfxId);
```
fig02, line 43: for functions outside of the file (that are accessed globally) we have functions.h. youll want to put these there.
the third one here func_80078884 is already in functions.h

Sirius902: Where exactly in there should I put them?

fig02: the file is in order by address. lots of functions will already have an entry there just commented out. if they dont then add it in

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
Sirius902: Alright, I managed to match it by using `!(player->stateFlags2 & 0x1000000)`. Is that good?

fig02: perfect

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
Sirius902: done

Sirius902:
Would ShotSun_CheckCollision be a better name for ShotSun_HyliaSunUpdate? I think it's the only function that actually uses the hitbox.

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -1,10 +1,29 @@
...
  * Overlay: ovl_Shot_Sun
- * Description: Song of Storms Fairy
+ * Description: Song of Storms Fairy spawner and Lake Hylia Sun hitbox
  */
 
 #include "z_shot_sun.h"
+#include <vt.h>
+
+// TODO: Remove when `EnItem00` gets moved to a header file
+struct EnItem00;
```
fig02, line 11: we talked a bit in discord about this and think it would be best if the header for enitem00 is moved to z64actor.h. are you ok with doing it in this pr?
Would also move the other header for the actor in en_a_keep.c

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +51,157 @@ const ActorInit Shot_Sun_InitVars = {
...
+            break;
+    }
+
+    Actor_Spawn(&globalCtx->actorCtx, globalCtx, ACTOR_EN_ELF, this->actor.initPosRot.pos.x,
+                this->actor.initPosRot.pos.y, this->actor.initPosRot.pos.z, 0, 0, 0, fairyParams);
+
+    Actor_Kill(&this->actor);
+}
+
+void ShotSun_StartFairyCountdown(ShotSun* this, GlobalContext* globalCtx) {
```
fig02, line 115: Perhaps this can be called ShotSun_TriggerFairy ?
this function is starting the countdown but its also playing the "you did it" jingle and spawning the little circle of sparkles as the fairy starts to spawn

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +51,157 @@ const ActorInit Shot_Sun_InitVars = {
...
+    ShotSun* this = THIS;
+    s32 params = this->actor.params & 0xFF;
+
+    if (params != 0x40 && params != 0x41) {
+        Collider_DestroyCylinder(globalCtx, &this->collider);
+    }
+}
+
+void ShotSun_FairyCountdown(ShotSun* this, GlobalContext* globalCtx) {
+    s32 params = this->actor.params & 0xFF;
```
fig02, line 92: Id go with ShotSun_SpawnFairy personally

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +51,157 @@ const ActorInit Shot_Sun_InitVars = {
...
+        }
+        if (this->unk_1A4[0] == 1) {
+            func_8010BD58(globalCtx, 1U);
+            this->unk_1A4[0] = 2;
+            return;
+        }
+
+        if (this->unk_1A4[0] == 2 && globalCtx->msgCtx.unk_E3EE == 4) {
+            if (((params == 0x40) && (*(u16*)(&globalCtx->msgCtx.unk_E3E8[4]) == 9)) ||
+                ((params == 0x41) && (*(u16*)(&globalCtx->msgCtx.unk_E3E8[4]) == 0xB))) {
```
fig02, line 151: is all of this pointer stuff and dereferencing needed? you should be able to just access those values normally

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +51,157 @@ const ActorInit Shot_Sun_InitVars = {
...
+            } else {
+                this->unk_1A4[0] = 0;
+            }
+            this->unk_1A4[0] = 0;
+        }
+    }
+}
+
+void ShotSun_HyliaSunUpdate(ShotSun* this, GlobalContext* globalCtx) {
+    Vec3s cylinderPos;
```
fig02, line 164: this one is tough to name. we want to try to keep the actionFunc names as actual actions. but this literally does update everything for the sun hitbox..
curious if anyone else has good ideas for an action name for this

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +51,157 @@ const ActorInit Shot_Sun_InitVars = {
...
+            globalCtx->csCtx.segment = SEGMENTED_TO_VIRTUAL(&D_02007020);
+            if (1) {}
+            gSaveContext.cutsceneTrigger = 1;
+        } else {
+            spawnPos.x = 700.0f;
+            spawnPos.y = -800.0f;
+            spawnPos.z = 7261.0f;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Shot_Sun/func_80BADE74.s")
+            collectible = (EnItem00*)Item_DropCollectible(globalCtx, &spawnPos, 0xE);
```
fig02, line 183: odd to me that this function doesnt just return an EnItem00, which would avoid the casting. 
@Roman971 thoughts?

Roman971: It should return a `EnItem00*`, but it doesn't currently because it's exposed in functions.h and the EnItem00 struct isn't exposed globally.

fig02: oooo right
sirius when you move enitem00 to z64_actor can you change this as well?

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
Sirius902: Yes, here it's loading a 16 bit value from unknown data left as a char array in the message context so I need to do this unless a proper field is put on the message context.

fig02: if you know the type then you can set the msgContext variables up correctly. would be much preferred to do that instead

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
Sirius902: Yep! Done.

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
Sirius902: Done. How does it look?

fig02: looks great thanks

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
Sirius902: How about ShotSun_CollideArrow or something like that? It does sound a little awkward to me but that's what it's really doing overall.

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
Sirius902: Got it.

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
fig02: thats the tricky thing it also spawns the item when collided with an arrow.. idk
im leaning toward _HandleSun which is kind of an action but not really. if anything we can rename it later when we have better ideas

Sirius902: So should I use HandleSun for now?

fig02: eh.. keep your name for now and see if anyone else who reviews has an idea.
I would make it UpdateHyliaSun though

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +33,157 @@ const ActorInit Shot_Sun_InitVars = {
...
+
+void ShotSun_HyliaSunUpdate(ShotSun* this, GlobalContext* globalCtx) {
+    Vec3s cylinderPos;
+    Player* player = PLAYER;
+    EnItem00* collectible;
+    s32 pad;
+    Vec3f spawnPos;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Shot_Sun/func_80BADDCC.s")
+    if ((this->collider.base.acFlags & 2) != 0) {
```
fig02, line 152: oh i must have missed this last time, this a flags cehck so you dont need != 0

Sirius902: Right, should I also change the if condition in ShotSun_TriggerFairy to be
`func_8005B198() == this->actor.type || this->timer`?

fig02: only for boolean values. in that case its actually checking the the timer to not be 0

Sirius902: Got it.

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +33,157 @@ const ActorInit Shot_Sun_InitVars = {
...
+            if (!(player->stateFlags2 & 0x1000000)) {
+                player->stateFlags2 |= 0x800000;
+                return;
+            } else {
+                this->unk_1A4[0] = 1;
+            }
+        }
+        if (this->unk_1A4[0] == 1) {
+            func_8010BD58(globalCtx, 1U);
+            this->unk_1A4[0] = 2;
```
fig02, line 127: im so sorry i keep seeing stuff that i missed in the earlier reviews lmao
this U is probably not needed

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +33,157 @@ const ActorInit Shot_Sun_InitVars = {
...
+            globalCtx->csCtx.segment = SEGMENTED_TO_VIRTUAL(&D_02007020);
+            if (1) {}
+            gSaveContext.cutsceneTrigger = 1;
+        } else {
+            spawnPos.x = 700.0f;
+            spawnPos.y = -800.0f;
+            spawnPos.z = 7261.0f;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Shot_Sun/func_80BADE74.s")
+            collectible = Item_DropCollectible(globalCtx, &spawnPos, 0xE);
```
fig02, line 165: the 0xE is ITEM00_MAGIC_LARGE from the item00 params enum that you moved to z64_actor

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +33,157 @@ const ActorInit Shot_Sun_InitVars = {
...
+        case 0x40:
+            fairyParams = 7;
+            break;
+        case 0x41:
+            fairyParams = 7;
+            break;
+    }
+
+    Actor_Spawn(&globalCtx->actorCtx, globalCtx, ACTOR_EN_ELF, this->actor.initPosRot.pos.x,
+                this->actor.initPosRot.pos.y, this->actor.initPosRot.pos.z, 0, 0, 0, fairyParams);
```
Roman971, line 92: Probably a good idea to write a comment about UB here, since fairyParams is not always set to a value before being used here. The function is never used in a context where it would happen originally, but it could end up causing issues if people aren't aware of this while modding.

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.h
```diff
@@ -10,9 +10,11 @@ typedef void (*ShotSunActionFunc)(struct ShotSun*, GlobalContext*);
 
 typedef struct ShotSun {
     /* 0x0000 */ Actor actor;
-    /* 0x014C */ char unk_14C[0x4C];
+    /* 0x014C */ ColliderCylinder collider;
     /* 0x0198 */ ShotSunActionFunc actionFunc;
-    /* 0x019C */ char unk_19C[0xC];
+    /* 0x019C */ Vec3s hitboxPos;
+    /* 0x01A2 */ s16 timer; // Frames until fairy spawns
+    /* 0x01A4 */ char unk_1A4[4];
```
Roman971, line 17: This shouldn't be a char array since it's only used as a byte. And the other 3 bytes afterwards would be struct padding.
```suggestion
    /* 0x01A4 */ u8 unk_1A4;
```

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
```diff
@@ -26,17 +33,157 @@ const ActorInit Shot_Sun_InitVars = {
...
+    s32 pad;
+    Vec3f spawnPos;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Shot_Sun/func_80BADDCC.s")
+    if (this->collider.base.acFlags & 2) {
+        func_80078884(NA_SE_SY_CORRECT_CHIME);
+        osSyncPrintf(VT_FGCOL(CYAN) "SHOT_SUN HIT!!!!!!!\n" VT_RST);
+        if (INV_CONTENT(SLOT_ARROW_FIRE) == ITEM_NONE) {
+            Actor_Spawn(&globalCtx->actorCtx, globalCtx, ACTOR_ITEM_ETCETERA, 700.0f, -800.0f, 7261.0f, 0, 0, 0, 7);
+            globalCtx->csCtx.segment = SEGMENTED_TO_VIRTUAL(&D_02007020);
```
Roman971, line 157:
```suggestion
            globalCtx->csCtx.segment = SEGMENTED_TO_VIRTUAL(D_02007020);
```

src/overlays/actors/ovl_Shot_Sun/z_shot_sun.c
Sirius902: I added a comment. Is it good?

Roman971: Looks alright to me. And it should be easy to find whenever we introduce a build flag to avoid UB.

PR #241, opened 2020-07-06
Disassemble more cutscenes

Thar0:
Addresses the backlog of not yet disassembled cutscene data that is currently in the repo. From here on cutscenes can be disassembled before merging.
Also fixes a bug with csdis where some commands would emit too many args.


fig02:
can you do demo_sa's cs data also? was merged recently

PR #243, opened 2020-07-08
z_skin_matrix.c decompiled (1 non matching, 1 non equivalent)

AdamKiddle:
`z_skin_matrix.c` matching except for:
- `func_800A73E0` non matching
- `func_800A7EC0` non matching and possibly non equivalent

`sys_matrix.c`:
-  Matched previously non matching function `Matrix_MtxFToMtx`


src/code/z_skin_matrix.c
```diff
@@ -1,42 +1,692 @@
...
+        sin2 = Math_Sins(roll);
+        cos2 = Math_Coss(roll);
+
+        yx = mf->yx;
+        zx = mf->zx;
+        mf->yx = (yx * cos2) + (zx * sin2);
+        mf->zx = (zx * cos2) - (yx * sin2);
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_skin_matrix/func_800A76A4.s")
+        if (1) {} // required to match
```
fig02, line 357: we will assume that anything dumb like this will be required for matching, i dont think we should point it out with comments. same for all the other ones below

src/code/z_skin_matrix.c
```diff
@@ -1,42 +1,692 @@
...
+
+    rx = mfA->wx;
+    ry = mfA->wy;
+    rz = mfA->wz;
+    rw = mfA->ww;
+    dest->ww = (cx * rx) + (cy * ry) + (cz * rz) + (cw * rw);
+}
+
+MtxF D_8012A4A0 = { 1.0f, 0.0f, 0.0f, 0.0f, 0.0f, 1.0f, 0.0f, 0.0f, 0.0f, 0.0f, 1.0f, 0.0f, 0.0f, 0.0f, 0.0f, 1.0f };
+
```
fig02, line 171: if you want to keep this formatted in rows and columns you can use // clang-format off and // clang-format on to temporarily disable the formatter between the two commens

src/code/z_skin_matrix.c
```diff
@@ -1,42 +1,692 @@
...
+                                        f32 dx, f32 dy, f32 dz) {
+    MtxF mft1;
+    MtxF mft2;
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_skin_matrix/func_800A7E70.s")
+    SkinMatrix_Translate(mf, dx, dy, dz);            // translation matrix
+    SkinMatrix_RotateRPY(&mft1, roll, pitch, yaw);   // rotation matrix
+    SkinMatrix_MtxFMtxFMult(mf, &mft1, &mft2);       // MtxFMtxFMult
+    SkinMatrix_Scale(&mft1, scaleX, scaleY, scaleZ); // scale matrix
+    SkinMatrix_MtxFMtxFMult(&mft2, &mft1, mf);       // MtxFMtxFMult
```
fig02, line 465: these comments seem redundant cause the name of the function being called tells you this info

src/code/z_skin_matrix.c
```diff
@@ -1,42 +1,692 @@
...
+
+    temp = src->ww * 0x10000;
+    m1[15] = (temp >> 0x10);
+    m2[15] = temp & 0xFFFF;
+}
+
+Mtx* func_800A7E70(GraphicsContext* gfxCtx, MtxF* src) {
+    Mtx* displayList = Graph_Alloc(gfxCtx, 0x40);
+
+    if (displayList == 0) {
```
fig02, line 577: is this a dlist? if it was it wouldnt be of type Mtx*
ID also make the comparison below for NULL instead of 0

src/code/z_skin_matrix.c
```diff
@@ -1,42 +1,692 @@
...
+    mf->wx = 0.0f;
+    mf->wy = 0.0f;
+    mf->wz = 0.0f;
+    mf->ww = 1.0f;
+}
+#else
 #pragma GLOBAL_ASM("asm/non_matchings/code/z_skin_matrix/func_800A7EC0.s")
+#endif
+
+// unknown
```
fig02, line 645: we'll assume that anything not documented is not understood yet, can take this out imo

src/code/z_skin_matrix.c
```diff
@@ -1,42 +1,734 @@
...
 #include <global.h>
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_skin_matrix/func_800A6E10.s")
+/**
+ * Multiplies a 4 component row vector [ src , 1 ] by the matrix mf and writes the resulting 4 components to xyzDest
+ * and wDest.
+ * \f[
+ *     \left[\texttt{xyzDest}, \texttt{wDest}\right] = \left[\texttt{src}, 1 \right] \cdot \left[mf\right]
+ * \f]
+ */
```
fig02, line 10: think we talked about this in discord but you can def remove this stuff. we're gonna prioritize how it looks in the code, not doxygen

include/functions.h
```diff
@@ -1297,22 +1297,22 @@ void func_800A663C(GlobalContext* globalCtx, PSkinAwb*, SkeletonHeader*, Animati
...
-void func_800A76A4(MtxF* dest, f32 x, f32 y, f32 z);
-void func_800A7704(MtxF* dest, s16 x, s16 y, s16 z);
-// ? func_800A7894(?);
-void func_800A7A24(MtxF* dest, f32 x, f32 y, f32 z);
-// ? func_800A7B04(?);
-// ? func_800A7B84(?);
-// ? func_800A7C60(?);
+void SkinMatrix_Scale(MtxF* mf, f32 xScale, f32 yScale, f32 zScale);
+void SkinMatrix_RotateRPY(MtxF* mf, s16 roll, s16 pitch, s16 yaw);
+void SkinMatrix_RotateYRP(MtxF* mf, s16 yaw, s16 roll, s16 pitch);
```
Roman971, line 1309: I could be wrong but some functions in here seem to be unused outside of the file?
If they aren't used globally they should be omitted from `functions.h`, just like `variables.h` is only meant for variables used globally.

src/code/z_skin_matrix.c
Thar0: I don't think this should be removed. It doesn't impact readability in code since there is readable comments above it. If the two were blended in some way it would make sense to prioritise code readability, but this is entirely separate. It's very helpful to have these for math-heavy functions as reading the maths straight out of the code can often be difficult and very time-consuming if the function is large.

fig02: i personally see it as clutter, but if it makes things alot easier to understand and is worth it, then sure ignore what i said

Roman971: It definitely impacts readability to some extent just by being there. That being said could it be written in a way that is readable as raw text, ideally over a single line?

src/code/z_skin_matrix.c
```diff
@@ -1,42 +1,734 @@
...
+
+    rx = mfA->wx;
+    ry = mfA->wy;
+    rz = mfA->wz;
+    rw = mfA->ww;
+    dest->ww = (cx * rx) + (cy * ry) + (cz * rz) + (cw * rw);
+}
+
+// clang-format off
+MtxF sGlobalIdentity = {
```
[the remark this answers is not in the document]
AdamKiddle: As a compromise (if you're okay with this), I will change these names to "Clear" to keep up with the codebase but since that name is very unconventional I will add/keep any comments referring to these as Identity matrices. I know if I came back in a few months and saw "MtxFClear" I wouldn't have a clue what that meant.

src/code/z_skin_matrix.c
```diff
@@ -1,42 +1,734 @@
...
+    my = mf->yz;
+    mz = mf->zz;
+    mw = mf->wz;
+    dest->z = mw + ((src->x * mx) + (src->y * my) + (src->z * mz));
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_skin_matrix/func_800A72FC.s")
+/**
+ * Matrix multiplication, dest = mfA * mfB.
+ * Note that mfB and dest may be the same matrix thanks to use of temp variables.
```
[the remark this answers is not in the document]
AdamKiddle: This comment shows the order of matrix multiplcation which is important. Or do you think it's implicit from the fact I named them A and B?

Also, for the second line I could change the comment to something like:
"dest and mfA should the same matrix"

src/code/z_skin_matrix.c
AdamKiddle: Put it on one line and removed the \left and \right tags

src/code/z_skin_matrix.c
```diff
@@ -1,42 +1,734 @@
...
+    0.0f, 0.0f, 1.0f, 0.0f,
+    0.0f, 0.0f, 0.0f, 1.0f,
+};
+// clang-format on
+
+/**
+ * Sets mf to a pointer to a global identity matrix in data.
+ */
+void SkinMatrix_GlobalIdentity(MtxF** mf) {
+    *mf = &sGlobalIdentity;
```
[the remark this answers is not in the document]
AdamKiddle: Changed it to "SkinMatrix_Clear" as the "setClear" and "getClear" may otherwise get confused as getting and setting the same variable.

src/code/z_skin_matrix.c
```diff
@@ -1,42 +1,734 @@
...
+}
+
+/**
+ * Sets the components of the matrix at mf to be the identity matrix.
+ */
+void SkinMatrix_SetToIdentity(MtxF* mf) {
```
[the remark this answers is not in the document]
AdamKiddle: Changed it to "SkinMatrix_Clear" as the "setClear" and "getClear" may otherwise get confused as getting and setting the same variable.

src/code/z_skin_matrix.c
```diff
@@ -1,42 +1,728 @@
...
+}
+#else
+#pragma GLOBAL_ASM("asm/non_matchings/code/z_skin_matrix/SkinMatrix_Invert.s")
+#endif
+
+/**
+ * Produces a matrix which scales x,y,z components of vectors or x,y,z rows of matrices (when applied on LHS)
+ */
+void SkinMatrix_SetScaling(MtxF* mf, f32 x, f32 y, f32 z) {
+    mf->xy = 0.0f;
```
fig02, line 316: the corresponding function in sys_matrix.c is simply called Matrix_Scale, is there any reason it needs to say scaling in this context?

fig02: i guess they arent really the same function, but im still wondering is Scale or SetScale would be better

AdamKiddle: I renamed it this due to Roman's suggestion (naming things "SetX" to clarify that these set an existing matrix rather than creating a new one like in sys_matrix).

Tha name "scaling" is the name of this type of matrix in the literature, I forgot that sys_matrix has a weird name for it lol. I'll set it back to "Scale" for now to be consistent with the codebase but would prefer if we changed all the matrix names to be consistent with the actual mathematical names for these objects

fig02: I see, yeah I just wasnt sure if thats the "proper" way to call it, as I know next to nothing about matrix math.
If anything, now is the time to change them in sys_matrix and make everything proper and consistent, but interested what others think.

asm/non_matchings/code/z_skin_matrix/SkinMatrix_Invert.s
```diff
@@ -17,7 +17,7 @@ glabel D_80140844
     .float 0.0005
 
 .text
-glabel func_800A73E0
+glabel Skin_Matrix_Invert
```
Thar0, line 20: Shouldn't this be SkinMatrix_Invert? Nonmatching builds will currently complain as the nonmatching version of this function is already SkinMatrix_Invert but all references are Skin_Matrix_Invert.
Rest looks good otherwise.

AdamKiddle: Good catch!

src/code/z_skin_matrix.c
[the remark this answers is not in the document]
Roman971: This comment still needs to be addressed.

src/code/z_skin_matrix.c
```diff
@@ -1,42 +1,734 @@
...
+// Saved register usage differences and probably regalloc
+/**
+ * Inverts a matrix using a slight modification of the Gauss-Jordan method
+ * (column operations instead of row operations).
+ * @returns 0 if successfully inverted
+ * @returns 2 if matrix non-invertible (0 determinant)
+ */
+#ifdef NON_MATCHING
+s32 func_800A73E0(MtxF* src, MtxF* dest) {
+    MtxF mf_copy;
```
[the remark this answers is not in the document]
Roman971: The other variables should probably be cleaned up for the same reason.

src/code/z_skin_matrix.c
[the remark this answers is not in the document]
AdamKiddle: I mentioned this on the previous comment (must've clicked resolved on that by accident) but I wanted to keep this comment as it's not obvious what "Clear" means in this context. I'll unresolve the previous one so you can see the reply there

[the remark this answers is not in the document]
Roman971: Ah, I was referring to the rest of the comment but I wasn't clear on that part, sorry.

src/code/z_skin_matrix.c
```diff
@@ -1,42 +1,728 @@
...
+// Saved register usage differences and probably regalloc
+s32 SkinMatrix_Invert(MtxF* src, MtxF* dest) {
+    MtxF mfCopy;
+    s32 i;
+    s32 k;
+    f32 temp1;
+    f32 temp2;
+    f32 diag_element;
+    s32 this_row;
+    s32 this_col;
```
fig02, line 252: not sure if this was mentioned in a previous comment and missed, or if this was overlooked in general, but there should be no snake_case variables. if theres anymore in the file change those too please :D

PR #249, opened 2020-07-12
Rename function with wrong memory address

shawlucas:
Renamed func_8008EF40 to func_8008EF30 in z_player_lib.c, because the function is actually at address 0x8008EF30.


shawlucas:
Ignore this, ill open a new one soon when i fix my repo

PR #255, opened 2020-07-14
Decomp BgYdanHasi

louist103:
Decomp BgYdanHasi and fix an issue with BgJyaLift's action func

PR #257, opened 2020-07-15
Add .editorconfig

L-P:
After reading [a comment on another PR](https://github.com/zeldaret/oot/pull/255/files#r455288401) I thought we'd need an editor-agnostic configuration file to avoid these kind of hard to catch mistakes.

See https://editorconfig.org/ for details.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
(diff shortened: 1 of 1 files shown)


Roman971:
I think we should keep this idea in mind, but I believe we concluded not to go with this for now, so we probably don't need to hold an open PR for it. We can always find it again whenever we decide to add this kind of file.

PR #268, opened 2020-07-18
Effect rodata migration

ethteck:


asm/non_matchings/overlays/effects/ovl_Effect_Ss_Blast/func_8099EF08.s
```diff
@@ -1,3 +1,22 @@
...
+    .balign 4
+
+glabel D_8099F148
+    .asciz "../z_eff_ss_blast.c"
+    .balign 4
+
+.late_rodata
+glabel D_8099F15C
+    .float 0.0025
+    .balign 4
```
Roman971, line 17: Floats should still not have any .balign after them. It would be nice to clean that up someday like we agreed back when they were added..

ethteck: Done :)

PR #281, opened 2020-07-20
ovl_Item_Etcetera OK

shawlucas:
its OK
also tried to name some things but I'm bad at naming things so they might not be good


shawlucas:
ignore this i messed up something

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.h
```diff
@@ -6,11 +6,31 @@
...
+typedef void (*ItemEtceteraDrawFunc)(struct ItemEtcetera*, GlobalContext*);
+
 typedef struct ItemEtcetera {
     /* 0x0000 */ Actor actor;
-    /* 0x014C */ char unk_14C[0x14];
+    /* 0x014C */ ItemEtceteraActionFunc actionFunc_14C;
+    /* 0x0150 */ s16 drawId;
+    /* 0x0152 */ s16 getItemId;
+    /* 0x0154 */ u8 objBankIndex;
+    /* 0x0155 */ char unk_155[0x3];
```
ethteck, line 18: I think this is padding and can be removed.

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
```diff
@@ -13,8 +13,17 @@
 void ItemEtcetera_Init(Actor* thisx, GlobalContext* globalCtx);
 void ItemEtcetera_Destroy(Actor* thisx, GlobalContext* globalCtx);
 void ItemEtcetera_Update(Actor* thisx, GlobalContext* globalCtx);
+void func_80B85C64(Actor* thisx, GlobalContext* globalCtx);
+void ItemEtcetera_DrawFireArrow(Actor* thisx, GlobalContext* globalCtx);
```
fig02, line 17: commented on this in another pr, but imo separate the og main funcs from the extra ones

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
```diff
@@ -26,29 +35,186 @@ const ActorInit Item_Etcetera_InitVars = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Item_Etcetera/ItemEtcetera_Update.s")
+// Indexes passed to the item table in z_draw.c
+static s16 sDrawItemTableIdx[] = {
+    0x0000, 0x0044, 0x002B, 0x0015, 0x0029, 0x002A, 0x0001, 0x005F, 0x006C, 0x006D, 0x006E, 0x0070, 0x0013, 0x0001,
+};
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Item_Etcetera/func_80B85C64.s")
+static s16 sGetEtceteraItemId[] = {
+    GI_BOTTLE,     GI_LETTER_RUTO, GI_SHIELD_HYLIAN, GI_QUIVER_40, GI_SCALE_SILVER, GI_SCALE_GOLD, GI_KEY_SMALL,
```
fig02, line 51: sEtceteraGetItemId maybe?
probably dont even need the word etcetera in there, sGetItemIDs is good enough imo

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
```diff
@@ -26,29 +35,186 @@ const ActorInit Item_Etcetera_InitVars = {
...
+    GI_ARROW_FIRE, GI_INVALID,     GI_INVALID,       GI_INVALID,   GI_INVALID,      GI_INVALID,    GI_INVALID,
+};
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Item_Etcetera/func_80B85CB8.s")
+void ItemEtcetera_SetupAction(ItemEtcetera* this, ItemEtceteraActionFunc actionFunc) {
+    this->actionFunc_15C = actionFunc;
```
fig02, line 56: any reason the offset was left on the field name?

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
```diff
@@ -26,29 +35,186 @@ const ActorInit Item_Etcetera_InitVars = {
...
+    osSyncPrintf("bank_ID = %d\n", objBankIndex);
+    if (objBankIndex < 0) {
+        __assert("0", "../z_item_etcetera.c", 241);
+    } else {
+        this->objBankIndex = objBankIndex;
+    }
+    this->drawId = sDrawItemTableIdx[type];
+    this->getItemId = sGetEtceteraItemId[type];
+    this->actionFunc_14C = func_80B85824;
+    this->currentDrawRoutine = ItemEtcetera_DrawFireArrow;
```
fig02, line 77: we never really call functions routines anywhere else. currentDrawFunc at least
is it important to specify current? just drawFunc might be good enough

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
```diff
@@ -26,29 +35,186 @@ const ActorInit Item_Etcetera_InitVars = {
...
+    if (func_8002F410(&this->actor, globalCtx) != 0) {
+        if ((this->actor.params & 0xFF) == 1) {
+            gSaveContext.eventChkInf[3] |= 2;
+            Flags_SetSwitch(globalCtx, 0xB);
+        }
+        Actor_Kill(&this->actor);
+    } else {
+        if (0) {} // Necessary to match
+        func_8002F434(&this->actor, globalCtx, this->getItemId, 30.0f, 50.0f);
+        if ((globalCtx->gameplayFrames & 0xD) == 0) {
```
fig02, line 141: think this is gonna be % 14 but could be wrong

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.h
```diff
@@ -6,11 +6,30 @@
...
+typedef void (*ItemEtceteraDrawFunc)(struct ItemEtcetera*, GlobalContext*);
+
 typedef struct ItemEtcetera {
     /* 0x0000 */ Actor actor;
-    /* 0x014C */ char unk_14C[0x14];
+    /* 0x014C */ ItemEtceteraActionFunc actionFunc_14C;
+    /* 0x0150 */ s16 drawId;
+    /* 0x0152 */ s16 getItemId;
+    /* 0x0154 */ u8 objBankIndex;
+    /* 0x0158 */ ActorFunc currentDrawRoutine; // gets set to this->actor.draw
```
fig02, line 18: yeah im thinking just drawFunc for the name, and the comment is implied at that point and can be removed

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
fig02: disregard

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
shawlucas: Well there is an actionFunc_15C and an actionFunc_14C. I can't find a pattern for when they exactly used which one for what things. Let me know if you have a better name though or if you notice a specific purpose for one of them

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
shawlucas: I think calling it current was to differentiate from just the regular this->actor.draw. I don't really know what else to call it but ill remove routine

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
shawlucas: It messes everything up if i try to use mod here

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.h
shawlucas: Okay I'll change to drawFunc

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
shawlucas: Actually i'll change to drawFunc

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
Roman971: `actionFunc_15C` is the actual `actionFunc`, while `actionFunc_14C` should be called something like `actionFuncAfterLoad` or `futureActionFunc`. And what you called `drawFunc` should also be called with the same pattern, since it also gets set after the object is loaded.

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
```diff
@@ -26,29 +35,186 @@ const ActorInit Item_Etcetera_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Item_Etcetera/func_80B85B6C.s")
+static s16 sObjectIds[] = {
+    OBJECT_GI_BOTTLE, OBJECT_GI_BOTTLE_LETTER, OBJECT_GI_SHIELD_2, OBJECT_GI_ARROWCASE, OBJECT_GI_SCALE,
+    OBJECT_GI_SCALE,  OBJECT_GI_KEY,           OBJECT_GI_M_ARROW,  OBJECT_GI_RUPY,      OBJECT_GI_RUPY,
+    OBJECT_GI_RUPY,   OBJECT_GI_RUPY,          OBJECT_GI_HEARTS,   OBJECT_GI_KEY,
+};
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Item_Etcetera/ItemEtcetera_Update.s")
+// Indexes passed to the item table in z_draw.c
+static s16 sDrawItemTableIdx[] = {
```
Roman971, line 46:
```suggestion
static s16 sDrawItemIndexes[] = {
```

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
```diff
@@ -26,29 +35,186 @@ const ActorInit Item_Etcetera_InitVars = {
...
+    0x0000, 0x0044, 0x002B, 0x0015, 0x0029, 0x002A, 0x0001, 0x005F, 0x006C, 0x006D, 0x006E, 0x0070, 0x0013, 0x0001,
+};
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Item_Etcetera/func_80B85C64.s")
+static s16 sGetItemId[] = {
```
Roman971, line 50:
```suggestion
static s16 sGetItemIds[] = {
```

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
```diff
@@ -26,29 +35,186 @@ const ActorInit Item_Etcetera_InitVars = {
...
+            Actor_SetScale(&this->actor, 0.5f);
+            this->actor.draw = NULL;
+            this->actor.shape.unk_08 = 50.0f;
+            break;
+        case 8:
+        case 9:
+        case 10:
+        case 11:
+        case 12:
+        case 13: // unused values?
```
Roman971, line 99: I believe these types are the ones used in the treasure chest minigame.
Looking at the draw function, you can see they are only drawn when the lens is active (actorCtx.unk_03 being set). And the object IDs + draw indexes correspond to (in order): Green Rupee, Blue Rupee, Red Rupee, Purple Rupee, Piece of Heart and Small Key.
I suggest naming types for these in this format: `ITEM_ETCETERA_RUPEE_GREEN_CMG`, `ITEM_ETCETERA_RUPEE_BLUE_CMG`, etc. CMG being short for chest minigame.

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
```diff
@@ -26,29 +35,186 @@ const ActorInit Item_Etcetera_InitVars = {
...
+        this->actionFunc_15C = ItemEtcetera_MoveFireArrowDown;
+    }
+}
+
+void ItemEtcetera_Update(Actor* thisx, GlobalContext* globalCtx) {
+    ItemEtcetera* this = THIS;
+    this->actionFunc_15C(this, globalCtx);
+}
+
+void func_80B85C64(Actor* thisx, GlobalContext* globalCtx) {
```
Roman971, line 205: Following my previous comment, I think this draw function can be named fairly accurately with something like this:
```suggestion
void ItemEtcetera_DrawThroughLens(Actor* thisx, GlobalContext* globalCtx) {
```

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
fig02: does nextActionFunc make sense instead of future?

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
```diff
@@ -26,29 +35,186 @@ const ActorInit Item_Etcetera_InitVars = {
...
+    ItemEtcetera* this = THIS;
+    if (globalCtx->actorCtx.unk_03 != 0) {
+        func_8002EBCC(&this->actor, globalCtx, 0);
+        func_8002ED80(&this->actor, globalCtx, 0);
+        func_800694A0(globalCtx, this->drawId);
+    }
+}
+
+void ItemEtcetera_DrawFireArrow(Actor* thisx, GlobalContext* globalCtx) {
```
Roman971, line 214: This function is used for all non-cmg items, not just fire arrows, so I suggest simply calling it "Draw", to go along with the other draw function.
```suggestion
void ItemEtcetera_Draw(Actor* thisx, GlobalContext* globalCtx) {
```

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.h
```diff
@@ -6,11 +6,30 @@
 
 struct ItemEtcetera;
 
+typedef void (*ItemEtceteraActionFunc)(struct ItemEtcetera*, GlobalContext*);
+typedef void (*ItemEtceteraDrawFunc)(struct ItemEtcetera*, GlobalContext*);
```
Roman971, line 10:
```suggestion
```

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
Roman971: It does not, because these functions are specifically being used after the object is loaded, and they will not do anything if you change them later on. "next" is normally used in cases where we know it will take over on the next frame, or the next event or whatever else makes sense in the given context, which isn't the case here.

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.c
fig02: Id go with the latter even though its long, just cause its not a common acronym that anyone will know what it means. Even with context of what the actor is doing imo itll be confusing

Roman971: CMG is fairly common or at least it's the most common one for this minigame afaik.

fig02: maybe in the rando scene? coming from speedrunning this is never abbreviated at all 
but ye either way I think itll be fine without

Roman971: Yeah mostly the rando scene. 
35+ characters for a name is a bit much but I guess it's not a huge issue.

fig02: will probably look nicer when we arent calling this thing ItemEtcetera 
but then again who knows what the readable name will be lol

src/overlays/actors/ovl_Item_Etcetera/z_item_etcetera.h
```diff
@@ -27,7 +26,13 @@ typedef enum {
...
     /* 0x05 */ ITEM_ETCETERA_SCALE_GOLD,
     /* 0x06 */ ITEM_ETCETERA_KEY_SMALL,
-    /* 0x07 */ ITEM_ETCETERA_ARROW_FIRE
+    /* 0x07 */ ITEM_ETCETERA_ARROW_FIRE,
+    /* 0x08 */ ITEM_ETCETERA_RUPEE_GREEN_CHEST_GAME,
+    /* 0x09 */ ITEM_ETCETERA_RUPEE_BLUE_CHEST_GAME,
+    /* 0x0A */ ITEM_ETCETERA_RUPEE_RED_CHEST_GAME,
+    /* 0x0B */ ITEM_ETCETERA_RUPEE_PURPLE_CHEST_GAME,
+    /* 0x0C */ ITEM_ETCETERA_HEART_PIECE_CHEST_GAME,
+    /* 0x0D */ ITEM_ETCETERA_SMALL_KEY_CHEST_GAME
```
Roman971, line 35: For consistency with `ITEM_ETCETERA_KEY_SMALL`:
```suggestion
    /* 0x0D */ ITEM_ETCETERA_KEY_SMALL_CHEST_GAME
```

PR #284, opened 2020-07-21
Ovl_Bg_Jya_Megami OK

gamestabled:


src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -21,31 +29,325 @@ const ActorInit Bg_Jya_Megami_InitVars = {
...
+}
+
+#ifdef NON_MATCHING
+// Equivalent, 1 stack placement incorrect
+void func_8089A1DC(GlobalContext* globalContext, Vec3f* burstOrigin, Vec3f* burstDepthX, s32 num, s32 maskIndex) {
+    Vec3f spB4;
+    s32 i;
+
+    for (i = 0; i < num; ++i) {
+        u32* dList = &D_0600B9F8;
```
fig02, line 138: dlists are always an array of Gfx[]
Also are you sure its a dlist? it gets passed into an effect so its a good guess, but i dont know if you can say for sure yet

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -9,7 +9,15 @@ void BgJyaMegami_Destroy(Actor* thisx, GlobalContext* globalCtx);
...
 
-/*
+void BgJyaMegami_SetupDetectLight(BgJyaMegami* this);
+void BgJyaMegami_DetectLight(BgJyaMegami* this, GlobalContext* globalContext);
+void BgJyaMegami_SetupExplosion(BgJyaMegami* this);
+void BgJyaMegami_Explosion(BgJyaMegami* this, GlobalContext* globalContext);
+
+extern u32 D_06005780;
+extern u32 D_06005C4C;
+extern u32 D_0600B9F8;
```
fig02, line 19: D_06005780 is a dlist, type Gfx[]
D_06005C4C is a dynapoly collision header, but it should be UNK_TYPE while we wait for bgcheck stuff
D_0600B9F8 is what i mentioned in a previous comment, imo you cant say for sure what it is yet unless you had a look at the object file or something, so UNK_TYPE

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -21,31 +29,325 @@ const ActorInit Bg_Jya_Megami_InitVars = {
...
+    gSPDisplayList(gfxCtx->polyOpa.p++, &D_06005780);
+    Graph_CloseDisps(dispRefs, globalContext->state.gfxCtx, "../z_bg_jya_megami.c", 720);
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Jya_Megami/func_8089A604.s")
+void BgJyaMegami_DrawExplosion(BgJyaMegami* this, GlobalContext* globalContext) {
+    GraphicsContext* gfxCtx;
+    BgJyaMegami_Piece* temp;
+    u32 i;
+    Gfx* dispRefs[5];
```
fig02, line 320: since we know open disps is part of a macro, it would be helpful if you make it work like the macro does while we wait for the macro to be implemented.
The pattern is that dispRefs should always be size 4. You may need to add a pad. And since it gets declared below the normal block at the top of the function, it should be last in your declaration order. With gfxCtx being on top.
My guess would be something like:

    s32 pad;
    BgJyaMegami_Piece* temp;
    u32 i;
    GraphicsContext* gfxCtx;
    Gfx* dispRefs[4];

fig02: and maybe just name the temp piece

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -21,31 +29,325 @@ const ActorInit Bg_Jya_Megami_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Jya_Megami/func_8089AD48.s")
+void BgJyaMegami_Draw(Actor* thisx, GlobalContext* globalContext) {
+    BgJyaMegami* this = THIS;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Jya_Megami/BgJyaMegami_Draw.s")
+    func_800628A4(0, &this->collider);
+    if (this->actionFunc == BgJyaMegami_Explosion) {
+        BgJyaMegami_DrawExplosion(this, globalContext);
+    } else {
+        BgJyaMegami_DrawDetectLight(this, globalContext);
```
fig02, line 351: confused by this name, what is it actually drawing? the mirror shield patter on the face?
Think it can probably be phrased better

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.h
```diff
@@ -6,9 +6,34 @@
...
+} BgJyaMegami_PieceInit; // size = 0x18
+
+typedef struct BgJyaMegami_Piece {
+    /* 0x00 */ Vec3f pos;
+    /* 0x0C */ Vec3f vel;
+    /* 0x18 */ s16 rotVelX;
+    /* 0x1A */ s16 rotVelY;
+} BgJyaMegami_Piece; // size = 0x1C
+
+static BgJyaMegami_PieceInit sPiecesInit[13];
```
fig02, line 26: I dont know if including BgJyaMegami_PieceInit and the sPiecesInit declaration in the header is worth it for the sake of using array count. but i want others opinions as well

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -21,31 +29,325 @@ const ActorInit Bg_Jya_Megami_InitVars = {
...
+    { { 14.759999f, -125.8f, -44.16f }, 0.2f, 0x0320, 0x0258, 20 },
+};
+
+s16 sScales[] = {
+    0x0005, 0x0008, 0x000B, 0x000E, 0x0011, 0x0014, 0x0017, 0x001A,
+};
+
+s16 sDurations[] = {
+    0x0012, 0x001A, 0x0022, 0x002A, 0x0032, 0x003C, 0x0046, 0x0050,
+};
```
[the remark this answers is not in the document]
gamestabled: I'm actually not 100% sure, I took some of these names from the prototype for func_80029E8C, which is an effect spawner. Maybe I should just leave them unnamed for now

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
gamestabled: as above, i named it based off the prototype for func_80029E8C, but yeah I can just leave it as unknown for now

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
gamestabled: didnt know about this macro part, but makes sense

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
gamestabled: there are two different actionFuncs and the main draw function goes into a sub draw function for each of them, so the naming scheme I used is BgJyaMegami_Draw[ActionFuncName]. I think for DrawExplosion this works well but maybe DrawDetectLight should just become DrawFace or DrawNormal? not sure

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.h
gamestabled: seems worth it to me for the sake of easy modding, but understandable either way

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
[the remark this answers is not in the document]
fig02: oh gotcha gotcha. yea the duration of effects is called "life" and will be decimal. but yea i think itd be better if it was just left unknown until the effect is done. im working on getting through all the effect overlays now anyway so itll be a matter of time FeelsOkMan

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
fig02: DrawFace works best if its drawing the face that detects the light imo

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.h
ethteck: I'm for it in theory, but I wish we had a more consistent approach to this

Roman971: Doing it this way is very confusing and unfortunate imo. How about having a define for the piece count, used as the size of both arrays?

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.h
```diff
@@ -4,11 +4,28 @@
 #include <ultra64.h>
 #include <global.h>
 
+#define BGJYAMEGAMI_NUMPIECES 13
```
gamestabled, line 7: I'm not sure if there's a naming scheme for defines like this, also it seems weird to me to have it in the header, but I do need it for the array in the actor struct.

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -21,31 +37,325 @@ const ActorInit Bg_Jya_Megami_InitVars = {
...
+        } else if (this->explosionTimer == temp2->delay) {
+            Math_Vec3f_Sum(&temp->pos, &temp2->unk_00, &sp8C);
+            sp8C.z += 10.0f;
+            func_8089A1DC(globalContext, &sp8C, &temp->vel, 4, 2);
+        }
+    }
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Jya_Megami/func_8089A1DC.s")
+    if ((this->explosionTimer % 4 == 0) && (this->explosionTimer > 30) && (this->explosionTimer < 80) &&
+        (this->explosionTimer > 40)) {
```
gamestabled, line 282: the redundancy of these checks makes me think that the constants 30 and 40 were originally defines, and because they are also magic constants elsewhere. 30 is the number of frames between the explosion beginning and the correct chime playing. 40 is the number of frames of light the face needs before it explodes (its "health"). Is it worth making these defines now?

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -21,31 +37,325 @@ const ActorInit Bg_Jya_Megami_InitVars = {
...
+        }
+    }
+}
+
+void BgJyaMegami_Init(Actor* thisx, GlobalContext* globalContext) {
+    BgJyaMegami* this = THIS;
+
+    BgJyaMegami_DynaPolySetup(this, globalContext, &D_06005C4C, 0);
+    BgJyaMegami_ColliderSetup(this, globalContext);
+    if (Flags_GetSwitch(globalContext, this->dyna.actor.params & 0x3F) != 0) {
```
fig02, line 179: can remove != 0 for the flag check

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -21,31 +37,325 @@ const ActorInit Bg_Jya_Megami_InitVars = {
...
+    }
+    this->explosionTimer = 0;
+}
+
+void BgJyaMegami_Explosion(BgJyaMegami* this, GlobalContext* globalContext) {
+    BgJyaMegami_Piece* temp;
+    u32 i;
+    Vec3f sp8C;
+    BgJyaMegami_PieceInit* temp2;
+    u32 pad;
```
fig02, line 249: not a huge deal but we recently changed our pads to be consistently s32 in the whole codebase

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
fig02: id say probably not for now, but curious what others think

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -21,31 +37,325 @@ const ActorInit Bg_Jya_Megami_InitVars = {
...
+    u32 temp;
+
+    temp = 0;
+    DynaPolyInfo_SetActorMove(&this->dyna, flags);
+    DynaPolyInfo_Alloc(collision, &temp);
+    this->dyna.dynaPolyId =
+        DynaPolyInfo_RegisterActor(globalContext, &globalContext->colCtx.dyna, &this->dyna.actor, temp);
+}
+
+void BgJyaMegami_ColliderSetup(BgJyaMegami* this, GlobalContext* globalContext) {
```
fig02, line 136: the naming scheme we use now is SetupDynapoly and the same can be used for collider as well

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -21,31 +37,325 @@ const ActorInit Bg_Jya_Megami_InitVars = {
...
+    Collider_InitJntSph(globalContext, &this->collider);
+    Collider_SetJntSph(globalContext, &this->collider, &this->dyna.actor, &sJntSphInit, &this->colliderItem);
+}
+
+void func_8089A1DC(GlobalContext* globalContext, Vec3f* burstOrigin, Vec3f* burstDepthX, s32 num, s32 maskIndex) {
+    Vec3f spB4;
+    s32 i;
+
+    for (i = 0; i < num; ++i) {
+        s32 temp_s1 = ((s16)(Math_Rand_ZeroOne() * 8.0f)) & sBitMasks[maskIndex];
```
ethteck, line 148: Not a huge deal, but temps used as an array index can be named `idx` which will distinguish them from other temps like the one right below. Also, the one called `temp` seems to be the named `u0` in the function it's being passed to, so I'd rename it to that.

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -21,31 +37,325 @@ const ActorInit Bg_Jya_Megami_InitVars = {
...
+    { { 34.04f, -61.72f, -37.04f }, 1.5f, 0x0258, 0x0258, 28 },
+    { { 59.48f, -38.399998f, -49.4f }, 1.8f, 0x01F4, 0x0000, 38 },
+    { { -19.04f, -112.24f, -35.120003f }, -1.6f, 0x012C, 0x0320, 22 },
+    { { 12.24f, -99.04f, -31.64f }, 1.8f, 0xFC18, 0xFE70, 24 },
+    { { 49.16f, -81.24f, -55.52f }, 2.4f, 0x02BC, 0x012C, 32 },
+    { { 14.759999f, -125.8f, -44.16f }, 0.2f, 0x0320, 0x0258, 20 },
+};
+
+// sEffectScales
+s16 D_8089B14C[] = {
```
Roman971, line 71: All these name comments shouldn't be merged in, so either rename the symbols, or just leave the comments out for later documentation if you aren't sure about them. (since having comments that will end up wrong or outdated is even worse than having wrong names imo)

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -21,31 +37,325 @@ const ActorInit Bg_Jya_Megami_InitVars = {
...
+s16 D_8089B15C[] = {
+    0x0012, 0x001A, 0x0022, 0x002A, 0x0032, 0x003C, 0x0046, 0x0050,
+};
+
+// sEffectRotSpeeds
+s16 D_8089B16C[] = {
+    0x0030, 0x002A, 0x0024, 0x0020, 0x001C, 0x0018, 0x0014, 0x0010,
+};
+
+s16 sBitMasks[] = {
```
Roman971, line 85: This name is too generic imo, it should probably describe what the masks are for.

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -21,31 +37,325 @@ const ActorInit Bg_Jya_Megami_InitVars = {
...
+    0x0001,
+    0x0003,
+    0x0007,
+};
+
+// sBurstDepthX
+static Vec3f D_8089B184 = {
+    0.0f,
+    0.0f,
+    0.8f,
```
Roman971, line 95: Omit trailing commas in vectors and such so they aren't formatted vertically. (and rerun the formatter afterwards ofc)
```suggestion
    0.8f
```

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -21,31 +37,325 @@ const ActorInit Bg_Jya_Megami_InitVars = {
...
+static UNK_PTR sLeftSideCrumbles[] = {
+    0x06001500, 0x06003500, 0x06003D00, 0x06004500, 0x06000500,
+};
+
+static Gfx* sDLists[] = {
+    0x06009928, 0x06009AC0, 0x06009C80, 0x06009DE8, 0x06009F60, 0x0600A0A8, 0x0600A278,
+    0x0600A418, 0x0600A568, 0x0600A6A0, 0x0600A7E0, 0x0600A978, 0x0600AAC8,
+};
+
+void BgJyaMegami_SetupDynapoly(BgJyaMegami* this, GlobalContext* globalContext, void* collision,
```
Roman971, line 124: "Init" fits better than "Setup" for init sub routines like these imo. Also fyi DynaPolyMoveFlag should not be used as a type, even tho it's currently done in a few places until a future cleanup (since we don't use enums as types in the codebase).
```suggestion
void BgJyaMegami_InitDynapoly(BgJyaMegami* this, GlobalContext* globalContext, void* collision,
```

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -21,31 +37,325 @@ const ActorInit Bg_Jya_Megami_InitVars = {
...
+    u32 temp;
+
+    temp = 0;
+    DynaPolyInfo_SetActorMove(&this->dyna, flags);
+    DynaPolyInfo_Alloc(collision, &temp);
+    this->dyna.dynaPolyId =
+        DynaPolyInfo_RegisterActor(globalContext, &globalContext->colCtx.dyna, &this->dyna.actor, temp);
+}
+
+void BgJyaMegami_SetupCollider(BgJyaMegami* this, GlobalContext* globalContext) {
```
Roman971, line 136:
```suggestion
void BgJyaMegami_InitCollider(BgJyaMegami* this, GlobalContext* globalContext) {
```

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -21,31 +37,325 @@ const ActorInit Bg_Jya_Megami_InitVars = {
...
+
+    Collider_InitJntSph(globalContext, &this->collider);
+    Collider_SetJntSph(globalContext, &this->collider, &this->dyna.actor, &sJntSphInit, &this->colliderItem);
+}
+
+void func_8089A1DC(GlobalContext* globalContext, Vec3f* burstOrigin, Vec3f* burstDepthX, s32 num, s32 maskIndex) {
+    Vec3f spB4;
+    s32 i;
+
+    for (i = 0; i < num; ++i) {
```
Roman971, line 147: Avoid pre-incrementing unless necessary, for consistency with the rest of the codebase. (relevant throughout the file)
```suggestion
    for (i = 0; i < num; i++) {
```

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.h
Roman971: Using `ARRAY_COUNT(this->pieces)` throughout the file is probably better than using a define.

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.h
```diff
@@ -4,11 +4,28 @@
...
+
 struct BgJyaMegami;
 
+typedef void (*BgJyaMegamiActionFunc)(struct BgJyaMegami*, GlobalContext*);
+
+typedef struct BgJyaMegami_Piece {
```
Roman971, line 13: `struct`s should only be named when necessary.
```suggestion
typedef struct {
```

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -9,7 +9,23 @@ void BgJyaMegami_Destroy(Actor* thisx, GlobalContext* globalCtx);
...
+void BgJyaMegami_SetupDetectLight(BgJyaMegami* this);
+void BgJyaMegami_DetectLight(BgJyaMegami* this, GlobalContext* globalContext);
+void BgJyaMegami_SetupExplosion(BgJyaMegami* this);
+void BgJyaMegami_Explosion(BgJyaMegami* this, GlobalContext* globalContext);
+
+extern Gfx D_06005780[];
+extern UNK_TYPE D_06005C4C;
+extern Gfx D_0600B9F8[];
+
+typedef struct BgJyaMegami_PieceInit {
```
Roman971, line 21:
```suggestion
typedef struct {
```

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.h
```diff
@@ -4,11 +4,28 @@
...
+    /* 0x00 */ Vec3f pos;
+    /* 0x0C */ Vec3f vel;
+    /* 0x18 */ s16 rotVelX;
+    /* 0x1A */ s16 rotVelY;
+} BgJyaMegami_Piece; // size = 0x1C
```
Roman971, line 18: Type names usually don't include any underscores.
```suggestion
} BgJyaMegamiPiece; // size = 0x1C
```

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -9,7 +9,23 @@ void BgJyaMegami_Destroy(Actor* thisx, GlobalContext* globalCtx);
...
+    /* 0x00 */ Vec3f unk_00;
+    /* 0x0C */ f32 velX;
+    /* 0x10 */ s16 rotVelX;
+    /* 0x12 */ s16 rotVelY;
+    /* 0x14 */ s16 delay;
+} BgJyaMegami_PieceInit; // size = 0x18
```
Roman971, line 27:
```suggestion
} BgJyaMegamiPieceInit; // size = 0x18
```

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -21,31 +37,325 @@ const ActorInit Bg_Jya_Megami_InitVars = {
...
+void BgJyaMegami_SetupDynapoly(BgJyaMegami* this, GlobalContext* globalContext, void* collision,
+                               DynaPolyMoveFlag flags) {
+    s32 pad;
+    u32 temp;
+
+    temp = 0;
+    DynaPolyInfo_SetActorMove(&this->dyna, flags);
+    DynaPolyInfo_Alloc(collision, &temp);
+    this->dyna.dynaPolyId =
+        DynaPolyInfo_RegisterActor(globalContext, &globalContext->colCtx.dyna, &this->dyna.actor, temp);
```
Roman971, line 133: This is relevant throughout the whole file but we use `globalCtx`, not `globalContext`
```suggestion
        DynaPolyInfo_RegisterActor(globalCtx, &globalCtx->colCtx.dyna, &this->dyna.actor, temp);
```

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -21,31 +37,325 @@ const ActorInit Bg_Jya_Megami_InitVars = {
...
+
+    this->actionFunc = BgJyaMegami_Explosion;
+    for (i = 0; i < BGJYAMEGAMI_NUMPIECES; ++i) {
+        Math_Vec3f_Copy(&this->pieces[i].pos, &this->dyna.actor.posRot.pos);
+        this->pieces[i].vel.x = sPiecesInit[i].velX;
+    }
+    this->explosionTimer = 0;
+}
+
+void BgJyaMegami_Explosion(BgJyaMegami* this, GlobalContext* globalContext) {
```
Roman971, line 244: "Explosion" isn't a proper action name (fyi we call actions with verbs that correspond to the action of the actor itself, so things like "Fly", "Run", "Wait", etc...). "Explode" seems already better but it might not be accurate to what the actor is actually doing, so be mindful about that.
```suggestion
void BgJyaMegami_Explode(BgJyaMegami* this, GlobalContext* globalContext) {
```

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
fig02: im fine with this change but it will need to be made everywhere, cause this pattern is used pretty consistently atm

mzxrules: Dynapoly or DynaPoly?

fig02: its Dynapoly everywhere in the codebase currently, but we want to discuss changing it, its something thats outside the scope of this pr

fig02: nvm its DynaPoly everywhere lmfao

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
```diff
@@ -21,31 +37,325 @@ const ActorInit Bg_Jya_Megami_InitVars = {
...
+
+void func_8089A1DC(GlobalContext* globalContext, Vec3f* burstOrigin, Vec3f* burstDepthX, s32 num, s32 maskIndex) {
+    Vec3f spB4;
+    s32 i;
+
+    for (i = 0; i < num; ++i) {
+        s32 temp_s1 = ((s16)(Math_Rand_ZeroOne() * 8.0f)) & sBitMasks[maskIndex];
+        s16 temp = ((temp_s1 < 5) && (Math_Rand_ZeroOne() < 0.7f)) ? 0x40 : 0x20;
+        func_80029E8C(globalContext, burstOrigin, burstDepthX, burstOrigin, -90, temp, D_8089B16C[temp_s1], 4, 0,
+                      D_8089B14C[temp_s1], 0, 5, D_8089B15C[temp_s1], -1, OBJECT_JYA_OBJ, &D_0600B9F8);
```
mzxrules, line 151: you can fix an incorrect type warning by removing the & on D_0600B9F8

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
gamestabled: Their purpose is pretty unclear to me, I think its best if i just leave it unnamed again for now.

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
gamestabled: so should I leave it as SetupDynaPoly for now? or make the switch (and same for SetupCollider below)

Roman971: As far as I can tell we use setup and init in 2 places respectively, so I think using init here is fine, and we will fix the others eventually.

src/overlays/actors/ovl_Bg_Jya_Megami/z_bg_jya_megami.c
Roman971: Alright, that's fine

Roman971 (approved):

ethteck:
I believe this is missing a header

fig02:
any idea which? the actor header is in the diff at least

PR #297, opened 2020-07-28
Math Sizes

louist103:
Added sizes to most variable types

### DIFF
--- a/include/z64math.h
+++ b/include/z64math.h
@@ -7,69 +7,69 @@
 
 typedef struct {
     f32 x, y;
-} Vec2f; //size = 0x08
+} Vec2f; // size = 0x08
 
 typedef struct {
     f32 x, y, z;
-} Vec3f; //size = 0x0C
+} Vec3f; // size = 0x0C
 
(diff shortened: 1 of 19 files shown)


include/z64math.h
```diff
@@ -51,25 +51,25 @@ typedef struct {
...
-} Cylinder16;
+} Cylinder16; //size = 0x0C
 
 typedef struct {
     f32 radius;
     f32 height;
     f32 yShift;
     Vec3f pos;
-} Cylinderf;
+} Cylinderf; //size = 0x10
```
Roman971, line 61:
```suggestion
} Cylinderf; //size = 0x18
```

include/z64math.h
```diff
@@ -7,39 +7,39 @@
 
 typedef struct {
     f32 x, y;
-} Vec2f;
+} Vec2f; //size = 0x08
```
Roman971, line 10: This suggestion applies to all size comments:
```suggestion
} Vec2f; // size = 0x08
```

PR #308, opened 2020-08-07
audio_playback OK

zelda2774:
and rename some of the other audio files to match sm64 -- seems good to reduce the number of `code_...` files in the repo.


src/code/audio_playback.c
```diff
@@ -149,11 +151,152 @@ void Audio_NoteDisable(Note* note) {
...
+    u8 bookOffset;
+    f32 scale;
+    s32 i;
+
+    for (i = 0; i < gAudioContext.gMaxSimultaneousNotes; i++) {
+        note = &gAudioContext.gNotes[i];
+        noteSubEu2 = &gAudioContext.gNoteSubsEu[gAudioContext.gNoteSubEuOffset + i];
+        playbackState = &note->playbackState;
+        if (playbackState->parentLayer != NO_LAYER) {
+            if ((u32) playbackState->parentLayer < 0x7fffffffU) {
```
fig02, line 175: are the u32 cast and U at the end of the number needed? Id also make all the F's uppercase in the number.

Other than that, make sure to run format.sh (after installing clang-format and clang-tidy).
Rest looks good to me

fig02: also our style generally is to have nothing in the args if it takes nothing, unless void is needed to match of course

zelda2774: The u32 cast is necessary -- it's converting a pointer to an integer. The U at the end of the number silences a signed->unsigned conversion warning and clarifies that an unsigned comparison is happening.

I'm surprised to hear that `()` is preferred, seems like it could hide errors (and will confuse mips_to_c). Apparently there's 99 `()` and 45 `(void)` in the codebase.

fig02: I didnt realize there was that much of a discrepancy, I guess we should agree on something soon and keep it consistent

zelda2774:
Closing in favor of #313 which includes all these changes

PR #316, opened 2020-08-10
Various repo improvements

ethteck:
OSX Compatibility:
* dSYM dirs added to .gitignore
* clean target flag order changed
* ZAP2 OSX fixes

Tool changes:
* extract_assets.py and extract_baserom.py +x
* permuter_settings.toml file that decomp_permuter will use to better work with macros

Also:
* ~~Explicitly list files requiring ido 5 in build/src/libultra_boot_O1/O2 dirs~~
* Error if the machine doesn't have mips-linux-gnu installed, remove mention of mips64-elf


Makefile
```diff
@@ -125,8 +125,15 @@ build/src/libultra_code/osStopTimer.o: OPTFLAGS := -O1
 build/src/libultra_code/llcvt.o: OPTFLAGS := -O1
 build/src/libultra_code/llcvt.o: MIPS_VERSION := -mips3 -32
 
-build/src/libultra_boot_O1/%.o: CC := $(CC_OLD)
-build/src/libultra_boot_O2/%.o: CC := $(CC_OLD)
```
Roman971, line 129: What's the reason behind this change?
The whole point of having subfolders right now is to avoid specifying files separately in the makefile. That may change eventually with a full restructure, but I don't see the point of only changing the makefile on its own.

ethteck: I have a theory that ido 5 is either not necessary or is at least not necessary for many files we think it is necessary for. For example, only one file currently needs it in the /O1 directory, despite it being used on every file in there currently.

I'm okay with removing this change since I see your point, but I thought it was a nice record of the investigation I did into this situation. Could I keep a comment saying which files need IDO 5?

Roman971: I see. I think it would be better to keep that kind of info in the discord or something like that

ethteck: Okay, that's fair

PR #319, opened 2020-08-13
Decompile PreRender.c

Random06457:
I figured I'd make a PR since it's been a while I haven't made any progress on this file.
2 functions left to decompile, the rest is matching.


src/code/PreRender.c
```diff
@@ -1,42 +1,615 @@
...
-#pragma GLOBAL_ASM("asm/non_matchings/code/PreRender/func_800C0F08.s")
+void PreRender_Destroy(PreRenderContext* this)
+{
+    ListAlloc_FreeAll(&this->alloc);
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/PreRender/func_800C0F28.s")
+void func_800C0F28(PreRenderContext* this, Gfx** gfxp, void* buf, void* bufSave)
+{
+    Gfx* gfx;
```
fig02, line 41: do we want to go with the normal convention of calling gfx a dList and dListHead or something? not sure if this context is different and warrants a different name

src/code/PreRender.c
```diff
@@ -1,42 +1,615 @@
...
+    *gfxp = gfx;
+}
+
+void func_800C1E9C(PreRenderContext* this, Gfx** gfxp)
+{
+    LogUtils_CheckNullPointer("this->zbuf_save", this->zbufSave, "../PreRender.c", 481);
+    LogUtils_CheckNullPointer("this->zbuf", this->zbuf, "../PreRender.c", 482);
+
+    if (!this->zbufSave || !this->zbuf)
+        return;
```
fig02, line 245: here and below there are some braces omitted on single line ifs which we dont do

src/code/PreRender.c
```diff
@@ -1,42 +1,615 @@
...
+    func_800C1B24(this, gfxp, this->fbuf, this->cvgSave);
+}
+
+void func_800C2118(PreRenderContext* this, Gfx** gfxp)
+{
+    func_800C0F28(this, gfxp, this->zbufSave, this->zbuf);
+}
+
+void func_800C213C(PreRenderContext* this, Gfx** gfxp)
+{
```
fig02, line 294: brace stuff, run formatter

src/code/PreRender.c
```diff
@@ -1,42 +1,615 @@
...
+        }
+        #endif
```
fig02, line 341: ive never seen us do a non matching inside a func like this. I guess it makes sense if youve isolated the problem, but seems weird

src/code/PreRender.c
```diff
@@ -1,42 +1,615 @@
...
+    func_800C0F28(this, gfxp, this->fbufSave, this->fbuf);
+}
+
+void func_800C24E0(PreRenderContext* this, Gfx** gfxp)
+{
+    func_800C1258(this, gfxp);
+}
+
+#if 0
+void func_800C2500(PreRenderContext *this, s32 x, s32 y) {
```
fig02, line 364: we've got ifdef NON_EQUIVALENT now

src/code/PreRender.c
Random06457: I've seen both being used. Not sure which name I should use, also maybe the OG name is good enough?

fig02: I prefer dList if it actually is a display list, gfx doesnt tell me much. but thats just me

Random06457: Gfx can only be a display list anyway.

src/code/PreRender.c
```diff
@@ -1,42 +1,638 @@
...
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/PreRender/func_800C0F08.s")
+void PreRender_Destroy(PreRenderContext* this) {
+    ListAlloc_FreeAll(&this->alloc);
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/PreRender/func_800C0F28.s")
+void func_800C0F28(PreRenderContext* this, Gfx** dListp, void* buf, void* bufSave) {
+    Gfx* dList;
```
fig02, line 36: the pattern we use in most places (at least afaik) is to have a dList variable for the whole display list and then dListHead for the front of it that you do ++ on
So dListp would be dList and Gfx* dList; would be dListHead.

Pretty sure this is the scheme we want to go with anyway, curious on others thoughts

src/code/PreRender.c
fig02: @Roman971 mentioned in discord we should probably not have non matchings in the middle of a function and I agree.
If you want to comment exactly what the issue is and what that fix is, feel free to do that tho

fig02: @Random06457 not sure if the above was missed, but i think this is all thats holding this up. id just put the whole function in a non matching instead of this part

fig02 (approved):

src/code/PreRender.c
```diff
@@ -1,42 +1,632 @@
...
+                        G_TD_CLAMP | G_TP_NONE | G_CYC_1CYCLE | G_PM_NPRIMITIVE,
+                    G_AC_NONE | G_ZS_PRIM | G_RM_VISCVG | G_RM_VISCVG2);
+    gDPSetScissor(dListHead++, G_SC_NON_INTERLACE, 0, 0, this->width, this->height);
+    gDPFillRectangle(dListHead++, 0, 0, this->width, this->height);
+    gDPPipeSync(dListHead++);
+
+    *dList = dListHead;
+}
+
+void func_800C20B4(PreRenderContext* this, Gfx** dList) {
```
Roman971, line 293: It looks like this function and other "draw" entrypoints of this system are called with a pointer to the head of a display buffer, not a dlist. Based on our previous discussion, we would call that `dispHeadp` or `dispHeadP`.

However, since this system doesn't know that fact, and it could very well work with a pointer the the head of a dlist, I think we should just have it as a generic `gfxp`, which becomes `gfx` when dereferencing it later on in other functions. Based on various strings in this file, they even had them as `glistpp`/`glistp`, which likely means a generic "pointer to gfx list" (or pointer to pointer for `pp`), and confirms the generic intent here.

src/code/PreRender.c
```diff
@@ -1,42 +1,632 @@
...
+    func_800C1AE8(this, dList, this->fbuf, this->fbufSave);
+}
+
+void func_800C1FA4(PreRenderContext* this, Gfx** dList) {
+    Gfx* dListHead;
+
+    dListHead = *dList;
+
+    gDPPipeSync(dListHead++);
+    gDPSetBlendColor(dListHead++, 0xFF, 0xFF, 0xFF, 0x08);
```
Roman971, line 280: Colors need to be updated to decimal throughout the file

src/code/PreRender.c
```diff
@@ -1,42 +1,632 @@
...
+    x = this->height;
+    add = 0x1000 / (this->width * 2);
+    uls = 0;
+    x2 = 0;
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/PreRender/func_800C1B24.s")
+    while (x > 0) {
+        lrs = this->width - 1;
+        if (x < add)
+            add = x;
```
Roman971, line 66: clang-tidy should be run on this file to fix style issues like the lack of braces here (`format.sh` runs it along with other things)

Roman971 (approved):

PR #321, opened 2020-08-14
Fix Bug with colliderinit.py

louist103:
MZX fixed a bug with colliderinit.py having the wrong size for ColliderTrisItemInit


fig02:
merging this as soon as jenkins is done, without review. its a one line change to a script that has already been confirmed to work

PR #336, opened 2020-08-21
z_en_js OK

fullgrowngaming:
Magic Carpet Salesman decomped and OK


src/overlays/actors/ovl_En_Js/z_en_js.c
```diff
@@ -21,35 +28,202 @@ const ActorInit En_Js_InitVars = {
...
-
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Js/func_80A895F4.s")
-
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Js/EnJs_Draw.s")
+static ColliderCylinderInit sCylinderInit = {
+    { COLTYPE_UNK10, 0x00, 0x11, 0x39, 0x10, COLSHAPE_CYLINDER },
+    { 0x00, { 0x00000000, 0x00, 0x00 }, { 0xFFCFFFFF, 0x00, 0x00 }, 0x00, 0x01, 0x01 },
+    { 30, 40, 0, { 0, 0, 0 } },
+};
+s32 D_80A896DC[] = { 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000 };
```
fig02, line 37: this is being used in a matrix mult vec3f call, so its just a vec3f with 0.0fs. the rest of the zeroes are padding.
imo it also makes more sense for this to be defined in the function with static instead of at the top of the file here

src/overlays/actors/ovl_En_Js/z_en_js.c
```diff
@@ -21,35 +28,202 @@ const ActorInit En_Js_InitVars = {
...
+
+extern SkeletonHeader D_06005EA0;
+extern AnimationHeader D_0600045C;
+extern AnimationHeader D_0600018C;
+
+void En_Js_SetupAction(EnJs* this, EnJsActionFunc actionfunc) {
+    this->actionfunc = actionfunc;
```
fig02, line 44: actionFunc for both the variable and the argument

src/overlays/actors/ovl_En_Js/z_en_js.c
```diff
@@ -21,35 +28,202 @@ const ActorInit En_Js_InitVars = {
...
+    En_Js_SetupAction(this, func_80A89304);
+    this->unk_284 = 0;
+    this->actor.gravity = -1.0f;
+    Actor_SpawnAttached(&globalCtx->actorCtx, &this->actor, globalCtx, ACTOR_EN_JSJUTAN, this->actor.posRot.pos.x,
+                        this->actor.posRot.pos.y, this->actor.posRot.pos.z, 0, 0, 0, 0);
+}
+
+void EnJs_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+    EnJs* this = THIS;
+    Collider_DestroyCylinder(globalCtx, &this->collider);
```
fig02, line 68: put a line break between the stack variables and the code

src/overlays/actors/ovl_En_Js/z_en_js.c
```diff
@@ -21,35 +28,202 @@ const ActorInit En_Js_InitVars = {
...
+        return 1;
+    } else {
+        this->actor.textId = arg2;
+        temp = this->actor.yawTowardsLink - this->actor.shape.rot.y;
+
+        if (temp >= 0) {
+            temp_2 = temp;
+        } else {
+            temp_2 = -temp;
+        }
```
fig02, line 85: this should be the ABS macro

src/overlays/actors/ovl_En_Js/z_en_js.c
```diff
@@ -21,35 +28,202 @@ const ActorInit En_Js_InitVars = {
...
+
+u8 func_80A88F64(EnJs* this, GlobalContext* globalCtx, u16 arg2) {
+    s16 temp;
+    s32 temp_2;
+
+    if (func_8002F194(&this->actor, globalCtx) != 0) {
+        return 1;
+    } else {
+        this->actor.textId = arg2;
+        temp = this->actor.yawTowardsLink - this->actor.shape.rot.y;
```
fig02, line 79: this temp can be called yawDiff. and you may be able to get rid of the extra temp with the abs macro

src/overlays/actors/ovl_En_Js/z_en_js.c
```diff
@@ -21,35 +28,202 @@ const ActorInit En_Js_InitVars = {
...
+}
+
+u8 func_80A88F64(EnJs* this, GlobalContext* globalCtx, u16 arg2) {
+    s16 temp;
+    s32 temp_2;
+
+    if (func_8002F194(&this->actor, globalCtx) != 0) {
+        return 1;
+    } else {
+        this->actor.textId = arg2;
```
fig02, line 78: you could name the arg textId

src/overlays/actors/ovl_En_Js/z_en_js.c
```diff
@@ -21,35 +28,202 @@ const ActorInit En_Js_InitVars = {
...
+        temp = this->actor.yawTowardsLink - this->actor.shape.rot.y;
+
+        if (temp >= 0) {
+            temp_2 = temp;
+        } else {
+            temp_2 = -temp;
+        }
+
+        if (temp_2 < 0x1801 && this->actor.xzDistFromLink < 100.0f) {
```
fig02, line 87: <= 1800 should work

src/overlays/actors/ovl_En_Js/z_en_js.c
```diff
@@ -21,35 +28,202 @@ const ActorInit En_Js_InitVars = {
...
+}
+
+void func_80A891C4(EnJs* this, GlobalContext* globalCtx) {
+    u8 temp_v0;
+
+    if ((func_8010BDBC(&globalCtx->msgCtx) == 4) && (func_80106BC8(globalCtx) != 0)) {
+        switch (globalCtx->msgCtx.choiceIndex) {
+            case 0: // yes
+                if (gSaveContext.rupees < 200) {
+                    // yes, but not enough rupees
```
fig02, line 140: this one is doesnt need a comment i think.
the comment on the cases are fine tho

src/overlays/actors/ovl_En_Js/z_en_js.c
```diff
@@ -21,35 +28,202 @@ const ActorInit En_Js_InitVars = {
...
+    if (func_8002F410(&this->actor, globalCtx) != 0) {
+        this->actor.attachedA = NULL;
+        En_Js_SetupAction(this, func_80A8910C);
+    } else {
+        func_8002F434(&this->actor, globalCtx, GI_BOMBCHUS_10, 10000.0f, 50.0f);
+    }
+}
+
+void func_80A891C4(EnJs* this, GlobalContext* globalCtx) {
+    u8 temp_v0;
```
fig02, line 134: leftover from mips2c? if its not needed remove, otherwise make it s32 pad, or s8 pad if it has to be a byte

src/overlays/actors/ovl_En_Js/z_en_js.c
```diff
@@ -21,35 +28,202 @@ const ActorInit En_Js_InitVars = {
...
+}
+
+void func_80A89294(EnJs* this) {
+    En_Js_SetupAction(this, func_80A891C4);
+    SkelAnime_ChangeAnim(&this->skelAnime, &D_0600018C, 1.0f, 0.0f, SkelAnime_GetFrameCount(&D_0600018C.genericHeader),
+                         2, -4.0f);
+}
+
+void func_80A89304(EnJs* this, GlobalContext* globalCtx) {
+    // 0x6077 - text for initial offer to buy something for 200 rupees
```
fig02, line 162: we're trying to stay away from comments describing text. eventually the whole text table will be in here and youll be able to look them up easily

src/overlays/actors/ovl_En_Js/z_en_js.c
```diff
@@ -21,35 +28,202 @@ const ActorInit En_Js_InitVars = {
...
+    return 0;
+}
+
+void EnJs_PostLimbDraw(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3s* rot, Actor* thisx) {
+    if (limbIndex == 12) {
+        Matrix_MultVec3f(D_80A896DC, &thisx->posRot2.pos);
+    }
+}
+void EnJs_Draw(Actor* thisx, GlobalContext* globalCtx) {
+    EnJs* this = THIS;
```
fig02, line 225: line break

fig02:
should clarify this doesn't mean everything that checks against !=0 should be treated that way, but in this file im fairly sure that every signle one is a boolean check. double check tho

src/overlays/actors/ovl_En_Js/z_en_js.c
fig02: you should be able to use the ABS macro directly in the conditional, but maybe not

PR #337, opened 2020-08-21
Ido Recomp

ethteck:
I've added -O2 ido 5/7 binaries to the repo. The Makefile will use the ido recompiled binaries if the environment `IDO_RECOMP` is set. 

Jenkins will use qemu-irix on master branch builds and the recomp for others.


Random06457:
Note that `/usr/lib/err.english.cc` is not being redirected on this version and thus error messages look like this : `ERROR MESSAGE 8ERROR MESSAGE 103`.

fig02:
Wait so error messages dont work at all?

Random06457:
`cfe` erros messages more specifically but I have a fix [here](https://github.com/Random06457/ido-static-recomp/commit/7798729ed20f9d025b2820c868f66f84fbdb423f). (I can send the compiled binaries in Dicord)

fig02:
yea coordinate with @ethteck to get this one in the pr instead

PR #343, opened 2020-08-23
z_lights.c OK

fig02:
- 7 functions matched
- Major structural and naming changes. I am very open to naming suggestions.


fig02:
update: krim matched the last non matching FeelsGoodMan

src/code/z_lights.c
```diff
@@ -1,190 +1,381 @@
...
+        light->l.dir[0] = params->dir.x;
+        light->l.dir[1] = params->dir.y;
+        light->l.dir[2] = params->dir.z;
     }
 }
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_lights/func_8007A474.s")
+void Lights_Update(Lights* lights, LightNode* head, Vec3f* vec) {
+    LightsUpdateFunc updateFuncs[] = { Lights_UpdatePoint, Lights_UpdateDirectional, Lights_UpdatePoint };
+    LightInfo* info;
```
fig02, line 145: After looking into things more, it appears that Lights_Update and Lights_Draw will do so for every light currently in the node buffer, even if those lights werent added by the system currently calling this function.
To put that another way, Kankyo and Actor lights will still get drawn if you nop their respective functions, but leave the last one in the chain in EfectsSs intact..

Should this instead be UpdateAll and DrawAll, or should that be left to an implementation detail that I can mention in a comment?

Because of this fact, im very confused why the buffer can have up to 32 nodes. It seems like the hard limitation is 7 lights in a LightsN no matter what..

fig02: I guess to be completely accurate, it would be Lights_UpdateAll which updates every light in the buffer and adds all of them to a LightsN group, and then Lights_Draw draws all of the lights in the LightsN (which again, has a hard limit of 7 lights at a given time for the whole game).

imo seeing Lights_UpdateAll followed by Lights_Draw would be pretty confusing. So Im not sure which naming scheme is better to use for the sake of clarity outside of the lighting system. 

No matter what Ill write comments explaining all of these details. But want help with the naming scheme that will be used externally

src/code/z_lights.c
```diff
@@ -1,190 +1,398 @@
...
+    gfxCtx = gfxCtxArg;
+    Graph_OpenDisps(dispRefs, gfxCtx, "../z_lights.c", 339);
+
+    gSPNumLights(gfxCtx->polyOpa.p++, lights->numLights);
+    gSPNumLights(gfxCtx->polyXlu.p++, lights->numLights);
+
+    i = 0;
+    light = &lights->l.l[0];
+
+    while (i < lights->numLights) {
```
ethteck, line 74: Did you try a for loop here? It might work better if possible

include/functions.h
```diff
@@ -975,29 +975,29 @@ void Health_InitData(GlobalContext* globalCtx);
...
+LightNode* Lights_FindBufSlot();
+void Lights_InitContext(GlobalContext* globalCtx, LightContext* lightCtx);
+void func_8007A698(LightContext* lightCtx, u8 arg1, u8 arg2, u8 arg3, s16 arg4, s16 arg5);
+void Lights_SetAmbientColor(LightContext* lightCtx, u8 red, u8 green, u8 blue);
+Lights* Lights_New(LightContext* lightCtx, GraphicsContext* gfxCtx);
+void Lights_ClearHead(GlobalContext* globalCtx, LightContext* lightCtx);
+LightNode* Lights_Insert(GlobalContext* globalCtx, LightContext* lightCtx, LightInfo* info);
+void Lights_Remove(GlobalContext* globalCtx, LightContext* lightCtx, LightNode* light);
+Lights* Lights_AllocAndDraw(GraphicsContext* gfxCtx, u8 ambientR, u8 ambientG, u8 ambientB, u8 numLights, u8 r, u8 g, u8 b,
+                      s8 x, s8 y, s8 z);
```
ethteck, line 997: I think the rest of the parameters should line up with the `GraphicsContext*...` on the previous line

src/code/z_lights.c
fig02: this loop was very picky and this is the only configuration that worked. roman and krim can both attest FeelsOkayMan

ethteck: gorcha, kk :)

fig02:
updated disps macros

src/code/z_lights.c
```diff
@@ -1,190 +1,392 @@
...
+        lights->l.l[i].l.dir[1] = y;
+        lights->l.l[i].l.dir[2] = z;
+    }
+
+    Lights_Draw(lights, gfxCtx);
+
+    return lights;
+}
+
+Lights* Lights_AllocAndInit(GraphicsContext* gfxCtx, u8 r, u8 g, u8 b) {
```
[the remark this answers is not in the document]
fig02: I defintiley see your reasoning why that makes sense here, but the wrapper is what gets used externally and is why I feel it should have the simple and effective name of New.
NewAmbient to me sounds like its only making a new ambient light, when its a new ambient + up to 7 other lights. I may be on board with a different name, but I really like Lights_New for this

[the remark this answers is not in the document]
fig02: maybe we can go for something like Setup for AllocAndInit, as it allocates the Lightsn and sets the ambient color?
I dont think the name of Lights_New has to indicate that its setting the ambient color, it feels like more of an implementation detail to me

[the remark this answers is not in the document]
Roman971: How about _NewWithGlobalAmbient instead of _NewAmbient then? Since there aren't many arguments it can afford to have an explicit name imo. And I think it's important to not hide this part, since we wouldn't expect a completely "new" lights to already have the ambient set.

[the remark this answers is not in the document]
fig02: so I decided to actually go ahead and make the system name changes for LightContext related functions.
It got me thinking, LightContext_NewLights could be the name for this function, and the fact that it sets the ambient color as that provided by light context can be an implementation detail.
While Lights_New is not a part of the LightContext system and can just be named that

What you think?

fig02:
@Roman971 made a bunch of changes based on our discussion. Changed quite a few function names so give the whole file another read-through when you have the time

src/code/z_lights.c
```diff
@@ -223,22 +216,31 @@ void func_8007A698(LightContext* lightCtx, u8 arg1, u8 arg2, u8 arg3, s16 numLig
...
 }
 
-LightNode* Lights_Insert(GlobalContext* globalCtx, LightContext* lightCtx, LightInfo* info) {
+/*
+ * Insert a new light node into the list pointed to by LightContext
+ *
+ * Note: Due to the limited number of slots in a Lights group, inserting too many nodes in the
+ * list may result in older entries not being bound to a Light when calling Lights_BindAll
+ */
+LightNode* LightContext_InsertNewNode(GlobalContext* globalCtx, LightContext* lightCtx, LightInfo* info) {
```
Roman971, line 243: Imo this function's purpose is to add a new "light" in the list (ie the light described by `LightInfo`). The fact that it inserts a `LightNode` that points to this `LightInfo` is just a mean to do that, not the purpose. In the end, a `LightNode` is just the structure used to create what is widely known as a **doubly linked list** of `LightInfo`.
So I suggest naming this function `LightContext_InsertLight` or something along those lines, and updating its description accordingly.

`LightContext_RemoveNode` would also have to renamed accordingly ofc (`LightContext_RemoveLight` for instance).
And it would probably be good to update the description for `Lights_BindAll` to make it clear it binds lights and not the nodes themselves, which are just the linked list implementation.

Roman971 (commented):
Looks good to me aside from this slight distinction that I think is really important

src/code/z_lights.c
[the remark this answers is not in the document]
Roman971: I think it's fine. The ambient color might be more than an implementation detail, but it's built into the system anyway so it's probably supposed to be the "normal" way to create lights.

src/code/z_lights.c
fig02: I felt it was more appropriate to make it clear that this is inserting a node into the light context list. Especially with the system name change, LightContext_InsertNode makes it clear to me that its inserting an item to the list that exists in light context

I do like LightContext_InsertLight more than just plain old "insert", cause "Light Context Insert" didnt really make sense at all. Ill probably go with this name change but wanted to explain my thought process

Roman971: Right, I understand why the name had to be changed. But my point is it's not inserting a light node (`LightNode`) in the list, it's inserting a `LightInfo` (which describes a light).
The node is just an intermediate structure used to link items in the list, not the item itself. You can think of it like a heap node VS the actual memory block it corresponds to.

fig02: ah yeah that makes alot more sense when put that way.
Im a little worried about the separation between an "abstract" light (represented by info) and a real light from a LightsN group not being clear with a name like this, but at least externally it makes sense to refer to it as a plain old light. And the LightContext system name makes it clear its the light context list

fig02:
made final name changes

Roman971 (approved):

PR #354, opened 2020-08-27
fix m2ctx incompatibility on MacOS

notyouraveragehooman:
A version of `cpp` that is installed with the standard command line tools for MacOS is somehow unable to parse comments that begin with `//`. Switching the preprocessor to `gcc` alleviates this issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Cache files
 __pycache__/
 .pyc
+.DS_Store
 
 # Text editor remnants
 .vscode/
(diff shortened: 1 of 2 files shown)


ethteck:
Thanks for submitting this. I like this change, and I think we should do the same thing to the main Makefile where we run cpp on the spec and undefined_syms and such, as this process has issues on my mac as well. Do you mind fixing it there too?

ethteck:
I'm going to go ahead and approve this, because I plan on making a mac compatibility PR soon which includes ido recomp binaries for osx. I'll make the changes I mentioned above there too.

PR #356, opened 2020-08-28
ovl_En_Box OK, documented

Dragorn421:
Long time user first time contributor, don't hesitate to make lots of comments :)

This is the chest actor, 100% OK

A bunch of things require more decompilation work in other areas to be understood, like why does this access message context or how does a struct member even change when the chest actor apparently isn't touching it.


fig02:
nice

in the spec youll want to remove the data and bss lines. and to change to the auto generated relocation data you want to replace the reloc line with 
include "build/src/overlays/actors/ovl_En_Box/ovl_En_Box_reloc.o"

after that remove all function and data .s files that are no longer in use. you can pass -d to tools/find_unused_asm.sh to do this

src/overlays/actors/ovl_En_Box/z_en_box.h
```diff
@@ -4,11 +4,84 @@
...
+typedef void (*EnBoxActionFunc)(struct EnBox*, GlobalContext*);
+
+typedef enum {
+    /*
+     set on init unless treasure flag is set
+     if clear, chest moves (Actor_MoveForward) (falls, likely)
+     ends up cleared from SWITCH_FLAG_FALL types when switch flag is set
+    */
+    ENBOX_STATE_IMMOBILE = 1 << 0,
+    /*
```
fig02, line 21: imo for movement flags it would make more sense for the prefix to be ENBOX_MOVE_ instead of "state"

src/overlays/actors/ovl_En_Box/z_en_box.h
```diff
@@ -4,11 +4,84 @@
...
 
+#define ENBOX_TREASURE_FLAG_UNK_MIN 20
+#define ENBOX_TREASURE_FLAG_UNK_MAX 32
+
 struct EnBox;
 
+typedef void (*EnBoxActionFunc)(struct EnBox*, GlobalContext*);
+
+typedef enum {
+    /*
```
fig02, line 15: is this and the 1FB enum used outside the actor? We only want to keep public things that we are exposing to other files in the header. So if not, they can go in the C file just under the THIS macro definition.
Im thinking type is the only enum that will end up being used externally for params.

src/overlays/actors/ovl_En_Box/z_en_box.h
```diff
@@ -4,11 +4,84 @@
...
+     only used for SWITCH_FLAG_FALL types
+     ends up "blinking" (set/clear every frame) once switch flag is set,
+        if some collision-related condition (?) is met
+     only used for signum of z rotation
+    */
+    ENBOX_STATE_FALL_ANGLE_SIDE = 1 << 2,
+    /*
+     when set, is cleared next EnBox_Update call and clip to the floor
+    */
+    ENBOX_STATE_FLAG_STICK_TO_GROUND_NEXT_FRAME = 1 << 4
```
fig02, line 37: maybe something a bit shorter like ENBOX_MOVE_GROUND or something. The specifics of it happening on the next frame can be explained in the comment

src/overlays/actors/ovl_En_Box/z_en_box.h
```diff
@@ -4,11 +4,84 @@
...
 typedef struct EnBox {
-    /* 0x0000 */ Actor actor;
-    /* 0x014C */ char unk_14C[0xB0];
+    /* 0x0000 */ DynaPolyActor dyna;
+    /* 0x0164 */ SkelAnime skelanime;
+    /*
+     initialized in init for types 3 and 8
+     linked to different animation delays
+    */
+    /* 0x01A8 */ s32 unk1A8;
```
fig02, line 71: to get rid of this blocky comment in the header, I think it can be shortened a considerable amount and put on the side with a // comment.
"// related to animation delays for types 3 and 8" or something like that

src/overlays/actors/ovl_En_Box/z_en_box.h
```diff
@@ -4,11 +4,84 @@
...
+    /* 0x01AC */ s32 unk_1AC;
+    /* 0x01B0 */ f32 unk1B0; // 0-1, rotation-related, apparently unused (in z_en_box.c at least)
+    /* 0x01B4 */ EnBoxActionFunc actionFunc;
+    /* 0x01B8 */ Vec3s limbDrawTable[5];
+    /* 0x01D6 */ Vec3s transitionDrawTable[5];
+    /* 0x01F4 */ s16 unk1F4;       // probably a frame count? used but never initialized!
+    /* 0x01F6 */ u8 movementFlags; // movement behavior, see EnBox_MovementFlags for values
```
fig02, line 78: variable name is pretty self explanatory i think, especially because the enum is called movement flags

src/overlays/actors/ovl_En_Box/z_en_box.h
```diff
@@ -4,11 +4,84 @@
...
+    /* 0x01F7 */ u8 alpha;         // transparency (255 opaque)
```
fig02, line 79: this is the case for all alphas, i dont think it needs a comment

src/overlays/actors/ovl_En_Box/z_en_box.h
```diff
@@ -4,11 +4,84 @@
...
+    /* 0x01F8 */ u8 switchFlag;    // switch flag to make the chest appear (for SWITCH_FLAG types)
+    /* 0x01F9 */ u8 type;          // size/behavior, see EnBox_Type for values
```
fig02, line 81: both of these are also named after their enums and dont need a comment imo

src/overlays/actors/ovl_En_Box/z_en_box.h
```diff
@@ -4,11 +4,84 @@
...
+    /* 0x01FA */ u8 unk1FA; // amount of frames some effect has been played (ice trap?)
+                            // (set to 0 in init, 100 if treasure flag is set), 0-100
```
fig02, line 83: this is indeed for the ice trap effect. Id call the enum iceEffTimer, and with a name like that it doesnt really need either of the comments. The values it gets init to can be easily found in the code

src/overlays/actors/ovl_En_Box/z_en_box.h
```diff
@@ -4,11 +4,84 @@
...
+    /* 0x01FB */ u8 unk1FB_state; // see EnBox_State1FB
```
fig02, line 84: we generally try to avoid calling things just a generic "state", but i havent read much of the code to offer a better alternative. I would probably just call it unk_1FB for now and name the header the same. 
The names will be synced up, so i also feel the comment is unnecessary

src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -21,47 +71,608 @@ const ActorInit En_Box_InitVars = {
...
+    accel->x = 0.0f;
+    accel->y = 0.0f;
+    accel->z = 0.0f;
+}
+
+/**
+ * Spawns effects randomly around the chest
+ * Used when the chest hits the ground after falling (FALL types)
+ */
+void EnBox_SpawnLandingEffects(EnBox* this, GlobalContext* globalCtx) {
```
fig02, line 233: this is for dust specifically, so you can change func names and docs accordingly.
might be worth naming the random movement function after effects or dust specifically

src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -21,47 +71,608 @@ const ActorInit En_Box_InitVars = {
...
+                                       this->dyna.actor.shape.rot.y, this->dyna.actor.shape.rot.z, 0xFFFF);
+                    func_800F5C64(0x92B);
+            }
+        }
+        osSyncPrintf("Actor_Environment_Tbox_On() %d\n", this->dyna.actor.params & 0x1F);
+        Flags_SetTreasure(globalCtx, this->dyna.actor.params & 0x1F);
+    } else {
+        player = PLAYER;
+        func_8002DBD0(&this->dyna.actor, &sp4C, &player->actor.posRot.pos);
+        if (-50.0f < sp4C.z && sp4C.z < 0.0f && fabsf(sp4C.y) < 10.0f && fabsf(sp4C.x) < 20.0f &&
```
fig02, line 443: const on the left is something we dont have a rule on, but yeah if it were me id flip it.

and the classic parens argument of course, this is so much harder for me to read personally rofl. but yeah FeelsOkayMan

src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -21,47 +71,608 @@ const ActorInit En_Box_InitVars = {
...
+void func_809C9EF8(EnBox* this, GlobalContext* globalCtx) {
+    Vec3f sp4c_3f;
+    Vec3f sp40_3f;
+    Vec3f sp34_3f;
+    f32 f0;
+
+    sp40_3f = D_809CA814;
+    sp34_3f = D_809CA820;
+    this->iceEffTimer++;
+    func_8002F974(&this->dyna.actor, 0x31F1);
```
fig02, line 508: sfx id

src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -21,47 +71,608 @@ const ActorInit En_Box_InitVars = {
...
+        } else {
+            sp4c_3f.x += f0 * 20.0f * Math_Sins(this->dyna.actor.posRot.rot.y + 0x4000);
+            sp4c_3f.z += f0 * 20.0f * Math_Coss(this->dyna.actor.posRot.rot.y + 0x4000);
+            f0 = Math_Rand_ZeroOne();
+            f0 = f0 + f0 - 1.0f;
+            sp40_3f.x = f0 * 1.6f * Math_Sins(this->dyna.actor.posRot.rot.y);
+            sp40_3f.y = 1.8f;
+            sp40_3f.z = f0 * 1.6f * Math_Coss(this->dyna.actor.posRot.rot.y);
+        }
+        func_8002AA44(globalCtx, &sp4c_3f, &sp40_3f, &sp34_3f, 0x96);
```
fig02, line 530: heres the prototype if you want to name these proper. if not id at least remove 3f from the name

`void EffectSsIceSmoke_Spawn(GlobalContext* globalCtx, Vec3f* pos, Vec3f* velocity, Vec3f* accel, s16 scale)`

scale in deimal would be appreciated

src/overlays/actors/ovl_En_Box/z_en_box.c
fig02, line 508: id call it iceSmokeTimer or even smokeTimer to keep it simple

src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -21,47 +71,608 @@ const ActorInit En_Box_InitVars = {
...
+        } else {
+            gSPDisplayList((*gfx)++, D_06001678);
+        }
+    }
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Box/func_809C94AC.s")
+Gfx* EnBox_EmptyDList(GraphicsContext* gfxCtx) {
+    Gfx* displayList;
+    Gfx* displayListHead;
```
fig02, line 591: we havent done this project wide yet, but we plan to go with dList as a name consistently at some point, if youd like to change it now

src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -21,47 +71,608 @@ const ActorInit En_Box_InitVars = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Box/func_809CA448.s")
+void EnBox_Draw(Actor* thisx, GlobalContext* globalCtx) {
+    GraphicsContext* gfxCtx;
+    EnBox* this = THIS;
+    Gfx* dispRefs[4];
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Box/func_809CA4A0.s")
+    gfxCtx = globalCtx->state.gfxCtx;
+    Graph_OpenDisps(dispRefs, globalCtx->state.gfxCtx, "../z_en_box.c", 1581);
```
fig02, line 650: can update this to use OPEN_DISPS

fig02 (commented):
sorry for the insanely late review on this. might be worth just skimming the file again yourself since this was your first file, and youve picked up on some of our conventions since then. I pointed out as much as I could see tho

src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -21,47 +71,608 @@ const ActorInit En_Box_InitVars = {
...
+    this->movementFlags = 0;
+    this->type = (thisx->params >> 12) & 0xF;
+    this->iceEffTimer = 0;
+    this->unk_1FB = ENBOX_STATE_0;
+    this->dyna.actor.gravity = -5.5f;
+    this->switchFlag = this->dyna.actor.posRot.rot.z;
+    this->dyna.actor.minVelocityY = -50.0f;
+
+    if (globalCtx) {} // helps the compiler store globalCtx into s1
+
```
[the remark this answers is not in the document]
Dragorn421: can't remove it no :(

fig02 (approved):

fig02:
still waiting on a roman review, but theres apparently a conflict in functions.h. probably from something that got merged today lol

src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -21,47 +71,598 @@ const ActorInit En_Box_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Box/func_809C8DC0.s")
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Box/func_809C8DC8.s")
+static AnimationHeader* D_809CA800[4] = { &D_0600024C, &D_06000128, &D_0600043C, &D_0600043C };
+
+static InitChainEntry sInitChain[] = {
+    ICHAIN_U8(unk_1F, 0, ICHAIN_STOP),
+};
+
+static s32 unused;
```
Roman971, line 81:
```suggestion
static s32 sUnused;
```

src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -21,47 +71,598 @@ const ActorInit En_Box_InitVars = {
...
+        }
+        EnBox_SetupAction(this, EnBox_WaitOpen);
+        this->movementFlags |= ENBOX_MOVE_IMMOBILE;
+        this->movementFlags |= ENBOX_MOVE_STICK_TO_GROUND;
+    }
+
+    this->dyna.actor.shape.rot.z = 0;
+    this->dyna.actor.posRot.rot.y += 0x8000;
+    this->dyna.actor.posRot.rot.z = this->dyna.actor.shape.rot.z;
+    this->dyna.actor.initPosRot.rot.z = this->dyna.actor.shape.rot.z;
```
Roman971, line 185:
```suggestion
    this->dyna.actor.posRot.rot.y += 0x8000;
    this->dyna.actor.initPosRot.rot.z = this->dyna.actor.posRot.rot.z = this->dyna.actor.shape.rot.z = 0;
```

src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -21,47 +71,598 @@ const ActorInit En_Box_InitVars = {
...
+    } else {
+        sfxId = 0;
+
+        if (func_800A56C8(&this->skelanime, 30.0f)) {
+            sfxId = NA_SE_EV_TBOX_UNLOCK;
+        } else if (func_800A56C8(&this->skelanime, 90.0f)) {
+            sfxId = NA_SE_EV_TBOX_OPEN;
+        }
+
+        if (sfxId) {
```
Roman971, line 481:
```suggestion
        if (sfxId != 0) {
```

src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -21,47 +71,598 @@ const ActorInit En_Box_InitVars = {
...
+
+void EnBox_Update(Actor* thisx, GlobalContext* globalCtx) {
+    EnBox* this = THIS;
+
+    if (this->movementFlags & ENBOX_MOVE_STICK_TO_GROUND) {
+        this->movementFlags &= ~ENBOX_MOVE_STICK_TO_GROUND;
+        EnBox_ClipToGround(this, globalCtx);
+    }
+
+    (this->actionFunc)(this, globalCtx);
```
Roman971, line 535:
```suggestion
    this->actionFunc(this, globalCtx);
```

src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -21,47 +71,598 @@ const ActorInit En_Box_InitVars = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Box/func_809C9630.s")
+    dListHead = Graph_Alloc(gfxCtx, sizeof(Gfx));
+    if (dListHead == NULL) {
+        __assert("gfxp != NULL", "../z_en_box.c", 1528);
+    }
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Box/func_809C9700.s")
+    dList = dListHead;
+    gSPEndDisplayList(dList++);
```
Roman971, line 592: It's the other way around, we call `dList` the pointer to the generated display list, while `dListHead` is a pointer to the head of the list we're currently appending to.
```suggestion
    dListHead = dList;
    gSPEndDisplayList(dListHead++);
```

src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -21,47 +71,598 @@ const ActorInit En_Box_InitVars = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Box/EnBox_Destroy.s")
+    if ((this->dyna.actor.params >> 5 & 0x7F) == 0x7C && this->actionFunc == EnBox_Open &&
+        this->skelanime.animCurrentFrame > 45 && this->iceSmokeTimer < 100) {
+        func_809C9EF8(this, globalCtx);
+    }
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Box/func_809C92F4.s")
+void EnBox_PostLimbDraw2(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3s* rot, Actor* thisx, Gfx** gfx) {
```
Roman971, line 559: The type is called "2" but that's just because there are 2 slightly different variants, so we usually just name the function without it.
```suggestion
void EnBox_PostLimbDraw(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3s* rot, Actor* thisx, Gfx** gfx) {
```

src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -21,47 +71,598 @@ const ActorInit En_Box_InitVars = {
...
+    f32 frameCount;
+    AnimationHeader* anim;
+    s32 linkAge;
+    s32 pad;
+    Vec3f sp4C;
+    Player* player;
+
+    this->alpha = 255;
+    this->movementFlags |= ENBOX_MOVE_IMMOBILE;
+    if (this->unk_1F4) { // unk_1F4 is modified by player code
```
Roman971, line 414:
```suggestion
    if (this->unk_1F4 != 0) { // unk_1F4 is modified by player code
```

src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -21,47 +71,598 @@ const ActorInit En_Box_InitVars = {
...
+    EnBox* this = THIS;
+    AnimationHeader* animHeader;
+    s32 dynaUnk;
+    f32 animFrameStart;
+    f32 animFrameCount;
+    s32 linkAge;
+
+    animFrameStart = 0.0f;
+    linkAge = gSaveContext.linkAge;
+    animHeader = D_809CA800[linkAge];
```
Roman971, line 112: FYI this is probably just the `(0, foo)` pattern and there is no actual temp
```suggestion
    animHeader = D_809CA800[(0, gSaveContext.linkAge)];
```

src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -4,14 +4,64 @@
...
+     ends up "blinking" (set/clear every frame) once switch flag is set,
+        if some collision-related condition (?) is met
+     only used for signum of z rotation
+    */
+    ENBOX_MOVE_FALL_ANGLE_SIDE = 1 << 2,
+    /*
+     when set, gets cleared next EnBox_Update call and clip to the floor
+    */
+    ENBOX_MOVE_STICK_TO_GROUND = 1 << 4
+} EnBox_MovementFlags;
```
Roman971, line 31: We don't really have any named flags right now so there is no standard, but imho flags should be defines, not enums with hardcoded values. It's also how decomp projects normally do it, or even just C codebases in general afaik.

src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -4,14 +4,64 @@
...
+
+typedef enum {
+    ENBOX_STATE_0, // waiting for player near / player available / player ? (IDLE)
+    ENBOX_STATE_1, // used only temporarily, maybe "player is ready" ?
+    ENBOX_STATE_2  // waiting for something message context-related
+} EnBox_StateUnk1FB;
```
Roman971, line 37:
```suggestion
} EnBoxStateUnk1FB;
```

src/overlays/actors/ovl_En_Box/z_en_box.h
```diff
@@ -4,13 +4,47 @@
...
+    /*  3 */ ENBOX_TYPE_SWITCH_FLAG_FALL_BIG,   // falling, appear on switch flag set
+    /*  4 */ ENBOX_TYPE_4,                      // big, drawn differently
+    /*  5 */ ENBOX_TYPE_SMALL,                  // same as ENBOX_TYPE_BIG_DEFAULT but small
+    /*  6 */ ENBOX_TYPE_6,                      // small, drawn differently
+    /*  7 */ ENBOX_TYPE_ROOM_CLEAR_SMALL,       // use room clear, store temp clear as perm clear
+    /*  8 */ ENBOX_TYPE_SWITCH_FLAG_FALL_SMALL, // falling, appear on switch flag set
+    /*  9 */ ENBOX_TYPE_9,                      // big, has something more to do with player and message context?
+    /* 10 */ ENBOX_TYPE_10,                     // like 9
+    /* 11 */ ENBOX_TYPE_SWITCH_FLAG_BIG         // big, appear on switch flag set
+} EnBox_Type;
```
Roman971, line 30:
```suggestion
} EnBoxType;
```

Roman971 (changes requested):

src/overlays/actors/ovl_En_Box/z_en_box.c
Dragorn421: Somehow that made the stack usage go up, but removing the `linkAge` temp thanks to `(0, gSaveContext.linkAge)` brought it back to matching. Nice

Roman971 (approved):

PR #358, opened 2020-08-29
attchaedA/attachedB renamed to parent/child

fig02:
and some cleanups related to the name change. including changing returns to true/false in related functions, and getting rid of a bunch of != 0's

func_8002F410 -> Actor_HasParent
func_8002F618 -> Actor_HasChild
func_8002F5A0 -> Actor_HasNoParent
func_8002F674 -> Actor_HasNoChild
Actor_SpawnAttached -> Actor_SpawnAsChild

sort of unrelated but I also changed func_80033640 to Actor_GetCollidedExplosive. This was incorrectly being called an attached actor in one of its usages, so this change was just to make that separation clear


src/code/z_actor.c
```diff
@@ -2804,18 +2804,18 @@ Actor* Actor_Spawn(ActorContext* actorCtx, GlobalContext* globalCtx, s16 actorId
     return actor;
 }
 
-Actor* Actor_SpawnAttached(ActorContext* actorCtx, Actor* attachedTo, GlobalContext* globalCtx, s16 actorId, f32 posX,
-                           f32 posY, f32 posZ, s16 rotX, s16 rotY, s16 rotZ, s16 params) {
+Actor* Actor_SpawnChild(ActorContext* actorCtx, Actor* parent, GlobalContext* globalCtx, s16 actorId, f32 posX,
```
Roman971, line 2807: Maybe `Actor_SpawnAsChild` would be better? Just a thought

fig02: yeah i like it

fig02: changed

Roman971:
Formatting needs to be updated but otherwise I think this looks good

fig02:
whoops. fixed

Roman971 (approved):

PR #359, opened 2020-08-29
BgMizuWater OK

kroemker:


src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,320 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+
+void BgMizuWater_ChangeWaterLevel(BgMizuWater* this, GlobalContext* globalCtx) {
+    u32 pad;
+    s16 params;
+    s32 waterLevelActionIndex;
+    WaterBox* waterBoxes;
+
+    waterBoxes = globalCtx->colCtx.stat.colHeader->waterBoxes;
+    switch(this->type)
+    {
```
ethteck, line 233: Please run format.sh on this to fix formatting issues.

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,308 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+                this->targetY = this->baseY;
+            }
+            if (Math_ApproxF(&this->actor.posRot.pos.y, this->targetY, 1.0f) != 0) {
+                globalCtx->unk_11D30[0] = 0;
+                this->actionFunc = BgMizuWater_Idle;
+            }
+            waterBoxes[8].unk_02 = this->actor.posRot.pos.y;
+            break;
+        case 4:
+            if (Flags_GetSwitch(globalCtx, this->switchTriggerID) != 0) {
```
ethteck, line 271: Flags_GetSwitch returns a boolean, so you can remove `!= 0` from these

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,308 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+    }
+
+    this->actionFunc(this, globalCtx);
+}
+
+void BgMizuWater_Draw(Actor* thisx, GlobalContext* globalCtx) {
+    BgMizuWater* this = THIS;
+    s32 gameplayFrames;
+    GraphicsContext* gfxCtx;
+    Gfx* dispRefs[4];
```
ethteck, line 329: We just merged a change to how we handle graphics functions, so you'll want to adapt your GFX funcs to the new style. See other actors for reference :)

kroemker:
I rebased the branch to get the new GFX changes and had to force push then. Don't know if this is the common way to do it, but it should work.

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,305 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+        bREG(0) = 0;
+    }
+    if ((Flags_GetSwitch(globalCtx, 0x1C)) && (switchID != 0x1C))
+        ret = 3;
+    else if ((Flags_GetSwitch(globalCtx, 0x1D)) && (switchID != 0x1D))
+        ret = 2;
+    else if ((Flags_GetSwitch(globalCtx, 0x1E)) && (switchID != 0x1E))
+        ret = 1;
+    else
+        ret = 0;
```
Dragorn421, line 81: Use braces for each block, like
```c
    if ((Flags_GetSwitch(globalCtx, 0x1C)) && (switchID != 0x1C)) {
        ret = 3;
    } else if (...) {
        ...
```

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,305 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+}
+
+void BgMizuWater_setWaterBoxesHeights(WaterBox* waterBoxes, s16 height) {
+    u32 i;
+
+    for (i = 0; i < 8; i += 4) {
+        waterBoxes[waterBoxIDs[i]].unk_02 = height;
+        waterBoxes[waterBoxIDs[i + 1]].unk_02 = height;
+        waterBoxes[waterBoxIDs[i + 2]].unk_02 = height;
+        waterBoxes[waterBoxIDs[i + 3]].unk_02 = height;
```
Dragorn421, line 93: does it match doing a more normal loop like
```c
    for (i = 0; i < 8; i ++) {
        waterBoxes[waterBoxIDs[i]].unk_02 = height;
    }
```
?

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,305 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+    }
+
+    this->actionFunc = BgMizuWater_Idle;
+}
+
+void BgMizuWater_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+}
+
+void BgMizuWater_Idle(BgMizuWater* this, GlobalContext* globalCtx) {
+    u32 pad;
```
Dragorn421, line 172: padding should be s32

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,305 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+            if (Flags_GetSwitch(globalCtx, this->switchTriggerID))
+                this->targetY = this->baseY + 110.0f;
+            else
+                this->targetY = this->baseY;
+            break;
+        case 4:
+            if (Flags_GetSwitch(globalCtx, this->switchTriggerID))
+                this->targetY = this->baseY + 160.0f;
+            else
+                this->targetY = this->baseY;
```
Dragorn421, line 208: Use braces here too `if (...) {...} else {...}`

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,305 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+                globalCtx->unk_11D30[0] = 0;
+                this->actionFunc = BgMizuWater_Idle;
+            }
+            waterBoxes[16].unk_02 = this->actor.posRot.pos.y;
+            break;
+    }
+
+    if (this->targetY < this->actor.posRot.pos.y) {
+        func_800AA000(0.0f, 0x78, 0x14, 0xA);
+        func_8002F948(&this->actor, NA_SE_EV_WATER_LEVEL_DOWN - SFX_FLAG);
```
Dragorn421, line 286: `& ~SFX_FLAG` makes more sense than `- SFX_FLAG` for a flag

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,305 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Mizu_Water/BgMizuWater_Destroy.s")
+    gSPMatrix(oGfxCtx->polyXlu.p++, Matrix_NewMtx(globalCtx->state.gfxCtx, "../z_bg_mizu_water.c", 749),
+              G_MTX_NOPUSH | G_MTX_LOAD | G_MTX_MODELVIEW);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Mizu_Water/func_8089FCF8.s")
+    gDPSetEnvColor(oGfxCtx->polyXlu.p++, 0xFF, 0xFF, 0xFF, 0x80);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Mizu_Water/func_8089FEC4.s")
+    gDPSetPrimColor(oGfxCtx->polyXlu.p++, 0, 0, 0xFF, 0xFF, 0xFF, 0x66);
```
Dragorn421, line 342: Use decimal for color/alpha, eg 255 instead of 0xFF

undefined_syms.txt
```diff
@@ -261,7 +264,7 @@ D_06002590 = 0x06002590;
 
 // z_bg_spot08_bakudankabe
 D_060039D4 = 0x060039D4;
-D_06003898 = 0x06003898; 
+D_06003898 = 0x06003898;
```
Dragorn421, line 267: Does this have to do with this PR?

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -9,7 +15,23 @@ void BgMizuWater_Destroy(Actor* thisx, GlobalContext* globalCtx);
...
-/*
+void BgMizuWater_Idle(BgMizuWater* this, GlobalContext* globalCtx);
+void BgMizuWater_ChangeWaterLevel(BgMizuWater* this, GlobalContext* globalCtx);
+
+extern UNK_TYPE D_06004B20;
+
+typedef struct {
+    s32 switchID;
+    s32 yDiff;
+} WaterLevel;
```
Dragorn421, line 26: Usually a switch flag is called `switchFlag`, may want to change `switchID` to that to keep it consistent.

Dragorn421, line 26: `yDiff` is the y relative to where the actor spawned (baseY), correct? `yDiff` sounds like it is relative to the current y instead.
I don't have a lot of ideas for naming it better, do you?

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,305 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
 };
-*/
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Mizu_Water/func_8089F8B0.s")
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Mizu_Water/func_8089F9D0.s")
+static u32 unk_808A0550 = 0;
+static f32 unk_808A0554 = 110.0f;
+
+static u32 waterBoxIDs[] = { 0x00000002, 0x00000003, 0x00000005, 0x00000007,
+                             0x0000000C, 0x00000014, 0x00000015, 0x00000016 };
```
Dragorn421, line 51: Does hexadecimal really make sense for indices?

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,305 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+
+static InitChainEntry sInitChain[] = {
+    ICHAIN_VEC3F(scale, 1, ICHAIN_STOP),
+};
+
+u32 BgMizuWater_getWaterLevelActionIndex(s16 switchID, GlobalContext* globalCtx) {
```
Dragorn421, line 57: The naming scheme is `SystemName_FuncName` so this lacks capitalization (same for `BgMizuWater_setWaterBoxesHeights`)

Also why `Action`? the result indexes into `waterLevels` so it makes more sense to just call it `waterLevelIndex`?

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,305 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+    if ((Flags_GetSwitch(globalCtx, 0x1C)) && (switchID != 0x1C))
+        ret = 3;
+    else if ((Flags_GetSwitch(globalCtx, 0x1D)) && (switchID != 0x1D))
+        ret = 2;
+    else if ((Flags_GetSwitch(globalCtx, 0x1E)) && (switchID != 0x1E))
+        ret = 1;
+    else
+        ret = 0;
+
+    return ret;
```
Dragorn421, line 83: you have to use a temp for the return value?

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,305 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+}
+
+void BgMizuWater_setWaterBoxesHeights(WaterBox* waterBoxes, s16 height) {
```
Dragorn421, line 86: Maybe only have "water boxes" plural and keep "height" singular?

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,305 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+                if (params != waterLevels[waterLevelActionIndex].switchID) {
+                    this->actor.params = waterLevels[waterLevelActionIndex].switchID;
+                    this->targetY = waterLevels[waterLevelActionIndex].yDiff + this->baseY;
+                }
+            }
+
+            if ((params != this->actor.params) && (params != 0))
+                Flags_UnsetSwitch(globalCtx, params);
+
+            if (Math_ApproxF(&this->actor.posRot.pos.y, this->targetY, 5.0f) != 0) {
```
Dragorn421, line 237: `Math_ApproxF` returns a boolean so removing the `!= 0` checks makes sense

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.h
```diff
@@ -6,9 +6,15 @@
...
+typedef void (*BgMizuWaterActionFunc)(struct BgMizuWater*, GlobalContext*);
+
 typedef struct BgMizuWater {
     /* 0x0000 */ Actor actor;
-    /* 0x014C */ char unk_14C[0x14];
+    /* 0x014C */ BgMizuWaterActionFunc actionFunc;
+    /* 0x0150 */ s32 type;
+    /* 0x0154 */ f32 targetY;
+    /* 0x0158 */ f32 baseY;
+    /* 0x015C */ s32 switchTriggerID; // only used for types 2-4
```
Dragorn421, line 17: `triggerSwitchFlag` ?

Dragorn421 (changes requested):
Nice, I didn't expect this to have a configuration, even hardcoded in a `WaterLevel` array.
Just out of curiosity, do you have an idea of what the different types do?

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,305 @@ const ActorInit Bg_Mizu_Water_InitVars = {
     (ActorFunc)BgMizuWater_Update,
     (ActorFunc)BgMizuWater_Draw,
 };
-*/
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Mizu_Water/func_8089F8B0.s")
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Mizu_Water/func_8089F9D0.s")
+static u32 unk_808A0550 = 0;
+static f32 unk_808A0554 = 110.0f;
```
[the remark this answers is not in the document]
kroemker: Yes, they are unused. Maybe they are from earlier stages of development since 110.0f is the also the y-difference used in type 3 (just speculation though).

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
kroemker: Yea, I'm not too happy with that name either. My idea with calling it `Action` was that it returns 0 if the water level doesn't have to be changed and a water level index otherwise, which is used for changing the water level (it also sets the switches based on `bREG(0)` and `bREG(1)`). But yeah, probably a bad name anyway.

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
kroemker: Yes, it is required.

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.h
kroemker: It's the switch ID for the types 2-4 for changing the water level, since type 0 checks all three switches `0x1C`, `0x1D`, `0x1E`. Maybe just `switchID` ?

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
fig02: we use - sfx flag for now. can talk about it more in discord if you like, but this is the convention for the time being

kroemker:
Afaik the only type used is type 0, which uses the three switches `0x1C`, `0x1D`, `0x1E` to determine the water level. Types 1-4 change the waterbox height by a hardcoded value, where type 1 doesn't change it at all, type 4 by 160 and type 2 and 3 in between. They also use a arbitrary switch (except type 0 which essentially does nothing) in contrast to type 0 that uses the 3 hardcoded switches.

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
kroemker: Not really, I'm not good at this naming stuff.

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.h
Dragorn421: I guess leaving out `trigger` is fine. All actors ever do with switch flags is set/clear/react to them anyway so I'd say it's good 👍

Dragorn421 (approved):

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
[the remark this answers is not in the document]
ethteck: I don't think we should necessarily name unused things like this. @Roman971 @fig02 ?

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -9,7 +15,23 @@ void BgMizuWater_Destroy(Actor* thisx, GlobalContext* globalCtx);
...
+void BgMizuWater_ChangeWaterLevel(BgMizuWater* this, GlobalContext* globalCtx);
+
+extern Gfx D_06004B20[];
+
+typedef struct {
+    s32 switchFlag;
+    s32 yDiff;
+} WaterLevel;
+
+static WaterLevel waterLevels[] = {
```
Roman971, line 28:
```suggestion
static WaterLevel sWaterLevels[] = {
```

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,307 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
     (ActorFunc)BgMizuWater_Draw,
 };
-*/
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Mizu_Water/func_8089F8B0.s")
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Mizu_Water/func_8089F9D0.s")
+static f32 unused1 = 0;
+static f32 unused2 = 110.0f;
+
+static u32 waterBoxIDs[] = { 2, 3, 5, 7, 12, 20, 21, 22 };
```
Roman971, line 50:
```suggestion
static f32 sUnused1 = 0;
static f32 sUnused2 = 110.0f;

static u32 sWaterBoxIndexes[] = { 2, 3, 5, 7, 12, 20, 21, 22 };
```

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,307 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+                Flags_SetSwitch(globalCtx, 0x1E);
+                break;
+        }
+        bREG(0) = 0;
+    }
+    if ((Flags_GetSwitch(globalCtx, 0x1C)) && (switchFlag != 0x1C)) {
+        ret = 3;
+    } else if ((Flags_GetSwitch(globalCtx, 0x1D)) && (switchFlag != 0x1D)) {
+        ret = 2;
+    } else if ((Flags_GetSwitch(globalCtx, 0x1E)) && (switchFlag != 0x1E)) {
```
Roman971, line 77:
```suggestion
    if (Flags_GetSwitch(globalCtx, 0x1C) && (switchFlag != 0x1C)) {
        ret = 3;
    } else if (Flags_GetSwitch(globalCtx, 0x1D) && (switchFlag != 0x1D)) {
        ret = 2;
    } else if (Flags_GetSwitch(globalCtx, 0x1E) && (switchFlag != 0x1E)) {
```

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,307 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+    WaterBox* waterBoxes;
+    s32 waterLevelActionIndex;
+
+    waterBoxes = globalCtx->colCtx.stat.colHeader->waterBoxes;
+    this->type = this->actor.params & 0xFF;
+    this->switchFlag = (this->actor.params >> 8) & 0xFF;
+    Actor_ProcessInitChain(&this->actor, sInitChain);
+    initialActorY = this->actor.posRot.pos.y;
+    this->baseY = initialActorY;
+    this->targetY = initialActorY;
```
Roman971, line 106: Any chance you could get rid of `initialActorY` like this? (it will probably have to be replaced by a pad)
```suggestion
    this->targetY = this->baseY =  this->actor.posRot.pos.y;
```

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,307 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+            if (Flags_GetSwitch(globalCtx, this->switchFlag)) {
+                this->actor.posRot.pos.y = this->baseY + 85.0f;
+            }
+            waterBoxes[6].unk_02 = this->actor.posRot.pos.y;
+            break;
+        case 3:
+            if (Flags_GetSwitch(globalCtx, this->switchFlag)) {
+                this->actor.posRot.pos.y = this->baseY + 110.0f;
+                if (1) {}
+                this->targetY = this->actor.posRot.pos.y;
```
Roman971, line 148: Any chance this works? (might still need the if (1) {} for other reasons). If it works you can apply the same idea in case 4 below.
```suggestion
                this->targetY = this->actor.posRot.pos.y = this->baseY + 110.0f;
```

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,307 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+    this->actionFunc = BgMizuWater_Idle;
+}
+
+void BgMizuWater_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+}
+
+void BgMizuWater_Idle(BgMizuWater* this, GlobalContext* globalCtx) {
+    s32 pad;
+    s32 waterLevelActionIndex;
+    s16 params;
```
Roman971, line 171: Considering the usage of this variable, I suggest naming it `prevSwitchFlag` or `oldSwitchFlag`

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,307 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+    }
+
+    if (this->targetY != this->actor.posRot.pos.y) {
+        this->actionFunc = BgMizuWater_ChangeWaterLevel;
+    }
+}
+
+void BgMizuWater_ChangeWaterLevel(BgMizuWater* this, GlobalContext* globalCtx) {
+    s32 pad;
+    s16 params;
```
Roman971, line 219: same here about changing it to `prevSwitchFlag` or `oldSwitchFlag`

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,307 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+            break;
+    }
+
+    this->actionFunc = BgMizuWater_Idle;
+}
+
+void BgMizuWater_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+}
+
+void BgMizuWater_Idle(BgMizuWater* this, GlobalContext* globalCtx) {
```
Roman971, line 168: This is just a suggestion but here is an idea for a more explicit name:
```suggestion
void BgMizuWater_WaitForAction(BgMizuWater* this, GlobalContext* globalCtx) {
```

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
```diff
@@ -21,19 +43,307 @@ const ActorInit Bg_Mizu_Water_InitVars = {
...
+        func_8002F948(&this->actor, NA_SE_EV_WATER_LEVEL_DOWN - SFX_FLAG);
+    } else if (this->targetY > this->actor.posRot.pos.y) {
+        func_800AA000(0.0f, 0x78, 0x14, 0xA);
+        func_8002F948(&this->actor, NA_SE_EV_WATER_LEVEL_DOWN - SFX_FLAG);
+    }
+}
+
+void BgMizuWater_Update(Actor* thisx, GlobalContext* globalCtx) {
+    BgMizuWater* this = THIS;
+    s32 y;
```
Roman971, line 297:
```suggestion
    s32 posY;
```

Roman971 (changes requested):

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
kroemker: Sadly no, `initialActorY` is required.

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.c
kroemker: It doesn't work. It's strange, but after the calculation of `this->baseY + 110.0f` the result is stored in `this->actor.posRot.pos.y` and then afterwards loaded again from there to store it in `this->targetY`. That's why the  `if (1) {}` is required.

Roman971 (approved):

src/overlays/actors/ovl_Bg_Mizu_Water/z_bg_mizu_water.h
```diff
@@ -6,10 +6,16 @@
...
 typedef struct BgMizuWater {
     /* 0x0000 */ Actor actor;
-    /* 0x014C */ char unk_14C[0x14];
-} BgMizuWater; // size = 0x0160
+    /* 0x014C */ BgMizuWaterActionFunc actionFunc;
+    /* 0x0150 */ s32 type;
+    /* 0x0154 */ f32 targetY;
+    /* 0x0158 */ f32 baseY;
+    /* 0x015C */ s32 switchFlag; // only used for types 2-4
+} BgMizuWater;                   // size = 0x0160
```
fig02, line 18: just a small thing, we dont run the formatter on headers cause it can do things like this with the comment. if you could move the size comment back to right after the name id appreciate it :D 
everything else looks good to me

fig02 (approved):

PR #367, opened 2020-08-30
ovl_Bg_Hidan_Dalm OK

Dragorn421:
FeelsRegallocMan


src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.c
```diff
@@ -21,17 +29,175 @@ const ActorInit Bg_Hidan_Dalm_InitVars = {
...
+    Collider_DestroyTris(globalCtx, &this->unk_16C);
+}
+
+void BgHidanDalm_WaitInteraction(BgHidanDalm* this, GlobalContext* globalCtx) {
+    Player* player = PLAYER;
+
+    if ((this->unk_16C.base.acFlags & 2) && !func_8008E988(globalCtx) &&
+        (player->swordAnimation == 22 || player->swordAnimation == 23)) {
+        this->unk_16C.base.acFlags &= ~(1 << 1);
+        if (this->unk_16C.list[0].body.bumperFlags & 2 || this->unk_16C.list[1].body.bumperFlags & 2) {
```
fig02, line 114: everywhere else in the codebase we have bit flags as hex. They will definitely becomes defines that show the shift value in the future, but they should probably stay hex for the time being 
(lots of these comments that im adding are going to be relevant for en_box too)

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.c
```diff
@@ -21,17 +29,175 @@ const ActorInit Bg_Hidan_Dalm_InitVars = {
...
+        this->dyna.actor.posRot.pos.z += 32.5f * Math_Coss(this->dyna.actor.posRot.rot.y);
+
+        func_8002DF54(globalCtx, &this->dyna.actor, 8);
+        this->dyna.actor.flags |= 1 << 4;
+        this->actionFunc = &BgHidanDalm_Shrink;
+        this->dyna.actor.bgCheckFlags &= ~(1 << 1);
+        this->dyna.actor.bgCheckFlags &= ~(1 << 3);
+        this->dyna.actor.speedXZ = 10.0f;
+        Flags_SetSwitch(globalCtx, this->switchFlag);
+        func_8002F7DC(&PLAYER->actor, NA_SE_IT_HAMMER_HIT);
```
fig02, line 129: youve got a player temp, can you use it? 
want to avoid inline macro usages for THIS and PLAYER as much as we can

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.c
```diff
@@ -21,17 +29,175 @@ const ActorInit Bg_Hidan_Dalm_InitVars = {
...
+            this->dyna.actor.posRot.rot.y -= 0x4000;
+        } else {
+            this->dyna.actor.posRot.rot.y += 0x4000;
+        }
+        this->dyna.actor.posRot.pos.x += 32.5f * Math_Sins(this->dyna.actor.posRot.rot.y);
+        this->dyna.actor.posRot.pos.z += 32.5f * Math_Coss(this->dyna.actor.posRot.rot.y);
+
+        func_8002DF54(globalCtx, &this->dyna.actor, 8);
+        this->dyna.actor.flags |= 1 << 4;
+        this->actionFunc = &BgHidanDalm_Shrink;
```
fig02, line 124: for function pointers and arrays you do not need &. also relevant in en_box

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.c
```diff
@@ -21,17 +29,175 @@ const ActorInit Bg_Hidan_Dalm_InitVars = {
...
+
+    unkPos.x = this->dyna.actor.posRot.pos.x;
+    unkPos.y = this->dyna.actor.posRot.pos.y + this->dyna.actor.scale.x * 160.0f;
+    unkPos.z = this->dyna.actor.posRot.pos.z;
+
+    for (i = 0; i != 4; i++) {
+        unkDir.x = 5.0f * Math_Sins(this->dyna.actor.posRot.rot.y + 0x8000) + (Math_Rand_ZeroOne() - 0.5f) * 5.0f;
+        unkDir.z = 5.0f * Math_Coss(this->dyna.actor.posRot.rot.y + 0x8000) + (Math_Rand_ZeroOne() - 0.5f) * 5.0f;
+        unkDir.y = (Math_Rand_ZeroOne() - 0.5f) * 1.5f;
+        func_80028B18(globalCtx, &unkPos, &unkDir, &D_808865A8);
```
fig02, line 156: i have this effect decompiled, if you wanted to name these args. Its the sparkle effect.
func_80028B18(GlobalContext* globalCtx, Vec3f* pos, Vec3f* velocity, Vec3f* accel)

you can also inline the definition for D_808865A8 as static in the function, which looks a bit more natural. 
We tend to try to keep static vars on top of the stack vars in the list. If what i explained isnt clear feel free to ask here or in discord

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.c
```diff
@@ -21,17 +29,175 @@ const ActorInit Bg_Hidan_Dalm_InitVars = {
...
+    }
+
+    this->dyna.actor.scale.y = this->dyna.actor.scale.z = this->dyna.actor.scale.x;
+
+    unkPos.x = this->dyna.actor.posRot.pos.x;
+    unkPos.y = this->dyna.actor.posRot.pos.y + this->dyna.actor.scale.x * 160.0f;
+    unkPos.z = this->dyna.actor.posRot.pos.z;
+
+    for (i = 0; i != 4; i++) {
+        unkDir.x = 5.0f * Math_Sins(this->dyna.actor.posRot.rot.y + 0x8000) + (Math_Rand_ZeroOne() - 0.5f) * 5.0f;
```
fig02, line 153: should be able to do < 4

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.c
```diff
@@ -21,17 +29,175 @@ const ActorInit Bg_Hidan_Dalm_InitVars = {
...
+        func_80028B18(globalCtx, &unkPos, &unkDir, &D_808865A8);
+    }
+}
+
+void BgHidanDalm_Update(Actor* thisx, GlobalContext* globalCtx) {
+    BgHidanDalm* this = THIS;
+
+    this->actionFunc(this, globalCtx);
+    Actor_MoveForward(thisx);
+    func_8002E4B4(globalCtx, thisx, 10.0f, 15.0f, 32.0f, 5);
```
fig02, line 165: does using &this->actor work? we prefer that when possible so that we're not mixing this and thisx usage.
This is relevant for the other main functions in this file.
If its needed to match for any of them though then dont worry about it

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.c
```diff
@@ -21,17 +29,175 @@ const ActorInit Bg_Hidan_Dalm_InitVars = {
...
+
+#define sTrisInit D_80886590
+ColliderTrisInit sTrisInit = {
+    { COLTYPE_UNK10, 0x00, 0x09, 0x00, 0x20, COLSHAPE_TRIS },
+    4,
+    sTrisItemInit,
+};
+
+#define sInitChain D_808865A0
+InitChainEntry sInitChain[] = {
```
fig02, line 60: is the define going to be local to this file? it would cause issues if not, just making sure.

generally we just dont rename the var and leave it as the D_ variant until the asm file isnt needed anymore.
But if this solution works its probably fine (@Roman971 thoughts?)

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.c
```diff
@@ -8,8 +8,16 @@ void BgHidanDalm_Init(Actor* thisx, GlobalContext* globalCtx);
...
 void BgHidanDalm_Draw(Actor* thisx, GlobalContext* globalCtx);
+#define BgHidanDalm_WaitInteraction func_80885F4C
+void BgHidanDalm_WaitInteraction(BgHidanDalm* this, GlobalContext* globalCtx);
+void BgHidanDalm_Shrink(BgHidanDalm* this, GlobalContext* globalCtx);
+void BgHidanDalm_UpdateCollider(BgHidanDalm* this);
+
+// object
+extern Gfx D_0600BBF0[];
+extern Gfx D_0600BDF0[];
+extern UNK_TYPE D_0600DA10;
```
fig02, line 19: I personally prefer to have the externs underneath the data, right before the functions start. Its not a hard requirement, but its what a majority of the actors in the code base do. 

I also dont think the object comment is needed, thats a given cause of segment 6 which is set up by the engine. Theyll be named to indicate that too eventually

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.c
```diff
@@ -8,8 +8,16 @@ void BgHidanDalm_Init(Actor* thisx, GlobalContext* globalCtx);
 void BgHidanDalm_Destroy(Actor* thisx, GlobalContext* globalCtx);
 void BgHidanDalm_Update(Actor* thisx, GlobalContext* globalCtx);
 void BgHidanDalm_Draw(Actor* thisx, GlobalContext* globalCtx);
+#define BgHidanDalm_WaitInteraction func_80885F4C
+void BgHidanDalm_WaitInteraction(BgHidanDalm* this, GlobalContext* globalCtx);
```
fig02, line 12: we like to keep a line break between the main functions (that take thisx) and the other ones.
also you can rename the function in the init asm file to get around needing the define

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.c
Dragorn421: unfortunately, no

Dragorn421:
I got rid of the `#define sInitChain D_808865A0`, I renamed the function in the .s and got rid of `sInitChain` names
Should have addressed everything else apart from the unskippable PLAYER

fig02 (approved):

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.h
```diff
@@ -6,9 +6,14 @@
 
 struct BgHidanDalm;
 
+typedef void (*BgHidanDalm_ActionFunc)(struct BgHidanDalm*, GlobalContext*);
```
Roman971, line 9:
```suggestion
typedef void (*BgHidanDalmActionFunc)(struct BgHidanDalm*, GlobalContext*);
```

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.c
```diff
@@ -9,7 +9,10 @@ void BgHidanDalm_Destroy(Actor* thisx, GlobalContext* globalCtx);
 void BgHidanDalm_Update(Actor* thisx, GlobalContext* globalCtx);
 void BgHidanDalm_Draw(Actor* thisx, GlobalContext* globalCtx);
 
-/*
+void BgHidanDalm_WaitInteraction(BgHidanDalm* this, GlobalContext* globalCtx);
+void BgHidanDalm_Shrink(BgHidanDalm* this, GlobalContext* globalCtx);
+void BgHidanDalm_UpdateCollider(BgHidanDalm* this);
```
Roman971, line 14:
```suggestion
```

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.c
```diff
@@ -21,17 +24,176 @@ const ActorInit Bg_Hidan_Dalm_InitVars = {
...
+
+    switchFlag = thisx->params >> 8; /*
+    // this instead of the line above solves half the regalloc issues
+    switchFlag = thisx->params;
+    switchFlag = switchFlag >> 8 & 0xFFFFFFFFFFFFFFFF;
+    //*/
+    this->switchFlag = switchFlag;   // can use thisx->params >> 8 without changes
+    thisx->params &= 0xFF;
+    // can use thisx->params >> 8 , seems to improve regalloc, but it changes codegen
+    if (Flags_GetSwitch(globalCtx, switchFlag & 0xFF)) {
```
Roman971, line 86: This should match:
```suggestion
    this->switchFlag = (thisx->params >> 8) & 0xFF;
    thisx->params &= 0xFF;

    if (Flags_GetSwitch(globalCtx, this->switchFlag)) {
```

For the explanation, the `& 0xFF` in the `Flags_GetSwitch` call is just an implicit `u8` cast. As a general rule, when something is read after it was just written, the compiler often optimizes out the reload, but it casts the value so it's actually equivalent to a load. So here, after writing inside `this->switchFlag`, it keeps the register and casts it to `u8` since that's the type of `this->switchFlag`, which looks like a variable but it's not.

As for the actual `& 0xFF` when writing `this->switchFlag`, it's optimized out since it stores it as a byte anyway, but it's important for regalloc.

And you can replace the `switchFlag` temp with a pad since that should still be needed for stack allocation.

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.c
```diff
@@ -21,17 +24,176 @@ const ActorInit Bg_Hidan_Dalm_InitVars = {
...
+
+    this->actionFunc(this, globalCtx);
+    Actor_MoveForward(&this->dyna.actor);
+    func_8002E4B4(globalCtx, &this->dyna.actor, 10.0f, 15.0f, 32.0f, 5);
+}
+
+/*
+Update vertices of collider tris based on the current matrix
+*/
+void BgHidanDalm_UpdateCollider(BgHidanDalm* this) {
```
Roman971, line 167:
```suggestion
/**
 * Update vertices of collider tris based on the current matrix
 */
void BgHidanDalm_UpdateCollider(BgHidanDalm* this) {
```

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.c
```diff
@@ -21,17 +24,176 @@ const ActorInit Bg_Hidan_Dalm_InitVars = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Hidan_Dalm/func_808862D8.s")
+    if (this->dyna.actor.params == 0) {
+        Gfx_DrawDListOpa(globalCtx, D_0600BBF0);
+    } else {
+        Gfx_DrawDListOpa(globalCtx, D_0600BDF0);
+    }
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Hidan_Dalm/BgHidanDalm_Draw.s")
+    if (this->actionFunc == &BgHidanDalm_WaitInteraction) {
```
Roman971, line 196:
```suggestion
    if (this->actionFunc == BgHidanDalm_WaitInteraction) {
```

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.h
```diff
@@ -6,9 +6,14 @@
...
+
 typedef struct BgHidanDalm {
-    /* 0x0000 */ Actor actor;
-    /* 0x014C */ char unk_14C[0x1B0];
+    /* 0x0000 */ DynaPolyActor dyna;
+    /* 0x0164 */ BgHidanDalm_ActionFunc actionFunc;
+    /* 0x0168 */ u8 switchFlag;
+    /* 0x016C */ ColliderTris unk_16C;
+    /* 0x018C */ ColliderTrisItem unk_18C[4];
```
Roman971, line 16:
```suggestion
    /* 0x016C */ ColliderTris collider;
    /* 0x018C */ ColliderTrisItem colliderItems[4];
```

Roman971 (changes requested):

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.c
Dragorn421: why?

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.c
Dragorn421: is that the format for documentation comments?

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.c
fig02: we only forward declare when necessary. this one is above is usage, so its not needed

src/overlays/actors/ovl_Bg_Hidan_Dalm/z_bg_hidan_dalm.c
Roman971: Yes

Roman971 (approved):

fig02 (approved):

PR #368, opened 2020-08-30
Po event

petrie911:
Bg_Po_Event matched and documented.


Dragorn421:
Nice, I noticed you commented out the data in .data.s instead of deleting the file
You should delete the file and remove it from the spec file
If the file is 100% OK you also want to use the compiled relocation data instead of the original
[See this other PR for how to edit the spec file](https://github.com/zeldaret/oot/commit/c9e85c4b8625386befb0bd801da69e35485c297c#diff-b979c2934ac0b4ba3f08dabfdd1b2299)

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+        sins *= 2.4f;
+        scaley = 1.818f;
+        coss *= 2.4f;
+    } else {
+        scaley = 1.0f;
+    }
+    for (i1 = 0; i1 < sTrisInit.count; i1++) {
+        item = &sTrisInit.list[i1];
+        if (1) {} // This section looks like a macro of some sort.
+        for (i2 = 0; i2 != 3; i2++) {
```
Dragorn421, line 108: can you use `i2 < 3` instead of `!= 3` ?

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+            vtxVec = &item->dim.vtx[i2];
+            sp9C[i2].x = (vtxVec->x * coss) + (this->dyna.actor.initPosRot.pos.x + (sins * vtxVec->z));
+            sp9C[i2].y = (vtxVec->y * scaley) + this->dyna.actor.initPosRot.pos.y;
+            sp9C[i2].z = this->dyna.actor.initPosRot.pos.z + (coss * vtxVec->z) - (vtxVec->x * sins);
+        }
+        func_800627A0(&this->collider1, i1, &sp9C[0], &sp9C[1], &sp9C[2]);
+    }
+    if (this->actorType != 4) {
+        if (this->actorIndex != 2) {
```
Dragorn421, line 117: can you merge those conditions with an and `&&` ?

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+            if (this->actorType == 2) {
+                phi_t2 = this->actorIndex;
+            } else {
+                phi_t2 = this->actorIndex + 2;
+            }
+            if (Actor_SpawnAsChild(&globalCtx->actorCtx, &this->dyna.actor, globalCtx, ACTOR_BG_PO_EVENT,
+                                   paintingsX[phi_t2], paintingsY[this->actorIndex], paintingsZ[this->actorIndex], 0,
+                                   this->dyna.actor.shape.rot.y + 0x8000, 0,
+                                   ((this->actorIndex + 1) << 0xC) + (this->actorType << 8) +
```
Dragorn421, line 126: I think `12` makes more sense than `0xC` for a bitshift offset?

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+                                       this->dyna.actor.params) == 0) {
```
Dragorn421, line 127: use NULL to check for null pointers such as the actor instance pointer returned by `Actor_SpawnAsChild`

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+                }
+                this->dyna.actor.parent = this->dyna.actor.child->child;
+                this->dyna.actor.child->child->child = &this->dyna.actor;
+            }
+        }
+    }
+    this->timer = 0;
+    if (this->actorType == 4) {
+        puzzleState = 0;
+        this->actionFunc = &BgPoEvent_AmyWait;
```
Dragorn421, line 144: no need to put a `&` in front for function pointers

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+    } else {
+        puzzleState = (s32)(Math_Rand_ZeroOne() * 3.0f) % 3;
+        this->actionFunc = &BgPoEvent_PaintingEmpty;
+    }
+}
+
+void BgPoEvent_InitBlocks(BgPoEvent* this, GlobalContext* globalCtx) {
+    Actor* thisx = &this->dyna.actor;
```
Dragorn421, line 152: Do you really have to use thisx over this?

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+    thisx->groundY = func_8003C9A4(&globalCtx->colCtx, &thisx->floorPoly, &dummy, thisx, &thisx->posRot.pos);
+    this->actionFunc = BgPoEvent_BlockWait;
+}
+
+void BgPoEvent_Init(Actor* thisx, GlobalContext* globalCtx) {
+    BgPoEvent* this = THIS;
+    ColliderTris* collider;
+
+    Actor_ProcessInitChain(&this->dyna.actor, sInitChain);
+    this->actorType = (thisx->params >> 8) & 0xF;
```
Dragorn421, line 191: can you avoid using thisx?

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+        puzzleState &= ~0x10;
+    }
+}
+
+void BgPoEvent_BlockShake(BgPoEvent* this, GlobalContext* globalCtx) {
+
+    DECR(this->timer);
+    if (this->timer < 15) {
+        this->dyna.actor.posRot.pos.x = this->dyna.actor.initPosRot.pos.x + 2.0f * ((this->timer % 3) - 1);
+        if (!(this->timer % 4)) {
```
Dragorn421, line 254: I think ` == 0` makes more sense when checking actual numerical values

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+            phi_v1 = temp_f6;
+            phi_a1 = temp_f10;
+            phi_a3 = this->dyna.actor.posRot.pos.x;
+            phi_t0 = this->dyna.actor.child->posRot.pos.x;
+        } else { // this->actorIndex == 1
+            phi_v1 = temp_f6;
+            phi_a1 = temp_f10;
+            phi_a3 = this->dyna.actor.child->posRot.pos.x;
+            phi_t0 = this->dyna.actor.posRot.pos.x;
+        }
```
Dragorn421, line 287: this is kinda weird, can you set  `phi_v1` and `phi_a1` outside of the sub-if/else,
or alternatively get rid of temp_f6 and temp_f10?

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+    }
+    if ((phi_v1 == phi_a1) && ((phi_t0 - phi_a3) == 60)) {
+        puzzleState |= (1 << this->actorIndex);
+    } else {
+        puzzleState &= ~(1 << this->actorIndex);
+    }
+}
+
+void BgPoEvent_BlockFall(BgPoEvent* this, GlobalContext* globalCtx) {
+    s32* temp1 = &firstFall;
```
Dragorn421, line 311: can this temp be inlined at all?

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+
+    this->dyna.actor.velocity.y++;
+    if (Math_ApproxF(&this->dyna.actor.posRot.pos.y, 433.0f, this->dyna.actor.velocity.y) != 0) {
```
Dragorn421, line 314: `Math_ApproxF` returns a bool so not having != 0 is ok

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+    Actor* newActor;
+
+    if (puzzleState == 0xF) {
+        this->actionFunc = BgPoEvent_BlockSolved;
+        if ((this->actorType == 0) && (this->actorIndex == 0)) {
+            newActor =
+                Actor_Spawn(&globalCtx->actorCtx, globalCtx, ACTOR_EN_PO_SISTERS, this->dyna.actor.posRot.pos.x + 30.0f,
+                            this->dyna.actor.posRot.pos.y - 30.0f, this->dyna.actor.posRot.pos.z + 30.0f, 0,
+                            this->dyna.actor.shape.rot.y, 0, this->dyna.actor.params + 0x300);
+            if (newActor != 0) {
```
Dragorn421, line 346: use NULL to check for null pointers

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+    if (2.0f < temp_f0) {
+        this->dyna.actor.speedXZ = 2.0f;
+    } else {
+        this->dyna.actor.speedXZ = temp_f0;
+    }
+    sp30 = Math_ApproxF(&blockPushDist, 20.0f, this->dyna.actor.speedXZ);
+    sp34 = this->direction * blockPushDist;
+    this->dyna.actor.posRot.pos.x = (Math_Sins(this->dyna.unk_158) * sp34) + this->dyna.actor.initPosRot.pos.x;
+    this->dyna.actor.posRot.pos.z = (Math_Coss(this->dyna.unk_158) * sp34) + this->dyna.actor.initPosRot.pos.z;
+    if (sp30 != 0) {
```
Dragorn421, line 405: `Math_ApproxF` result -> bool

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+
+void BgPoEvent_BlockPush(BgPoEvent* this, GlobalContext* globalCtx) {
+    f32 sp34;
+    s32 sp30;
+    Player* player = PLAYER;
+    f32 temp_f0;
+
+    this->dyna.actor.speedXZ += 0.1f;
+    temp_f0 = this->dyna.actor.speedXZ;
+    if (2.0f < temp_f0) {
```
Dragorn421, line 396: constants on the right hand side? idk

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+    }
+    if (Math_ApproxF(&this->dyna.actor.posRot.pos.y, 493.0f, 1.0f) != 0) {
+        if (Math_ApproxUpdateScaledS(&this->dyna.actor.shape.rot.z, this->dyna.actor.posRot.rot.z - 0x4000, 0x400) !=
+            0) {
+            this->actorIndex = (this->actorIndex + 1) % 4;
+            this->actionFunc = BgPoEvent_BlockFall;
+            puzzleState = 0;
+            if (this->actorType == 1) {
+                this->timer += 10;
+                this->timer = (this->timer >= 121) ? 120 : this->timer;
```
Dragorn421, line 442: does `this->timer > 120` work?

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+    Player* player = PLAYER;
+    s32 phi_v0;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Po_Event/func_808A7138.s")
+    DECR(this->timer);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Po_Event/func_808A7238.s")
+    if ((this->timer == 0) || ((thisx->xzDistFromLink < 150.0f) && (thisx->yDistFromLink < 50.0f)) ||
+        (((func_8002DD78(player) != 0) && (thisx->xzDistFromLink < 320.0f)) &&
+         ((this->actorIndex != 2) ? (thisx->yDistFromLink < 100.0f) : (thisx->yDistFromLink < 0.0f)) &&
```
Dragorn421, line 535: I would remove many brackets but there's at least one which definitely has no purpose
`((func_8002DD78(player) != 0) && (thisx->xzDistFromLink < 320.0f))` -> `(func_8002DD78(player) != 0) && (thisx->xzDistFromLink < 320.0f)`

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+         (func_8002DFC8(thisx, 0x2000, globalCtx) != 0))) {
+        /*This last conditional is checking if
+            1) Link is holding a ranged weapon
+            2) Link is too close in the xz plane
+            3) Link is too close in the y direction. The painting
+               under the balcony allows him to be closer.
+            4) Link is within 45 degrees of facing the painting. */
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Po_Event/func_808A72AC.s")
+        if ((thisx->parent != NULL) || (thisx->child != NULL)) {
```
Dragorn421, line 544: can you merge this if into its parent? making an even bigger condition 😄 
It would allow to remove the return and wrap the rest of the function in an `else` clause though

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+            this->timer = 0;
+            Audio_PlayActorSound2(thisx, NA_SE_EN_PO_LAUGH);
+            this->actionFunc = &BgPoEvent_PaintingVanish;
+            return;
+        }
+    }
+    if ((this->collider1.base.acFlags & 2) == 0) {
+        return;
+    }
```
Dragorn421, line 553: can you negate the condition and put the rest of the function under it?

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+    BgPoEvent* this = THIS;
+    f32 pad;
+    f32 pad2;
+    Vec3f sp58;
+    Vec3f sp4C;
+    f32 temp_f0;
+    u8 phi_a0;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Po_Event/BgPoEvent_Update.s")
+    OPEN_DISPS(globalCtx->state.gfxCtx, "../z_bg_po_event.c", 0x5C9);
```
Dragorn421, line 620: line numbers make more sense in decimal

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +48,601 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+    func_80093D18(globalCtx->state.gfxCtx);
+    if ((this->actorType == 3) || (this->actorType == 2)) {
+        if (this->actionFunc == BgPoEvent_PaintingEmpty) {
+            phi_a0 = 0xFF;
+        } else if (this->actionFunc == BgPoEvent_PaintingPresent) {
+            phi_a0 = 0;
+        } else {
+            phi_a0 = this->timer;
+        }
+        gDPSetEnvColor(oGfxCtx->polyOpa.p++, 0xFF, 0xFF, 0xFF, phi_a0);
```
Dragorn421, line 630: we use decimal for colors

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.h
```diff
@@ -6,9 +6,17 @@
...
 typedef struct BgPoEvent {
-    /* 0x0000 */ Actor actor;
-    /* 0x014C */ char unk_14C[0xFC];
+    /* 0x0000 */ DynaPolyActor dyna;
+    /* 0x0164 */ BgPoEventActionFunc actionFunc;
+    /* 0x0168 */ u8 actorType;
+    /* 0x0169 */ u8 actorIndex;
+    /* 0x016A */ s8 direction;
+    /* 0x016C */ s16 timer;
+    /* 0x0170 */ ColliderTris collider1;
```
Dragorn421, line 18: why the 1 in collider1?

Dragorn421 (changes requested):
👍 nice
Do you have hindsights on how this actor works? the `thisx->child->child->child->child = thisx;` scares me although it looks like a great meme

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
petrie911: You'd think that, but it actually doesn't match. For modulo operations with a power of 4, setting it equal to zero and using logical not aren't the same thing. In the former, the compiler knows it doesn't have to care about the sign of the result, but it doesn't know that with logical not.

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
petrie911: No, but it's a lot shorter than having this->dyna.actor everywhere and the stack padding is needed.

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
petrie911: For some reason, at most one of those three params statements can use this->dyna.actor. If two or more do, it fails to match. I suppose I could leave the two thisx in the actortype and actorindex statements and switch the rest to this->dyna.actor

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
petrie911: Yes, and that's probably how it was written. super ugly.

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
petrie911: Not that I could figure. You can replace one, but if you replace both the compiler loads the pointer to firstFall first, then loads its value from the pointer. This makes very little sense.

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
petrie911: I'm not sure of this one. bitshifts are a very binary concept.

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +41,576 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+            amy =
+                Actor_Spawn(&globalCtx->actorCtx, globalCtx, ACTOR_EN_PO_SISTERS, this->dyna.actor.posRot.pos.x + 30.0f,
+                            this->dyna.actor.posRot.pos.y - 30.0f, this->dyna.actor.posRot.pos.z + 30.0f, 0,
+                            this->dyna.actor.shape.rot.y, 0, this->dyna.actor.params + 0x300);
+            if (amy != NULL) {
+                func_800800F8(globalCtx, 0xC62, 0x1E, amy, 0);
+            }
+            func_80078884(0x4802);
+            gSaveContext.timer1State = 0xA;
+            return;
```
Dragorn421, line 330:
```suggestion
```

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +41,576 @@ const ActorInit Bg_Po_Event_InitVars = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Po_Event/func_808A7568.s")
+    this->actionFunc(this, globalCtx);
+    if ((this->actionFunc == BgPoEvent_AmyWait) || (this->actionFunc == BgPoEvent_PaintingPresent)) {
+        CollisionCheck_SetAC(globalCtx, &globalCtx->colChkCtx, &this->collider.base);
+    }
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Po_Event/func_808A75B8.s")
+Gfx* displayLists[] = { D_060075A0, D_060079E0, D_06006830, D_06006D60, D_06007230 };
```
Dragorn421, line 577: since you put that here can it also be moved into Draw?

Dragorn421 (changes requested):
details

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
petrie911: No. In fact, since there's no inlines it doesn't even need to be moved down, so I could put it at the top with the rest of the data if that's preferred.

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +41,573 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+
+void BgPoEvent_Init(Actor* thisx, GlobalContext* globalCtx) {
+    static InitChainEntry sInitChain[] = {
+        ICHAIN_VEC3F_DIV1000(scale, 1000, ICHAIN_STOP),
+    };
+    GlobalContext* globalCtx2 = globalCtx;
+    BgPoEvent* this = THIS;
+    
+    Actor_ProcessInitChain(&this->dyna.actor, sInitChain);
+    this->actorType = (thisx->params >> 8) & 0xF;
```
Dragorn421, line 180: This can easily be confused with the `ActorType` enum, maybe just `type`?

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +41,573 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+            puzzleState = ((BgPoEvent*)this->dyna.actor.child)->actorIndex;
+        }
+    } else if (this->dyna.actor.parent != NULL) {
+        puzzleState = ((BgPoEvent*)this->dyna.actor.parent)->actorIndex;
+    } else if (this->dyna.actor.child != NULL) {
+        puzzleState = ((BgPoEvent*)this->dyna.actor.child)->actorIndex;
+    } else {
+        return 0;
+    }
+    return 1;
```
Dragorn421, line 469:
```suggestion
        return false;
    }
    return true;
```

Dragorn421 (changes requested):

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +47,573 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+    this->timer--;
+    sp54.x = (Math_Sins(this->dyna.actor.shape.rot.y) * 5.0f) + this->dyna.actor.posRot.pos.x;
+    sp54.y = Math_Rand_CenteredFloat(66.0f) + this->dyna.actor.posRot.pos.y;
+    sp54.z = Math_Rand_CenteredFloat(50.0f) + this->dyna.actor.posRot.pos.z;
+    if (this->timer >= 0) {
+        if (this->type == 2) {
+            func_8002A6B8(globalCtx, &sp54, &sZeroVec, &sZeroVec, 0x64, 0, 0xFF, 0xFF, 0x96, 0xAA, 0xFF, 0x00, 0x00, 1, 9,
+                          1);
+        } else {
+            func_8002A6B8(globalCtx, &sp54, &sZeroVec, &sZeroVec, 0x64, 0, 0xC8, 0xFF, 0xFF, 0xAA, 0x32, 0x64, 0xFF, 1, 9,
```
fig02, line 559: i can take care of this on my end if you want, but here is the prototype for this effect

`
void EffectSsDeadDb_Spawn(GlobalContext* globalCtx, Vec3f* pos, Vec3f* velocity, Vec3f* accel, s16 scale, s16 scaleStep, s16 primR, s16 primG, s16 primB, s16 primA, s16 envR, s16 envG, s16 envB, s16 unused, s32 arg14, s16 playSound)
`
sp54 is the position for the effect, and the 1 at the end is a boolean for if it should play a burning disintegrating sound. so colors/scale/scaleStep as decimal and that last thing as true would help

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +47,573 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+    this->actionFunc(this, globalCtx);
+    if ((this->actionFunc == BgPoEvent_AmyWait) || (this->actionFunc == BgPoEvent_PaintingPresent)) {
+        CollisionCheck_SetAC(globalCtx, &globalCtx->colChkCtx, &this->collider.base);
+    }
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Po_Event/func_808A780C.s")
+void BgPoEvent_Draw(Actor* thisx, GlobalContext* globalCtx) {
+    static Gfx* sDisplayLists[] = { D_060075A0, D_060079E0, D_06006830, D_06006D60, D_06007230 };
+    s32 pad;
```
fig02, line 583: s

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +47,573 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+    BgPoEvent* this = THIS;
+    u8 opacity;
```
fig02, line 585: we refer to this is alpha everywhere else so we should keep that consistent

fig02 (commented):
nice job, a few things

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +47,573 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+void BgPoEvent_Init(Actor* thisx, GlobalContext* globalCtx) {
+    static InitChainEntry sInitChain[] = {
+        ICHAIN_VEC3F_DIV1000(scale, 1000, ICHAIN_STOP),
+    };
+    s32 pad;
+    BgPoEvent* this = THIS;
+    
+    Actor_ProcessInitChain(&this->dyna.actor, sInitChain);
+    this->type = (thisx->params >> 8) & 0xF;
+    this->index = (thisx->params >> 0xC) & 0xF;
```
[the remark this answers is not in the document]
petrie911: it is.

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +47,573 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+            sp9C[i2].z = this->dyna.actor.initPosRot.pos.z + (coss * vtxVec->z) - (vtxVec->x * sins);
+        }
+        func_800627A0(&this->collider, i1, &sp9C[0], &sp9C[1], &sp9C[2]);
+    }
+    if ((this->type != 4) && (this->index != 2)) {
+        phi_t2 = (this->type == 2) ? this->index : this->index + 2;
+        newPainting = Actor_SpawnAsChild(
+            &globalCtx->actorCtx, &this->dyna.actor, globalCtx, ACTOR_BG_PO_EVENT, sPaintingsX[phi_t2],
+            sPaintingsY[this->index], sPaintingsZ[this->index], 0, this->dyna.actor.shape.rot.y + 0x8000, 0,
+            ((this->index + 1) << 0xC) + (this->type << 8) + this->dyna.actor.params);
```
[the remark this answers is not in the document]
petrie911: I seem to have run it on the whole repo inadvertently. this will add some changed files.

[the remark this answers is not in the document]
fig02: thats fine, I encourage it cause people forget to run it all the time

fig02 (approved):

fig02:
still waiting on a roman review before a merge, but just a heads up that theres an undefined syms conflict

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +47,574 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+    },
+};
+
+static ColliderTrisInit sTrisInit = {
+    { COLTYPE_UNK10, 0x00, 0x09, 0x00, 0x20, COLSHAPE_TRIS },
+    2,
+    sTrisItemsInit,
+};
+
+static u8 blockPosAtRest = 0;
```
Dragorn421, line 68:
```suggestion
static u8 sBlockPosAtRest = 0;
```

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +47,574 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+            } else {
+                func_8002DF54(globalCtx, &PLAYER->actor, 7);
+            }
+        }
+        this->direction = 0;
+        this->actionFunc = BgPoEvent_BlockStop;
+    }
+}
+
+void BgPoEvent_BlockStop(BgPoEvent* this, GlobalContext* globalCtx) {
```
Dragorn421, line 319: I guess "stop" stands for "stop falling" but I don't think it's the best name, the action is the main one where the player gets to push the blocks around (BlockPush), managing to solve it (BlockSolved) or not (BlockReset?)
so this is where I'd call this BlockMain but maybe BlockWaitInteract works too?

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +47,574 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+        EffectSsDeadDb_Spawn(globalCtx, &pos, &sZeroVec, &sZeroVec, 170, 0, 200, 255, 100, 170, 0, 255, 0, 1, 9, true);
+    } else if (sPuzzleState == 0x20) {
+        Actor_Kill(&this->dyna.actor);
+    } else {
+        DECR(this->timer);
+    }
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Po_Event/func_808A6BC8.s")
+s32 BgPoEvent_NextPicture(BgPoEvent* this) {
```
Dragorn421, line 462: just like BlockCheck isnt an action but has the Block prefix, should this be `PaintingNext`?

Dragorn421 (changes requested):

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +47,574 @@ const ActorInit Bg_Po_Event_InitVars = {
...
+};
+
+void BgPoEvent_Init(Actor* thisx, GlobalContext* globalCtx) {
+    s32 pad;
+    BgPoEvent* this = THIS;
+
+    Actor_ProcessInitChain(thisx, sInitChain);
+    this->type = (thisx->params >> 8) & 0xF;
+    this->index = (thisx->params >> 0xC) & 0xF;
+    this->dyna.actor.params &= 0x3F;
```
Roman971, line 189:
```suggestion
    thisx->params &= 0x3F;
```

src/overlays/actors/ovl_Bg_Po_Event/z_bg_po_event.c
```diff
@@ -21,47 +47,574 @@ const ActorInit Bg_Po_Event_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Po_Event/func_808A780C.s")
+void BgPoEvent_Draw(Actor* thisx, GlobalContext* globalCtx) {
+    static Gfx* displayLists[] = { D_060075A0, D_060079E0, D_06006830, D_06006D60, D_06007230 };
+    s32 pad;
+    BgPoEvent* this = THIS;
+    u8 alpha;
+    Vec3f sp58;
+    Vec3f sp4C;
+    f32 sp48;
+    f32 pad44;
```
Roman971, line 590:
```suggestion
    s32 pad2;
```

Roman971 (approved):

PR #369, opened 2020-08-30
audio_seqplayer, the easy half

zelda2774:
Audio_SequenceChannelProcessScript and Audio_SequencePlayerProcessSequence are 1000 asm instructions each and consist of large switches, so are probably going to take forever. So I figure this might be a good place to stop to avoid potential merge conflicts. (This PR touches almost every line of z64audio.h; once it gets merged that file should be far more stable.)


src/code/audio_seqplayer.c
```diff
@@ -1,74 +1,1077 @@
...
+    seqPlayer->muteVolumeScale = 0.5f;
+
+    for (i = 0; i < 0x10; i++) {
+        Audio_SequenceChannelInit(seqPlayer->channels[i]);
+    }
+}
+
+void func_800EC734(s32 player) {
+    SequenceChannel *seqChannel;
+    SequencePlayer *seqPlayer;
```
fig02, line 1022: run the formatter on all files real quick, please and ty 
(itll catch star on the right and move it left)

fig02: also.. this is very obviously a different context so it might not be an issue, but we use the word "player" to refer to link/the player everywhere in this codebase. do you think itd be good to make the separation?

zelda2774: Formatting fixed. Confusion over "player" is a good point; I'll rename player -> seqPlayerIndex in one place. I think seqPlayer is fairly clear however, especially in context.

fig02: we commoly use the abbreviation Idx throughout the code for index, if you wanted to shorten it. Its not a hard requirement, but it might look nice

zelda2774: Surely, done for a bunch of function parameters. (But left the struct members.) Idx is not super commonly used, I'll note:
```
$ rg Index | wc -l
2200
$ rg Idx | wc -l
290
```

fig02: maybe it just feels like that as of late lmao. Maybe something we want to standardize? @Roman971

Roman971: Never been a fan of "idx", mostly because it feels like a variant of "id", which can easily lead to confusion between an id and an index which are often vastly different. We should probably discuss it further tho.

ethteck: I like "idx" just to throw my hat into the pile (or whatever the phrase is), but I'm not against "index" either I guess

Roman971 (approved):

PR #370, opened 2020-08-31
Match more z_fbdemo functions

Thar0:
- `TransitionUnk_InitData`, `TransitionUnk_SetData`, `TransitionUnk_Update` OK, however it seems like there should be array accesses involved but it always causes regalloc issues or additional shifts.
- Progress on `TransitionUnk_InitGraphics`, control flow is correct but there are a lot of small reoderings all over that makes it hard to verify functional equivalence.
- Very small cleanups in `TransitionFade_Update`, nothing major unfortunately.


Roman971 (approved):
I guess if we really need all this pointer arithmetic...

fig02 (approved):

PR #374, opened 2020-09-05
z_bg_menkuri_eye OK

fullgrowngaming:
In this part in Update 
```if (!Flags_GetSwitch(globalCtx, this->actor.params)) {
        if (this->unk_14C != -1) {
            if (this->unk_14C != 0) {
                this->unk_14C -= 1;
            }
            if (this->unk_14C == 0) {
                this->unk_14C = -1;
                D_8089C1A0 -= 1;
            }
        }
    }
```
I tried DECR for the whole thing and comparing it to == 0, but it didn't match. It does match for just the `if (this->unk_14C != 0)` part, but the Discord told me we're not really using macros in that sort of situation.

Also, later on in Update, I tried to fit in a CLAMP_MAX, but I could not get it to match.


src/overlays/actors/ovl_Bg_Menkuri_Eye/z_bg_menkuri_eye.c
```diff
@@ -27,11 +26,113 @@ const ActorInit Bg_Menkuri_Eye_InitVars = {
...
+        temp = ABS((s16)(this->collider.base.ac->posRot.rot.y - this->actor.shape.rot.y));
+        if (temp > 0x5000) {
+            this->collider.base.acFlags &= ~0x2;
+            if (this->unk_14C == -1) {
+                Audio_PlayActorSound2(&this->actor, NA_SE_EN_AMOS_DAMAGE);
+                D_8089C1A0 += 1;
+                if (D_8089C1A0 > 4) {
+                    D_8089C1A0 = 4;
+                } else {
+                    D_8089C1A0 = D_8089C1A0;
```
Dragorn421, line 99: Is this no-effect statement required to match? Is that what CLAMP_MAX wouldn't work for?
Maybe `D_8089C1A0 = D_8089C1A0 > 4 ? 4 : D_8089C1A0` would work?

src/overlays/actors/ovl_Bg_Menkuri_Eye/z_bg_menkuri_eye.c
```diff
@@ -27,11 +26,113 @@ const ActorInit Bg_Menkuri_Eye_InitVars = {
...
+                }
+            }
+            this->unk_14C = 0x1A0;
```
Dragorn421, line 102: `unk_14C` looks like a timer so maybe frame amount in base 10 (decimal) makes more sense?

src/overlays/actors/ovl_Bg_Menkuri_Eye/z_bg_menkuri_eye.c
```diff
@@ -27,11 +26,113 @@ const ActorInit Bg_Menkuri_Eye_InitVars = {
...
+    func_80093D84(globalCtx->state.gfxCtx);
+    if (Flags_GetSwitch(globalCtx, this->actor.params)) {
+        gDPSetEnvColor(oGfxCtx->polyXlu.p++, 0xC8, 0x00, 0x00, 0xFF);
+    } else {
+        if (this->unk_14C == -1) {
+            gDPSetEnvColor(oGfxCtx->polyXlu.p++, 0xC8, 0x00, 0x00, 0x00);
+        } else {
+            gDPSetEnvColor(oGfxCtx->polyXlu.p++, 0xC8, 0x00, 0x00, 0xFF);
+        }
+    }
```
Dragorn421, line 129: Can you merge this if chain? Like `if {} else if {} else {}`

src/overlays/actors/ovl_Bg_Menkuri_Eye/z_bg_menkuri_eye.h
```diff
@@ -8,9 +8,9 @@ struct BgMenkuriEye;
 
 typedef struct BgMenkuriEye {
     /* 0x0000 */ Actor actor;
-    /* 0x014C */ char unk_14C[0x4];
-    /* 0x0150 */ Collider collider;
-    /* 0x0168 */ char unk_168[0x48];
+    /* 0x014C */ s16 unk_14C;
+    /* 0x0150 */ ColliderJntSph collider;
+    /* 0x0170 */ ColliderJntSphItem colliderItem;
```
Dragorn421, line 13: May want to make `colliderItem` an array of length 1 (and rename it to `colliderItems`), even with length 1 it's handled as an array (if you do that the `&` prefixing `colliderItem` in `Collider_SetJntSph` in Init can be removed)

src/overlays/actors/ovl_Bg_Menkuri_Eye/z_bg_menkuri_eye.h
```diff
@@ -8,9 +8,9 @@ struct BgMenkuriEye;
 
 typedef struct BgMenkuriEye {
     /* 0x0000 */ Actor actor;
-    /* 0x014C */ char unk_14C[0x4];
-    /* 0x0150 */ Collider collider;
-    /* 0x0168 */ char unk_168[0x48];
+    /* 0x014C */ s16 unk_14C;
```
Dragorn421, line 11: I think `unk_14C` is something like `framesUntilDisable`, it gets set to some positive value when `D_8089C1A0` (which I assume is the amount of eyes "enabled") is incremented, ticks down, and decrements `D_8089C1A0` when it reaches 0 again

src/overlays/actors/ovl_Bg_Menkuri_Eye/z_bg_menkuri_eye.c
```diff
@@ -27,11 +26,113 @@ const ActorInit Bg_Menkuri_Eye_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Menkuri_Eye/BgMenkuriEye_Update.s")
+    OPEN_DISPS(globalCtx->state.gfxCtx, "../z_bg_menkuri_eye.c", 292);
+    func_80093D84(globalCtx->state.gfxCtx);
+    if (Flags_GetSwitch(globalCtx, this->actor.params)) {
+        gDPSetEnvColor(oGfxCtx->polyXlu.p++, 0xC8, 0x00, 0x00, 0xFF);
+    } else {
+        if (this->unk_14C == -1) {
+            gDPSetEnvColor(oGfxCtx->polyXlu.p++, 0xC8, 0x00, 0x00, 0x00);
+        } else {
+            gDPSetEnvColor(oGfxCtx->polyXlu.p++, 0xC8, 0x00, 0x00, 0xFF);
```
Dragorn421, line 127: We use decimal for colors and alpha (eg 255 instead of 0xFF)

Dragorn421 (changes requested):
Good job 👍

src/overlays/actors/ovl_Bg_Menkuri_Eye/z_bg_menkuri_eye.c
fullgrowngaming: That worked, and actually, somehow I got CLAMP_MAX to work this time

src/overlays/actors/ovl_Bg_Menkuri_Eye/z_bg_menkuri_eye.c
```diff
@@ -27,11 +26,113 @@ const ActorInit Bg_Menkuri_Eye_InitVars = {
...
+            }
+            if (this->unk_14C == 0) {
+                this->unk_14C = -1;
+                D_8089C1A0 -= 1;
+            }
+        }
+    }
+    if (this->collider.base.acFlags & 2) {
+        temp = ABS((s16)(this->collider.base.ac->posRot.rot.y - this->actor.shape.rot.y));
+        if (temp > 0x5000) {
```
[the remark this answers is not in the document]
fullgrowngaming: (turns out the temp wasn't required oops)

ethteck (approved):

Dragorn421 (approved):

Roman971 (approved):

fig02 (approved):

PR #381, opened 2020-09-09
audio_heap: 1 non-matching

zelda2774:


src/code/audio_heap.c
```diff
@@ -1,106 +1,1388 @@
 #include <ultra64.h>
 #include <global.h>
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/audio_heap/func_800DDE20.s")
+#define ALIGN16(val) (((val) + 0xF) & ~0xF)
```
Dragorn421, line 4: this is already defined in `macros.h` (the exact same way)

Dragorn421 (commented):
I'm reading this having no clue how audio works in OoT nor in general, and it's preventing me from appreciating the large amount of work this is
If you have any reference for me and others to read I would appreciate it a lot

zelda2774:
> If you have any reference for me and others to read I would appreciate it a lot

I've mainly been basing this off https://github.com/n64decomp/sm64/blob/master/src/audio/heap.c (and https://github.com/n64decomp/sm64/blob/master/src/audio/internal.h for structs) -- many functions have been direct copy-paste. (The naming there isn't great, but staying close to sm64 is useful for now.) There's no good overview that I know of unfortunately, but it mainly revolves around managing m64 format players/channels/layers (seqplayer.c; m64 being a MIDI-like Turing Compete language, see https://github.com/sauraen/seq64, https://en.wikipedia.org/wiki/Music_tracker, https://hackmd.io/opEB-OmxRa26P8h8pA-x7w), cached audio data and DMA:ing stuff from memory (load.c), managing memory for m64/instrument data/notes (heap.c), notes (playback.c), special effects (effects.c), and synthesizing actual sound (synthesis.c). This subsystem runs on a separate thread and the game interfaces to it via other files that I know nothing about.

Makefile
```diff
@@ -207,8 +207,8 @@ build/src/overlays/%.o: src/overlays/%.c
 	@$(OBJDUMP) -d $@ > $(@:.o=.s)
 
 build/src/%.o: src/%.c
-	$(CC) -c $(CFLAGS) $(MIPS_VERSION) $(OPTFLAGS) -o $@ $^
 	$(CC_CHECK) $^
+	$(CC) -c $(CFLAGS) $(MIPS_VERSION) $(OPTFLAGS) -o $@ $^
```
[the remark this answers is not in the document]
zelda2774: Oops, will revert. The reason for changing is that GCC produces nicer error messages in case of compile failures.

include/z64audio.h
```diff
@@ -90,23 +90,67 @@ typedef struct {
...
-    /*?0x01 */ u8 loaded;
+    /* 0x00 */ u8 unk_bits2 : 2;
+    /* 0x01 */ u32 bits24 : 24;
     /*?0x04 */ u8* sampleAddr;
     /* 0x08 */ AdpcmLoop* loop;
     /*?0x0C */ AdpcmBook* book;
-    /*?0x10 */ u32 sampleSize; // never read. either 0 or 1 mod 9, depending on padding
-} AudioBankSample;
+    /*?0x01 */ // u8 loaded;
+    /*?0x10 */ // u32 sampleSize; // never read. either 0 or 1 mod 9, depending on padding
```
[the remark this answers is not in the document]
zelda2774: probably not, I'll remove them. `book` is still `?`; I guess I can de-`?` sampleAddr (it's at least known to be a pointer)

isghj5:
`UnkInstrument* Audio_GetUnkInstrument()`

I think this is sfx, the sequence format OOT/MM uses has a list of instruments/drums and sfx, but the sfx list is almost never used. Seq64 shows a list of sfx in the audiobank format. 

From seq64's wiki "Banks contain three different kinds of sounds: instruments, drums, and SFX. (I have not yet found any games which use the SFX category [in-game sound effects are usually implemented as instruments in the zeroth bank], but the bank format supports it so SEQ64 does as well.)"

zelda2774:
> UnkInstrument* Audio_GetUnkInstrument()
>
> I think this is sfx, the audiobank format OOT/MM uses has a list of instruments/drums and sfx, but the sfx list is almost never used. Seq64 shows a list of sfx in the audiobank tab.

Good catch! This sounds very believable; I've made the change.

src/code/audio_heap.c
```diff
@@ -1,106 +1,1386 @@
...
+    s32 i;
+    f32 tmp[16];
+
+    tmp[0] = (f32)(arg1 * 262159.0f);
+    tmp[8] = (f32)(arg0 * 262159.0f);
+    tmp[1] = (f32)((arg1 * arg0) * 262159.0f);
+    tmp[9] = (f32)(((arg0 * arg0) + arg1) * 262159.0f);
+
+    for (i = 2; i < 8; i++) {
+        //! @bug they probably meant to store the value to tmp[i] and tmp[8 + i]
```
fig02, line 614: code looks good to me, only thing i would change is we try to stay away from referring to the devs directly. would phrase it as like "value was probably meant to be stored to tmp[i] and tmp[8 + i]"

fig02 (approved):

src/code/audio_heap.c
```diff
@@ -1,106 +1,1386 @@
...
+        return NULL;
+    }
+    gAudioContext.unk_2D60[sp18].poolIndex = poolIdx;
+    gAudioContext.unk_2D60[sp18].id = id;
+    gAudioContext.unk_2D60[sp18].size = size;
+    // @bug missing return? "ret" is in v0, but doing an explicit return
+    // uses an additional register.
+#ifdef AVOID_UB
+    return ret;
+#endif
```
Roman971, line 1016: We don't have any notion of AVOID_UB in oot (for now at least) so this should probably be omitted?

fig02 (approved):

Roman971 (approved):

PR #383, opened 2020-09-11
ovl_En_Part

maekclena:


src/overlays/actors/ovl_En_Part/z_en_part.c
```diff
@@ -21,23 +32,305 @@ const ActorInit En_Part_InitVars = {
...
+static Vec3f D_80ACF1E0 = { 0.0f, -1.5f, 0.0f };
+static Vec3f D_80ACF1EC = { 0.0f, 0.0f, 0.0f };
+static EnPartActionFunc sActionFuncs[] = { func_80ACDDE8, func_80ACE13C, func_80ACE5B8, func_80ACE5C8, func_80ACE7E8 };
+
+extern UNK_PTR D_06001300[];
+extern UNK_PTR D_06001700[];
+extern UNK_PTR D_06001900[];
+extern UNK_PTR D_06001B00[];
+extern UNK_PTR D_06001F00[];
+extern UNK_PTR D_06002100[];
```
Dragorn421, line 49: Given their usage I would think those are textures (whicharrays of UNK_TYPE is better for I think) but I have no idea what object this ends up using so I can't check

(this actor must be the most janky for its object management, which comes from the spawn z rotation in a hardcoded if case in Actor_Spawn)

src/overlays/actors/ovl_En_Part/z_en_part.c
```diff
@@ -21,23 +32,305 @@ const ActorInit En_Part_InitVars = {
...
+
+        if ((this->actor.bgCheckFlags & 1) || (this->actor.posRot.pos.y <= this->actor.groundY)) {
+            this->action = 4;
+            this->actor.speedXZ = 0.0f;
+            this->actor.gravity = 0.0f;
+            this->actor.velocity.y = 0.0f;
+        }
+
+        if (this->actor.params == 13) {
+            if ((this->actor.parent != NULL) && (this->actor.parent->update == NULL)) {
```
Dragorn421, line 140: what about `&&`ing these?

src/overlays/actors/ovl_En_Part/z_en_part.c
```diff
@@ -21,23 +32,305 @@ const ActorInit En_Part_InitVars = {
...
+    this->action = 3;
+}
+
+void func_80ACE5C8(EnPart* this, GlobalContext* globalCtx) {
+    Player* player = PLAYER;
+    Vec3f velocity;
+    Vec3f accel;
+    u8 invincibilityTimer;
+
+    this->delay -= 1;
```
Dragorn421, line 199: `this->delay--;` ?

src/overlays/actors/ovl_En_Part/z_en_part.c
```diff
@@ -21,23 +32,305 @@ const ActorInit En_Part_InitVars = {
...
+        return;
+    }
+
+    velocity = D_80ACF1D4;
+    accel = D_80ACF1E0;
+
+    if (sqrt(this->actor.xyzDistFromLinkSq) <= 40.0f) {
+        invincibilityTimer = player->invincibilityTimer;
+        if (player->invincibilityTimer <= 0) {
+            if (player->invincibilityTimer < -39) {
```
Dragorn421, line 211: What about `<= -40` instead of `< -39`

Dragorn421 (changes requested):
well what this actor does is very mysterious to me, looks like an effect but it's not?

maekclena:
This actor handles colored flames when enemies die, stalfos' bones flying and probably other similar death effects.

src/code/z_actor.c
```diff
@@ -3120,7 +3120,7 @@ s32 func_8003305C(Actor* actor, struct_80032E24* arg1, GlobalContext* globalCtx,
                                                     mtx->wy, mtx->wz, 0, 0, objBankIndex, params);
         if (spawnedEnPart != NULL) {
             func_800D20CC(&arg1->unk_00[arg1->unk_08], &spawnedEnPart->actor.shape.rot, 0);
-            spawnedEnPart->unk_150 = arg1->unk_0C[arg1->unk_08];
+            spawnedEnPart->displayList = arg1->unk_0C[arg1->unk_08];
```
Dragorn421, line 3123: You can also change `arg1->unk_0C` to be of type `Gfx**`
(in `struct_80032E24`)

src/overlays/actors/ovl_En_Part/z_en_part.c
```diff
@@ -21,23 +26,307 @@ const ActorInit En_Part_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Part/func_80ACE13C.s")
+void EnPart_Draw(Actor* thisx, GlobalContext* globalCtx) {
+    EnPart* this = THIS;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Part/func_80ACE5B8.s")
+    OPEN_DISPS(globalCtx->state.gfxCtx, "../z_en_part.c", 647);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Part/func_80ACE5C8.s")
+    if (thisx->params > 0) {
+        Matrix_RotateZ(this->rotation, MTXMODE_APPLY);
```
Dragorn421, line 297: based on this, I would rename `rotation` to `rotationZ` or `rotZ` for shorts

src/overlays/actors/ovl_En_Part/z_en_part.c
```diff
@@ -21,23 +26,307 @@ const ActorInit En_Part_InitVars = {
...
+    func_80029724(globalCtx, &this->actor.posRot.pos, &velocity, &accel, 20,
+                  (s32)((Math_Rand_ZeroOne() * 5.0f + 12.0f) * 2), -1, 10, 0);
+    Audio_PlayActorSound2(&this->actor, NA_SE_EN_MONBLIN_GNDWAVE - SFX_FLAG);
+}
+
+void func_80ACE7E8(EnPart* this, GlobalContext* globalCtx) {
+    static Vec3f D_80ACF1EC = { 0.0f, 0.0f, 0.0f };
+
+    Vec3f zero = D_80ACF1EC;
+    f32 dist;
```
Dragorn421, line 226: I would rename `dist` to `diffsSum` or something like that because `Math_SmoothScaleMaxMinF` returns the absolute difference between the computed value and the target, and `dist` is summing position differences with a rotation difference

src/code/z_actor.c
```diff
@@ -4081,11 +4081,11 @@ EnPart* func_800358DC(Actor* actor, Vec3f* spawnPos, Vec3s* spawnRot, Vec3f* arg
...
-        spawnedEnPart->unk_150 = arg8;
-        spawnedEnPart->unk_14C = 2;
-        spawnedEnPart->unk_14E = arg4;
-        spawnedEnPart->unk_154 = arg3->y;
-        spawnedEnPart->unk_158 = arg3->z;
+        spawnedEnPart->displayList = arg8;
+        spawnedEnPart->action = 2;
+        spawnedEnPart->delay = arg4;
+        spawnedEnPart->rotation = arg3->y;
+        spawnedEnPart->rotationSpeed = arg3->z;
```
Dragorn421, line 4088: may be neat to create a struct for `arg3`, or just make it a `f32` array, I don't think `Vec3f` is appropriate now that we know what it's used for

You can also update `func_800358DC`'s prototype (types and names of arguments) according to your findings

src/overlays/actors/ovl_En_Part/z_en_part.h
```diff
@@ -6,13 +6,15 @@
...
+
 typedef struct EnPart {
     /* 0x000 */ Actor actor;
-    /* 0x14C */ u8 unk_14C;
-    /* 0x14E */ s16 unk_14E;
-    /* 0x150 */ s32 unk_150;
-    /* 0x154 */ f32 unk_154;
-    /* 0x158 */ f32 unk_158;
+    /* 0x14C */ u8 action;
+    /* 0x14E */ s16 delay;
```
Dragorn421, line 14: I think this actor uses `delay` like "when `delay` reaches 0, stop doing things"
I'm not sure I like "delay" because that often implies "wait to do something" rather than "do something until"

For one of the actions (4, `func_80ACE7E8`) "stop doing things" doesn't mean "remove the actor" like for the others (it still draws), so `lifetime` isn't a good name
I'm not very inspired at the moment

Dragorn421 (changes requested):

src/overlays/actors/ovl_En_Part/z_en_part.h
maekclena: From my tests, delay was used as "wait until spawning effect". See also `func_80ACE13C` where all the function calls are effects.

Dragorn421: But doesn't `EnPart_Draw` get called meanwhile? It draws `this->displayList`

maekclena: I tested and it seems, at least for stalchild, that `EnPart_Draw` draws the bones. Since `delay` only matter for the flames, a good name might be `effectDelay`.

Dragorn421 (approved):

src/overlays/actors/ovl_En_Part/z_en_part.c
```diff
@@ -21,23 +26,307 @@ const ActorInit En_Part_InitVars = {
...
+void func_80ACE13C(EnPart* this, GlobalContext* globalCtx) {
+    static Vec3f D_80ACF1B0 = { 0.0f, 0.0f, 0.0f };
+    static Vec3f D_80ACF1BC = { 0.0f, 0.0f, 0.0f };
+    static Vec3f D_80ACF1C8 = { 0.0f, 0.0f, 0.0f };
+
+    s32 i;
+    Vec3f pos;
+    Vec3f velocity = D_80ACF1B0;
+    Vec3f accel = D_80ACF1BC;
+    Vec3f zero = D_80ACF1C8;
```
Roman971, line 117:
```suggestion
    s32 i;
    Vec3f pos;
    Vec3f velocity = { 0.0f, 0.0f, 0.0f };
    Vec3f accel = { 0.0f, 0.0f, 0.0f };
    Vec3f zero = { 0.0f, 0.0f, 0.0f };
```

src/overlays/actors/ovl_En_Part/z_en_part.c
```diff
@@ -21,23 +26,307 @@ const ActorInit En_Part_InitVars = {
...
+    velocity.x = Math_Rand_CenteredFloat(16.0f);
+    func_80029724(globalCtx, &this->actor.posRot.pos, &velocity, &accel, 20,
+                  (s32)((Math_Rand_ZeroOne() * 5.0f + 12.0f) * 2), -1, 10, 0);
+    Audio_PlayActorSound2(&this->actor, NA_SE_EN_MONBLIN_GNDWAVE - SFX_FLAG);
+}
+
+void func_80ACE7E8(EnPart* this, GlobalContext* globalCtx) {
+    static Vec3f D_80ACF1EC = { 0.0f, 0.0f, 0.0f };
+
+    Vec3f zero = D_80ACF1EC;
```
Roman971, line 225:
```suggestion
    Vec3f zero = { 0.0f, 0.0f, 0.0f };;
```

src/overlays/actors/ovl_En_Part/z_en_part.c
```diff
@@ -21,23 +26,307 @@ const ActorInit En_Part_InitVars = {
...
+    u8 invincibilityTimer;
+
+    this->effectDelay--;
+    if (this->effectDelay == 0) {
+        Actor_Kill(&this->actor);
+        return;
+    }
+
+    velocity = D_80ACF1D4;
+    accel = D_80ACF1E0;
```
Roman971, line 197: If you use an else here (or just a new block), you can inline the vector definitions so you don't need to copy static variables (see my previous comments)

src/overlays/actors/ovl_En_Part/z_en_part.c
```diff
@@ -21,23 +26,307 @@ const ActorInit En_Part_InitVars = {
...
+    if (this->actor.parent->colChkInfo.health != 0) {
+        Actor_Kill(&this->actor);
+    }
+}
+
+void EnPart_Update(Actor* thisx, GlobalContext* globalCtx) {
+    static EnPartActionFunc sActionFuncs[] = { func_80ACDDE8, func_80ACE13C, func_80ACE5B8, func_80ACE5C8,
+                                               func_80ACE7E8 };
+
+    EnPart* this = THIS;
```
Roman971, line 258: I would personally leave this array of functions outside the file (right above it), but if you want to keep it inside the function I suggest formatting it like this (note the trailing comma):
```suggestion
    static EnPartActionFunc sActionFuncs[] = {
        func_80ACDDE8, func_80ACE13C, func_80ACE5B8, func_80ACE5C8, func_80ACE7E8,
    };
    EnPart* this = THIS;
```

src/overlays/actors/ovl_En_Part/z_en_part.h
```diff
@@ -6,13 +6,15 @@
...
+
 typedef struct EnPart {
     /* 0x000 */ Actor actor;
-    /* 0x14C */ u8 unk_14C;
-    /* 0x14E */ s16 unk_14E;
-    /* 0x150 */ s32 unk_150;
-    /* 0x154 */ f32 unk_154;
-    /* 0x158 */ f32 unk_158;
+    /* 0x14C */ u8 action;
+    /* 0x14E */ s16 effectDelay;
```
Roman971, line 14: I could be wrong but I think this is just the timer for the whole actor
```suggestion
    /* 0x14E */ s16 timer;
```

src/overlays/actors/ovl_En_Part/z_en_part.c
```diff
@@ -21,23 +26,307 @@ const ActorInit En_Part_InitVars = {
...
+    if (this->effectDelay == 0) {
+        Actor_Kill(&this->actor);
+        return;
+    }
+
+    velocity = D_80ACF1D4;
+    accel = D_80ACF1E0;
+
+    if (sqrt(this->actor.xyzDistFromLinkSq) <= 40.0f) {
+        invincibilityTimer = player->invincibilityTimer;
```
Roman971, line 200: Considering the usage, I suggest naming this "prev" or "old"
```suggestion
        prevInvincibilityTimer = player->invincibilityTimer;
```

Roman971 (commented):
Good job with this actor. I just have a few possible improvements

Roman971 (approved):

fig02:
After you merge effects in, theres a good amount of args that will make more sense in decimal. Most of them are named so it will be clear what they should be. Ill give a more in-depth review after that :D

src/overlays/actors/ovl_En_Part/z_en_part.c
```diff
@@ -21,23 +26,299 @@ const ActorInit En_Part_InitVars = {
...
+                break;
+            case 4:
+                for (i = 7; i >= 0; i--) {
+                    pos.x = this->actor.posRot.pos.x + Math_Rand_CenteredFloat(60.0f);
+                    pos.y = this->actor.posRot.pos.y + this->actor.shape.unk_08 * this->actor.scale.y +
+                            Math_Rand_CenteredFloat(50.0f);
+                    pos.z = this->actor.posRot.pos.z + Math_Rand_CenteredFloat(60.0f);
+                    velocity.y = Math_Rand_ZeroOne() + 1.0f;
+                    EffectSsDtBubble_SpawnColorProfile(globalCtx, &pos, &velocity, &accel, Math_Rand_S16Offset(80, 100),
+                                                       25, 0, 1);
```
fig02, line 152: randXZ (the last arg there) is a bool so you can make that `true`
I also prefer to call `zero` `zeroVec` but I leave that up to you

src/overlays/actors/ovl_En_Part/z_en_part.c
```diff
@@ -21,23 +26,299 @@ const ActorInit En_Part_InitVars = {
...
+            }
+        }
+    }
+
+    sActionFuncs[this->action](this, globalCtx);
+}
+
+Gfx* func_80ACEAC0(GraphicsContext* gfxCtx, u8 primR, u8 primG, u8 primB, u8 envR, u8 envG, u8 envB) {
+    Gfx* displayList;
+    Gfx* displayListHead;
```
fig02, line 270: we've started going with dList and dList head as the abbreviations for these, but we havent made the change project wide yet. dont have to change it now if you dont want, but just letting you know

src/overlays/actors/ovl_En_Part/z_en_part.c
```diff
@@ -1,3 +1,9 @@
+/*
+ * File: z_en_part.c
+ * Overlay: ovl_En_Part
+ * Description: Particle emitter for enemies' death
```
fig02, line 4: I prefer the term "effects" over particles, cause some of the effects do stuff like making sounds. and other things that arent really particles.
I would say Effect Spawner instead of Particle Emitter but thats just me :D

fig02 (approved):

PR #385, opened 2020-09-12
Added ability to provide ichaindis a symbol

FryJay:
It will now accept either a symbol or ROM address.  Also gave it ability to look up actor var names.


FryJay:
I refactored the sym_info.py script a bit to split off the function that looks up the ROM address into it's own file that can be imported.  This allowed me to use an import into my script instead of awkwardly running the other script.

FryJay:
Well not "my" script but the one I was working on :P

tools/ichaindis.py
```diff
@@ -19,11 +20,53 @@
     'ICHAIN_VEC3S',
 ]
 
+Z64_ACTOR_PATH = "../include/z64actor.h"
+
+def get_rom_address(offset):
+    # Run the sym_info script and parse out the ROM address
+    stream = os.popen('cd .. && python3 sym_info.py ' + offset)
```
ethteck, line 27: I would prefer if our python scripts called each others' functions directly rather than spawning a new python process like this

tools/ichaindis.py
```diff
@@ -19,11 +20,53 @@
     'ICHAIN_VEC3S',
 ]
 
+Z64_ACTOR_PATH = "../include/z64actor.h"
```
ethteck, line 23: Correct me if I'm wrong, but doesn't this assume running the script from a certain directory? I think we should make paths relative to the script's directory, not the current user directory. See `m2ctx.py` for an example

ethteck (changes requested):
Thanks for this, it's a great improvement.

Roman971:
I strongly disagree with using scripts like `sym_info.py` for this kind of purpose. It's not what it's meant to be used for at all, and it means both scripts become dependent/related to each other, which is bad practice imo. There is also the concern of `sym_info.py` using the build map under the hood, which can very well be wrong or shifted when you run it.

Either way, I thought we already had a python module for this very purpose (`tools/overlayhelps/filemap.py` iirc), used for colliders and such. Assuming it could be used for init chains, I think it would be better to just use that.

tools/ichaindis.py
FryJay: Weird.  I had already made this change but put it in the wrong branch.  I've applied the changes now.  I couldn't convince Python to include a script from the parent folder so I refactored the sym_info.py script to move the ROM lookup function into its own file in the tools folder.

tools/ichaindis.py
FryJay: Yeah it does, I've fixed it now to make it relative to the script's location so if the script ever moves it will need to be updated but it should work even if called from another folder.

FryJay:
I'll take a look at filemap.py.  I had refactored sym_info.py to move the ROM address lookup function to a standalone file in the tools folder but if such a function already exists I'll undo that.

tools/ichaindis.py
ethteck: Thanks! :)

Roman971:
`sym_info.py` (just like `first_diff.py`) is designed to be portable and fully working on its own, so making it dependent on the tools folder would be pretty unfortunate imo.

And to be clear, I agree with the idea behind these changes, I just have concerns about the implementation.

FryJay:
Sounds good.  I'll take a look at alternate approaches and undo the changes to the sym_info script.

FryJay:
I fixed it up to use filemap.py and restored sym_info.py.

tools/ichaindis.py
```diff
@@ -19,11 +21,54 @@
...
+
+def get_rom_address(offset):
+
+    if "D_" in offset:
```
Dragorn421, line 28:
```suggestion
    if offset.startswith("D_"):
```

Dragorn421 (commented):
I would add a `--no-names` option (`parser.add_argument('--no-names', help='Disable using names from the actor struct')`) option just in case something goes wrong once (be it reading the file or using it), as a backup solution

Should the name reading filter out types that aren't basic types (`s16`, `u32`, ...) like `PosRot`?

tools/ichaindis.py
FryJay: Fixed.

FryJay:
I've added the switch you suggested.  I'm not sure I understand what you're saying about types though.  Are you suggesting I add a check to just skip basic types when generating the list of variables from Actor?

Dragorn421:
Would be nice to not compute `actor_variable_names` either when `args.no_names` is set, to bypass the actor members names logic completely (again, if there ever is a failure, as an immediate temporary workaround)

With what I said about basic types, I meant that eg `posRot` as a member doesn't make sense anyway (though I guess it would match) so may as well leave it as unk_24 for the user to figure out
Not a big issue anyway, doesn't look like ichains are used to access the decomp's actor struct's substructs

FryJay:
Made that change too to only generate the names list if needed.

fig02:
its an important distinction to make yeah, cause for example the Vec3f ichain type is expecting the var name for the whole vector and not just the .x component
I dont know if stuff like that was accounted for here or not. It might be a good idea to. But yeah I agree with having the ability to just output unk offsets if the situation ever called for it

FryJay:
It should retrieve the vector's name if it's owned by the Actor struct.  It won't attempt to go any deeper than what is there.  Though I do now realize that the way it's implemented won't handle arrays such as unk_CC[2] and will include the subscript in the name.

ethteck (approved):

Dragorn421 (approved):
Can be improved but my local version with hardcoded names was good enough for me, so  I'll agree there's not much need to waste time improving this further

fig02 (approved):

Roman971 (approved):
The part about getting names from the actor struct will most likely break in various ways in the future when things change and no one remembers to update the script, but I guess when that happens we'll just have to deal with it.

PR #386, opened 2020-09-12
Match and fully document bg_ddan_jd

mzxrules:


spec
```diff
@@ -669,7 +669,6 @@ endseg
 beginseg
     name "ovl_Bg_Ddan_Jd"
     include "build/src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.o"
-    include "build/data/overlays/actors/z_bg_ddan_jd.data.o"
     include "build/data/overlays/actors/z_bg_ddan_jd.reloc.o"
```
Dragorn421, line 672: use the built relocation data since it's matching :)

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -15,11 +15,13 @@ void BgDdanJd_Destroy(Actor* thisx, GlobalContext* globalCtx);
...
 void BgDdanJd_Draw(Actor* thisx, GlobalContext* globalCtx);
 
-void func_80870B88(BgDdanJd* this, GlobalContext* globalCtx);
-void func_80870D2C(BgDdanJd* this, GlobalContext* globalCtx);
-void func_80870F00(BgDdanJd* this, GlobalContext* globalCtx);
+void BgDdanJd_Idle(BgDdanJd* this, GlobalContext* globalCtx);
+void BgDdanJd_MoveEffects(BgDdanJd* this, GlobalContext* globalCtx);
+void BgDdanJd_Move(BgDdanJd* this, GlobalContext* globalCtx);
+
+extern Gfx D_060037B8;
```
Dragorn421, line 22: make this a Gfx array, and drop the subsequently irrelevant `&`s

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -31,17 +33,157 @@ const ActorInit Bg_Ddan_Jd_InitVars = {
...
+    Actor_ProcessInitChain(thisx, &sInitChain);
+    DynaPolyInfo_SetActorMove(this, DPM_PLAYER);
+    DynaPolyInfo_Alloc(&D_06003CE0, &colHeader);
+    this->dyna.dynaPolyId = DynaPolyInfo_RegisterActor(globalCtx, &globalCtx->colCtx.dyna, thisx, colHeader);
+    this->idleTimer = IDLE_FRAMES;
+    this->state = STATE_GO_BOTTOM;
+
+    // @bug: does not check if actor.params < 0x40. This will cause inconsistent behavior if the bound switch flag state
+    // changes when the actor is loaded vs not loaded
+    if (Flags_GetSwitch(globalCtx, this->dyna.actor.params) != 0) {
```
Dragorn421, line 70: I'm not sure I understand the issue here
if it's just that `this->dyna.actor.params` should be `this->dyna.actor.params & 0x3F` I don't think that should be considered a bug, all variables have an implicit format

the `!= 0` can be dropped

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -31,17 +33,157 @@ const ActorInit Bg_Ddan_Jd_InitVars = {
...
+
+// Implements the platform's idle state
+void BgDdanJd_Idle(BgDdanJd* this, GlobalContext* globalCtx) {
+
+    if (this->idleTimer != 0) {
+        this->idleTimer--;
+    }
+
+    // if this is the platform that rises all the way to the top, and the switch state has changed to 1
+    if (this->stepSpeed == 1) {
```
Dragorn421, line 91: `1` is `DEFAULT_STEP_SPEED`
isn't the comment inconsistent with that define's name?

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -31,17 +33,157 @@ const ActorInit Bg_Ddan_Jd_InitVars = {
...
+        if (this->dyna.actor.params < 0x40) {
+            if (Flags_GetSwitch(globalCtx, this->dyna.actor.params)) {
```
Dragorn421, line 93: ` && ` these three?

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -31,17 +33,157 @@ const ActorInit Bg_Ddan_Jd_InitVars = {
...
+    if (this->stepSpeed == 1 && this->dyna.actor.params < 0x40 &&
+        Flags_GetSwitch(globalCtx, this->dyna.actor.params) != 0) {
+        this->stepSpeed = SHORTCUT_STEP_SPEED;
+        this->state = STATE_GO_MIDDLE_FROM_BOTTOM;
+        this->dyna.actor.posRot.pos.y = this->dyna.actor.initPosRot.pos.y + MOVE_HEIGHT_MIDDLE;
+        this->idleTimer = 0;
+        this->actionFunc = BgDdanJd_Idle;
+        func_800800F8(globalCtx, 0xBF4, -0x63, &this->dyna.actor, 0);
+    } else {
+        if (Math_ApproxF(&this->dyna.actor.posRot.pos.y, this->targetHeight, (f32)this->stepSpeed)) {
```
Dragorn421, line 174: `else if`

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -31,17 +33,157 @@ const ActorInit Bg_Ddan_Jd_InitVars = {
...
+            this->actionFunc = BgDdanJd_Idle;
+        }
+    }
+    BgDdanJd_MoveEffects(this, globalCtx);
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Ddan_Jd/BgDdanJd_Update.s")
+void BgDdanJd_Update(Actor* thisx, GlobalContext* globalCtx) {
+    BgDdanJd* this = THIS;
+    this->actionFunc(thisx, globalCtx);
```
Dragorn421, line 184: pass `this` instead of `thisx` if possible?

undefined_syms.txt
```diff
@@ -916,3 +916,7 @@ D_0600018C = 0x0600018C;
 // z_en_dy_extra
 D_0601BFB0 = 0x0601BFB0;
 D_0601C160 = 0x0601C160;
+
+// z_bg_ddan_jd
+D_06003CE0 = 0x06003CE0; //collision
+D_060037B8 = 0x060037B8; //dlist
```
Dragorn421, line 922: pretty sure any comments should go in code since undefined_syms.txt is temporary
besides those comments arent very useful

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -31,17 +33,157 @@ const ActorInit Bg_Ddan_Jd_InitVars = {
...
+
+#define MOVE_HEIGHT_MIDDLE 140.0f
+#define MOVE_HEIGHT_TOP 700.0f
+
+#define IDLE_FRAMES 100
+
+// since stepSpeed also determines whether the platform is the shortcut to the top of the dungeon, these must be
+// different values to work correctly
+#define DEFAULT_STEP_SPEED 1
+#define SHORTCUT_STEP_SPEED 5
```
Dragorn421, line 54: not sold on calling it "shortcut" since it's just a gameplay intent rather than what the actor actually does (rise higher, two levels)

Dragorn421 (changes requested):

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
mzxrules: The bug is a bit complex to explain: 

The implementation of Flags_GetSwitch doesn't prevent a leftshift that is bigger than the wordsize/smaller than 0 from occurring, which I think is  technically undefined behavior in C. On MIPS architecture, the right operand's 5 least significant bits are used to determine how much to left shift by, meaning that there's an implicit mod 32 on the right operand. If you pass in a flag value of 0x20 or higher to Flags_GetSwitch, it will return a temp flag state equal to flag%32. 

Now, to differentiate between the shortcut platform and the normal platform, the game uses the `stepState` variable's state; if stepState is 5, then the actor instance is always treated as the shortcut platform, but if `stepState` is 1, then it checks to see if params < 0x40 to determine if it should become the shortcut platform everywhere else... except here.

So let's say someone decided to spawn an instance where params = 0x0100; this ties the instance to temp switch flag 0x00, and let's say temp switch 0x00 is currently off.  The platform height will alternate from lowest point to middle point.

if you hit a switch or something that changes temp switch 0x00 to on while in the same room, the platform's behavior won't change, but if you leave the room and return, suddenly the init gets run again and now it's going all the way to the top because it doesn't do the range check on params

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
fig02: in this case it should definitely be this, cause action funcs take the casted actor

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
mzxrules: i feel like gameplay intent is fine because the actor is a set piece designed to be used specifically in Dodongo's Cavern

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -15,11 +15,13 @@ void BgDdanJd_Destroy(Actor* thisx, GlobalContext* globalCtx);
 void BgDdanJd_Update(Actor* thisx, GlobalContext* globalCtx);
 void BgDdanJd_Draw(Actor* thisx, GlobalContext* globalCtx);
 
-void func_80870B88(BgDdanJd* this, GlobalContext* globalCtx);
-void func_80870D2C(BgDdanJd* this, GlobalContext* globalCtx);
-void func_80870F00(BgDdanJd* this, GlobalContext* globalCtx);
+void BgDdanJd_Idle(BgDdanJd* this, GlobalContext* globalCtx);
+void BgDdanJd_MoveEffects(BgDdanJd* this, GlobalContext* globalCtx);
+void BgDdanJd_Move(BgDdanJd* this, GlobalContext* globalCtx);
```
Dragorn421, line 20:
```suggestion
void BgDdanJd_Idle(BgDdanJd* this, GlobalContext* globalCtx);
void BgDdanJd_Move(BgDdanJd* this, GlobalContext* globalCtx);
```
unneeded, defined before used

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -31,17 +33,152 @@ const ActorInit Bg_Ddan_Jd_InitVars = {
...
+    DynaPolyInfo_SetActorMove(thisx, DPM_PLAYER);
+    DynaPolyInfo_Alloc(&D_06003CE0, &colHeader);
+    this->dyna.dynaPolyId = DynaPolyInfo_RegisterActor(globalCtx, &globalCtx->colCtx.dyna, thisx, colHeader);
+    this->idleTimer = IDLE_FRAMES;
+    this->state = STATE_GO_BOTTOM;
+
+    // @bug: Missing check for actor.params < 0x40. This will cause inconsistent behavior if params >= 0x40 and the
+    // bound switch state is turned on while in the same room, as the shortcut behavior won't become enabled until the
+    // actor is reloaded.
+    if (Flags_GetSwitch(globalCtx, this->dyna.actor.params)) {
```
Dragorn421, line 71: this "bug" is also present in other actors without the `@bug` mention

I really dont think lack of input sanitization should be considered a bug when the input is static game data
especially with the consequences not being particularly harmful.

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -31,17 +33,152 @@ const ActorInit Bg_Ddan_Jd_InitVars = {
...
+        this->stepSpeed = SHORTCUT_STEP_SPEED;
+    } else {
+        this->stepSpeed = DEFAULT_STEP_SPEED;
+    }
+    this->actionFunc = BgDdanJd_Idle;
+}
+
+void BgDdanJd_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+    BgDdanJd* this = THIS;
+    DynaPolyInfo_Free(globalCtx, &globalCtx->colCtx.dyna, this->dyna.dynaPolyId);
```
Dragorn421, line 81:
```suggestion
    BgDdanJd* this = THIS;
    
    DynaPolyInfo_Free(globalCtx, &globalCtx->colCtx.dyna, this->dyna.dynaPolyId);
```

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -31,17 +33,152 @@ const ActorInit Bg_Ddan_Jd_InitVars = {
...
+}
+
+// Implements the platform's idle state
```
Dragorn421, line 84: not a very useful comment

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -31,17 +33,152 @@ const ActorInit Bg_Ddan_Jd_InitVars = {
...
+void BgDdanJd_Idle(BgDdanJd* this, GlobalContext* globalCtx) {
+
+    if (this->idleTimer != 0) {
+        this->idleTimer--;
+    }
+
+    // if this is the platform that rises all the way to the top, and the switch state has turned on
+    if (this->stepSpeed == DEFAULT_STEP_SPEED && this->dyna.actor.params < 0x40 &&
```
Dragorn421, line 92: `this is the platform that rises all the way to the top` vs `stepSpeed == DEFAULT_STEP_SPEED` is inconsistent

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -31,17 +33,152 @@ const ActorInit Bg_Ddan_Jd_InitVars = {
...
+}
+
+// Handles dust particles and sfx when moving
+void BgDdanJd_MoveEffects(BgDdanJd* this, GlobalContext* globalCtx) {
+    Vec3f dustPos;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Ddan_Jd/func_80870B88.s")
+    // Generate random dust particles at the platform's base.
+    dustPos.y = this->dyna.actor.initPosRot.pos.y;
+    if ((globalCtx->gameplayFrames & 1) != 0) {
```
Dragorn421, line 138:
```suggestion
    if (globalCtx->gameplayFrames & 1) {
```

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -31,17 +33,152 @@ const ActorInit Bg_Ddan_Jd_InitVars = {
...
+        this->state = STATE_GO_MIDDLE_FROM_BOTTOM;
+        this->idleTimer = 0;
+        this->dyna.actor.posRot.pos.y = this->dyna.actor.initPosRot.pos.y + MOVE_HEIGHT_MIDDLE;
+        func_800800F8(globalCtx, 0xBF4, -0x63, &this->dyna.actor, 0);
+    }
+    if (this->idleTimer == 0) {
+        this->idleTimer = IDLE_FRAMES;
+        if (this->state == STATE_GO_BOTTOM) {
+            this->state = STATE_GO_MIDDLE_FROM_BOTTOM;
+            this->targetHeight = this->dyna.actor.initPosRot.pos.y + MOVE_HEIGHT_MIDDLE;
```
Dragorn421, line 104: I would rename `targetHeight` to `targetY` since "height" in `MOVE_HEIGHT_MIDDLE` is relative to the initial Y, while targetHeight is not
(height == relative, Y == absolute)

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -31,17 +33,152 @@ const ActorInit Bg_Ddan_Jd_InitVars = {
...
+        Audio_PlayActorSound2(&this->dyna.actor, NA_SE_EV_PILLAR_MOVE_STOP);
+        this->actionFunc = BgDdanJd_Idle;
+    }
+    BgDdanJd_MoveEffects(this, globalCtx);
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Ddan_Jd/BgDdanJd_Update.s")
+void BgDdanJd_Update(Actor* thisx, GlobalContext* globalCtx) {
+    BgDdanJd* this = THIS;
+    this->actionFunc(this, globalCtx);
```
Dragorn421, line 179:
```suggestion
    BgDdanJd* this = THIS;
    
    this->actionFunc(this, globalCtx);
```

Dragorn421 (changes requested):

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
mzxrules: `this->dyna.actor.params < 0x40` tests if `this` is the platform that will rise to the highest level when the switch state is set to `on`.
`this->stepSpeed == DEFAULT_STEP_SPEED && Flags_GetSwitch()` combined tests if the switch state has been changed to `on` since the last time it was checked. This is because on init the platform's stepSpeed will be set to DEFAULT_STEP_SPEED if the switch is in it's `off` state.

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
mzxrules: i'll double check, but i think this change causes a non-matching

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -31,17 +32,150 @@ const ActorInit Bg_Ddan_Jd_InitVars = {
...
+        func_800800F8(globalCtx, 0xBF4, -0x63, &this->dyna.actor, 0);
+    }
+    if (this->idleTimer == 0) {
+        this->idleTimer = IDLE_FRAMES;
+        if (this->state == STATE_GO_BOTTOM) {
+            this->state = STATE_GO_MIDDLE_FROM_BOTTOM;
+            this->targetY = this->dyna.actor.initPosRot.pos.y + MOVE_HEIGHT_MIDDLE;
+        } else if (this->state == STATE_GO_MIDDLE_FROM_BOTTOM) {
+            // If platform goes all the way to the top
+            if (this->ySpeed != DEFAULT_Y_SPEED) {
```
Dragorn421, line 105: Sorry to insist but I don't see how this makes sense

```c
    // if this is the platform that rises all the way to the top, and the switch state has just changed to on
    if (this->ySpeed == DEFAULT_Y_SPEED
```
and later
```c
            // If platform goes all the way to the top
            if (this->ySpeed != DEFAULT_Y_SPEED) {
```

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -31,17 +32,150 @@ const ActorInit Bg_Ddan_Jd_InitVars = {
...
+void BgDdanJd_MoveEffects(BgDdanJd* this, GlobalContext* globalCtx) {
+    Vec3f dustPos;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Ddan_Jd/func_80870B88.s")
+    // Generate random dust particles at the platform's base.
+    dustPos.y = this->dyna.actor.initPosRot.pos.y;
+    if (globalCtx->gameplayFrames & 1) {
+        dustPos.x = this->dyna.actor.posRot.pos.x + 65.0f;
+        dustPos.z = Math_Rand_CenteredFloat(110.0f) + this->dyna.actor.posRot.pos.z;
+        func_80033480(globalCtx, &dustPos, 5.0f, 1, 0x14, 0x3C, 1);
```
Dragorn421, line 138:
```suggestion
        func_80033480(globalCtx, &dustPos, 5.0f, 1, 20, 60, 1);
```
those two arguments are scale-related so decimal is ok
also for other `func_80033480` calls

Dragorn421 (changes requested):

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -31,17 +32,150 @@ const ActorInit Bg_Ddan_Jd_InitVars = {
...
+    // if this is the platform that rises all the way to the top, and the switch state has just changed to on
+    if (this->ySpeed == DEFAULT_Y_SPEED && this->dyna.actor.params < 0x40 &&
+        Flags_GetSwitch(globalCtx, this->dyna.actor.params)) {
+        this->ySpeed = SHORTCUT_Y_SPEED;
+        this->state = STATE_GO_MIDDLE_FROM_BOTTOM;
+        this->dyna.actor.posRot.pos.y = this->dyna.actor.initPosRot.pos.y + MOVE_HEIGHT_MIDDLE;
+        this->idleTimer = 0;
+        this->actionFunc = BgDdanJd_Idle;
+        func_800800F8(globalCtx, 0xBF4, -0x63, &this->dyna.actor, 0);
+    } else if (Math_ApproxF(&this->dyna.actor.posRot.pos.y, this->targetY, (f32)this->ySpeed)) {
```
Roman971, line 166:
```suggestion
    } else if (Math_ApproxF(&this->dyna.actor.posRot.pos.y, this->targetY, this->ySpeed)) {
```

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -31,17 +32,150 @@ const ActorInit Bg_Ddan_Jd_InitVars = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Ddan_Jd/BgDdanJd_Destroy.s")
+static InitChainEntry sInitChain[] = {
+    ICHAIN_VEC3F_DIV1000(scale, 100, ICHAIN_STOP),
+};
+
+#define STATE_GO_BOTTOM 0
+#define STATE_GO_MIDDLE_FROM_BOTTOM 1
+#define STATE_GO_MIDDLE_FROM_TOP 2
+#define STATE_GO_TOP 3
```
Roman971, line 43: This should probably be an enum

Roman971 (changes requested):

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
Roman971: I agree, I don't think it's necessary.

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
mzxrules: Then let's follow the logic.

In `BgDdanJd_Init`, your initial state is set to `STATE_GO_BOTTOM` and ySpeed set to `DEFAULT_Y_SPEED` or `SHORCUT_Y_SPEED`, depending on whether Flags_GetSwitch returns switch off or switch on respectively. If `ySpeed == SHORTCUT_Y_SPEED` at this stage, then the actor will **ALWAYS** exhibit the shortcut behavior. You can trace BgDdanJd_Idle's logic to see this to be true: 

1. `STATE_GO_BOTTOM -> STATE_GO_MIDDLE_FROM_BOTTOM`
2. `this->ySpeed != DEFAULT_Y_SPEED  /* true */ -> STATE_GO_TOP`
3. `STATE_GO_TOP -> STATE_GO_MIDDLE_FROM_TOP`
4. repeat 2 and 3 indefinitely

Now, if BgDdanJd_Init sets ySpeed to `DEFAULT_Y_SPEED`, then it will exhibit the bottom <-> middle behavior, **BUT** it can be either...

1. A platform that **ALWAYS** exhibits the bottom <-> middle behavior regardless of switch state
2. The shortcut platform, which swaps to the shortcut behavior on switch flag on.

As such, in the first comment ySpeed == DEFAULT_Y_SPEED on it's own is not enough to determine whether the platform is the shortcut platform or not; it's only by checking params that we can determine that yes, this is the platform that can go all the way to the top.

In the second comment, ySpeed != DEFAULT_Y_SPEED, so the platform is the shortcut platform because it exhibits the shortcut behavior.

Dragorn421: Thanks, I get it now, so the first comment should state "if this is the platform that can become a shortcut" (refering to the actor variable, not to the `ySpeed` value?), and the second "if this platform is the enabled shortcut" or something like that?

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -31,17 +32,152 @@ const ActorInit Bg_Ddan_Jd_InitVars = {
...
+static InitChainEntry sInitChain[] = {
+    ICHAIN_VEC3F_DIV1000(scale, 100, ICHAIN_STOP),
+};
+
+enum BgDdanJdState {
+    STATE_GO_BOTTOM,
+    STATE_GO_MIDDLE_FROM_BOTTOM,
+    STATE_GO_MIDDLE_FROM_TOP,
+    STATE_GO_TOP,
+};
```
Roman971, line 45: This is the style we typically use for enums:
```suggestion
typedef enum {
    STATE_GO_BOTTOM,
    STATE_GO_MIDDLE_FROM_BOTTOM,
    STATE_GO_MIDDLE_FROM_TOP,
    STATE_GO_TOP,
} BgDdanJdState;
```

Roman971 (approved):
Seems good to me code-wise

fig02 (approved):

PR #395, opened 2020-09-16
Decompiled ovl_Bg_Spot18_Shutter

ethteck:


src/overlays/actors/ovl_Bg_Spot18_Shutter/z_bg_spot18_shutter.c
```diff
@@ -21,21 +32,114 @@ const ActorInit Bg_Spot18_Shutter_InitVars = {
...
+        } else {
+            this->actionFunc = func_808B971C;
+            func_800800F8(globalCtx, 0x107D, 0x8C, &this->dyna.actor, 0);
+        }
+    }
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Spot18_Shutter/func_808B95B8.s")
+void func_808B9698(BgSpot18Shutter* this, GlobalContext* globalCtx) {
+    if (Math_ApproxF(&this->dyna.actor.posRot.pos.y, this->dyna.actor.initPosRot.pos.y + 180.0f, 1.44f) != 0) {
```
Dragorn421, line 112: `Math_ApproxF` returns a boolean

Dragorn421 (changes requested):
Cool,
This is used in goron city for blocking Darunia's room at both ages (with different switch flags), and the shop as child at night (if I'm reading the scene setups correctly, that sounds a bit weird)
The game scenes always use 0 for the higher variable byte so the `(this->dyna.actor.params >> 8) & 1 == 1` case which seems to be a sideways-sliding door looks like unused code :(

Dragorn421 (approved):

Roman971 (approved):

fig02 (approved):

PR #396, opened 2020-09-16
All Effect Overlays and z_effect_soft_sprite_old_init.c OK

fig02:
Most effects are documented but it definitely needs to be revisited at a later time.
If anyone wants to make sense of the dust draw flags and help me name the 9 million dust wrapper functions, feel free.

An important change beyond just effects stuff here is that Color_RGBA is now the non-union type. You should always use this unless you need the u32 aligned one (which is pretty rare).


src/code/z_effect_soft_sprite_old_init.c
```diff
@@ -214,255 +328,718 @@ void EffectSsBomb2_SpawnLayered(GlobalContext* globalCtx, Vec3f* pos, Vec3f* vel
...
+/**
+ * Spawn a lightning shock effect
+ *
+ * param determines where the ligntning should go
+ * 0: dont attach to any actor. spawns at the position specified by pos
+ * 1: spawn at one of Player's body parts, chosen at random
+ * 2: spawn at one of Phantom Ganon's body parts, chosen at random
+ */
+void EffectSsFhgFlash_SpawnLightning(GlobalContext* globalCtx, Actor* actor, Vec3f* pos, s16 scale, u8 param) {
     EffectSsFhgFlashInitParams initParams;
```
fig02, line 777: note to self, gonna rename this EffectSsFhgFlash_SpawnShock later (and related param names)

fig02:
player is merged FeelsOkayMan

src/overlays/effects/ovl_Effect_Ss_G_Splash/z_eff_ss_g_splash.c
```diff
@@ -1,40 +1,152 @@
...
+
+void EffectSsGSplash_Draw(GlobalContext* globalCtx, u32 index, EffectSs* this) {
+    s16 texIdx;
+
+    switch (this->rType) {
+        case 0:
+            texIdx = this->rTexIdx / 0x64;
+            if (texIdx >= 8) {
+                texIdx = 7;
+            }
```
Roman971, line 117: Relevant for the other cases
```suggestion
            texIdx = this->rTexIdx / 100;
            if (texIdx > 7) {
                texIdx = 7;
            }
```

src/code/z_effect_soft_sprite_old_init.c
```diff
@@ -1,38 +1,99 @@
...
+#include "overlays/effects/ovl_Effect_Ss_Dead_Db/z_eff_ss_dead_db.h"
+#include "overlays/effects/ovl_Effect_Ss_Dead_Dd/z_eff_ss_dead_dd.h"
+#include "overlays/effects/ovl_Effect_Ss_Dead_Ds/z_eff_ss_dead_ds.h"
 #include "overlays/effects/ovl_Effect_Ss_Dead_Sound/z_eff_ss_dead_sound.h"
+#include "overlays/effects/ovl_Effect_Ss_Ice_Smoke/z_eff_ss_ice_smoke.h"
+
+static Vec3f sZeroVec = { 0.0f, 0.0f, 0.0f };
+
+// effects that use this draw function are responsible for making sure their reg usage lines up with those listed here
+#define rTexIdx regs[0]
```
Roman971, line 43: I thought we concluded to have these in a common header? Otherwise they will very easily get out of sync.
If you're worried about name conflicts, we can have a specific prefix for these shared regs, like `gReg` instead of just `r` for instance. And if they have a proper prefix they can even be in z64effect.h with no issues for now.

fig02: yeah I had it in the header when it was still an enum and moved it after we changed it cause it made more sense to me. I do agree that it should probably be there tho
and yeah thats good thinking for name conflicts, thanks

src/overlays/effects/ovl_Effect_Ss_Bomb2/z_eff_ss_bomb2.c
```diff
@@ -30,170 +28,170 @@ EffectSsInit Effect_Ss_Bomb2_InitVars = {
...
-static UNK_PTR D_8099FF00[] = {
-    0x04007F80, 0x04008780, 0x04008F80, 0x04009780, 0x04009F80, 0x0400A780, 0x0400AF80, 0x0400B780,
-};
-
 extern Gfx D_0400BF80[];
 extern Gfx D_0400BFE8[];
 extern Gfx D_0400C040[];
 
 u32 EffectSsBomb2_Init(GlobalContext* globalCtx, u32 index, EffectSs* this, void* initParamsx) {
+    static void* drawFuncs[] = {
```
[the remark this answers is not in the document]
fig02: yeah id be down for a typedef. should I make one for both draw and update and use it in the EffectSs struct?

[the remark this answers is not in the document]
Roman971: Yeah I think so, feel free to make one for init as well if you want

src/overlays/effects/ovl_Effect_Ss_Stick/z_eff_ss_stick.c
```diff
@@ -1,40 +1,79 @@
...
+    /* 0x00 */ s16 objectID;
+    /* 0x04 */ Gfx* displayList;
+} StickDrawInfo;
+
+u32 EffectSsStick_Init(GlobalContext* globalCtx, u32 index, EffectSs* this, void* initParamsx) {
+    StickDrawInfo drawInfo[] = {
+        { OBJECT_LINK_BOY, 0x0602BA38 },   // adult, broken sword
+        { OBJECT_LINK_CHILD, 0x06006CC0 }, // child, broken stick
+    };
+    StickDrawInfo* ageInfoEntry = gSaveContext.linkAge + drawInfo;
```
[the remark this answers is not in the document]
fig02: doenst work unfortunately. the issue is the order of the regs when it adds the offset to the array base. the only way I could swap it is to do pointer math, which makes me think its what they did originally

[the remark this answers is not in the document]
Roman971: This pattern also affected the operation order in some cases so idk why it wouldn't here.
And I highly doubt they would actually write "age + array" to get a pointer.

src/code/z_effect_soft_sprite_old_init.c
fig02: do you think another name would be better to avoid conflicting with the g global prefix? Im trying to think of an alternative and cant lmao
if its obvious that this isnt a variable then its prob fine, but im not sure

fig02: `rgName` ?

Roman971: `rg` is probably fine

src/overlays/effects/ovl_Effect_Ss_Stick/z_eff_ss_stick.c
```diff
@@ -1,40 +1,79 @@
...
+    EffectSsStickInitParams* initParams = (EffectSsStickInitParams*)initParamsx;
+    Vec3f pos;
+
+    this->rObjBankIdx = Object_GetIndex(&globalCtx->objectCtx, ageInfoEntry->objectID);
+    this->gfx = ageInfoEntry->displayList;
+    pos = initParams->pos;
+    this->pos = pos;
+    this->vec = pos;
```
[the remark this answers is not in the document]
fig02: works :D
knew there was no way they made a useless copy like that. its definitely this too cause every case has accel first, which is backwards. they were totally chained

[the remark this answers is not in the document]
fig02: actually rip i cant do it in `EffectSsGMagma2_Init`
this one has velocity before accel, and it does the copy a good bit before the actual store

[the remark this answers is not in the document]
fig02: nvm got it to work by rearranging some stuff. it wasnt a chained assignment but was able to remove the temp

fig02:
changes made

Roman971 (approved):

src/overlays/effects/ovl_Effect_Ss_Stone1/z_eff_ss_stone1.c
```diff
@@ -6,35 +6,77 @@
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/effects/ovl_Effect_Ss_Stone1/EffectSsStone1_Init.s")
+typedef struct {
+    /* 0x00 */ void* texture;
+    /* 0x04 */ Color_RGBA8 primColor;
+    /* 0x08 */ Color_RGBA8 envColor;
+} EffStoneDrawInfo;
+
+static EffStoneDrawInfo sDrawInfo[] = {
+    { 0x04029A90, { 200, 0, 0, 255 }, { 0, 0, 0, 255 } },
```
Kenix3, line 27: Why did you decide to not add a symbol for this gameplay keep data?

Kenix3 (changes requested):
One small thing to try out. I doubt IDO can do this optimization, though.

The other is a question for why you didn't add a sym to gameplay keep data.

src/overlays/effects/ovl_Effect_Ss_Stone1/z_eff_ss_stone1.c
fig02: in oot at least, we only add stuff to undefined syms if its needed for proper code generation. It will get a proper symbol/name when objects are handled proper.
I know this is different in mm tho. Roz and I had a conversation about this pretty recently in mm-decomp-help

src/code/z_effect_soft_sprite_old_init.c
```diff
@@ -171,26 +227,73 @@ void func_80028A54(GlobalContext* globalCtx, f32 randScale, Vec3f* srcPos) {
...
+    Math_Vec3f_Copy(&initParams.velocity, velocity);
+    Math_Vec3f_Copy(&initParams.accel, accel);
+    initParams.life = life;
+    initParams.updateMode = 1;
+    initParams.rotSpeed = 0x1518;
+    initParams.yaw = Math_Rand_ZeroOne() * 16384.0f;
+    initParams.scale = scale;
+    Color_RGBA8_Copy(&initParams.primColor, primColor);
+    Color_RGBA8_Copy(&initParams.envColor, envColor);
+    initParams.alphaStep = (-(255.0f / initParams.life)) + (-(255.0f / initParams.life));
```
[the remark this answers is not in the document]
fig02: neither worked unfortunately. that would have been nice though

Kenix3 (approved):

PR #402, opened 2020-09-19
ovl_En_Door

Thar0:
2 nonmatchings, regalloc only.
Mostly documented.

Left the actor params structure in a comment in the header. If we don't want this I can remove it but since it is purely documentation I would hope that if we want to do this at this stage we can.


src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +39,328 @@ const ActorInit En_Door_InitVars = {
...
+    Vec3f sp2C;
+    s8 numKeys;
+    s16 phi_v0;
+    s16 temp_v0;
+
+    doorType = (this->actor.params >> 7) & 7;
+    func_8002DBD0(&this->actor, &sp2C, &player->actor.posRot.pos);
+    if (this->unk_191 != 0) {
+        this->actionFunc = EnDoor_Open;
+        SkelAnime_ChangeAnimPlaybackStop(&this->skelAnime, D_809FCECC[this->unk_190], (((s32)player->stateFlags1 * 0x10) < 0) ? 0.75f : 1.5f);
```
Dragorn421, line 210:
```suggestion
        SkelAnime_ChangeAnimPlaybackStop(&this->skelAnime, D_809FCECC[this->unk_190], (player->stateFlags1 & 0x8000000) ? 0.75f : 1.5f);
```
maybe

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +39,328 @@ const ActorInit En_Door_InitVars = {
...
+            this->actionFunc = EnDoor_IdleAjar;
+        }
+    }
+}
+#else
+#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Door/EnDoor_Idle.s")
+#endif
+
+void EnDoor_WaitForCheck(EnDoor* this, GlobalContext* globalCtx) {
+    if (func_8002F194(&this->actor, globalCtx)) { // if actor is checked
```
Dragorn421, line 247: "checked"?

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +39,328 @@ const ActorInit En_Door_InitVars = {
...
+
+void EnDoor_Talk(EnDoor* this, GlobalContext* globalCtx) {
+    if (func_8002F334(&this->actor, globalCtx)) {
+        // Return to waiting to be checked once text is dismissed
+        this->actionFunc = EnDoor_WaitForCheck;
+    }
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Door/EnDoor_Destroy.s")
+void EnDoor_CheckRange(EnDoor* this, GlobalContext* globalCtx) {
```
Dragorn421, line 262: the CheckRange name is kind of confusing with the unrelated "Check" from eg `EnDoor_WaitForCheck`?

I think action functions should be named to describe the state they're in, not what they're going to do next
for this one I'd suggest `EnDoor_AjarPlayerFar` or `EnDoor_AjarOpen`? (the latter being consistent with following suggestions)

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +39,328 @@ const ActorInit En_Door_InitVars = {
...
+    if (this->actor.xzDistFromLink < DOOR_SLAM_RANGE) {
+        this->actionFunc = EnDoor_Shut;
+    }
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Door/func_809FC41C.s")
+void EnDoor_IdleAjar(EnDoor* this, GlobalContext* globalCtx) {
```
Dragorn421, line 268: `EnDoor_AjarOpening`?

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +39,328 @@ const ActorInit En_Door_InitVars = {
...
+    if (this->actor.xzDistFromLink < DOOR_SLAM_RANGE) {
+        this->actionFunc = EnDoor_Shut;
+    } else if (Math_ApproxUpdateScaledS(&this->actor.posRot.rot.y, -0x1800, 0x100) != 0) {
```
Dragorn421, line 271:
```suggestion
    } else if (Math_ApproxUpdateScaledS(&this->actor.posRot.rot.y, -0x1800, 0x100)) {
```
true if target value reached

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +39,328 @@ const ActorInit En_Door_InitVars = {
...
+        this->actionFunc = EnDoor_CheckRange;
+    }
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Door/func_809FC5D0.s")
+void EnDoor_Shut(EnDoor* this, GlobalContext* globalCtx) {
```
Dragorn421, line 276: `EnDoor_AjarClosing`?

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +39,328 @@ const ActorInit En_Door_InitVars = {
...
+    if (Math_ApproxUpdateScaledS(&this->actor.posRot.rot.y, 0, 0x700) != 0) {
```
Dragorn421, line 277:
```suggestion
    if (Math_ApproxUpdateScaledS(&this->actor.posRot.rot.y, 0, 0x700)) {
```

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +39,328 @@ const ActorInit En_Door_InitVars = {
...
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Door/func_809FC878.s")
+void EnDoor_Open(EnDoor* this, GlobalContext* globalCtx) {
+    s32 i;
+    s32 numEffects;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Door/func_809FC8C0.s")
+    if (DECR(this->lockTimer) == 0) { // Wait for the lock to finish if applicable
+        if (SkelAnime_FrameUpdateMatrix(&this->skelAnime) != 0) {
```
Dragorn421, line 287:
```suggestion
        if (SkelAnime_FrameUpdateMatrix(&this->skelAnime)) {
```

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +39,328 @@ const ActorInit En_Door_InitVars = {
...
+            // If done, return to the idle action and free the player
+            this->actionFunc = EnDoor_Idle;
+            this->unk_191 = 0;
+        } else if (func_800A56C8(&this->skelAnime, sDoorAnimOpenFrames[this->unk_190]) != 0) {
```
Dragorn421, line 291:
```suggestion
        } else if (func_800A56C8(&this->skelAnime, sDoorAnimOpenFrames[this->unk_190])) {
```

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +39,328 @@ const ActorInit En_Door_InitVars = {
...
+            Audio_PlayActorSound2(&this->actor, (globalCtx->sceneNum == SCENE_HAKADAN || 
+                globalCtx->sceneNum == SCENE_HAKADANCH || 
+                globalCtx->sceneNum == SCENE_HIDAN) ? NA_SE_EV_IRON_DOOR_OPEN : NA_SE_OC_DOOR_OPEN);
+            if (this->skelAnime.animPlaybackSpeed < 1.5f) {
+                numEffects = (s32) (Math_Rand_ZeroOne() * 30.0f) + 50;
+                for (i = 0; i < numEffects; i++) {
+                    func_800293E4(globalCtx, &this->actor.posRot.pos, 60.0f, 100.0f, 50.0f, 0.15f);
+                }
+            }
+        } else if (func_800A56C8(&this->skelAnime, sDoorAnimCloseFrames[this->unk_190]) != 0) {
```
Dragorn421, line 302:
```suggestion
        } else if (func_800A56C8(&this->skelAnime, sDoorAnimCloseFrames[this->unk_190])) {
```

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +39,328 @@ const ActorInit En_Door_InitVars = {
...
+            *dList = (ABS(phi_v0_2) < 0x4000) ? temp_a2[0] : temp_a2[1];
+        } else {
+            phi_v0 = this->unk_192;
+            if (transitionEntry->sides[0].room != this->actor.room) {
+                phi_v0 ^= 1;
+            }
+            *dList = temp_a2[phi_v0];
+        }
+    }
+    return 0;
```
Dragorn421, line 340:
```suggestion
    return false;
```
the return value is a "skipDrawingLimb" boolean

src/overlays/actors/ovl_En_Door/z_en_door.h
```diff
@@ -4,14 +4,57 @@
...
+    /* 0x00 */ s16 sceneNum;
+    /* 0x02 */ u8 displaylistIdx;
+    /* 0x04 */ s16 objectId;
+} EnDoorInfo;
+
+typedef enum {
+    /* 0x00 */ DOOR_0,          // loads rooms
+    /* 0x01 */ DOOR_LOCKED,     // small key locked door
+    /* 0x02 */ DOOR_1,          // loads rooms
+    /* 0x03 */ DOOR_NOLOAD,     // doesn't load rooms, used for doors paired with scene transition polygons
```
Dragorn421, line 35: NOLOAD is a bit confusing imo, makes me think about "don't load door"
maybe `DOOR_EXIT` / `DOOR_SCENE_EXIT` / `DOOR_EXIT_SCENE`? exit as in "scene exit list" (cf `func_800990F0`)

but actually, can this type be used for a same-room "transition"?

src/overlays/actors/ovl_En_Door/z_en_door.h
```diff
@@ -4,14 +4,57 @@
...
+ */
+
+typedef struct {
+    /* 0x00 */ s16 sceneNum;
+    /* 0x02 */ u8 displaylistIdx;
+    /* 0x04 */ s16 objectId;
+} EnDoorInfo;
+
+typedef enum {
+    /* 0x00 */ DOOR_0,          // loads rooms
```
Dragorn421, line 32: maybe make it clear that 0, 1 and 7 are the same "default" behavior?
one could assume they're just unknown behavior especially with keeping digits

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +39,328 @@ const ActorInit En_Door_InitVars = {
...
+    { SCENE_HIDAN,     1, OBJECT_HIDAN_OBJECTS },
+    { SCENE_MIZUSIN,   2, OBJECT_MIZU_OBJECTS },
+    { SCENE_HAKADAN,   3, OBJECT_HAKA_DOOR },
+    { SCENE_HAKADANCH, 3, OBJECT_HAKA_DOOR },
+    { 0xFFFF,          0, OBJECT_GAMEPLAY_KEEP },
+    { 0xFFFF,          4, OBJECT_GAMEPLAY_FIELD_KEEP },
+};
+
+InitChainEntry D_809FCEC4[] = {
+    ICHAIN_U8(unk_1F, 0x0, ICHAIN_CONTINUE),
```
Dragorn421, line 56:
```suggestion
    ICHAIN_U8(unk_1F, 0, ICHAIN_CONTINUE),
```

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -10,12 +10,24 @@
...
+
 void EnDoor_Init(Actor* thisx, GlobalContext* globalCtx);
 void EnDoor_Destroy(Actor* thisx, GlobalContext* globalCtx);
 void EnDoor_Update(Actor* thisx, GlobalContext* globalCtx);
 void EnDoor_Draw(Actor* thisx, GlobalContext* globalCtx);
 
-/*
+void EnDoor_SetupType(EnDoor* this, GlobalContext* globalCtx);
+
+void EnDoor_Idle(EnDoor* this, GlobalContext* globalCtx);
```
Dragorn421, line 23:
```suggestion
void EnDoor_SetupType(EnDoor* this, GlobalContext* globalCtx);
void EnDoor_Idle(EnDoor* this, GlobalContext* globalCtx);
```

Dragorn421 (changes requested):
Good job!
I'm a bit confused by some of this code since it's interacting with fresh player code
I mainly have naming-related comments

src/overlays/actors/ovl_En_Door/z_en_door.c
Thar0: This actions waits for the initial check before transitioning into `EnDoor_Talk` for the remainder of the check, so checked doesn't really make sense here since `EnDoor_WaitForChecked` implies to me that it waits until the check has ended, which isn't the case.

src/overlays/actors/ovl_En_Door/z_en_door.h
Thar0: I hoped to remove 2 (previously 1, but that was a typo) and 7 and leave gaps in the enum with a comment saying `// unimplemented` or something to that affect but when searching for `ACTOR_EN_DOOR` uses in scenes it appears that at least 2 is used, so I left them in for completeness. It should be clear enough that they refer to the same behavior as they are just numbered versions of the same name.

src/overlays/actors/ovl_En_Door/z_en_door.c
Thar0: This worked but didn't affect matching unfortunately, but it's definitely a lot cleaner.

src/overlays/actors/ovl_En_Door/z_en_door.h
```diff
@@ -4,14 +4,57 @@
...
+ * |------------------|-------|-------------|---------------------------------
+ * | 0 0 0 0 0 0      | 0 0 0 | 0           | 0 0 0 0 0 0 
+ * | 6                | 3     | 1           | 6
+ * |
+ * 
+ * Transition Index     1111110000000000
+ * Type                 0000001110000000
+ * Double Door          0000000001000000
+ * Switch Flag          0000000000111111
+ * Text id - 0x0200     0000000000111111
```
Dragorn421, line 21: Would be nice to comment on:

"Transition Index" being set by the actor engine
(cf `Actor_SpawnTransitionActors`, it has `(i << 0xA) + transitionActor->params`)
so, those should be zeros in the scene transition actors list (relevant scene header command `func_80098C68`)

"Switch Flag" is for type DOOR_LOCKED only

"Text id - 0x0200" is for type DOOR_TALKING only

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +40,329 @@ const ActorInit En_Door_InitVars = {
...
+    transitionEntry = &globalCtx->transitionActorList[(u16)this->actor.params >> 0xA];
+    if (transitionEntry->id < 0) {
+        transitionEntry->id = -transitionEntry->id;
+    }
+}
+
+void EnDoor_SetupType(EnDoor* this, GlobalContext* globalCtx) {
+    s32 doorType;
+
+    if (Object_IsLoaded(&globalCtx->objectCtx, this->currentObject)) {
```
Dragorn421, line 162: maybe `requiredObject`(`Id`) is better than currentObject, because the "true" current object is `Actor.objectBankIndex` while `EnDoor.currentObject` is the object that should be actually used

Dragorn421 (changes requested):

src/overlays/actors/ovl_En_Door/z_en_door.c
Thar0: `EnDoor_WaitForTalk` doesn't really work since the A button says `Check` and not `Speak` within the check range. I could rename the talk related names to make them check related instead.

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +40,329 @@ const ActorInit En_Door_InitVars = {
...
+// Regalloc only
+void EnDoor_Idle(EnDoor* this, GlobalContext *globalCtx) {
+    Player* player = PLAYER;
+    s32 doorType;
+    Vec3f sp2C;
+    s8 numKeys;
+    s16 phi_v0;
+    s16 temp_v0;
+
+    doorType = (this->actor.params >> 7) & 7;
```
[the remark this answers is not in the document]
fig02: disgusting Kappa

src/overlays/actors/ovl_En_Door/z_en_door.c
Dragorn421: I like Talk better, gets the point of using dialogue across much better
And I don't really care about the in-game UI saying Check tbh. As far as code goes, talking to Anju 🐔 or to a door 🚪 or to Talon through a door doesn't make a difference

Thar0: While the text side of the code is identical, the actor itself handles none of that so really the actor should make it clear which of Check and Speak is in use. None of the other internal text machinery features whatsoever so isn't relevant to actor documentation.

Dragorn421: Well I guess if "check" is supposed to ring the bell of ~"checkable spot" then it would make sense, but since I even started on this by asking what "check" was I guess I won't be convinced about using it

"check" is just way too generic for me

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +40,347 @@ const ActorInit En_Door_InitVars = {
     (ActorFunc)EnDoor_Update,
     (ActorFunc)EnDoor_Draw,
 };
-*/
+
+/**
+ * Controls which object and display lists to use in a given scene
+ */
+EnDoorInfo D_809FCEA0[] = {
```
Dragorn421, line 47: you can make the data `static`

src/overlays/actors/ovl_En_Door/z_en_door.h
```diff
@@ -4,14 +4,57 @@
...
+ * Actor Parameters
+ *
+ * |                  |       |             |
+ * | Transition Index | Type  | Double Door | Switch Flag OR Text Id - 0x0200
+ * |------------------|-------|-------------|---------------------------------
+ * | 0 0 0 0 0 0      | 0 0 0 | 0           | 0 0 0 0 0 0
+ * | 6                | 3     | 1           | 6
+ * |
+ *
+ * Transition Index     1111110000000000    Set by the actor engine when the door is spawned
```
Dragorn421, line 17: it's or `|`'d, not set, so it being 0 initially matters, idk if that's worth pressing on though

Dragorn421 (changes requested):

src/overlays/actors/ovl_En_Door/z_en_door.c
Thar0: Not all data can be made `static` if it is used in the `NON_MATCHING`s as the version of asm-processor we currently use does not deal with static symbols

src/overlays/actors/ovl_En_Door/z_en_door.h
Thar0: The instance space is zero'd out with a call to `Lib_MemSet` on spawn so this isn't worth specifying in particular.

Dragorn421: I meant if one was to use this documentation for building a transition actors list in a scene header

Still, what's the point of `Lib_MemSet`? the actor.params is set afterwards anyway?

Dragorn421 (approved):

src/overlays/actors/ovl_En_Door/z_en_door.h
Thar0: MemSet ensures no data that was previously allocated in that space will cause it to be nonzero. This concern seems like the kind of thing that will be covered when macros for getting and setting actor params are introduced in the future, perhaps at that stage a note can be left on the setter for the transition index that it is for internal use only. That should cover the issue entirely as once they are introduced no actor params should be edited without reference to these macros.

Dragorn421: I think MemSet is completely irrelevant for what I'm talking about
https://github.com/zeldaret/oot/blob/14191307e96cc915fa5a67704eba16421cf4deb2/src/code/z_actor.c#L2818
I was mistaken it's a `+` not a `|` but the issue is unchanged

Whatever. macros should indeed eliminate the concern

src/overlays/actors/ovl_En_Door/z_en_door.c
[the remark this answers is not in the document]
Roman971: Surely that's not a serious suggestion... right?

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +40,342 @@ const ActorInit En_Door_InitVars = {
...
+
+void EnDoor_Init(Actor* thisx, GlobalContext* globalCtx) {
+    s32 pad;
+    s32 pad1;
+    EnDoorInfo* objectInfo;
+    s32 i;
+    s32 objIndex;
+    f32 xOffset;
+    f32 zOffset;
+    EnDoor* this = THIS;
```
petrie911, line 102:
```suggestion
void EnDoor_Init(Actor* thisx, GlobalContext* globalCtx2) {
    GlobalContext* globalCtx = globalCtx2;
    EnDoor* this = THIS;
    EnDoorInfo* objectInfo;
    s32 i;
    s32 objIndex;
    f32 xOffset;
    f32 zOffset;
```
This matches and lets you get rid of ` if (!globalCtx->sceneNum) {}` below

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +40,342 @@ const ActorInit En_Door_InitVars = {
...
+    SkelAnime_Init(globalCtx, &this->skelAnime, &D_0400FF78, &D_0400E758, this->limbDrawTable,
+                   this->transitionDrawTable, 5);
+    for (i = 0; i < ARRAY_COUNT(D_809FCEA0) - 2; i++, objectInfo++) {
+        if (globalCtx->sceneNum == objectInfo->sceneNum) {
+            break;
+        }
+    }
+    if (i >= ARRAY_COUNT(D_809FCEA0) - 2 && Object_GetIndex(&globalCtx->objectCtx, OBJECT_GAMEPLAY_FIELD_KEEP) >= 0) {
+        objectInfo++;
+        if (!globalCtx->sceneNum) {}
```
petrie911, line 115:
```suggestion
```

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +40,342 @@ const ActorInit En_Door_InitVars = {
...
+    TransitionActorEntry* transitionEntry;
+    Gfx** temp_a2;
+    s32 pad2;
+    s16 phi_v0_2;
+    s32 phi_v0;
+    EnDoor* this = THIS;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Door/func_809FC998.s")
+    if (limbIndex == 4) {
+        temp_a2 = &D_809FCEE4[(s8)this->displaylistIdx];
```
petrie911, line 338:
```suggestion
        temp_a2 = D_809FCEE4[(s8)this->displaylistIdx];
```
this & is unneeded and causes a type warning. Also is the (s8) cast needed? Seem to be a lot of stray casts around.

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +40,342 @@ const ActorInit En_Door_InitVars = {
...
+    25,
+    25,
+};
+
+static u8 sDoorAnimCloseFrames[] = {
+    60,
+    70,
+    60,
+    70,
+};
```
petrie911, line 81:
```suggestion
static u8 sDoorAnimOpenFrames[] = {
    25,
    25,
    25,
    25
};

static u8 sDoorAnimCloseFrames[] = {
    60,
    70,
    60,
    70
};
```
We probably don't need format.sh putting these all on different lines. Might be true of `D_809FCECC` as well.

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +40,342 @@ const ActorInit En_Door_InitVars = {
...
+    { SCENE_HIDAN, 1, OBJECT_HIDAN_OBJECTS },
+    { SCENE_MIZUSIN, 2, OBJECT_MIZU_OBJECTS },
+    { SCENE_HAKADAN, 3, OBJECT_HAKA_DOOR },
+    { SCENE_HAKADANCH, 3, OBJECT_HAKA_DOOR },
+    // KEEP objects should remain last and in this order
+    { -1, 0, OBJECT_GAMEPLAY_KEEP },
+    { -1, 4, OBJECT_GAMEPLAY_FIELD_KEEP },
+};
+
+InitChainEntry D_809FCEC4[] = {
```
petrie911, line 57: This can be static and be sInitChain now.

src/overlays/actors/ovl_En_Door/z_en_door.c
petrie911: But it can be static now.

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +40,342 @@ const ActorInit En_Door_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Door/EnDoor_Update.s")
+    if (this->actor.objBankIndex == this->requiredObject) {
+        OPEN_DISPS(globalCtx->state.gfxCtx, "../z_en_door.c", 910);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Door/func_809FCC3C.s")
+        func_80093D18(globalCtx->state.gfxCtx);
+        SkelAnime_Draw(globalCtx, this->skelAnime.skeleton, this->skelAnime.limbDrawTbl, EnDoor_OverrideLimbDraw, NULL,
+                       &this->actor);
+        if (this->actor.posRot.rot.y != 0) {
+            if (!this) {}
```
petrie911, line 367:
```suggestion
            if(1) {}
```
It doesn't have to be `this`.

petrie911 (commented):
Unfortunately I couldn't find a solution for EnDoor_Idle, and neither could the permuter. I have some other suggestions, though.

src/code/z_actor.c
```diff
@@ -3509,16 +3509,16 @@ void func_80033F54(GlobalContext* globalCtx, s32 arg1, s32 arg2) {
...
 
     OPEN_DISPS(globalCtx->state.gfxCtx, "../z_actor.c", 8265);
 
     Matrix_Translate(0.0f, entry->unk_08, 500.0f, MTXMODE_APPLY);
     Matrix_Get(&spB0);
 
-    temp1 = sinf(entry->unk_00 - var) * -(10 - arg1) * 0.1f * entry->unk_04;
-    temp2 = cosf(entry->unk_00 - var) * (10 - arg1) * 0.1f * entry->unk_04;
+    temp1 = sinf(entry->unk_00 - var) * -(10 - frame) * 0.1f * entry->unk_04;
+    temp2 = cosf(entry->unk_00 - var) * (10 - frame) * 0.1f * entry->unk_04;
```
fig02, line 3521: what does frame mean here? frame count in the opening animation or something? I havent read the code its called from yet, but it may be worth being more specific anyway

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +40,330 @@ const ActorInit En_Door_InitVars = {
...
+    }
+
+    // Double doors
+    if (this->actor.params & 0x40) {
+        EnDoor* attached;
+
+        xOffset = Math_Coss(this->actor.shape.rot.y) * 30.0f;
+        zOffset = Math_Sins(this->actor.shape.rot.y) * 30.0f;
+        attached = (EnDoor*)Actor_SpawnAsChild(&globalCtx->actorCtx, &this->actor, globalCtx, ACTOR_EN_DOOR,
+                                               this->actor.posRot.pos.x + xOffset, this->actor.posRot.pos.y,
```
fig02, line 128: since we've changed the name of attached to parent and child i guess we can change this temp name. what is the actor though? the second door? might be worth saying otherDoor or something, since the fact that its a child of the other door doesnt mean too much

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +40,330 @@ const ActorInit En_Door_InitVars = {
...
+    } else {
+        // allows the player to "check" the door if within 40 units
+        func_8002F2CC(&this->actor, globalCtx, DOOR_CHECK_RANGE);
+    }
+}
+
+void EnDoor_Check(EnDoor* this, GlobalContext* globalCtx) {
+    if (func_8002F334(&this->actor, globalCtx)) {
+        // Return to waiting to be checked once text is dismissed
+        this->actionFunc = EnDoor_WaitForCheck;
```
fig02, line 253: this will be very clear after that func is named, can prob remove comment

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +40,330 @@ const ActorInit En_Door_InitVars = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Door/func_809FC878.s")
+void EnDoor_Open(EnDoor* this, GlobalContext* globalCtx) {
+    s32 i;
+    s32 numEffects;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Door/func_809FC8C0.s")
+    if (DECR(this->lockTimer) == 0) { // Wait for the lock to finish if applicable
+        if (SkelAnime_FrameUpdateMatrix(&this->skelAnime)) {
+            // If done, return to the idle action and free the player
```
fig02, line 283: all of these can go too imo, its pretty clear from the code

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +40,330 @@ const ActorInit En_Door_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Door/func_809FCC3C.s")
+        func_80093D18(globalCtx->state.gfxCtx);
+        SkelAnime_Draw(globalCtx, this->skelAnime.skeleton, this->skelAnime.limbDrawTbl, EnDoor_OverrideLimbDraw, NULL,
+                       &this->actor);
+        if (this->actor.posRot.rot.y != 0) {
+            if (1) {}
+            if (this->actor.posRot.rot.y > 0) {
+                gSPDisplayList(oGfxCtx->polyOpa.p++, D_0400EE00);
+            } else {
+                gSPDisplayList(oGfxCtx->polyOpa.p++, D_0400ECB8);
```
fig02, line 359: these can be updated to use POLY macros

fig02 (commented):
a few things, sorry for the wait

src/code/z_actor.c
Thar0: It's the animation frame, yes. If more clarity is desired I could change it to `animFrame`?

fig02: i would personally, but its probably pretty clear from context too. whatever you feel is best

fig02 (approved):

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +40,317 @@ const ActorInit En_Door_InitVars = {
...
+            *dList = (ABS(phi_v0_2) < 0x4000) ? temp_a2[0] : temp_a2[1];
+        } else {
+            phi_v0 = this->unk_192;
+            if (transitionEntry->sides[0].room != this->actor.room) {
+                phi_v0 ^= 1;
+            }
+            *dList = temp_a2[phi_v0];
+        }
+    }
+    return false;
```
Roman971, line 330:
```suggestion
    return 0;
```

src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -27,29 +40,317 @@ const ActorInit En_Door_InitVars = {
...
+    this->displaylistIdx = objectInfo->displaylistIdx;
+    objIndex = Object_GetIndex(&globalCtx->objectCtx, objectInfo->objectId);
+    if (objIndex < 0) {
+        Actor_Kill(&this->actor);
+        return;
+    }
+
+    this->requiredObject = objIndex;
+    this->displaylistIdx = objectInfo->displaylistIdx;
+    if (this->actor.objBankIndex == this->requiredObject) {
```
Roman971, line 110: I suggest naming `requiredObject` with something that makes it clear it's a bank index and not an object id.
For instance:
```suggestion
    if (this->actor.objBankIndex == this->requiredObjBankIndex) {
```

fig02: id also rename that variable above to `dListIndex` as we've been trying to make that abbreviation more consistent

fig02:
changes made

Roman971 (approved):

PR #403, opened 2020-09-20
ovl_En_Horse_Normal

Sirius902:
Okay, so I did hear that there might be opposition to having a collider named `bodyCollider` but I'm not really sure what to call it then. Any preference for what it should be called? Also one more thing, I'm not sure exactly how to phrase the clone comment in `EnHorseNormal_Draw`. I would like to say it draws a clone but that wouldn't be accurate since it also handles collisions for the clone.


src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
```diff
@@ -21,45 +48,635 @@ const ActorInit En_Horse_Normal_InitVars = {
...
+};
+
+// Unused
+static EnHorseNormalUnkStruct2 D_80A6D4B8 = { ARRAY_COUNT(D_80A6D470), D_80A6D470 };
+
+void func_80A6B250(EnHorseNormal* this) {
+    static s32 D_80A6D4C0[] = { 0, 16 };
+
+    if (D_80A6D4C0[this->unk_200] < this->skin.skelAnime.animCurrentFrame &&
+        ((this->unk_200 != 0) || !(D_80A6D4C0[1] < this->skin.skelAnime.animCurrentFrame))) {
```
Dragorn421, line 118:
```suggestion
        ((this->unk_200 != 0) || this->skin.skelAnime.animCurrentFrame <= D_80A6D4C0[1])) {
```
I think that's a simpler way to write it but it may not match like that?

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
```diff
@@ -21,45 +48,635 @@ const ActorInit En_Horse_Normal_InitVars = {
...
+        Audio_PlaySoundGeneral(NA_SE_EV_HORSE_WALK, &this->actor.projectedPos, 4, &D_801333E0, &D_801333E0,
+                               &D_801333E8);
+        this->unk_200 += 1;
+        if (this->unk_200 >= 2) {
```
Dragorn421, line 122:
```suggestion
        this->unk_200++;
        if (this->unk_200 >= ARRAY_COUNT(D_80A6D4C0)) {
```
since it indexes into `D_80A6D4C0` the 2 must be the length

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
```diff
@@ -21,45 +48,635 @@ const ActorInit En_Horse_Normal_InitVars = {
...
+            if (this->actor.posRot.rot.z != 7) {
+                Actor_Kill(&this->actor);
+                return;
+            }
+        } else if (this->actor.posRot.rot.z != 5) {
+            Actor_Kill(&this->actor);
+            return;
+        }
+        this->actor.shape.rot.z = 0;
+        this->actor.initPosRot.rot.z = this->actor.posRot.rot.z = this->actor.shape.rot.z;
```
Dragorn421, line 196:
```suggestion
        this->actor.initPosRot.rot.z = this->actor.posRot.rot.z = this->actor.shape.rot.z = 0;
```
doesn't match?

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
```diff
@@ -21,45 +48,635 @@ const ActorInit En_Horse_Normal_InitVars = {
...
+        }
+    } else if (globalCtx->sceneNum == SCENE_MALON_STABLE) {
+        if (!gSaveContext.nightFlag) {
+            Actor_Kill(&this->actor);
+        } else {
+            func_800A663C(globalCtx, &this->skin, &D_06009FAC, &D_06004580);
+            SkelAnime_ChangeAnimDefaultStop(&this->skin.skelAnime, sAnimations[this->animationIdx]);
+            func_80A6C6B0(this);
+        }
+        return;
```
Dragorn421, line 214:
```suggestion
            Actor_Kill(&this->actor);
            return;
        } else {
            func_800A663C(globalCtx, &this->skin, &D_06009FAC, &D_06004580);
            SkelAnime_ChangeAnimDefaultStop(&this->skin.skelAnime, sAnimations[this->animationIdx]);
            func_80A6C6B0(this);
            return;
        }
```
More readable imo than sharing the return statement

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
```diff
@@ -21,45 +48,635 @@ const ActorInit En_Horse_Normal_InitVars = {
...
+}
+
+void EnHorseNormal_FollowPath(EnHorseNormal* this, GlobalContext* globalCtx) {
+    Path* path = &globalCtx->setupPathList[this->actor.params & 0xF];
+    Vec3s* pointPos = SEGMENTED_TO_VIRTUAL(path->points);
+    f32 dx;
+    f32 dz;
+    s32 pad;
+
+    pointPos += this->waypoint;
```
Dragorn421, line 262:
```suggestion
    Vec3s* pointPos = &SEGMENTED_TO_VIRTUAL(path->points)[this->waypoint];
    f32 dx;
    f32 dz;
    s32 pad;

```
if this makes it less readable please ignore, my reasoning is that it avoids pointPos being used as a temp but no big deal (may also not match?)

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
```diff
@@ -21,45 +48,635 @@ const ActorInit En_Horse_Normal_InitVars = {
...
+    dx = pointPos->x - this->actor.posRot.pos.x;
+    dz = pointPos->z - this->actor.posRot.pos.z;
+    Math_SmoothScaleMaxMinS(&this->actor.posRot.rot.y, Math_atan2f(dx, dz) * 10430.378f, 10, 2000, 1);
```
Dragorn421, line 265:
```suggestion
    Math_SmoothScaleMaxMinS(&this->actor.posRot.rot.y, Math_atan2f(dx, dz) * (0x8000 / M_PI), 10, 2000, 1);
```
`0x8000` may need to be `(f32) 0x8000` for matching (`10430.378f` is also somewhere else in the file)

also, use hexadecimal for 10/2000, we use hex for rotations (and yeah they sometimes didnt)

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
```diff
@@ -21,45 +48,635 @@ const ActorInit En_Horse_Normal_InitVars = {
...
+    this->actor.shape.rot.y = this->actor.posRot.rot.y;
+    if (SQ(dx) + SQ(dz) < 600.0f) {
+        this->waypoint += 1;
```
Dragorn421, line 268:
```suggestion
        this->waypoint++;
```

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
```diff
@@ -21,45 +48,635 @@ const ActorInit En_Horse_Normal_InitVars = {
...
+    if (SkelAnime_FrameUpdateMatrix(&this->skin.skelAnime)) {
+        SkelAnime_ChangeAnim(&this->skin.skelAnime, sAnimations[this->animationIdx], func_80A6B30C(this), 0.0f,
+                             SkelAnime_GetFrameCount(&sAnimations[this->animationIdx]->genericHeader), 2, 0.0f);
+        func_80A6BCEC(this);
+    }
+}
+
+void func_80A6BBAC(EnHorseNormal* this) {
+    this->type = 0;
+    this->animationIdx += 1;
```
Dragorn421, line 283:
```suggestion
    this->animationIdx++;
```

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
```diff
@@ -21,45 +48,635 @@ const ActorInit En_Horse_Normal_InitVars = {
...
+                Audio_PlaySoundGeneral(NA_SE_EV_HORSE_GROAN, &this->unk_204, 4, &D_801333E0, &D_801333E0, &D_801333E8);
+            } else if (phi_t0 == 3) {
+                Audio_PlaySoundGeneral(NA_SE_EV_HORSE_NEIGH, &this->unk_204, 4, &D_801333E0, &D_801333E0, &D_801333E8);
+            } else {
+                func_80A6BCEC(this);
+            }
+            SkelAnime_ChangeAnim(&this->skin.skelAnime, sAnimations[this->animationIdx], func_80A6B30C(this), 0.0f,
+                                 SkelAnime_GetFrameCount(&sAnimations[this->animationIdx]->genericHeader), 2, -3.0f);
+            return;
+        }
```
Dragorn421, line 408: wait for someone else's input on this but unfortunately I would put an else here instead of an early return 😅

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
```diff
@@ -21,45 +48,635 @@ const ActorInit En_Horse_Normal_InitVars = {
...
+            func_800A63CC(&this->actor, globalCtx, &this->skin, 0, 0, 1, 0, 3);
+            this->cloneCollider.dim.pos.x = clonePos.x;
+            this->cloneCollider.dim.pos.y = clonePos.y;
+            this->cloneCollider.dim.pos.z = clonePos.z;
+            CollisionCheck_SetOC(globalCtx, &globalCtx->colChkCtx, &this->cloneCollider.base);
+            func_80094044(globalCtx->state.gfxCtx);
+            gDPSetPrimColor(oGfxCtx->polyXlu.p++, 0, 0, 0, 0, 0, 255);
+            Matrix_Translate(clonePos.x, clonePos.y, clonePos.z, 0);
+            temp_f0_4 = (1.0f - (distFromGround * 0.01f)) * this->actor.shape.unk_10;
+            Matrix_Scale(this->actor.scale.x * temp_f0_4, 1.0f, this->actor.scale.z * temp_f0_4, 1);
```
Dragorn421, line 671:
```suggestion
            Matrix_Translate(clonePos.x, clonePos.y, clonePos.z, MTXMODE_NEW);
            temp_f0_4 = (1.0f - (distFromGround * 0.01f)) * this->actor.shape.unk_10;
            Matrix_Scale(this->actor.scale.x * temp_f0_4, 1.0f, this->actor.scale.z * temp_f0_4, MTXMODE_APPLY);
```
and in most other Matrix_ functions too

Dragorn421 (changes requested):
Poggers horses in Hyrule field (the unused data) 🏇 

I don't understand the clone thing, in `EnHorseNormal_Draw` is `func_800A6330` drawing the "main"/base horse and the rest of the function is drawing a clone at hardcoded locations? Is the clone static? Is there cloning logic in code apart from drawing and the bonus collider?

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
Sirius902: Yeah, it doesn't match like that.

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
Sirius902: I went ahead and used hex for the angles but unfortunately neither `(0x80000 / M_PI)` or `((f32)0x80000 / M_PI)` match.

Sirius902:
@Dragorn421 The way it works is that some horses are given a `type` of `3` which means they will be static (non-moving), marked as clone-able, and will do some fun idle animations. In the draw function, it first draws the original horse (I'm not sure exactly what function does that) then will check if `type` is `3`. If so it will, in the draw function, draw another horse (the clone) at hard-coded positions dependent on the original horse's position with the same animation / animation frame as the original as well as playing the same sounds the original is playing. Also in the draw function, it will move and use a collider from the original horse to act as the hitbox for the clone horse.

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
Sirius902: It doesn't match.

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
```diff
@@ -21,45 +42,635 @@ const ActorInit En_Horse_Normal_InitVars = {
...
+    this->unk_21C = 0;
+    this->unk_21E = 0;
+    this->actor.flags |= 0x30;
+    this->actor.speedXZ = 0.0f;
+    this->unk_218 = 0.0f;
+    SkelAnime_ChangeAnim(&this->skin.skelAnime, sAnimations[this->animationIdx], func_80A6B30C(this), 0.0f,
+                         SkelAnime_GetFrameCount(&sAnimations[this->animationIdx]->genericHeader), 2, 0.0f);
+}
+
+void EnHorseNormal_WaitClone(EnHorseNormal* this, GlobalContext* globalCtx) {
```
Dragorn421, line 482: naming action functions is hard but from what I saw "Wait" is used when the action will eventually change due to a timer or events (switch flag, colliding)
The name also implies "waiting for clone" imo instead of "I am cloned and idling"? "ClonedIdle"?

But does this function actually do something cloning-related? idk if "Clone" should be included in the name only because the only type using this function is the cloned type
can't really tell the difference from `EnHorseNormal_Wait` though because of the unks

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
```diff
@@ -21,45 +42,635 @@ const ActorInit En_Horse_Normal_InitVars = {
...
+    this->animationIdx = 0;
+    this->unk_21C = 0;
+    this->unk_21E = 0;
+    this->actor.speedXZ = 0.0f;
+    this->unk_218 = 0.0f;
+    SkelAnime_ChangeAnim(&this->skin.skelAnime, sAnimations[this->animationIdx], func_80A6B30C(this), 0.0f,
+                         SkelAnime_GetFrameCount(&sAnimations[this->animationIdx]->genericHeader), 2, 0.0f);
+}
+
+void EnHorseNormal_Wait(EnHorseNormal* this, GlobalContext* globalCtx) {
```
Dragorn421, line 451: same comment as WaitClone about "Wait"

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.h
```diff
@@ -6,9 +6,31 @@
...
+    /* 0x0210 */ char unk_210[0x08];
+    /* 0x0218 */ f32 unk_218;
+    /* 0x021C */ s16 unk_21C;
+    /* 0x021E */ s16 unk_21E;
+    /* 0x0220 */ f32 unk_220;
+    /* 0x0224 */ char unk_224[0x04];
+    /* 0x0228 */ ColliderCylinder bodyCollider;
+    /* 0x0274 */ ColliderJntSph headCollider;
+    /* 0x0294 */ ColliderJntSphItem headColliderItems[1];
+    /* 0x02D4 */ ColliderCylinder cloneCollider;
```
Dragorn421, line 31: I kind of see why `cloneCollider` is clone-only related (updated in Draw for type 3 only),
but I don't see why `bodyCollider`/`headCollider` are body/head?

Dragorn421 (changes requested):
Thanks for the explanation, it's clearer now

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.h
Sirius902: The original horses (not cloned) have two hitboxes: a cylinder for the body and a sphere for the head. On the other hand, the clones have one giant cylinder hitbox.

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
Sirius902: I think we may need some input from others on this one. I do agree that `Idle` would make more sense (to me at least) though @fig02 said we don't usually use the word `Idle` for naming action functions. Also, I think we usually put a verb first in the name so `IdleCloned` might be better. Everyone please let me know your thoughts.

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.h
Dragorn421: But how do you know they are for head and body?

Sirius902: A mixture of gz hitbox viewer and testing in debug rom changing the sizes of the colliders.

Dragorn421 (approved):
There are comments I brought up which are still unresolved (waiting for others' input), but I'm fine with it as is.

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
```diff
@@ -21,45 +50,633 @@ const ActorInit En_Horse_Normal_InitVars = {
...
+    EnHorseNormal_WaitClone,       EnHorseNormal_FollowPath,
+};
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Horse_Normal/func_80A6BC00.s")
+void EnHorseNormal_Update(Actor* thisx, GlobalContext* globalCtx) {
+    EnHorseNormal* this = THIS;
+    s32 pad;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Horse_Normal/func_80A6BC48.s")
+    sActionFuncs[this->type](this, globalCtx);
```
Roman971, line 539: We normally that the `action`
```suggestion
    sActionFuncs[this->action](this, globalCtx);
```

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
```diff
@@ -1,15 +1,44 @@
...
+    EnHorseNormalUnkStruct1* items;
+} EnHorseNormalUnkStruct2;
+
+typedef enum {
+    /* 0x00 */ HORSE_NULL, // Cycles through animations in sAnimations
+    /* 0x01 */ HORSE_WANDER,
+    /* 0x02 */ HORSE_STATIC,
+    /* 0x03 */ HORSE_STATIC_CLONEABLE,
+    /* 0x04 */ HORSE_FOLLOW_PATH
+} EnHorseNormalType;
```
Roman971, line 29: Based on the names you chose for action functions:
```suggestion
typedef enum {
    /* 0x00 */ HORSE_CYCLE_ANIMATIONS,
    /* 0x01 */ HORSE_WANDER,
    /* 0x02 */ HORSE_WAIT,
    /* 0x03 */ HORSE_WAIT_CLONE,
    /* 0x04 */ HORSE_FOLLOW_PATH
} EnHorseNormalAction;
```

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
```diff
@@ -21,45 +50,633 @@ const ActorInit En_Horse_Normal_InitVars = {
...
+        }
+        func_80A6CC88(globalCtx, this, &clonePos);
+        SkinMatrix_SetScaleRotateYRPTranslate(&skinMtx, this->actor.scale.x, this->actor.scale.y, this->actor.scale.z,
+                                              this->actor.shape.rot.x, cloneRotY, this->actor.shape.rot.z, clonePos.x,
+                                              (this->actor.shape.unk_08 * this->actor.scale.y) + clonePos.y,
+                                              clonePos.z);
+        mtx1 = SkinMatrix_MtxFToNewMtx(globalCtx->state.gfxCtx, &skinMtx);
+        if (mtx1 == NULL) {
+            return;
+        } else {
```
Roman971, line 659: No else here imo

src/overlays/actors/ovl_En_Horse_Normal/z_en_horse_normal.c
Roman971: I agree, the blocks look equivalent to me so it should probably be an else

Roman971 (approved):
Looks good to me code-wise

fig02 (approved):

PR #364, opened 2020-08-30
Decomp Heishi4 All OK

louist103:

PR #404, opened 2020-09-20
Update diff.py asm-differ to commit 9d79eb9 2020-09-08

Dragorn421:
Update to https://github.com/simonlindholm/asm-differ/blob/9d79eb9f539e5fa58fe63c62862661c5d9ce0d27/diff.py

Also fixes the diff.py regression introduced by #364 

Since 22th june and #220 it seems there were mainly fixes, the big new feature being `-3` for "three-way diffs" which shows the previous built asm in a third column

I've used the version before the two latest commits on a few files without issues

### DIFF
--- a/diff.py
+++ b/diff.py
@@ -1,35 +1,9 @@
 #!/usr/bin/env python3
 import sys
-import re
-import os
-import ast
-import argparse
-import subprocess
-import difflib
-import string
-import itertools
-import threading
-import queue
-import time
-
 
 def fail(msg):
     print(msg, file=sys.stderr)
     sys.exit(1)
-
-
-MISSING_PREREQUISITES = (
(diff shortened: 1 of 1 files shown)


fig02:
has stack differences being highlighted as regalloc been fixed? Im ok with still updating even if it hasnt, but that was an issue we were having with one of the latest upgrades. But ive lost track of which is which lmao

Dragorn421:
"Yes" then!

<details>
<summary>
Click for demo
</summary>

`./diff.py -mwo3 Player_IsBurningStickInRange` with OK then removing `s32 pad;` from the function:



</details>

fig02 (approved):

Roman971 (approved):

PR #409, opened 2020-09-22
Move functions/vars into z64audio.h, and don't include it from z64.h

zelda2774:
... plus some small drive-by cleanups.

Reduces the size of preprocessed code for a randomly picked file from 10901 lines to 10129, which hopefully might speed up builds. Let's see in CI!


zelda2774:
image

Bottom one is this PR. Excluding outliers that are probably due to several PRs being built at once, it doesn't seem like this makes a very big difference, interestingly enough. Probably helps a little with permuter C serialization times, but perhaps not to the extent where it would be worth taking this.

zelda2774:
Hm, the Setup step seems to be taking a fair chunk of CI time (roughly half). Looking at just the Build step it's more like 1m51s -> 1m46s, i.e. the ~7% reduction in lines did result in a ~7% compile time win, which makes this more promising.

src/overlays/actors/ovl_En_Zl2/z_en_zl2.c
```diff
@@ -367,34 +366,24 @@ void func_80B4EF64(EnZl2* this, s16 arg1, s32 arg2) {
...
         }
 
-        zero = 0;
         if (arg2 == 2) {
             if ((this->action == 5) || (this->action == 30)) {
                 animCurrentFrame = this->skelAnime.animCurrentFrame;
                 unk_278 = this->unk_278;
                 temp_t0 = (s32)((3500.0f * animCurrentFrame) / unk_278) + phi_a0;
                 if (temp_t0 >= temp_v1) {
-                    if (phi_v0 < zero) {
```
ethteck, line 377: 🤦‍♂️ wow...thank you. (I decomped this)

ethteck (approved):
Anything that further organizes the codebase is a plus in my book, regardless of build time changes.

fig02:
There was some discussion in discord about how separating things this early can be hurtful to our decomp process. If we havent come to an agreement on that yet we should, so that we can figure out where to go from here

fig02:
are we 100% sure that the stuff moved is only internal audio stuff?
As dragorn mentioned in discord, i dont want a scenario where people will have to go searching for the unusual place to find things because of this early split. But if it def is self contained then I will compromise and go with this plan

fig02:
Its been quite a while since we discussed this, but my opinion is still the same that we should probably wait on doing this. I understand the advantage of the improved build, but think its too early to start splitting stuff personally

I would personally like to close this and revisit at a later time, but want to hear others opinions

ethteck:
I'm okay with waiting, but there are some useful zl2 changes here that we should add in another PR in that case. I can do that if we close this

fig02:
sure yeah i agree

fig02:
we've decided on closing this for now and revisiting at a later time

PR #411, opened 2020-09-23
Bg_Hidan_Curtain

petrie911:
Flame Circles


src/overlays/actors/ovl_Bg_Hidan_Curtain/z_bg_hidan_curtain.c
```diff
@@ -21,25 +46,194 @@ const ActorInit Bg_Hidan_Curtain_InitVars = {
...
+    Actor_SetHeight(&this->actor, 20.0f);
+    this->type = (thisx->params >> 0xC) & 0xF; // needed for matching
+    if (this->type >= 7) {
+        // Type is not set
+        osSyncPrintf("Error : object のタイプが設定されていない(%s %d)(arg_data 0x%04x)\n", "../z_bg_hidan_curtain.c", 352, this->actor.params);
+        Actor_Kill(&this->actor);
+        return;
+    }
+
+    this->isSmall = ((this->type == 2) || (this->type == 4)) ? 1 : 0;
```
Dragorn421, line 65: since `isSmall` is used as an index rather than as a boolean maybe naming it like `sizeIdx` makes more sense
Or rather `paramsIdx` since it indexes into `sHCParams` and the size is only the end result of the params

src/overlays/actors/ovl_Bg_Hidan_Curtain/z_bg_hidan_curtain.c
```diff
@@ -21,25 +46,194 @@ const ActorInit Bg_Hidan_Curtain_InitVars = {
...
+    func_80061ED4(&thisx->colChkInfo, NULL, &sCcInfoInit);
+    if (this->type == 0) {
+        this->actionFunc = BgHidanCurtain_WaitForClear;
+    } else {
+        this->actionFunc = BgHidanCurtain_WaitForSwitchOn;
+        if ((this->type == 4)||(this->type == 5)) {
+            this->actor.posRot.pos.y = this->actor.initPosRot.pos.y - sp24->riseDist;
+        }
+    }
+    if(((this->type == 1) && Flags_GetTreasure(globalCtx, this->chestFlag))
```
Dragorn421, line 92:
```suggestion
    if(((this->type == 1) && Flags_GetTreasure(globalCtx, this->treasureFlag))
```

src/overlays/actors/ovl_Bg_Hidan_Curtain/z_bg_hidan_curtain.c
```diff
@@ -21,25 +46,194 @@ const ActorInit Bg_Hidan_Curtain_InitVars = {
...
+    this->type = (thisx->params >> 0xC) & 0xF; // needed for matching
+    if (this->type >= 7) {
+        // Type is not set
+        osSyncPrintf("Error : object のタイプが設定されていない(%s %d)(arg_data 0x%04x)\n", "../z_bg_hidan_curtain.c", 352, this->actor.params);
+        Actor_Kill(&this->actor);
+        return;
+    }
+
+    this->isSmall = ((this->type == 2) || (this->type == 4)) ? 1 : 0;
+    sp24 = &sHCParams[this->isSmall];
```
Dragorn421, line 66:
```suggestion
    curParams = &sHCParams[this->isSmall];
```
or smth

src/overlays/actors/ovl_Bg_Hidan_Curtain/z_bg_hidan_curtain.c
```diff
@@ -21,25 +46,194 @@ const ActorInit Bg_Hidan_Curtain_InitVars = {
...
+    this->collider.dim.pos.z = this->actor.posRot.pos.z;
+    this->collider.dim.radius = sp24->radius;
+    this->collider.dim.height = sp24->height;
+    Collider_CylinderUpdate(&this->actor, &this->collider);
+    func_80061ED4(&thisx->colChkInfo, NULL, &sCcInfoInit);
+    if (this->type == 0) {
+        this->actionFunc = BgHidanCurtain_WaitForClear;
+    } else {
+        this->actionFunc = BgHidanCurtain_WaitForSwitchOn;
+        if ((this->type == 4)||(this->type == 5)) {
```
Dragorn421, line 88:
```suggestion
        if ((this->type == 4) || (this->type == 5)) {
```

src/overlays/actors/ovl_Bg_Hidan_Curtain/z_bg_hidan_curtain.c
```diff
@@ -21,25 +46,194 @@ const ActorInit Bg_Hidan_Curtain_InitVars = {
...
+            this->actor.posRot.pos.y = (2.0f * this->actor.initPosRot.pos.y) - sp34->riseDist - this->actor.posRot.pos.y;
+        }
+        
+        this->actionFunc(this, globalCtx2);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Hidan_Curtain/func_80885604.s")
+        if ((this->type == 4) || (this->type == 5)) {
+            this->actor.posRot.pos.y = (2.0f * this->actor.initPosRot.pos.y) - sp34->riseDist - this->actor.posRot.pos.y;
+        }
+        temp_f2 = ( sp34->riseDist - (this->actor.initPosRot.pos.y - this->actor.posRot.pos.y)) /sp34->riseDist;
```
Dragorn421, line 204:
```suggestion
        temp_f2 = (sp34->riseDist - (this->actor.initPosRot.pos.y - this->actor.posRot.pos.y)) / sp34->riseDist;
```

src/overlays/actors/ovl_Bg_Hidan_Curtain/z_bg_hidan_curtain.c
```diff
@@ -21,25 +46,194 @@ const ActorInit Bg_Hidan_Curtain_InitVars = {
...
+            } else {// this->type == 1
+                this->timer = 300;
+            }
+            this->actionFunc = BgHidanCurtain_Timer;
+        }
+    }
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Hidan_Curtain/func_80885514.s")
+void BgHidanCurtain_Timer(BgHidanCurtain *this, GlobalContext *globalCtx) {
```
Dragorn421, line 172: maybe this could be named more explicitly like `WaitTurnOn`? idk

Dragorn421 (changes requested):

Roman971 (changes requested):
The format script needs to be run on this file. I'll wait for that before reviewing so my comments don't go outdated.

src/overlays/actors/ovl_Bg_Hidan_Curtain/z_bg_hidan_curtain.c
```diff
@@ -21,25 +45,201 @@ const ActorInit Bg_Hidan_Curtain_InitVars = {
...
+    if (this->type > 6) {
+        // Type is not set
+        osSyncPrintf("Error : object のタイプが設定されていない(%s %d)(arg_data 0x%04x)\n", "../z_bg_hidan_curtain.c",
+                     352, this->actor.params);
+        Actor_Kill(&this->actor);
+        return;
+    }
+
+    this->sizeIndex = ((this->type == 2) || (this->type == 4)) ? 1 : 0;
+    hcParams = &sHCParams[this->sizeIndex];
```
Dragorn421, line 66: I still think `paramsIndex` makes more sense because
> it indexes into sHCParams and the size is only the end result of the params

the params also include more than just "size" unless you mean "subtype" by "size" but that's confusing imo

src/overlays/actors/ovl_Bg_Hidan_Curtain/z_bg_hidan_curtain.c
```diff
@@ -21,25 +45,201 @@ const ActorInit Bg_Hidan_Curtain_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Hidan_Curtain/func_80885434.s")
+        if ((this->type == 4) || (this->type == 5)) {
+            this->actor.posRot.pos.y =
+                (2.0f * this->actor.initPosRot.pos.y) - hcParams->riseDist - this->actor.posRot.pos.y;
+        }
+        riseStatus =
+            (hcParams->riseDist - (this->actor.initPosRot.pos.y - this->actor.posRot.pos.y)) / hcParams->riseDist;
+        this->alpha = 255.0f * riseStatus;
+        if (this->alpha > 50) {
+            this->collider.dim.height = hcParams->height * riseStatus;
```
Dragorn421, line 212: I would use `riseProgress` (more percentage-like) rather than `riseStatus` (more "A or B"-like)

src/overlays/actors/ovl_Bg_Hidan_Curtain/z_bg_hidan_curtain.h
```diff
@@ -6,9 +6,26 @@
...
+	/* 0x04 */ f32 scale;
+	/* 0x08 */ f32 riseDist;
+	/* 0x0C */ f32 riseSpeed;
+} BgHidanCurtainParams; // size = 0x10
+
 typedef struct BgHidanCurtain {
     /* 0x0000 */ Actor actor;
-    /* 0x014C */ char unk_14C[0x58];
+    /* 0x014C */ BgHidanCurtainActionFunc actionFunc;
+	/* 0x0150 */ u8 type;
```
Dragorn421, line 22: weird formatting here

src/overlays/actors/ovl_Bg_Hidan_Curtain/z_bg_hidan_curtain.c
```diff
@@ -21,25 +45,201 @@ const ActorInit Bg_Hidan_Curtain_InitVars = {
...
+            this->actionFunc = BgHidanCurtain_WaitForCutscene;
+            func_800800F8(globalCtx, 3360, 60, &this->actor, 0);
+            this->timer = 30;
+        } else {
+            this->actionFunc = BgHidanCurtain_TurnOff;
+        }
+    }
+}
+
+void BgHidanCurtain_WaitForCutscene(BgHidanCurtain* this, GlobalContext* globalCtx) {
```
Dragorn421, line 123: I don't think that function should be named for what it is intended to do (wait for the cutscene to end) when all it does is wait on the timer like `WaitForTimer`
Maybe call this `WaitOffTimer` and rename `WaitForTimer` to `WaitOnTimer`? `TurnOffAfterTimer`/`TurnOnAfterTimer`?
Or just get rid of `Timer` in the name idk

Dragorn421 (changes requested):

src/overlays/actors/ovl_Bg_Hidan_Curtain/z_bg_hidan_curtain.c
```diff
@@ -21,25 +45,201 @@ const ActorInit Bg_Hidan_Curtain_InitVars = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Hidan_Curtain/BgHidanCurtain_Destroy.s")
+void BgHidanCurtain_Init(Actor* thisx, GlobalContext* globalCtx) {
+    s32 pad;
+    BgHidanCurtain* this = THIS;
+    BgHidanCurtainParams* hcParams;
+
+    osSyncPrintf("Curtain (arg_data 0x%04x)\n", this->actor.params);
+    Actor_SetHeight(&this->actor, 20.0f);
+    this->type = (thisx->params >> 0xC) & 0xF; // needed for matching
```
Roman971, line 56: I don't think the comment is necessary here
```suggestion
    this->type = (thisx->params >> 0xC) & 0xF;
```

src/overlays/actors/ovl_Bg_Hidan_Curtain/z_bg_hidan_curtain.c
```diff
@@ -21,25 +45,201 @@ const ActorInit Bg_Hidan_Curtain_InitVars = {
...
+            func_800800F8(globalCtx, 3360, 60, &this->actor, 0);
+            this->timer = 30;
+        } else {
+            this->actionFunc = BgHidanCurtain_TurnOff;
+        }
+    }
+}
+
+void BgHidanCurtain_WaitForCutscene(BgHidanCurtain* this, GlobalContext* globalCtx) {
+    if (this->timer-- == 0) {
```
Roman971, line 124: Does the `--` have to be inline? Otherwise it shouldn't be imo

src/overlays/actors/ovl_Bg_Hidan_Curtain/z_bg_hidan_curtain.c
```diff
@@ -21,25 +45,201 @@ const ActorInit Bg_Hidan_Curtain_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Hidan_Curtain/func_80885554.s")
+    gDPSetPrimColor(oGfxCtx->polyXlu.p++, 0x80, 0x80, 255, 220, 0, this->alpha);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Hidan_Curtain/func_80885594.s")
+    gDPSetEnvColor(oGfxCtx->polyXlu.p++, 255, 0, 0, 0);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Hidan_Curtain/func_80885604.s")
+    gSPSegment(oGfxCtx->polyXlu.p++, 0x08,
+               Gfx_TwoTexScroll(globalCtx->state.gfxCtx, 0, this->texScroll & 0x7F, 0, 0x20, 0x40, 1, 0,
+                                (u8)(this->texScroll * -0xF), 0x20, 0x40));
```
Roman971, line 237: This is slightly better imo (if it matches)
```suggestion
                                (this->texScroll * -0xF) & 0xFF, 0x20, 0x40));
```

src/overlays/actors/ovl_Bg_Hidan_Curtain/z_bg_hidan_curtain.h
```diff
@@ -6,9 +6,26 @@
 
 struct BgHidanCurtain;
 
+typedef void (*BgHidanCurtainActionFunc)(struct BgHidanCurtain*, GlobalContext*);
+
+typedef struct BgHidanCurtainParams  {
```
Roman971, line 11:
```suggestion
typedef struct {
```

src/overlays/actors/ovl_Bg_Hidan_Curtain/z_bg_hidan_curtain.h
```diff
@@ -6,9 +6,26 @@
...
+	/* 0x00 */ s16 radius;
+	/* 0x02 */ s16 height;
+	/* 0x04 */ f32 scale;
+	/* 0x08 */ f32 riseDist;
+	/* 0x0C */ f32 riseSpeed;
+} BgHidanCurtainParams; // size = 0x10
```
Roman971, line 17: This struct should probably be only in the C file since it's strictly internal

src/overlays/actors/ovl_Bg_Hidan_Curtain/z_bg_hidan_curtain.c
Dragorn421: How would you write something equivalent without inlining the -- ?

Roman971: I was thinking it could probably be written differently since there isn't any function call and such here but I guess probably not. Unfortunate

Roman971 (approved):

fig02 (approved):

PR #414, opened 2020-09-25
BgSpot03Taki

Electrk:


src/overlays/actors/ovl_Bg_Spot03_Taki/z_bg_spot03_taki.h
```diff
@@ -6,9 +6,16 @@
...
 typedef struct BgSpot03Taki {
-    /* 0x0000 */ Actor actor;
-    /* 0x014C */ char unk_14C[0x2C];
+    /* 0x0000 */ DynaPolyActor dyna;
+    /* 0x0164 */ BgSpot03TakiActionFunc actionFunc;
+    /* 0x0168 */ s16 unk_168;
+    /* 0x016A */ u8 unk_16A;
+    /* 0x016C */ u16 switchFlag;
+    /* 0x0170 */ f32 unk_170;
+    /* 0x0174 */ u8 unk_174;
```
mzxrules, line 18:
```suggestion
    /* 0x0168 */ s16 timer;
    /* 0x016A */ u8 state;
    /* 0x016C */ u16 switchFlag;
    /* 0x0170 */ f32 waterfallOpeningAlpha;
    /* 0x0174 */ u8 bufferIndex;
```

src/overlays/actors/ovl_Bg_Spot03_Taki/z_bg_spot03_taki.c
```diff
@@ -27,15 +31,141 @@ const ActorInit Bg_Spot03_Taki_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Spot03_Taki/func_808ADAE0.s")
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Spot03_Taki/BgSpot03Taki_Init.s")
+static InitChainEntry sInitChain[] = {
+    ICHAIN_VEC3F_DIV1000(scale, 100, ICHAIN_STOP),
+};
+
+extern UNK_TYPE D_06000C98;
+extern Vtx* D_06000800[];
+extern Vtx* D_06000990[];
```
mzxrules, line 41: D_06000800 and D_06000990 are identical Vtx data for the waterfall. D_06000800 is buffer 0, D_06000990 is buffer 1

src/overlays/actors/ovl_Bg_Spot03_Taki/z_bg_spot03_taki.c
```diff
@@ -27,15 +31,141 @@ const ActorInit Bg_Spot03_Taki_InitVars = {
...
+    BgSpot03Taki* this = THIS;
+
+    DynaPolyInfo_Free(globalCtx, &globalCtx->colCtx.dyna, this->dyna.dynaPolyId);
+}
+
+void func_808ADEF0(BgSpot03Taki* this, GlobalContext* globalCtx) {
+    if (this->unk_16A == 0) {
+        if (Flags_GetSwitch(globalCtx, this->switchFlag) != 0) {
+            this->unk_16A = 2;
+            this->unk_168 = 0x28;
```
mzxrules, line 81: unk_168 is a frame timer, so assignments should be converted to decimal numbers

src/overlays/actors/ovl_Bg_Spot03_Taki/z_bg_spot03_taki.h
mzxrules: Additionally, you could create `#define`s for `unk_16A`'s different values, putting them near the top of the .c file. Something like this should do:

    #define STATE_CLOSED 0
    #define STATE_OPENING_IDLE 1
    #define STATE_OPENING_ANIMATED 2
    #define STATE_OPENED 3
    #define STATE_CLOSING 4

Electrk: I'm 99% sure unk_174 is a boolean, especially considering the toggle near the end of Draw:

```c
// ...
    CLOSE_DISPS(globalCtx->state.gfxCtx, "../z_bg_spot03_taki.c", 358);

    this->unk_174 = this->unk_174 == 0;
// ...
```

src/overlays/actors/ovl_Bg_Spot03_Taki/z_bg_spot03_taki.c
Electrk: So should I rename these, or is this just a remark?

mzxrules: I'd just add a comment, since we are still figuring out a naming scheme for object file variables

src/overlays/actors/ovl_Bg_Spot03_Taki/z_bg_spot03_taki.h
mzxrules: this line doesn't represent the whole picture. Take a closer look at where the variable's state is being read:

In `func_808ADEF0` (called by the update function), it's passed into arg1 of `func_808ADAE0`:

    func_808ADAE0(this, this->unk_174);

`func_808ADAE0` uses it's state to select which vtx data to update, choosing either `D_06000800`  (0) or `D_06000990` (1). Later on, the draw routine uses it to select which Vertex list to pass to the RCP for it's next draw task.

    if (this->unk_174 == 0) {
        gSPVertex(oGfxCtx->polyXlu.p++, &D_06000800, 25, 0);
    } else {
        gSPVertex(oGfxCtx->polyXlu.p++, &D_06000990, 25, 0);
    }

Once again, it uses `unk_174` to choose either `D_06000800`  (0) or `D_06000990` (1). It doesn't make sense to assign true/false to these Vtx data.

One last thing is that you might be able to simplify `this->unk_174 = this->unk_174 == 0;` to `this->unk_174 != this->unk_174;`

Electrk: > Once again, it uses unk_174 to choose either D_06000800 (0) or D_06000990 (1). It doesn't make sense to assign true/false to these Vtx data.

It makes perfect sense and you show so in the first sentence. unk_174 is always either 0 or 1 - true or false.

In C, 0 is false and everything that is not 0 is true, which it does consistently throughout the file. If unk_174 is 0, it does stuff with D_06000800, and if unk_174 is anything else, it does stuff with D_06000990. It only ever has two states.

Not only that, but it sets itself to a boolean value in that toggle line I pointed out. `this->unk_174 == 0` evaluates to a boolean value, which it sets unk_174 to be.

I don't get how you can say it doesn't make sense to be a boolean when it only ever has two states and literally sets itself to a boolean value in a function.

mzxrules: does it make sense to you to call D_06000800 the false Vtx data, and D_06000990 the true Vtx data?

Electrk: > 
> 
> Additionally, you could create `#define`s for `unk_16A`'s different values, putting them near the top of the .c file. Something like this should do:
> 
> ```
> #define STATE_CLOSED 0
> #define STATE_OPENING_IDLE 1
> #define STATE_OPENING_ANIMATED 2
> #define STATE_OPENED 3
> #define STATE_CLOSING 4
> ```

What about using an enum like @Roman971 [suggested](https://github.com/zeldaret/oot/pull/386#discussion_r495467791) for the Dodongo's Cavern platform?

src/overlays/actors/ovl_Bg_Spot03_Taki/z_bg_spot03_taki.c
```diff
@@ -27,15 +31,141 @@ const ActorInit Bg_Spot03_Taki_InitVars = {
...
+    ICHAIN_VEC3F_DIV1000(scale, 100, ICHAIN_STOP),
+};
+
+extern UNK_TYPE D_06000C98;
+extern Vtx* D_06000800[];
+extern Vtx* D_06000990[];
+extern Gfx* D_06000B20[];
+extern Gfx* D_06001580[];
+
+void func_808ADAE0(BgSpot03Taki* this, s32 arg0) {
```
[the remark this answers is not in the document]
Electrk: Thinking of calling this function something like BgSpot03Taki_ApplyOpeningAlpha since it's not _setting_ the alpha, just applying it to the current vertex buffer. Or is it not kosher to be renaming functions right now?

src/overlays/actors/ovl_Bg_Spot03_Taki/z_bg_spot03_taki.h
mzxrules: yea, I would do the same.

src/overlays/actors/ovl_Bg_Spot03_Taki/z_bg_spot03_taki.c
```diff
@@ -10,12 +10,15 @@
 
 #define THIS ((BgSpot03Taki*)thisx)
 
+void BgSpot03Taki_ApplyOpeningAlpha(BgSpot03Taki* this, s32 bufferIndex);
+
```
Roman971, line 14: We usually don't forward declare functions unless it's necessary. So in this case the function is defined before it's used so it doesn't have to be forward declared.
```suggestion
```

src/overlays/actors/ovl_Bg_Spot03_Taki/z_bg_spot03_taki.c
```diff
@@ -27,15 +30,142 @@ const ActorInit Bg_Spot03_Taki_InitVars = {
...
+// These are identical vertex data for the waterfall.
+extern Vtx* D_06000800[]; // Vertex buffer 0
+extern Vtx* D_06000990[]; // Vertex buffer 1
+
+extern Gfx* D_06000B20[];
+extern Gfx* D_06000BC0[];
+extern Gfx* D_06001580[];
+
+void BgSpot03Taki_ApplyOpeningAlpha(BgSpot03Taki* this, s32 bufferIndex) {
+    int i;
```
Roman971, line 49:
```suggestion
    s32 i;
```

src/overlays/actors/ovl_Bg_Spot03_Taki/z_bg_spot03_taki.h
```diff
@@ -4,11 +4,26 @@
...
+    WATERFALL_CLOSED,
+    WATERFALL_OPENING_IDLE,
+    WATERFALL_OPENING_ANIMATED,
+    WATERFALL_OPENED,
+    WATERFALL_CLOSING,
+} BgSpot03TakiState;
+
 struct BgSpot03Taki;
 
+typedef void (*BgSpot03TakiActionFunc)(struct BgSpot03Taki*, struct GlobalContext*);
```
Roman971, line 17:
```suggestion
typedef void (*BgSpot03TakiActionFunc)(struct BgSpot03Taki*, GlobalContext*);
```

src/overlays/actors/ovl_Bg_Spot03_Taki/z_bg_spot03_taki.c
```diff
@@ -27,15 +30,142 @@ const ActorInit Bg_Spot03_Taki_InitVars = {
...
+    gSPMatrix(oGfxCtx->polyXlu.p++, Matrix_NewMtx(globalCtx->state.gfxCtx, "../z_bg_spot03_taki.c", 325),
+              G_MTX_NOPUSH | G_MTX_LOAD | G_MTX_MODELVIEW);
+
+    func_80093D84(globalCtx->state.gfxCtx);
+
+    gSPSegment(
+        oGfxCtx->polyXlu.p++, 0x08,
+        Gfx_TwoTexScroll(globalCtx->state.gfxCtx, 0, 0, gameplayFrames * 5, 64, 64, 1, 0, gameplayFrames * 5, 64, 64));
+
+    gSPDisplayList(oGfxCtx->polyXlu.p++, &D_06000B20);
```
Roman971, line 146: The `&` is omitted for arrays (and for function pointers). And this is relevant in the rest of the function.
```suggestion
    gSPDisplayList(oGfxCtx->polyXlu.p++, D_06000B20);
```

Roman971 (changes requested):
Well done on this actor. Just a few things

src/overlays/actors/ovl_Bg_Spot03_Taki/z_bg_spot03_taki.h
Electrk: fixed but fyi there are a few files that make this same mistake

src/overlays/actors/ovl_Bg_Spot03_Taki/z_bg_spot03_taki.c
Electrk: Is this true for `DynaPolyInfo_Alloc` too? Almost all the uses of this function use `&`

Roman971: It's only true for arrays (eg. display lists are of type `Gfx[]`) but the argument of `DynaPolyInfo_Alloc` isn't an array so this doesn't apply.

Electrk: Right

Roman971 (approved):

fig02 (approved):

fig02:
undefined syms conflict, will merge when resolved 
(these are really easy to fix will take 2 secs)

fig02:
fixed conflict, merging

PR #256, opened 2020-07-14
z_bgcheck.c, 800430A0.c, 80043480.c

mzxrules:
still wip but close to done. Making a pull request now to make people aware of it.

Edit: I've stopped actively trying to match functions, and have begun work documenting each function to the best of my ability in order to get a big picture sense of what everything is doing. 

Edit2: Implemented a breaking change that require modifications to ZAP. Waiting on Jack to finish his modifications before pushing my necessary changes

PR #418, opened 2020-09-26
code/z_bgcheck.c: decomp DynaPolyInfo_GetActor(), func_8003EBF8() and func_8003EC50()

Peppers1612:
A few functions that I encountered while working on ObjTimeblock. I thought I could bring them as-is, since they are so small.


fig02:
this whole file is an open PR already https://github.com/zeldaret/oot/pull/256
thank you though

Peppers1612:
Good thing I didn't spend more than a few minutes on those. Thanks for letting me know.

PR #426, opened 2020-10-01
BGM enum

petrie911:
Since we'll need it at some point, I made an enum for BGMs like we have for SFX out of the Cloudmodding listing. Names use filenames like SFX, though I had to make up ones for the Hyrule Field segments.


fig02 (approved):
My opinion on this has changed over time, setting BGM is not something that happens all that often so it wouldnt be too troublesome to introduce now I think.

That said, there are probably some usages of this enum that need to be added into this pr

fig02:
nvm, gonna close for now. seems best to wait until the code for parsing bgm commands is decompiled

PR #428, opened 2020-10-01
z_en_kakasi.c OK

i82orbom:
To whoever reviews this PR, there's other z_en_kakasiX.c PRs open, they're very similar in content as there was a lot of copy & pasta


src/overlays/actors/ovl_En_Kakasi/z_en_kakasi.h
```diff
@@ -6,9 +6,28 @@
...
+    /* 0x0194 */ u8 unk_194;
+    /* 0x0195 */ char unk_195;
+    /* 0x0196 */ s16 unk_196;
+    /* 0x0198 */ s16 unk_198;
+    /* 0x019A */ s16 unk_19A;
+    /* 0x019C */ s16 unk_19C[4];
+    /* 0x01A4 */ s16 unk_1A4;
+    /* 0x01A6 */ s16 unk_1A6;
+    /* 0x01A8 */ s16 unk_1A8;
+    /* 0x01AA */ char unk_1AA[0x2];
```
i82orbom, line 24: This seems to be required to match

src/overlays/actors/ovl_En_Kakasi/z_en_kakasi.c
```diff
@@ -21,27 +43,301 @@ const ActorInit En_Kakasi_InitVars = {
...
+    Player* player = PLAYER;
+    s16 angleTowardsLink;
+    s16 absAngleTowardsLink;
+
+    func_80A8F28C(this);
+    SkelAnime_FrameUpdateMatrix(&this->skelanime);
+    this->camId = -1;
+    if (func_8002F194(&this->actor, globalCtx)) {
+        if (this->unk_196 == 5) {
+            this->actionFunc = &func_80A8F9C8;
```
Dragorn421, line 196:
```suggestion
            this->actionFunc = func_80A8F9C8;
```
no need for `&` for function pointers

also in  a few other places in the file

src/overlays/actors/ovl_En_Kakasi/z_en_kakasi.h
Dragorn421: It probably is an unused s16, it can't be 2 bytes of padding in the middle of s16s
leaving it as char[] is ok

src/overlays/actors/ovl_En_Kakasi/z_en_kakasi.h
```diff
@@ -6,9 +6,28 @@
...
 
+typedef void (*EnKakasiFunc)(struct EnKakasi*, GlobalContext*);
+
 typedef struct EnKakasi {
     /* 0x0000 */ Actor actor;
-    /* 0x014C */ char unk_14C[0xC0];
+    /* 0x014C */ EnKakasiFunc actionFunc;
+    /* 0x0150 */ SkelAnime skelanime;
+    /* 0x0194 */ u8 unk_194;
+    /* 0x0195 */ char unk_195;
```
Dragorn421, line 16: this `char unk_195` can be removed though,
```c
    /* 0x0194 */ u8 unk_194; // ends at 0x0195
    /* 0x0196 */ s16 unk_196; // 2-bytes-wide type, aligned to nearest multiple of 2 aka 0x0196
```

Dragorn421 (changes requested):

Dragorn421 (approved):

src/overlays/actors/ovl_En_Kakasi/z_en_kakasi.c
```diff
@@ -21,27 +43,301 @@ const ActorInit En_Kakasi_InitVars = {
...
+            this->unk_196 = 5;
+        }
+    }
+    this->actionFunc = func_80A8F75C;
+}
+
+void func_80A8F75C(EnKakasi* this, GlobalContext* globalCtx) {
+    Player* player = PLAYER;
+    s16 angleTowardsLink;
+    s16 absAngleTowardsLink;
```
fig02, line 189: imo better to be specific and say yaw rather than just angle

fig02 (approved):

src/overlays/actors/ovl_En_Kakasi/z_en_kakasi.c
```diff
@@ -21,27 +43,301 @@ const ActorInit En_Kakasi_InitVars = {
...
+    this->actionFunc = func_80A8F660;
+}
+
+void func_80A8F28C(EnKakasi* this) {
+    this->unk_1A4 = 0;
+    this->skelanime.animPlaybackSpeed = 0.0f;
+    this->unk_1A8 = this->unk_1AC = 0x0;
+
+    Math_SmoothDownscaleMaxF(&this->skelanime.animCurrentFrame, 0.5f, 1.0f);
+    Math_SmoothScaleMaxMinS(&this->actor.shape.rot.x, this->rot.x, 0x5, 0x2710, 0x0);
```
Roman971, line 79: I can understand why we should probably use hex for rotations (tho I don't really agree with it), but I don't think we should use it for things like 0x0 or 0x5.

Roman971: Mostly saying this now so we don't start going in the wrong direction. I know you changed it because of a previous review comment, so sorry about that.

i82orbom: I also find weird that rotations tbh, so small values decimal, big values hex? what's a big value, though?

Roman971: I'd say 10 or above, since that's where hex starts making a difference other than the leading 0x

i82orbom: 👍

Roman971 (approved):

src/overlays/actors/ovl_En_Kakasi/z_en_kakasi.c
Dragorn421: yeah usually any hex value < 10 is written without the 0x prefix, not just rotations

rotations in hex make sense imo because the usual values like 90° are nice in hex (`0x4000`)
So you understand easier what `0x2710` (here) is, somewhere between 45° (`0x2000`) and ~70° (`0x3000`)
where `10000 = 0x2710` is to be compared to `8192 = 0x2000` and `12288 = 0x3000` which are way less nice values in decimal

it's just like radians, I have no clue what `0.1` radians is, but you know `pi / 4` radians is 45°

PR #431, opened 2020-10-02
Decompile z_sram.c

Random06457:
2 non-matchings left.

Additionaly, since I couldn't find anything better I left some original names which are not ideal. Therefor I'd like some input on it.

Names that might need to change:

Struct fields:
- `SaveSub1C.sceneDataId` (accessed though a "SCENE_DATA_ID" macro)
- `SaveContext.memory` (original name)
- `SaveContext.memory.information` (original name)

offsetof macros:
- `SAVECT`, `NAME`, `SAVE_64DD`, `LIFE`, `ITEM`, `HEART` (original names)
- `HEALTH` (not an original name)

sizeof macros:
- `SAVE_MAX` (original name)
- `SAVE_INFO_SIZE` (not an original name)


fig02:
some preliminary naming ideas. Im not suggesting these be the final names, should be heavily discussed:

`memory` substruct -> `save`. This is the part of save context that actually gets restored when you load a save.
`information` -> `info` just to shorten it until we have a better idea for something less generic

I cant think of a name for sub1C even tho it seems like theres a clear pattern to that data..

`gSaveContext` -> something else. Can name it after sram maybe? Session was another idea because everything beyond `memory` is not saved between game sessions. CommonData is also a meme suggestion but it is data that is common across gamestates..

`SAVE_MAX` -> `SAVE_SIZE` assuming that we change `memory` to `save`. The thing currently named `SAVE_SIZE` can become `SESSION_SIZE` or whatever we end up calling the whole SaveContext.

`SAVE_OFF` -> `SAVE_OFFSET` it doesnt make it much longer and makes it more clear. Same goes for all of the temps in the code called `off` imo just go for the whole word

Some function naming ideas. I personally would like to move away from the "Slot" terminology, not sure how others feel

`Sram_InitSaveSlotDefault` -> `Sram_SetupNewSave` I personally would like to make it clear that its a new blank save not just default. I also think that this and the debug one can be considered more of a setup that gets called by InitSave, but im not 100% on it.
`Sram_InitSaveSlotDebug` -> Sram_SetupDebugSave`
`Sram_OpenSaveSlot` -> `Sram_OpenSave`
`Sram_WriteSaveSlot` -> `Sram_WriteSave` this could also technically be SaveSession or whatever we end up calling the SaveContext since it technically writes the whole thing, then only reads back the "save" part.
`Sram_LoadSaveSlots` -> `Sram_LoadAllSaves` maybe make opening a singlular save
`Sram_InitSaveSlot` -> `Sram_InitSave` this func calls the respective setup func for new or debug save
`Sram_ClearSaveSlot` -> `Sram_EraseSave` erase is what the button is called in file select, but clear is good too.
`Sram_CopySaveSlot` -> `Sram_CopySave` 

Looking at this list you might even be able to just remove the word Save from all of them.. i havent made up my mind on anything yet, just putting all of my thoughts on paper FeelsOkayMan

fig02:
Also, is it possible to move the nested sub structs out of the bigger save context like we do with other things? For me that would be preferable I think. And imo a z64Save or z64Sram.h is warranted

fig02:
I implemented all of my ideas and submitted a pull request to your branch. The diff is kinda big cause I merged master to fix conflicts and it has all of the libultra cleanup changes lol

Prob easier to read from here:
https://github.com/fig02/oot/blob/z_sram/src/code/z_sram.c
https://github.com/fig02/oot/blob/z_sram/include/z64save.h

fig02 (approved):

src/code/z_sram.c
```diff
@@ -1,25 +1,884 @@
...
+    /* 0x18 */ s32 unk_18;    // increments with numDays, gets reset by goron for bgs and one other use
+    /* 0x1C */ SaveInfo info; // "information"
+} Save;                       // size = 0x1354
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_sram/func_800A8A20.s")
+#define SAVE_PLAYER_DATA (*((SavePlayerData*)&gSaveContext.newf))
+#define SAVE_INFO (*((SaveInfo*)&gSaveContext.newf))
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_sram/func_800A9258.s")
+#define FIELD_SIZE(type, field) sizeof(((type*)0)->field)
```
Roman971, line 74: Not sure I get the need for this macro, why not just use sizeof on the type instead?

src/code/z_sram.c
```diff
@@ -1,25 +1,884 @@
...
+#define HEALTH_CAP OFFSETOF(SaveContext, healthCapacity)
+#define QUEST OFFSETOF(SaveContext, inventory.questItems)
+#define DEFENSE OFFSETOF(SaveContext, inventory.defenseHearts)
+#define HEALTH OFFSETOF(SaveContext, health)
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_sram/func_800A9A9C.s")
+#define SLOT_OFFSET(idx) (SLOT_SIZE * idx + 0x20)
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_sram/func_800A9AD0.s")
+#define CALC_CHECKSUM(i, j, ptr, sum)         \
```
Roman971, line 92: Really not a fan of putting code inside macros like that, unless it's necessary ofc, but maybe that's just me.

src/code/z_sram.c
```diff
@@ -1,25 +1,884 @@
...
+    }
+
+    gSaveContext.entranceIndex = 0xCD;
+    gSaveContext.magicLevel = 0;
+    gSaveContext.sceneFlags[5].swch = 0x40000000;
+}
+
+#ifdef NON_MATCHING
+// regalloc differences at the end
+/**
```
Roman971, line 304: Imo it's better to have function level documentation outside the NON_MATCHING block since it's relevant regardless.

src/code/z_sram.c
```diff
@@ -1,25 +1,884 @@
...
+
+/**
+ *  For all 3 slots, verify that the checksum is correct. If corrupted, attempt to load a backup save.
+ *  If backup is also corrupted, default to a new save (or debug save for slot 0 on debug rom).
+ *
+ *  After verifying all 3 saves, pass relevant data to File Select to be displayed.
+ */
+#ifdef NON_MATCHING
+// There's a problem with how "offset" is loaded
+void Sram_VerifyAndLoadAllSaves(FileChooseContext* fileChoose, Sram* sram) {
```
Roman971, line 511:
```suggestion
void Sram_VerifyAndLoadAllSaves(FileChooseContext* fileChooseCtx, Sram* sram) {
```

include/z64player.h
```diff
@@ -5,6 +5,14 @@
 
 struct Player;
 
+typedef enum {
+    /* 0x00 */ PLAYER_SWORD_NONE,
+    /* 0x01 */ PLAYER_SWORD_KOKIRI,
+    /* 0x02 */ PLAYER_SWORD_MASTER,
+    /* 0x03 */ PLAYER_SWORD_LONG,
+    /* 0x04 */ PLAYER_SWORD_MAX
+} PlayerSword;
```
Roman971, line 14: As far as I know, there is nothing to consider as player sword values. The other "equipment" enums dedicated to player are associated to `currentShield`/`currentTunic`/`currentBoots`/`currentMask` in the Player struct (not the actual equip values stored in save context, which is a very important difference), but `currentSword` is just the item ID so there is no use for a player sword enum.

src/overlays/actors/ovl_En_Md/z_en_md.c
```diff
@@ -339,7 +339,8 @@ u16 EnMd_GetTextKokiriForest(GlobalContext* globalCtx, EnMd* this) {
         return 0x1034;
     }
 
-    if ((CUR_EQUIP_VALUE(EQUIP_SHIELD) == 1) && (CUR_EQUIP_VALUE(EQUIP_SWORD) == 1)) {
+    if ((CUR_EQUIP_VALUE(EQUIP_SHIELD) == PLAYER_SHIELD_DEKU) &&
+        (CUR_EQUIP_VALUE(EQUIP_SWORD) == PLAYER_SWORD_KOKIRI)) {
```
Roman971, line 343: This isn't quite right since player enums don't necessarily correspond to equip values and it's not what they are intended for. We don't have enums for equip values right now, tho maybe we should add them, but that's a different story and would probably better kept for a separate set of changes.

src/overlays/actors/ovl_En_Kz/z_en_kz.c
```diff
@@ -71,7 +71,7 @@ u16 EnKz_GetTextNoMaskAdult(GlobalContext* globalCtx, EnKz* this) {
 
     if (INV_CONTENT(ITEM_TRADE_ADULT) >= ITEM_FROG) {
         if (!(gSaveContext.infTable[19] & 0x200)) {
-            if (CHECK_OWNED_EQUIP(EQUIP_TUNIC, 2)) {
+            if (CHECK_OWNED_EQUIP(EQUIP_TUNIC, PLAYER_TUNIC_ZORA)) {
```
Roman971, line 74: For the same reason, the player enum isn't supposed to be used here.
Also note that it's a bit confusing right now due to the name of its arguments but `CHECK_OWNED_EQUIP` doesn't actually take an "equip value". It takes an index of the equipment to check for, with 0 being the first one (so 0 for kokiri tunic, etc). Whereas, for instance, `CUR_EQUIP_VALUE` returns 0 for none equipped (1 for kokiri tunic, etc).

src/overlays/actors/ovl_En_Kz/z_en_kz.c
```diff
@@ -245,7 +245,7 @@ void func_80A9CB18(EnKz* this, GlobalContext* globalCtx) {
                 this->actor.textId = CHECK_QUEST_ITEM(QUEST_SONG_SERENADE) ? 0x4045 : 0x401A;
                 player->actor.textId = this->actor.textId;
             } else {
-                this->actor.textId = CHECK_OWNED_EQUIP(EQUIP_TUNIC, 2) ? 0x401F : 0x4012;
+                this->actor.textId = CHECK_OWNED_EQUIP(EQUIP_TUNIC, PLAYER_TUNIC_ZORA) ? 0x401F : 0x4012;
```
Roman971, line 248: Same here

include/z64save.h
```diff
@@ -0,0 +1,191 @@
+#ifndef _Z64SRAM_H_
+#define _Z64SRAM_H_
```
Roman971, line 2:
```suggestion
#ifndef _Z64SAVE_H_
#define _Z64SAVE_H_
```

fig02:
most of these comments are for my changes so ill implement them, but ill tag random in anything that i cant weigh in on

src/code/z_sram.c
fig02: @Random06457 thoughts, i dont really mind either way. if file select is going to be doing a checksum calc the same way the macro might be useful to have, but idk if it will

src/code/z_sram.c
fig02: @Random06457

include/z64.h
```diff
@@ -1291,6 +1116,95 @@ typedef struct {
...
+    /* 0x1CAC4 */ f32 windowRotX;
+    /* 0x1CAC8 */ u16 kbdButtonIdx;
+    /* 0x1CACA */ u16 unk_1CACA;
+    /* 0x1CACC */ u16 kbdCharBoxA;
+    /* 0x1CACE */ s16 kbdCharIdx;
+    /* 0x1CAD0 */ s16 kbdCharX;
+    /* 0x1CAD2 */ s16 kbdCharY;
+    /* 0x1CAD4 */ s16 newFileNameCharCount;
+    /* 0x1CAD6 */ u16 unk_1CAD6[3];
+} FileChooseContext; // size = 0x1CADC
```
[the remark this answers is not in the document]
fig02: I think its fine to keep, ive made some changes to it already in my file select progress but its a good starting point and wont affect anything else

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -9177,10 +9177,11 @@ void Player_Init(Actor* thisx, GlobalContext* globalCtx) {
         this->stateFlags3 &= ~0x40;
     }
 
-    if (gSaveContext.unk_13C0 != 0) {
-        unk_13C0 = gSaveContext.unk_13C0;
-        Audio_PlayActorSound2(&this->actor, unk_13C0);
-        gSaveContext.unk_13C0 = 0;
+    if (gSaveContext.entranceSound != 0) {
+        entranceSound = gSaveContext.entranceSound;
+        // this appears to be bugged and doesnt play properly. not sure why yet
```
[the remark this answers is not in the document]
fig02: I understand your point for sure but its more of a reminder to figure out what is up with it. based on the code that we do have, it should absolutely be working, but it doesnt. so unless something in audio is preventing it, theres some issue with playing sounds in init that we dont understand yet, and I think leaving a remark to remind us to look into it cant hurt.

not sure i see a scenario where it ends up confusing anyone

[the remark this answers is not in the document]
Roman971: Saying that it doesn't work is my main issue. For all we know we may just be testing it under the wrong conditions, so I don't think it's fair to say it's bugged until we can confirm it in code.

src/code/z_sram.c
Roman971: Also I could be wrong but it looks like there are other slightly different loops to calculate checksums in the file that don't fall under this same macro so that's another reason to not have it imo, since it can't be consistent.

src/code/z_sram.c
Random06457: This was just because I initially made unnamed structs within structs but now that everything is split up properly in different structs, there's no need to keep this.

fig02:
```
#define SAVE_SIZE sizeof(Save)
#define INFO_SIZE sizeof(SaveInfo)
```
these do indeed work fine. probably no use for a define at this point right?

Separate but related questions. The Slot size is basically sizeof SaveContext plus some additional padding to round it up. Do we want to define it like `#define SLOT_SIZE sizeof(SaveContext) + 0x28` or just hardcode the number?
Similarly, SRAM size is 6 slots (3 main, 3 backup) plus some extra padding. So do we want to do 
`#define SRAM_SIZE (SLOT_SIZE * 6) + 0x620` ?

include/z64.h
```diff
@@ -1221,7 +1046,7 @@ typedef struct GlobalContext {
     /* 0x01C24 */ ActorContext actorCtx;
     /* 0x01D64 */ CutsceneContext csCtx; // "demo_play"
     /* 0x01DB4 */ SoundSource soundSources[16];
-    /* 0x01F74 */ SubGlobalContext1F74 sub_1F74;
+    /* 0x01F74 */ Sram sram;
```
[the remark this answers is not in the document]
fig02: this one is weird to me. I view srams "context" as the save context, and not as this struct with one u8 pointer in it, but I could just be looking at it from the wrong angle.
Is it even required to be a struct?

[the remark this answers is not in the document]
Roman971: It has to be a struct yeah, since it's the struct used by the sram system (you can see it as its `this` basically). It doesn't have much but I would still consider it as a usual context.

src/code/z_sram.c
Roman971: The 0x8000 bytes that make SRAM_SIZE is hardware dependent so it definitely shouldn't be relative to the size of slots. 
As for SLOT_SIZE, I think it's better to have it as `sizeof(SaveContext) + 0x28`, for now at least.

fig02: alright, sounds good.
I know that file select needs SRAM_SIZE, and it may need slot and checksum size as well. Where would be a good place for them to live? save.h or does it not fit there

Roman971: `z64.h` seems fine to me for the time being.

fig02:
changes made

include/z64.h
```diff
@@ -1104,9 +928,11 @@ typedef struct {
 } EntranceEntry;
 
 typedef struct {
-    /* 0x00 */ void* read_buff;
+    /* 0x00 */ u8* readBuff;
 } Sram; // size = 0x4
```
Roman971, line 932:
```suggestion
} SramContext; // size = 0x4
```

Roman971 (commented):
Last thing but the rest looks good to me

Roman971 (approved):

PR #433, opened 2020-10-03
code_800F7260

Thar0:
- 1 non-equivalent but is relatively close, but not close enough to verify equivalence due to a lot of regalloc induced by a few reorderings.
 - 1 left in asm, too many nested loops for me currently.
 - The rest of the file is matching.

The name `gSoundBanks` is taken from sm64 audio, and isn't really related to the file `Audiobank`. Could this possibly cause any confusion?


src/code/code_800F7260.c
```diff
@@ -1,40 +1,639 @@
...
+                        phi_v0->sfxId = D_8016E2F8[i];
+                        phi_v0->pos = a1;
+                        phi_v0->unk_8 = a2;
+                        phi_v0->unk_C = a3;
+                        phi_v0->unk_10 = a4;
+                        phi_v0->unk_14 = a5;
+                        D_801333A0++;
+                        phi_v0 = &D_8016C9A0[D_801333A0];
+                    }
+                    i = 10;
```
krimtonz, line 148: Interesting this was the method of breaking out of the loop

src/code/code_800F7260.c
```diff
@@ -1,40 +1,639 @@
...
+        phi_a0 = false;
+        entry = &D_8016C9A0[i];
+        switch (arg0) {
+            case 0:
+                if (SFX_BANK_MASK(entry->sfxId) == SFX_BANK_MASK(arg1->unk_28)) {
+                    phi_a0 = true;
+                }
+                break;
+            case 1:
+                if (SFX_BANK_MASK(entry->sfxId) == SFX_BANK_MASK(arg1->unk_28) && &entry->pos->x == arg1->posX) {
```
krimtonz, line 177:
```suggestion
                if (SFX_BANK_MASK(entry->sfxId) == SFX_BANK_MASK(arg1->unk_28) && (&entry->pos->x == arg1->posX)) {
```
Normally I view this as a preference thing, but with the & for the pointer on entry it looks a bit cluttered

krimtonz (commented):
Looks pretty good to me, just a couple of minor things

src/code/code_800F7260.c
Thar0: sm64 audio has a bunch of these reimplemented breaks too, seems like something to watch out for in other audio files as well

src/code/code_800F7260.c
Thar0: I see what you mean with the `&& &` business, brackets seem like a good call here.

krimtonz (approved):

src/code/code_800F7260.c
krimtonz: Does it actually set i to 10? Or just breaks?

Thar0: It does yeah, it sets it to 10 then increments the loop counter as if it were about to continue, only stopping since it doesn't satisfy the condition to continue the loop. Replacing it with a proper `break` ruins codegen all over the function.

krimtonz: I'd like to know what was going through their heads when they decided that

src/code/code_800F7260.c
```diff
@@ -1,40 +1,639 @@
...
+ * 5 : Ocarina Bank         size 3
+ * 6 : Voice Bank           size 5
+ */
+SoundBankEntry* gSoundBanks[7] = {
+    D_8016BAD0, D_8016BC80, D_8016BEC0, D_8016C2E0, D_8016C6A0, D_8016C820, D_8016C8B0,
+};
+
+u8 sBankSizes[ARRAY_COUNT(gSoundBanks)] = {
+    9, 12, 22, 20, 8, 3, 5,
+};
```
Dragorn421, line 73:
```suggestion
SoundBankEntry* gSoundBanks[7] = {
    D_8016BAD0, D_8016BC80, D_8016BEC0, D_8016C2E0, D_8016C6A0, D_8016C820, D_8016C8B0,
};

static u8 sBankSizes[ARRAY_COUNT(gSoundBanks)] = {
    ARRAY_COUNT(D_8016BAD0), ARRAY_COUNT(D_8016BC80), ARRAY_COUNT(D_8016BEC0), ARRAY_COUNT(D_8016C2E0), ARRAY_COUNT(D_8016C6A0), ARRAY_COUNT(D_8016C820), ARRAY_COUNT(D_8016C8B0),
};
```
`static` since you used the `s` prefix I guess this isn't used outside the file?

src/code/code_800F7260.c
```diff
@@ -1,40 +1,639 @@
...
+extern Struct_800F7CEC D_8016E1B8[7][3]; // total size = 0xA8
+extern Struct_800F8EA0 D_8016E270[];
+extern Struct_8013331C* D_8013331C[9];
+
+void func_800F7260(u16 arg0) {
+    u8 bankId;
+
+    for (bankId = 0; bankId < ARRAY_COUNT(gSoundBanks); bankId++) {
+        if (arg0 & 1) {
+            D_8016E264[bankId] = 1;
```
Dragorn421, line 105: imo `bankId` should be called an index (eg `bankIdx`) (indexes into an array) rather than an id which sounds more like map/dictionary/hashtable

src/code/code_800F7260.c
```diff
@@ -1,40 +1,639 @@
...
+                        gSoundBanks[phi_s5][phi_a1].unk_26 = temp_a2->unk_2;
+                        gSoundBanks[phi_s5][phi_a1].unk_24 = temp_a2->unk_0;
+                    } else if (gSoundBanks[phi_s5][phi_a1].unk_2A == 5) {
+                        gSoundBanks[phi_s5][phi_a1].unk_2A = 4;
+                    }
+                    phi_a1 = 0;
+                }
+            } else
+                return;
+        }
```
Dragorn421, line 277:
```suggestion
            } else {
                return;
            }
        }
```

src/code/code_800F7260.c
```diff
@@ -1,40 +1,639 @@
...
+void func_800F8480(u8 bankId) {
+    u8 bankIndex;
+    SequenceChannel* seqChannel;
+    SoundBankEntry* entry;
+    u8 i;
+
+    for (i = 0; i < D_80130578[D_801333CC][bankId]; i++) {
+        bankIndex = D_8016E1B8[bankId][i].unk_4;
+        if (bankIndex != 0xFF) {
+            entry = &gSoundBanks[bankId][bankIndex];
```
Dragorn421, line 349: `bankIndex` -> `entryIndex`?

src/code/code_800F7260.c
```diff
@@ -1,40 +1,639 @@
...
+    f32 temp_f0;
+    Struct_800F8EA0* temp_v0;
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/code_800F7260/func_800F8884.s")
+    if (arg2 == 0) {
+        arg2++;
+    }
+    temp_v0 = &D_8016E270[arg0];
+    temp_f0 = arg1 / 127.0f;
+    temp_v0->unk_4 = temp_f0;
```
Dragorn421, line 554: is this temp_f0 temp really mandatory (and temp_v0?)

include/z64audio.h
```diff
@@ -742,4 +742,47 @@ typedef struct {
...
+    /* 0x18 */ u32*     unk_18;
+    /* 0x1C */ f32      unk_1C;
+    /* 0x20 */ u32      unk_20;
+    /* 0x24 */ u8       unk_24; 
+    /* 0x26 */ u16      unk_26;
+    /* 0x28 */ u16      unk_28;     // "flag"
+    /* 0x2A */ u8       unk_2A;
+    /* 0x2B */ u8       unk_2B;
+    /* 0x2C */ u8       prev;       // prev bank index
+    /* 0x2D */ u8       next;       // next bank index
```
Dragorn421, line 761:
```suggestion
    /* 0x2C */ u8       prev;       // prev entry index
    /* 0x2D */ u8       next;       // next entry index
```

Dragorn421 (changes requested):
I'm not convinced about the usage of "bank", I usually saw it used as a synonym for segment (as in "segmented address")
I have no clue how audio works though
I guess it can change later when more audio is decompiled and documented

Any clue what `SoundBankEntry` is? Is it a command that is overwritten when eg actors play sounds or parameters for playing sounds, or some kind of middle ground between the two?

src/code/code_800F7260.c
Thar0: Other code files such as `game.c` use the `s` prefix without making the variables `static` so I'm also reluctant to do that currently as we might be doing that at a later stage.

fig02: yeah the `s` prefix doesnt mean `static` keyword, its two separate uses for the same word. can just treat it as local vs global

src/code/code_800F7260.c
Thar0: Both were removable it turns out

Thar0:
> Any clue what `SoundBankEntry` is? Is it a command that is overwritten when eg actors play sounds or parameters for playing sounds, or some kind of middle ground between the two?

I'm not entirely sure but the rough equivalent in sm64 audio is `SoundCharacteristics`, so I believe it's the parameters of a sound request added through calls to public functions such as `Audio_PlaySoundGeneral`, which is indirectly linked to the actor system through wrappers such as `Audio_PlayActorSound2` in `z_actor.c` but is not limited to actors in itself e.g. being called directly to queue the sound that plays when advancing text.

fig02 (approved):

krimtonz (approved):

src/code/code_800F7260.c
Thar0: After thinking about it for a while I think I'll hold off on naming changes in this PR unless there are very strong disagreements with current names and leave it to others who will do more audio than I will to rename when they learn more about the system than is currently known through other files.

Roman971 (approved):
I have no idea what's going on in there for now, so I'll trust yall with this.

PR #437, opened 2020-10-04
Make Gamestates use thisx for entry point functions

fig02:
do we have plans to make stuff in z_play use `this` when referring to itself instead of `globalCtx`? I was going to do that with this PR but wasnt sure if it was something we want to do later, like when we rename GlobalContext for example


fig02:
realized i missed `sample`. will do that today

fig02:
fixed sample

src/code/z_sample.c
```diff
@@ -83,7 +83,8 @@ void Sample_LoadTitleStatic(SampleContext* this) {
     DmaMgr_SendRequest1(this->staticSegment, _title_staticSegmentRomStart, size, "../z_sample.c", 164);
 }
 
-void Sample_Init(SampleContext* this) {
+void Sample_Init(GameState* thisx) {
+    SampleContext* this = (SampleContext*)thisx;
     this->state.main = Sample_Main;
```
Roman971, line 88:
```suggestion
    SampleContext* this = (SampleContext*)thisx;

    this->state.main = Sample_Main;
```

src/code/z_prenmi.c
```diff
@@ -39,17 +39,21 @@ void PreNMI_Draw(PreNMIContext* prenmiCtx) {
     CLOSE_DISPS(gfxCtx, "../z_prenmi.c", 112);
 }
 
-void PreNMI_Main(PreNMIContext* prenmiCtx) {
+void PreNMI_Main(GameState* thisx) {
+    PreNMIContext* prenmiCtx = (PreNMIContext*)thisx;
```
Roman971, line 43: Maybe `this` should be used in this file, if you want to do that

fig02: I can do that. I would like to do it for z_play as well, what do you think

Roman971: Really not sure about z_play.c tho, that's globalCtx so I think it would be confusing right now

fig02: I see it as no different than the other gamestates in terms of how confusing it is, but its not that big of a deal. we can change it when we rename globalCtx

fig02:
pushed changes. i got all the gamestates right? I missed sample before cause it doesnt ever get set via `SET_NEXT_GAMESTATE` which is what i used to search lol

Roman971 (approved):
Yeah I think that should be all of them

ethteck (approved):

PR #438, opened 2020-10-04
Morpha (ovl_Boss_Mo)

petrie911:
Morita's second greatest creation, matched and documented.


undefined_syms.txt
```diff
@@ -1285,6 +1285,16 @@ D_06000498 = 0x06000498;
...
+// z_boss_mo
+D_06000F20 = 0x06000F20;
+D_06000F70 = 0x06000F70;
+D_06000FC8 = 0x06000FC8;
+D_06001010 = 0x06001010;
+D_06004C50 = 0x06004C50;
+D_06006700 = 0x06006700;
+D_06006838 = 0x06006838;
+D_06007C00 = 0x06007C00;
+
```
louist103, line 1297: If the object is decompiled these can go.

louist103 (changes requested):
Nice job on this big one.  Just some small changes.

louist103 (approved):

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -22,299 +111,2944 @@ const ActorInit Boss_Mo_InitVars = {
...
+                    this->cameraEye = camera1->eye;
+                    this->cameraAt = camera1->at;
+                    this->cameraYaw = Math_FAtan2F(this->cameraEye.x - this->actor.world.pos.x,
+                                                   this->cameraEye.z - this->actor.world.pos.z);
+                    this->cameraYawRate = 0;
+                    goto tent_shake; // I think this is genuinely the correct code
+                }
+            }
+            break;
+        tent_shake:
```
Dragorn421, line 770:
```suggestion
                } else {
                    break;
                }
            } else {
                break;
            }
```
this doesn't change codegen but it changes regalloc somehow. just ignore this suggestion, putting it here if someone wanna try figure it out

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -22,299 +111,2944 @@ const ActorInit Boss_Mo_InitVars = {
...
+                this->actor.world.rot.y = this->actor.yawTowardsPlayer + 0x8000;
+                this->damageFlashTimer = 15;
+                Audio_PlayActorSound2(&this->actor, NA_SE_EN_MOFER_CORE_DAMAGE);
+                this->actor.colChkInfo.health -= damage;
+                this->hitCount++;
+                if ((s8)this->actor.colChkInfo.health <= 0) {
+                    if (((sMorphaTent1->csCamera == 0) && (sMorphaTent2 == NULL)) ||
+                        ((sMorphaTent1->csCamera == 0) && (sMorphaTent2 != NULL) && (sMorphaTent2->csCamera == 0))) {
+                        Enemy_StartFinishingBlow(globalCtx, &this->actor);
+                        Audio_QueueSeqCmd(NA_BGM_STOP | (1 << 0x10));
```
Dragorn421, line 1748:
```suggestion
                        Audio_QueueSeqCmd(0x100100FF);
```
same with the other `Audio_QueueSeqCmd(NA_BGM_STOP | (1 << 0x10))`

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -22,299 +111,2944 @@ const ActorInit Boss_Mo_InitVars = {
...
+        gDPSetPrimColor(POLY_XLU_DISP++, 0x80, 0x80, 255, 255, 255, (s8)this->baseAlpha);
+
+        func_8002ED80(&this->actor, globalCtx, 0);
+
+        gSPDisplayList(POLY_XLU_DISP++, SEGMENTED_TO_VIRTUAL(gMorphaCoreMembraneDL));
+
+        gDPPipeSync(POLY_XLU_DISP++);
+
+        gDPSetEnvColor(POLY_XLU_DISP++, 0, 220, 255, 128);
+        if (this->damageFlashTimer & 1) {
```
Dragorn421, line 2590:
```suggestion
        if (this->damageFlashTimer % 2 != 0) {
```

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.h
```diff
@@ -6,11 +6,196 @@
...
+    /*  9 */ MO_CORE_WATER_TEX1_X,
+    /* 10 */ MO_CORE_WATER_TEX1_Y,
+    /* 11 */ MO_CORE_WATER_TEX2_X,
+    /* 12 */ MO_CORE_WATER_TEX2_Y,
+    /* 13 */ MO_CORE_WATER_LEVEL,
+    /* 14 */ MO_CORE_FLOAT_MAX
+} BossMoCoreF32Var;
+
+#define MO_SHORT_MAX MAX(MO_TENT_SHORT_MAX, MO_CORE_SHORT_MAX)
+#define MO_FLOAT_MAX MAX(MO_TENT_FLOAT_MAX, MO_CORE_FLOAT_MAX)
```
Dragorn421, line 104: are all these enums and defines wip for swork/fwork stuff? maybe remove it if you don't plan to use it in this PR

Dragorn421 (changes requested):
I don't have much to say
great work petrie

and lol louis what did you do with that PR, I think this is going to be a record for amount of PR contributors (?)

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
petrie911: Yeah I had that before I used the goto method.

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.h
```diff
@@ -6,11 +6,101 @@
...
+    /* 0x0156 */ u8 tentSpawnPos;
+    /* 0x0158 */ s16 actionState;
+    /* 0x015A */ s16 movementTimer;
+    /* 0x015C */ s16 rippleTimer;
+    /* 0x015E */ s16 damageFlashTimer;
+    /* 0x0160 */ s16 invincibilityTimer;
+    /* 0x0162 */ s16 baseTex1x;
+    /* 0x0164 */ union{s16 baseTex1y; s16 corePos;};
+    /* 0x0166 */ union{s16 baseTex2x; s16 drawShadow;};
+    /* 0x0168 */ union{s16 baseTex2y; s16 waitUnderwater;};
```
fig02, line 26: like barinade, we should probably go the #define route for these unions

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -22,299 +128,2943 @@ const ActorInit Boss_Mo_InitVars = {
...
+void BossMo_SetupTentacle(BossMo* this, GlobalContext* globalCtx) {
+    this->actionFunc = BossMo_Tentacle;
+    this->actionState = MO_TENT_WAIT;
+    this->timers[0] = 50 + (s16)Rand_ZeroFloat(20.0f);
+}
+
+void BossMo_Tentacle(BossMo* this, GlobalContext* globalCtx) {
+    // Because of the large number of variables, I have marked the real
+    // stack variables that cannot be reordered and the pads currently
+    // not being used as temps.
```
fig02, line 410: were these comments added when the function was non matching? now that its matched its probably not important to keep I think

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -22,299 +128,2943 @@ const ActorInit Boss_Mo_InitVars = {
...
+            } else if (this->timers[0] > 30) {
+                rippleCount = 5;
+            } else if (this->timers[0] > 20) {
+                rippleCount = 8;
+            } else {
+                rippleCount = 3;
+            }
+            for (indS1 = 0; indS1 < rippleCount; indS1++) {
+                rand_f = Rand_ZeroFloat(50.0f);
+                rand_angle = Rand_ZeroFloat(0x10000);
```
fig02, line 586: camel case for names

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -22,299 +128,2943 @@ const ActorInit Boss_Mo_InitVars = {
...
+                    }
+                }
+            }
+            break;
+        case MO_TENT_ATTACK:
+            this->actor.flags |= 0x1000000;
+            func_80078914(&this->tentTipPos, NA_SE_EN_MOFER_ATTACK - SFX_FLAG);
+            Math_ApproachF(&this->waterLevelMod, -5.0f, 0.1f, 0.4f);
+            for (indS1 = 0; indS1 < 41; indS1++) {
+                Math_ApproachF(&this->tentStretch[indS1].y,
```
fig02, line 653: <= 40 maybe makes more sense in this context, not sure

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -22,299 +128,2943 @@ const ActorInit Boss_Mo_InitVars = {
...
+                    func_80064520(globalCtx, &globalCtx->csCtx);
+                    this->csCamera = Gameplay_CreateSubCamera(globalCtx);
+                    Gameplay_ChangeCameraStatus(globalCtx, MAIN_CAM, CAM_STAT_WAIT);
+                    Gameplay_ChangeCameraStatus(globalCtx, this->csCamera, CAM_STAT_ACTIVE);
+                    this->cameraEye = camera1->eye;
+                    this->cameraAt = camera1->at;
+                    this->cameraYaw = Math_FAtan2F(this->cameraEye.x - this->actor.world.pos.x,
+                                                   this->cameraEye.z - this->actor.world.pos.z);
+                    this->cameraYawRate = 0;
+                    goto tent_shake; // I think this is genuinely the correct code
```
fig02, line 783: i agree, and would probably not worry about including a comment about it

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -22,299 +128,2943 @@ const ActorInit Boss_Mo_InitVars = {
...
+                    sp120 = this->tentPos[this->meltIndex];
+                    sp120.x += Rand_CenteredFloat(30.0f);
+                    sp120.y += Rand_CenteredFloat(30.0f);
+                    sp120.z += Rand_CenteredFloat(30.0f);
+                    BossMo_SpawnStillDroplet(globalCtx->specialEffects, &sp120, Rand_ZeroFloat(0.1f) + .2f);
+                }
+                this->meltIndex++;
+            }
+            Math_ApproachF(&this->cutScale, 0.0, 1.0f, 0.2f);
+            if ((this->meltIndex >= 41) || (this->timers[0] == 0)) {
```
fig02, line 879: > 40 ? again not sure on context, just seems like a nicer number to me lol

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -22,299 +128,2943 @@ const ActorInit Boss_Mo_InitVars = {
...
+            Math_ApproachF(&this->tentMaxStretch, 0, 0.5f, 0.02f);
+            Math_ApproachF(&this->tentMaxAngle, 0.5f, 1.0f, 0.01);
+            Math_ApproachF(&this->tentSpeed, 320.0f, 1.0f, 50.0f);
+            if (this->timers[0] == 0) {
+                this->actor.flags &= ~1;
+                Math_ApproachF(&this->baseAlpha, 0.0, 1.0f, 5.0f);
+                for (indS1 = 0; indS1 < 40; indS1++) {
+                    if (sMorphaTent2->tentSpawnPos) {}
+                    indT5 = Rand_ZeroFloat(20.9f);
+                    VEC_SET(spFC, 0, 0, 0);
```
fig02, line 922: is vec_set needed to match? if not we should just set it normally

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -22,299 +128,2943 @@ const ActorInit Boss_Mo_InitVars = {
...
+    osSyncPrintf("Core_Damage_check 終わり ！！\n");
+    osSyncPrintf(VT_RST);
+}
+
+void BossMo_Core(BossMo* this, GlobalContext* globalCtx) {
+    static f32 coreBulge[11] = {
+        0.1f, 0.15f, 0.2f, 0.3f, 0.4f, 0.43f, 0.4f, 0.3f, 0.2f, 0.15f, 0.1f,
+    };
+    // Because of the large number of temps, I have indicated which named temps
+    // are not actually on the stack and can be reordered. All pads are unused.
```
fig02, line 1828: same here, this level of detail about the stack variables doesnt seem needed anymore if the function is matching

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -22,299 +128,2943 @@ const ActorInit Boss_Mo_InitVars = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Boss_Mo/func_80925480.s")
+void BossMo_Unknown(void) {
+    // Called every frame by UpdateCore, but nothing sets the BREGs.
+    // Testing in PJ64 indicates this function effectively does nothing.
+    static Vec3f zeroVec = { 0.0f, 0.0f, 0.0f };
+    static u16 unkSfx[532] = {
+        0x0800, 0x0800, 0x0800, 0x0801, 0x0802, 0x0803, 0x0804, 0x0805, 0x0806, 0x0807, 0x0808, 0x080B, 0x080A, 0x080B,
+        0x080C, 0x080D, 0x080F, 0x0810, 0x0810, 0x0811, 0x0812, 0x0813, 0x0814, 0x0815, 0x0816, 0x0817, 0x0818, 0x081B,
+        0x081A, 0x081B, 0x081D, 0x081F, 0x0820, 0x0820, 0x0821, 0x0822, 0x0823, 0x0824, 0x0825, 0x0826, 0x0827, 0x0828,
```
fig02, line 3023: as annoying as it is to do, these should prob use the sfx enum right?
does the script work on it? lol

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
fig02: lol github ate my greater than sign
either way tho i figured out later on that 41 is the number of core parts right. in which case disregard

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
petrie911: It was mostly there because gotos are considered harmful, but seems like this project isn't particularly hardline on that.

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
petrie911: The script won't work on it as-is because it only replaces stuff in function calls. I'll...see what I can do.

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
fig02: yeah we are very skeptical of them at least, but in this case it seems to be what it should be. gotos def have their place

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
petrie911: This is mostly a note for future cleanup. Basically, a bunch of the temps are probably split across the pads. It's also to clarify what all the comments next to the stack variables mean.

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -1,15 +1,121 @@
...
 
 #define FLAGS 0x00000035
 
 #define THIS ((BossMo*)thisx)
 
+#define MO_WATER_LEVEL(globalCtx) globalCtx->colCtx.colHeader->waterBoxes[0].ySurface
+
+#define HAS_LINK(tent) ((tent != NULL) && ((tent->actionState == MO_TENT_GRAB) || (tent->actionState == MO_TENT_SHAKE)))
+
+typedef struct BossMoEffect {
```
Roman971, line 20:
```suggestion
typedef struct {
```

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -22,299 +128,2943 @@ const ActorInit Boss_Mo_InitVars = {
...
+            return true;
+        }
+        if (fabsf(pos->z - -180.0f) < 90.0f + margin) {
+            return true;
+        }
+    }
+    return false;
+}
+
+void BossMo_SpawnRipples(BossMoEffect* effect, Vec3f* pos, f32 scale, f32 vMaxScale, s16 maxAlpha, s16 partLimit,
```
Roman971, line 187: Using `vMaxScale` here seems confusing, was it intended?
Also the function only spawns 1 effect so it should probably be named singular to avoid any potential confusion.

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -22,299 +128,2943 @@ const ActorInit Boss_Mo_InitVars = {
...
+static s16 sGrabRot[41] = {
+    0, 0, 0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0, 0, 0, -5, -5, -5,
+    0, 5, 10, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20, 0, 0, 0, 0,  0,
+};
+static s16 sAttackRot[41] = {
+    0, 5, 6, 7, 8, 8, 7, 6, 6, 2, 2, 2, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+};
+
+static InitChainEntry sInitChain[4] = {
```
Roman971, line 292:
```suggestion
static InitChainEntry sInitChain[] = {
```

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -22,299 +128,2943 @@ const ActorInit Boss_Mo_InitVars = {
...
+    u16 i;
+
+    Actor_ProcessInitChain(&this->actor, sInitChain);
+    ActorShape_Init(&this->actor.shape, 0.0f, NULL, 0.0f);
+    if (this->actor.params != BOSSMO_TENTACLE) {
+        Flags_SetSwitch(globalCtx, 0x14);
+        sMorphaCore = this;
+        MO_WATER_LEVEL(globalCtx) = this->waterLevel = MO_WATER_LEVEL(globalCtx);
+        globalCtx->unk_11D30[0] = 0xA0;
+        globalCtx->specialEffects = &sEffects;
```
Roman971, line 337:
```suggestion
        globalCtx->specialEffects = sEffects;
```

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -22,299 +128,2943 @@ const ActorInit Boss_Mo_InitVars = {
...
+        for (i = 0; i < ARRAY_COUNT(sEffects); i++) {
+            sEffects[i].type = 0;
```
Roman971, line 339:
```suggestion
            sEffects[i].type = MO_FX_NONE;
```

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -22,299 +128,2943 @@ const ActorInit Boss_Mo_InitVars = {
...
+        0x0000, 0x0020, 0x6800, 0x6800, 0x6800, 0x6800, 0x6800, 0x6800, 0x6800, 0x6801, 0x6801, 0x681C, 0x6802, 0x6803,
+        0x6814, 0x6804, 0x6804, 0x6804, 0x6804, 0x6805, 0x6805, 0x6808, 0x6807, 0x6808, 0x6808, 0x6809, 0x6809, 0x680B,
+        0x680C, 0x680D, 0x6841, 0x6842, 0x6840, 0x6841, 0x6842, 0x6840, 0x6850, 0x30D0, 0x38D1, 0x38D2, 0x38D3, 0x38D4,
+        0x38D5, 0x38D6, 0x38D7, 0x3927, 0x38D9, 0x30DA, 0x30DB, 0x3889, 0x39EF, 0x38E1, 0x38E2, 0x38E4, 0x38E5, 0x38E6,
+        0x387A, 0x38E7, 0x38E8, 0x38EC, 0x38ED, 0x38EE, 0x38EF, 0x30F0, 0x30F1, 0x38F2, 0x38F3, 0x39EC, 0x38F4, 0x30F5,
+        0x38F6, 0x38F7, 0x38FC, 0x38FD, 0x387B, 0x3900, 0x3901, 0x3902, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000,
+    };
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Boss_Mo/func_80925C18.s")
+    if (BREG(0x20) != 0) {
```
Roman971, line 3061: Ideally decimal should be used for all these REGs

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -22,299 +128,2943 @@ const ActorInit Boss_Mo_InitVars = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Boss_Mo/func_80924D70.s")
+    Matrix_Pop();
+    CLOSE_DISPS(gfxCtx, "../z_boss_mo.c", 7482);
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Boss_Mo/func_80925480.s")
+void BossMo_Unknown(void) {
+    // Called every frame by UpdateCore, but nothing sets the BREGs.
+    // Testing in PJ64 indicates this function effectively does nothing.
```
Roman971, line 3018: Did you test it while actually setting the BREGs from the reg editor? I would expect this function to be used for sfx testing during development, and we should probably name and/or comment it as such.

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -1,15 +1,121 @@
...
+    /* 0x24 */ u8 type;
+    /* 0x25 */ u8 timer;
+    /* 0x26 */ u8 stopTimer;
+    /* 0x28 */ s16 unk_28; // unused?
+    /* 0x2A */ s16 alpha;
+    /* 0x2C */ s16 rippleMode;
+    /* 0x2E */ s16 maxAlpha;
+    /* 0x30 */ f32 scale;
+    /* 0x34 */ f32 moFxFloat1;
+    /* 0x38 */ f32 moFxFloat2;
```
Roman971, line 33: Wouldn't it make more to define these as an array, at least to avoid the few `(*effect).` that are required in the file?

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -1,15 +1,121 @@
...
+    /* 0x3C */ Vec3f* targetPos;
+} BossMoEffect; // size = 0x40
+
 void BossMo_Init(Actor* thisx, GlobalContext* globalCtx);
 void BossMo_Destroy(Actor* thisx, GlobalContext* globalCtx);
+void BossMo_UpdateCore(Actor* thisx, GlobalContext* globalCtx);
 void BossMo_Update(Actor* thisx, GlobalContext* globalCtx);
+void BossMo_DrawCore(Actor* thisx, GlobalContext* globalCtx);
 void BossMo_Draw(Actor* thisx, GlobalContext* globalCtx);
```
Roman971, line 42: The "default" update and draw functions are only used for tentacles, right? If so I think they should be named UpdateTent and DrawTent to make it clear.

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
Roman971: Also not sure we need these comments, unless there is a cleanup planned soon, because otherwise they are just going to stay for a long time while not serving a real purpose. And I'd almost argue anyone who goes into changing the variable layout of a function should be ready to deal with stack issues to keep it matching.

Dragorn421 (changes requested):
would it help that someone handles these reviews for you? sorry for kind of forcing it on you

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
fig02: agreed here, assuming this is the case

fig02 (approved):

Roman971 (approved):
The work/fwork stuff was better with defines instead of enums imo, but it's probably something we can figure out later when we do more documentation passes

PR #441, opened 2020-10-06
Oshihiki

petrie911:
Push blocks, the best actor, are now documented. Some of the functions might need better names, so I'm open to suggestions on them.


src/overlays/actors/ovl_Obj_Oshihiki/z_obj_oshihiki.c
```diff
@@ -27,63 +67,661 @@ const ActorInit Obj_Oshihiki_InitVars = {
...
+    s32 pad;
+    ColHeader* colHeader = NULL;
+    s32 pad2;
+
+    DynaPolyInfo_SetActorMove(&this->dyna, moveFlag);
+    DynaPolyInfo_Alloc(collision, &colHeader);
+    this->dyna.dynaPolyId =
+        DynaPolyInfo_RegisterActor(globalCtx, &globalCtx->colCtx.dyna, &this->dyna.actor, colHeader);
+
+    if (this->dyna.dynaPolyId == 0x32 /*BGACTOR_MAX*/) {
```
Roman971, line 135: The proper define will be used everywhere once it's introduced, so I think there is no need to put a name we haven't decided to use in a comment. (Relevant in all other places there is a comment included with temporary names from other PRs)
```suggestion
    if (this->dyna.dynaPolyId == 50) {
```

src/overlays/actors/ovl_Obj_Oshihiki/z_obj_oshihiki.c
```diff
@@ -27,63 +67,661 @@ const ActorInit Obj_Oshihiki_InitVars = {
...
+    switch (BLOCK_TYPE) {
+        case SMALL_SWITCH_OFF:
+        case MEDIUM_SWITCH_OFF:
+        case SMALL_SWITCH_ON:
+        case MEDIUM_SWITCH_ON:
+            return 1;
+            break;
+        case LARGE_SWITCH_OFF:
+        case LARGE_SWITCH_ON:
+            return strength > 0;
```
Roman971, line 164:
```suggestion
            return strength >= PLAYER_STR_BRACELET;
```

src/overlays/actors/ovl_Obj_Oshihiki/z_obj_oshihiki.c
```diff
@@ -27,63 +67,661 @@ const ActorInit Obj_Oshihiki_InitVars = {
...
+            break;
+        case HUGE_SWITCH_OFF:
+        case HUGE_SWITCH_ON:
+            return strength >= 2;
```
Roman971, line 168:
```suggestion
            return strength >= PLAYER_STR_SILVER_G;
```

src/overlays/actors/ovl_Obj_Oshihiki/z_obj_oshihiki.c
```diff
@@ -27,63 +67,661 @@ const ActorInit Obj_Oshihiki_InitVars = {
...
+        case HUGE_SWITCH_OFF:
+        case SMALL_SWITCH_ON:
+        case MEDIUM_SWITCH_ON:
+        case LARGE_SWITCH_ON:
+        case HUGE_SWITCH_ON:
+            ObjOshihiki_InitDynapoly(this, globalCtx, &D_05004E98, 1);
+            return;
+    }
+    // Error : type cannot be determined
+    osSyncPrintf("Error : タイプが判別できない(%s %d)(arg_data 0x%04x)\n", "../z_obj_oshihiki.c", 444,
```
Roman971, line 255: This should probably be a default instead of the other case being an early return.

src/overlays/actors/ovl_Obj_Oshihiki/z_obj_oshihiki.c
```diff
@@ -27,63 +67,661 @@ const ActorInit Obj_Oshihiki_InitVars = {
...
+
+    paramsColorIdx = (this->dyna.actor.params >> 6) & 3;
+
+    for (i = 0; i < ARRAY_COUNT(sScenes); i++) {
+        if (sScenes[i] == globalCtx->sceneNum) {
+            break;
+        }
+    }
+
+    if (i > 8) {
```
Roman971, line 296:
```suggestion
    if (i >= ARRAY_COUNT(sColors)) {
```

src/overlays/actors/ovl_Obj_Oshihiki/z_obj_oshihiki.c
```diff
@@ -27,63 +67,661 @@ const ActorInit Obj_Oshihiki_InitVars = {
...
+void ObjOshihiki_Init(Actor* thisx, GlobalContext* globalCtx) {
+    s32 pad;
+    ObjOshihiki* this = THIS;
+    s32 switchFlag;
+
+    ObjOshihiki_CheckType(this, globalCtx);
+
+    switchFlag = this->dyna.actor.params >> 8;
+
+    if (((u8)switchFlag >= 0) && ((u8)switchFlag <= 0x3F)) {
```
Roman971, line 317: Any chance switchFlag could be a `u8`? or the `& 0xFF` is part of the line above where it extracts it from params?

src/overlays/actors/ovl_Obj_Oshihiki/z_obj_oshihiki.c
```diff
@@ -27,63 +67,661 @@ const ActorInit Obj_Oshihiki_InitVars = {
...
+    f32 cs = Math_Coss(angle);
+    s32 i;
+
+    for (i = 0; i < 4; i++) {
+        Vec3f faceVtx;
+        Vec3f faceVtxNext;
+        Vec3f posResult; // dummy value
+        Vec3f faceVtxOffset;
+        u32 bgId;               // dummy value
+        CollisionPoly* outPoly; // dummy value
```
Roman971, line 458: Imo there is no need to specify that the function doesn't end up reading these values. Or at least if you want to specify it, it should probably be part of the variable name. But I think it's better to just leave it available so whoever works on this function later can make use of them without having to rename them.
```suggestion
        Vec3f posResult;
        Vec3f faceVtxOffset;
        u32 bgId;
        CollisionPoly* outPoly;
```

src/overlays/actors/ovl_Obj_Oshihiki/z_obj_oshihiki.c
```diff
@@ -27,63 +67,661 @@ const ActorInit Obj_Oshihiki_InitVars = {
...
+        case SCENE_BMORI1:
+        case SCENE_HIDAN:
+        case SCENE_MIZUSIN:
+        case SCENE_JYASINZOU:
+        case SCENE_HAKADAN:
+        case SCENE_MEN:
+            gDPSetEnvColor(oGfxCtx->polyOpa.p++, this->color.r, this->color.g, this->color.b, 255);
+            break;
+        default:
+            gDPSetEnvColor(oGfxCtx->polyOpa.p++, mREG(0xD), mREG(0xE), mREG(0xF), 255);
```
Roman971, line 721: Regs are normally numbered as decimal
```suggestion
            gDPSetEnvColor(oGfxCtx->polyOpa.p++, mREG(13), mREG(14), mREG(15), 255);
```

src/overlays/actors/ovl_Obj_Oshihiki/z_obj_oshihiki.c
petrie911: The temp turns out to be unnecessary, but the (u8) casts are.

src/overlays/actors/ovl_Obj_Oshihiki/z_obj_oshihiki.h
```diff
@@ -1,16 +1,33 @@
...
+    /* 0x017C */ f32 direction; // dyna unk_150 stored here
+    /* 0x0180 */ s32 floorBgIds[5];
+    /* 0x0194 */ CollisionPoly* floorPolys[5];
+    /* 0x01A8 */ f32 floorHeights[5];
+    /* 0x01BC */ s16 highestFloor;
+    /* 0x01BE */ u8 cantMove;
+    /* 0x01C0 */ struct ObjOshihiki* blockUnder;
+    /* 0x01C4 */ f32 underDistX;
+    /* 0x01C8 */ f32 underDistZ;
+    /* 0x01CC */ UNK_PTR texture;
```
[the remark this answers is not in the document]
petrie911: UNK_PTR is just a macro for void* anyways, isn't it?

src/overlays/actors/ovl_Obj_Oshihiki/z_obj_oshihiki.c
Roman971: The u8 casts are very likely fake since it should be `& 0xFF` instead, so I would try to get rid of them however you can

src/overlays/actors/ovl_Obj_Oshihiki/z_obj_oshihiki.h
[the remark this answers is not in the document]
Roman971: It is, but the intent of the macro is to declare that we don't know what type of pointer it is, which we do here since it's a texture.

src/overlays/actors/ovl_Obj_Oshihiki/z_obj_oshihiki.h
```diff
@@ -6,11 +6,49 @@
...
+    /* 1 */ PUSHBLOCK_MEDIUM_START_ON,
+    /* 2 */ PUSHBLOCK_LARGE_START_ON,
+    /* 3 */ PUSHBLOCK_HUGE_START_ON,
+    /* 4 */ PUSHBLOCK_SMALL_START_OFF,
+    /* 5 */ PUSHBLOCK_MEDIUM_START_OFF,
+    /* 6 */ PUSHBLOCK_LARGE_START_OFF,
+    /* 7 */ PUSHBLOCK_HUGE_START_OFF
+} PushBlockTypes;
+
+#define PUSHBLOCK_ON_SCENE       ( 1 << 0 )
```
Roman971, line 20: No need for spaces imo, tho we don't really have a choice anyway if we want to be consistent since clang-format will get rid of them (eg. when we have them in C files).
```suggestion
#define PUSHBLOCK_ON_SCENE       (1 << 0)
```

src/overlays/actors/ovl_Obj_Oshihiki/z_obj_oshihiki.c
petrie911: I'll see what I can do, but it may end up being `((switchFlag & 0xFF) >= 0) && ((switchFlag & 0xFF) <= 0x3F)`

petrie911: I think the best I'm doing is `& 0xFF` without a temp. There was no nice looking way to use a temp and have the regalloc work. Also I'm pretty sure no temp is used for the flag in the switch check because it's clearly a macro.

Roman971 (approved):

fig02 (approved):

PR #443, opened 2020-10-10
Decompile z_kanfont

Thar0:
`func_8006EE50` moved from the end of `z_kaleido_setup.c` into `z_kanfont.c` as it is only called in `z_message_PAL.c` leading me to believe it makes more sense in this file instead. Majora's Mask's version of `z_kanfont.c` also has such a stub function as well.


src/code/z_kanfont.c
```diff
@@ -1,7 +1,54 @@
...
+
+void Kanfont_LoadOrderedFont(Font* font) {
+    s32 size;
+    s32 jj;
+    s32 fontStatic;
+    u32 fontBuf;
+    s32 codePointIndex;
+    s32 fontBufIndex;
+    s32 off;
+
```
fig02, line 33: code looks good to me, only personal peeve of mine is I prefer `offset`, its not much longer and is more explicit

fig02 (approved):

NEstelami (approved):

src/code/z_kanfont.c
```diff
@@ -1,7 +1,54 @@
...
+                        &_nes_font_staticSegmentRomStart[character * FONT_CHAR_TEX_SIZE], FONT_CHAR_TEX_SIZE,
+                        "../z_kanfont.c", 93);
+}
+
+void Kanfont_LoadMessageBoxEndIcon(Font* font, u16 icon) {
+    DmaMgr_SendRequest1(font->iconBuf, &_message_staticSegmentRomStart[4 * 0x1000 + icon * FONT_CHAR_TEX_SIZE],
+                        FONT_CHAR_TEX_SIZE, "../z_kanfont.c", 100);
+}
+
+void Kanfont_LoadOrderedFont(Font* font) {
```
Roman971, line 25: I think if we're taking out the "kan" in the struct/defines we might as well take it out from the system name as well
```suggestion
void Font_LoadOrderedFont(Font* font) {
```

include/z64.h
```diff
@@ -670,9 +670,38 @@ typedef struct {
...
+    MESSAGE_ICON_TRIANGLE,
+    MESSAGE_ICON_SQUARE,
+    MESSAGE_ICON_ARROW
+} MessageBoxIcon;
+
+typedef struct {
+    /* 0x0000 */ u32  offset;
+    /* 0x0004 */ u32  size;
+    /* 0x0008 */ u8   xy;
+} MessageData;
```
Roman971, line 683: What's the reason behind this struct? I think we should avoid sub structs unless they serve a real purpose.
Also what's `xy`? I don't see any usage of it here. And actually a lot of things in the font struct seem unused. Same goes for the box icon enum.

Thar0: The `MessageData` struct is from `z_message_PAL` progress, of which I copied part of my `MessageContext` struct from (everything before `textboxSegment`) when starting the file. It's possibly removable now given that it's also the start of the `Font` struct. These fields are confirmed but I can remove them if they are unwanted at this time. The box icon enum comes from inspecting the textures in `message_static`, they are supposed to be used in the `icon` argument of `LoadMessageBoxEndIcon`.

Roman971: I see. In that case, I think it would be better to leave only the stuff necessary for z_kanfont.c and such, if possible. This way we can review the other stuff once we can see where it's used in a future message PR.

Roman971 (approved):

PR #452, opened 2020-10-14
BgJyaBombchuiwa OK

louist103:
Decomp and lightly document BgJyaBombchuiwa.

PR #464, opened 2020-10-27
Decompile the last of libultra_code_O2

Thar0:
- Decompiles `osPfsFileState`
- Moved `guMtxL2F` to `asm/` as it is handwritten like `guMtxF2L`

One file left to name, but there's not a lot to go off in terms of its contents.


Roman971 (approved):

include/functions.h
```diff
@@ -2529,7 +2529,7 @@ s32 __osCheckPackId(OSPfs* pfs, __OSPackId* temp);
 s32 __osGetId(OSPfs* pfs);
 s32 __osCheckId(OSPfs* pfs);
 s32 __osPfsRWInode(OSPfs* pfs, __OSInode* inode, u8 flag, u8 bank);
-// ? func_80102FA0(?);
+// ? guMtxL2F(?);
```
fig02, line 2532: can probably add the prototype here since its used externally.

also, zap2 changes? did we downgrade on accident at some point or something

Thar0: #452 downgraded ZAP2 it looks like?

fig02: rip lmfao, thanks

fig02 (approved):

PR #466, opened 2020-10-29
ovl_demo_effect 100% OK

Kenix3:
Documentation is also complete


src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -8,7 +8,91 @@ void DemoEffect_Init(Actor* thisx, GlobalContext* globalCtx);
...
+s32 DemoEffect_CheckCsAction(DemoEffect* this, GlobalContext* globalCtx, s32 csActionCompareId);
+f32 DemoEffect_InterpolateCsFrames(GlobalContext* globalCtx, s32 csActionId);
+void DemoEffect_InitPositionFromCsAction(DemoEffect* this, GlobalContext* globalCtx, s32 csActionIndex);
+void DemoEffect_FaceToCsEndpoint(DemoEffect* this, Vec3f startPos, Vec3f endPos);
+void DemoEffect_MoveToCsEndpoint(DemoEffect* this, GlobalContext* globalCtx, s32 csActionId, s32 shouldUpdateFacing);
+void DemoEffect_MoveTowardTarget(Vec3f targetPos, DemoEffect* this, f32 speed);
+void DemoEffect_MoveGetItem(DemoEffect* this, GlobalContext* globalCtx, s32 csActionId, f32 speed);
+
+// gameplay_keep
+extern Gfx lightBall[];
```
petrie911, line 73: Are we not leaving these as D_ symbols?

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +104,2126 @@ const ActorInit Demo_Effect_InitVars = {
     (ActorFunc)DemoEffect_Update,
     NULL,
 };
-*/
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Effect/func_80970F50.s")
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Effect/func_80970F58.s")
+// Code only matches when this is an array, but the second element isn't used. This variable assures only one jewel will
+// play SFX
+s16 sfxJewelId[] = { 0, 0 };
```
petrie911, line 110: Because data is word-aligned, you don't even need the second element. Also all of these need to be static and start with the s prefix.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +104,2126 @@ const ActorInit Demo_Effect_InitVars = {
...
+    /* 0x15 */ OBJECT_GI_JEWEL,
+    /* 0x16 */ OBJECT_GI_JEWEL,
+    /* 0x17 */ OBJECT_GI_M_ARROW,
+    /* 0x18 */ OBJECT_EFC_TW,
+    /* 0x19 */ OBJECT_EFC_TW
+};
+
+u8 timewarpVertexSizeIndices[] = { 1, 1, 2, 0, 1, 1, 2, 0, 1, 2, 0, 2, 1, 0, 1, 0, 2, 0, 2, 2, 0 };
+
+Color_RGB8 jewelSparkleColors[5][2] = { { { 0xFF, 0xFF, 0xFF }, { 0x64, 0xFF, 0x00 } },
```
petrie911, line 144: Colors are in decimal.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +104,2126 @@ const ActorInit Demo_Effect_InitVars = {
...
+
+/**
+ * Initializes information for Jewel/Spritual Stone actors.
+ */
+void DemoEffect_InitJewel(GlobalContext* globalCtx, DemoEffect* this) {
+    this->initDrawFunc = &DemoEffect_DrawJewel;
+    if (LINK_IS_CHILD) {
+        this->initUpdateFunc = &DemoEffect_UpdateJewelChild;
+    } else {
+        this->initUpdateFunc = &DemoEffect_UpdateJewelAdult;
```
petrie911, line 177:
```suggestion
        this->initUpdateFunc = DemoEffect_UpdateJewelChild;
    } else {
        this->initUpdateFunc = DemoEffect_UpdateJewelAdult;
```
Functions are their own pointer, so these aren't needed.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +104,2126 @@ const ActorInit Demo_Effect_InitVars = {
...
+            break;
+
+        case Demo_Effect_Blue_Orb:
+            this->initDrawFunc = &DemoEffect_DrawBlueOrb;
+            this->initUpdateFunc = &DemoEffect_UpdateBlueOrbGrow;
+            this->blueOrb.opacity = 255;
+            this->blueOrb.scale = 5;
+            this->blueOrb.rotation = 0;
+            Actor_SetScale(&this->actor, 0.05f);
+            this->primXluColor[0] = 0xBC;
```
petrie911, line 256: Are you sure these aren't `RGB8`s?

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +104,2126 @@ const ActorInit Demo_Effect_InitVars = {
...
+                this->actor.posRot.pos.y, this->actor.posRot.pos.z, 0, 0, 0, Demo_Effect_Light_Ring_Triforce);
+
+            if (lightRing) {
+                Actor_SetScale(&lightRing->actor, 0.4f);
+            }
+            break;
+
+        case Demo_Effect_Medal_Fire:
+            DemoEffect_InitGetItem(this);
+            this->getItem.giIndex = 12;
```
petrie911, line 459: Does it make sense to have an enum for giIndex?

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +104,2126 @@ const ActorInit Demo_Effect_InitVars = {
...
+    s32 effectType;
+    s32 lightEffect;
+    s32 objectIndex;
+    DemoEffect* crystalLight;
+    DemoEffect* lightRing;
+
+    effectType = GET_EFFECT_TYPE(&this->actor);
+    lightEffect = GET_LIGHT_EFFECT_COLOR_PARAM(&this->actor);
+
+    osSyncPrintf("\x1b[36m no = %d\n\x1b[m", effectType);
```
petrie911, line 221: `\x` means VT macros. You can find them in `vt.h`

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +104,2126 @@ const ActorInit Demo_Effect_InitVars = {
...
+                    }
+                    break;
+
+                case 11:
+                    if (globalCtx->csCtx.frames == 350) {
+                        Audio_PlayActorSound2(&this->actor, NA_SE_IT_DM_FLYING_GOD_DASH);
+                    }
+                    break;
+
+                default:
```
petrie911, line 1222: Is `default: break;` required to match?

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +104,2126 @@ const ActorInit Demo_Effect_InitVars = {
...
+    f32 xzSpherical;
+
+    distance = frameDivisor * sqrtf(SQ(endPos.x - startPos.x) + SQ(endPos.y - startPos.y) + SQ(endPos.z - startPos.z));
+
+    this->actor.posRot.pos.x = radius * cosf(degrees * (M_PI / 180.0f));
+    this->actor.posRot.pos.y = distance;
+    this->actor.posRot.pos.z = radius * sinf(degrees * (M_PI / 180.0f));
+
+    xPos = this->actor.posRot.pos.x;
+    ySpherical = (this->actor.posRot.pos.y * cosf(rotation.x * (M_PI / 32768.0f))) -
```
petrie911, line 1473: `32768.0f` should be fine as `0x8000`

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +104,2126 @@ const ActorInit Demo_Effect_InitVars = {
...
+    }
+
+    if (startPos.x != endPos.x || startPos.y != endPos.y || startPos.z != endPos.z) {
+        this->jewelCsRotation.x = atan2f(endPos.z - startPos.z, -(endPos.x - startPos.x)) * (32768.0f / M_PI);
+        this->jewelCsRotation.y = Math_Vec3f_Yaw(&startPos, &endPos);
+    }
+
+    this->jewelCsRotation.z += 1024;
+
+    degrees += this->jewelCsRotation.z * (45.0f / 8192.0f);
```
petrie911, line 1537: this makes a little more sense as `360.0f / 0x10000`

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +104,2126 @@ const ActorInit Demo_Effect_InitVars = {
...
+    if (radius > 30.0f) {
+        radius = 30.0f;
+    }
+
+    if (startPos.x != endPos.x || startPos.y != endPos.y || startPos.z != endPos.z) {
+        this->jewelCsRotation.x = atan2f(endPos.z - startPos.z, -(endPos.x - startPos.x)) * (32768.0f / M_PI);
+        this->jewelCsRotation.y = Math_Vec3f_Yaw(&startPos, &endPos);
+    }
+
+    this->jewelCsRotation.z += 1024;
```
petrie911, line 1535:
```suggestion
    this->jewelCsRotation.z += 0x400;
```
Angles are hex.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +104,2126 @@ const ActorInit Demo_Effect_InitVars = {
...
+void DemoEffect_UpdateJewelChild(DemoEffect* this, GlobalContext* globalCtx) {
+    s32 hasCmdAction;
+    Actor* thisx = &this->actor;
+
+    this->jewel.timer++;
+
+    if (globalCtx->csCtx.state && globalCtx->csCtx.npcActions[this->csActionId]) {
+        switch (globalCtx->csCtx.npcActions[this->csActionId]->action) {
+            case 3:
+                if (gSaveContext.eventChkInf[4] & 2048) {
```
petrie911, line 1624:
```suggestion
                if (gSaveContext.eventChkInf[4] & 0x800) {
```
Bitflags are definitely in hex. Possibly should be `(1 << 11)`.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +104,2126 @@ const ActorInit Demo_Effect_InitVars = {
...
+            if (!hasCmdAction) {
+                this->effectFlags |= 1;
+                return;
+            }
+        }
+    }
+
+    thisx->shape.rot.y += 1024;
+    DemoEffect_PlayJewelSfx(this, globalCtx);
+    this->effectFlags &= 65534;
```
petrie911, line 1669:
```suggestion
    this->effectFlags &= ~1;
```

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +104,2126 @@ const ActorInit Demo_Effect_InitVars = {
...
+    scroll = globalCtx->gameplayFrames;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Effect/func_809723C4.s")
+    OPEN_DISPS(globalCtx->state.gfxCtx, "../z_demo_effect.c", 2737);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Effect/func_809726AC.s")
+    if (!DemoEffect_CheckCsAction(this, globalCtx, 2)) {
+        if (gSaveContext.entranceIndex == 0x00A0) {
+            if (gSaveContext.sceneSetupIndex == 4) {
+                if (globalCtx->csCtx.frames < 681) {
```
petrie911, line 1878:
```suggestion
                if (globalCtx->csCtx.frames <= 680) {
```

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +104,2126 @@ const ActorInit Demo_Effect_InitVars = {
...
+    gDPSetEnvColor(POLY_XLU_DISP++, this->envXluColor[0], this->envXluColor[1], this->envXluColor[2], 0xFF);
+    gSPSegment(POLY_XLU_DISP++, 8,
+               Gfx_TwoTexScroll(globalCtx->state.gfxCtx, 0, (scroll * 6) % 1024, 256 - ((scroll * 16) % 256) - 1, 256,
+                                64, 1, (scroll * 4) % 512, 128 - ((scroll * 12) % 128) - 1, 128, 32));
+    CLOSE_DISPS(globalCtx->state.gfxCtx, "../z_demo_effect.c", 3172);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Effect/func_80974EB4.s")
+    if (limbIndex == 0) {
+        LimbTransform* transforms = skelCuve->transforms;
+        transforms->scale.x = 1024;
```
petrie911, line 2139: Chained assignment?

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.h
```diff
@@ -6,11 +6,134 @@
...
+    /* 0x018C */ s16 csActionId;
+    /* 0x018E */ Vec3s jewelCsRotation;
+    /* 0x0194 */ DemoEffectFunc initUpdateFunc;
+    /* 0x0198 */ DemoEffectFunc initDrawFunc;
+    /* 0x019C */ DemoEffectFunc updateFunc;
 } DemoEffect; // size = 0x01A0
 
-extern const ActorInit Demo_Effect_InitVars;
+// These names come from the objects that correspond to this actor type.
+typedef enum Demo_Effect_Effect_Type {
```
petrie911, line 89: I believe enum entries should be all caps.

petrie911 (commented):
Nice, this is a big one. I've left some suggestions of things I noticed while looking through it.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
Kenix3: Ah I wasn't aware that was a thing in OOT. Thank you!

I'll let Fig and/or Roman chime in here. I don't see a reason to remove good names

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
Kenix3: You're right on alignment and good catch on the static. I dunno why it only matches as an array, though.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
Kenix3: Good catch, you're right.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
Kenix3: I think that they are, but Fig and Roman told me the preference would be to make them arrays. They may not have known the full context, so I guess we'll let them chime in here now that the whole file is ready.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
Kenix3: Yeah I like that better.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
Kenix3: Isn't this -2? 65535 would be -1

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
petrie911: Static variables seem to just do their own thing sometimes. I can't explain it.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
Kenix3: Agreed. Good catch!

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
petrie911: Not `-2`, `~1`. Bitwise negation.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
Kenix3: Yes, but I think this is out of scope for DemoEffect. It's an index into the gi draw table in z_draw, and that's not fully documented yet.

petrie911: Ah, ok.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
Kenix3: Ah, my mistake. My phone is displaying it like a -1. You are right!

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
Kenix3: Why `M_PI / 0x8000` as opposed to `M_PI / 32768.0f` or `M_PI / 32768`

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
Kenix3: Why not `360 / 65536`?
Agreed that including 360 in the number if better.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
Kenix3: NOT OK :D

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.h
Kenix3: Let's let Fig and/or Roman comment here then. :)

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
petrie911: The current standard has binary angles in hex. Since 32768 here represents a binary angle, 0x8000 makes more sense.

Kenix3:
All of your requested changes are in @petrie911, except for things that I want @fig02 and/or @Roman971 to comment on. Thank you!

spec
```diff
@@ -1514,7 +1514,6 @@ endseg
 beginseg
     name "ovl_Demo_Effect"
     include "build/src/overlays/actors/ovl_Demo_Effect/z_demo_effect.o"
-    include "build/data/overlays/actors/z_demo_effect.data.o"
     include "build/data/overlays/actors/z_demo_effect.reloc.o"
```
ethteck, line 1517: You'll want to update the reloc entry as well. See other matched actors for an example

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.h
ethteck: I agree enum names should be in all caps, and I also am not a fan of "Effect_Effect". I know you didn't ask for my opinion but yee

ethteck (changes requested):

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
fig02: for things from objects specifically we havent really been naming them. feel like thats better tackled when the objects are decompiled and we arent using undefined syms anymore

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +105,2115 @@ const ActorInit Demo_Effect_InitVars = {
...
+    u32 scroll = this->jewel.timer;
+
+    OPEN_DISPS(globalCtx->state.gfxCtx, "../z_demo_effect.c", 2543);
+
+    if (!DemoEffect_CheckCsAction(this, globalCtx, 1)) {
+        // Necessary to match
+        if (1) {}
+
+        if (!(this->effectFlags & (1 << 0))) {
+            globalCtx2 = globalCtx;
```
fig02, line 1727: can this globalCtx2 assignment be done at the top of the function?

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +105,2115 @@ const ActorInit Demo_Effect_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Effect/func_809719C0.s")
+    parent = (DemoEffect*)this->actor.parent;
+    scroll = globalCtx->gameplayFrames & 0xFFFF;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Effect/func_809719F0.s")
+    OPEN_DISPS(globalCtx->state.gfxCtx, "../z_demo_effect.c", 2634);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Effect/func_80971A28.s")
+    if (parent) {
+        gDPSetPrimColor(POLY_XLU_DISP++, 128, 128, 255, 255, 170, parent->triforceSpot.crystalLightOpacity);
```
fig02, line 1792: != NULL

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +105,2115 @@ const ActorInit Demo_Effect_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Effect/func_80971BBC.s")
+    func_80093D84(globalCtx->state.gfxCtx);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Effect/func_80971DFC.s")
+    gSPSegment(POLY_XLU_DISP++, 8,
+               Gfx_TwoTexScroll(globalCtx->state.gfxCtx, 0, (scroll * 2) % 512, 512 - (scroll % 512) - 1, 128, 128, 1,
+                                512 - ((scroll * 2) % 512) - 1, 0, 64, 64));
+    Matrix_Push();
+    Matrix_RotateY(0.0f, MTXMODE_APPLY);
+    Matrix_RotateX((11.0 * M_PI) / 180.0, MTXMODE_APPLY); // Doubles and not floats needed here
```
fig02, line 1804: comment not needed, we assume if you left it out it was done on purpose

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +105,2115 @@ const ActorInit Demo_Effect_InitVars = {
...
+        func_80093D18(globalCtx->state.gfxCtx);
+        func_8002EBCC(&this->actor, globalCtx, 0);
+        Matrix_Pull();
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Effect/func_809727B0.s")
+        this->godLgt.rotation++;
+        if (this->godLgt.rotation > 120) {
+            // if block necessary to match
+            if (1) {
+                this->godLgt.rotation = 0;
```
fig02, line 1900: the if 1 cant be on its own line?

fig02 (commented):
Great job on this

90% of my comments are on the usage of `if (!var)` and `if (var == 0)` and such.
We only use the former for things that can definitely be considered a boolean. So ive requested that many pointer checks be changed explicitly to `!= NULL`.
There are also some cases like `if (csCtx.state)` being used for values that can be more than true and false, which ive noted should be changed to != 0 as well.

I noted every case I came across to make it easier to implement the changes, but I may have missed some

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
fig02: iirc in order to use rgb8 you had to do some weird union? in which case the array is def preferred

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.h
```diff
@@ -6,11 +6,135 @@
...
+    /* 0x017B */ u8 envXluColor[3];
+    /* 0x017E */ u8 primOpaColor[3];
+    /* 0x0181 */ u8 envOpaColor[3];
+    /* 0x0184 */ union {
+        struct {
+            u8 timer;
+        } fireBall;
+        struct {
+            u8 opacity;
+            u8 scale;
```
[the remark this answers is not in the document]
Kenix3: I like this compromise between having a union and putting more info in the struct itself. I'll make this change!

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +105,2115 @@ const ActorInit Demo_Effect_InitVars = {
...
+    }
+
+    radius = 50.0f * frameDivisor;
+    // CLAMP_MAX macro is not OK
+    if (radius > 30.0f) {
+        radius = 30.0f;
+    }
+
+    if (startPos.x != endPos.x || startPos.y != endPos.y || startPos.z != endPos.z) {
+        this->jewelCsRotation.x = atan2f(endPos.z - startPos.z, -(endPos.x - startPos.x)) * (32768.0f / M_PI);
```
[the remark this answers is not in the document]
Kenix3: > 
> 
> 32768

Actually it's a rotation, so it's one where I missed 0x8000 / M_PI :)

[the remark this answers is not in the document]
petrie911: 0x8000

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
Kenix3: I wouldn't need a union in this case I believe.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.h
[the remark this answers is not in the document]
Kenix3: Great compromise! I agree.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +105,2115 @@ const ActorInit Demo_Effect_InitVars = {
...
+            crystalLight = (DemoEffect*)Actor_SpawnAsChild(
+                &globalCtx->actorCtx, &this->actor, globalCtx, ACTOR_DEMO_EFFECT, this->actor.posRot.pos.x,
+                this->actor.posRot.pos.y, this->actor.posRot.pos.z, 0, 0, 0, DEMO_EFFECT_CRYSTAL_LIGHT);
+
+            if (crystalLight) {
+                Actor_SetScale(&crystalLight->actor, 0.6f);
+            }
+
+            // Necessary to match. Otherwise has function argument reordering.
+            if (globalCtx) {}
```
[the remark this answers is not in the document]
Kenix3: This works

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +105,2115 @@ const ActorInit Demo_Effect_InitVars = {
...
+            Actor_SetScale(&this->actor, 0.14f);
+            return;
+        }
+
+        Actor_SetScale(&this->actor, 84 * 0.001f);
+        return;
+    }
+
+    if (gSaveContext.sceneSetupIndex == 5 || gSaveContext.sceneSetupIndex == 4 ||
+        (gSaveContext.entranceIndex == 0x0324 && !((gSaveContext.eventChkInf[12] & (1 << 9))))) {
```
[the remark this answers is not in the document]
Kenix3: Gotcha. I don't see a reason to undo this change, but I understand that it isn't a requirement right now.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +105,2115 @@ const ActorInit Demo_Effect_InitVars = {
...
+void DemoEffect_UpdateLightRingShrinking(DemoEffect* this, GlobalContext* globalCtx) {
+    if (this->lightRing.timer < this->lightRing.timerIncrement) {
+        Actor_Kill(&this->actor);
+        this->lightRing.timer = 0;
+    } else {
+        this->lightRing.timer -= this->lightRing.timerIncrement;
+    }
+
+    if (this->lightRing.timer < 256) {
+        if (this->lightRing.timer >= 225) {
```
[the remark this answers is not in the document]
Kenix3: Probably because the other one is already a >= that makes sense

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +105,2115 @@ const ActorInit Demo_Effect_InitVars = {
...
+    this->actor.posRot.pos.y += (targetPos.y - this->actor.posRot.pos.y) * speed;
+    this->actor.posRot.pos.z += (targetPos.z - this->actor.posRot.pos.z) * speed;
+}
+
+/**
+ * Initializes Jewel colors.
+ */
+void DemoEffect_InitJewelColor(DemoEffect* this) {
+    u8 jewelType = this->jewel.type;
+
```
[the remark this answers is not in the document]
Kenix3: Should I change this now or leave it? I named just about everything based on the related object model file name.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +105,2115 @@ const ActorInit Demo_Effect_InitVars = {
...
+    primColor.g = (sparkleColors + 0)->g;
+    primColor.b = (sparkleColors + 0)->b;
+    envColor.r = (sparkleColors + 1)->r;
+    envColor.g = (sparkleColors + 1)->g;
+    envColor.b = (sparkleColors + 1)->b;
+    primColor.a = 0;
+
+    while (i < spawnerCount) {
+        velocity.x = (Math_Rand_ZeroOne() - 0.5f) * 1.5f;
+        velocity.z = (Math_Rand_ZeroOne() - 0.5f) * 1.5f;
```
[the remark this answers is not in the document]
Kenix3: It does. I don't see why it matters here, but I changed it! :)

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
Kenix3: I tried that before and it did not work. I tried a different permutation by putting the if(1) {} afterward and it still matches.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +105,2115 @@ const ActorInit Demo_Effect_InitVars = {
...
+/**
+ * This update function waits until the associate object is loaded.
+ * Once the object is loaded, it will copy over the initUpdateFunc/initDrawFunc funcs to be active.
+ * They are copied to actor.draw and updateFunc.
+ * initUpdateFunc/initDrawFunc are set during initialization and are NOT executed.
+ */
+void DemoEffect_Wait(DemoEffect* this, GlobalContext* globalCtx) {
+    if (Object_IsLoaded(&globalCtx->objectCtx, this->initObjectBankIndex)) {
+        this->actor.objBankIndex = this->initObjectBankIndex;
+        this->actor.draw = (ActorFunc)this->initDrawFunc;
```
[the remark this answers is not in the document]
Kenix3: Removing it generates compiler warnings. We actually need to change the type of initDrawFunc to match the type of draw.

ethteck (approved):

Roman971 (commented):
There are still a few things left to address in previous comments, but these are my suggestions

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
Roman971: My opinion would also be to wait until we have decided on a naming scheme, like we've been doing in every other file so far.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
Roman971: Even if you play around with the order? That's a bit surprising

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
Roman971: I would say to use `Color_RGB8` if you can. Array is only preferred over a union of both.

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +105,2099 @@ const ActorInit Demo_Effect_InitVars = {
...
+    effectType = GET_EFFECT_TYPE(&this->actor);
+    lightEffect = GET_LIGHT_COLOR_PARAM(&this->actor);
+
+    osSyncPrintf(VT_SGR("36") " no = %d\n" VT_SGR(""), effectType);
+
+    objectIndex = sEffectTypeObjects[effectType] == 1
+                      ? 0
+                      : Object_GetIndex(&globalCtx->objectCtx, sEffectTypeObjects[effectType]);
+
+    osSyncPrintf(VT_SGR("36") " bank_ID = %d\n" VT_SGR(""), objectIndex);
```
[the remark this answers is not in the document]
fig02: theres vt_fmt.py in tools as well to make this easier

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +105,2099 @@ const ActorInit Demo_Effect_InitVars = {
...
+    OPEN_DISPS(globalCtx->state.gfxCtx, "../z_demo_effect.c", 2842);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Effect/func_80972B4C.s")
+    if (!DemoEffect_CheckCsAction(this, globalCtx, 1)) {
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Effect/func_80972E3C.s")
+        if (this->light.flicker == 0) {
+            this->light.flicker = 1;
+        } else {
+            disp = (u32)lightBall;
```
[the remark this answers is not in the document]
fig02: it does, but I dont think its real. the disp and opacity temps dont seem real either so this just needs fixing in general. I cant get anything after messing with it for a little bit. gonna leave a note

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
```diff
@@ -20,121 +105,2099 @@ const ActorInit Demo_Effect_InitVars = {
...
+ * Draw function for the Crystal Light Actor.
+ */
+void DemoEffect_DrawCrystalLight(Actor* thisx, GlobalContext* globalCtx) {
+    DemoEffect* this = THIS;
+    DemoEffect* parent;
+    u32 scroll;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Effect/func_809719C0.s")
+    parent = (DemoEffect*)this->actor.parent;
+    scroll = globalCtx->gameplayFrames & 0xFFFF;
```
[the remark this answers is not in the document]
fig02: cant get it to work

src/overlays/actors/ovl_Demo_Effect/z_demo_effect.c
fig02: y = 1024
z = x = 1024 


was the only way i got this to match lol. its better than setting it to the var tho imo

fig02:
review changes implemented

Roman971 (approved):

PR #468, opened 2020-10-31
Document Collision_Check

petrie911:
Did a documentation pass on collision_check.

 - All functions named and public functions added to `functions.h`. Functions cleaned up, variables named, and typedefs added to make the function arrays more manageable.

 - Decomped the last undecomped function (`Collider_DrawPoly`), although unfortunately it has a small regalloc error. Cleaned up the other non-matching function (now named `CollisionCheck_CylSideVsLineSeg`) enough to know what it does, though I'm still not sure it's equivalent.

 - Renamed and refactored `Collider` structs to no longer use "body" and to be a little more consistent with each other. Generally tried to move away from "type" and "flags" as fields because of how common those are.

 - Through in-game testing, reviewing `z_collision_check`'s code, and scraping almost all collision data from the ROM, determined the purposes of (most of) the collision data fields, and created macros for their flags.

 - Added pre-migrated collision data to most undecomped actors. There are a few that don't have it due to doing weird things with their initialization that messed with the automator, and I can't guarantee I got everything, but it should be convenient for future decompilations.

 - Updated `colliderinit.py` to use the new macros. It also will now give the element list automatically when you decompile a JntSph or Tris collider. This is less useful than it could be given the above, but it's there.

 - Various cleanups in actors that I stumbled across while implementing the flag macros.

 - formatted the whole repo, which seems to have made small adjustments to a bunch of things.

Not done:

 - dFlag macros. The damage flags are documented, but which macros we actually want is less clear, as they come in many combinations.

 - Naming the element types. The data indicates there are 7 element types (I added an 8th to be on the safe side), but so far only 5 have shown up in the code and with only vague ideas of what they mean. Hopefully we'll see more of them. A similar situation applies to the element OC flags.

 - Detailed comments on the workings of functions. Not really sure where to go with this one. `z_collision_check` tends to be either self-documenting or nasty math whose only purpose is to determine where to put the hitmark.

 - Actually matching `CollisionCheck_CylSideVsLineSeg` and `Collider_DrawPoly`. The latter is just permuter-resistant regalloc, but the former still has some ways to go. It seems to have the same matching issue as `Math3D_CylVsLineSeg`, which makes sense as they use the same code.

 - That weird `.bss` issue. I have no idea what that even is.


include/z64collision_check.h
```diff
@@ -9,245 +9,345 @@
...
-    /* 0x0C */ ColliderTrisItemInit* list;
+    /* 0x0C */ ColliderTrisElementInit* elements;
 } ColliderTrisInit_Set3; // size = 0x10
 
 typedef struct {
     /* 0x00 */ Vec3f quad[4];
     /* 0x30 */ Vec3s dcMid; // midpoint of vectors d, c
     /* 0x36 */ Vec3s baMid; // midpoint of vectors b, a
-    /* 0x3C */ f32 unk_3C;
+    /* 0x3C */ f32 acDist;
```
mzxrules, line 198: could you add a comment on what this is?

petrie911: Player quads use it to keep track of which collider element they hit was closest to them and only apply hit effects to that one. If a player quad has an AC collision with an element closer than acDist, it stores the new dist to acDist and resets the previous collider.

src/code/z_collision_check.c
```diff
@@ -2720,244 +2745,247 @@ void func_80061EFC(CollisionCheckInfo* info, DamageTable* damageTable, Collision
...
-    static s32 (*D_8011E018[4])(GlobalContext*, CollisionCheckContext*, Collider*, Vec3f*, Vec3f*) = {
-        CollisionCheck_generalLineOcCheck_JntSph,
-        CollisionCheck_generalLineOcCheck_Cyl,
+// Checks if the line segment ab intersects any OC colliders, excluding those attached to actors
+// on the exclusion list. Returns true if there are any intersections and false otherwise.
+s32 CollisionCheck_GeneralLineOcCheck(GlobalContext* globalCtx, CollisionCheckContext* colChkCtx, Vec3f* a, Vec3f* b,
+                                      Actor** exclusions, s32 numExclusions) {
+    static ColChkLineFunc ocLineCheckFuncs[4] = {
+        CollisionCheck_GeneralLineOcCheck_JntSph,
+        CollisionCheck_GeneralLineOcCheck_Cyl,
```
fig02, line 2895: agane

src/code/z_collision_check.c
```diff
@@ -2969,33 +2997,36 @@ static s8 sBssDummy12;
...
         }
     }
 }
 
-void func_80062A28(GlobalContext* globalCtx, Vec3f* v) {
+// Spawns red blood droplets.
+// No actor has a collision type that spawns red blood.
+void CollisionCheck_SpawnRedBlood(GlobalContext* globalCtx, Vec3f* v) {
     static EffectSparkInit D_8015CF10;
-    s32 sp24;
```
fig02, line 2994: should be simple enough to name this blood or something

src/code/z_collision_check.c
```diff
@@ -3084,10 +3117,11 @@ void func_80062B80(GlobalContext* globalCtx, Vec3f* v) {
...
-    Effect_Add(globalCtx, &sp24, EFFECT_SPARK, 0, 1, &D_8015D3D8);
+    Effect_Add(globalCtx, &pIndex, EFFECT_SPARK, 0, 1, &D_8015D3D8);
 }
 
-void func_80062CD4(GlobalContext* globalCtx, Vec3f* v) {
+// Spawns streaks of light from hits against solid objects
+void CollisionCheck_SpawnShieldParticles(GlobalContext* globalCtx, Vec3f* v) {
     static EffectShieldParticleInit init = {
         16,
         { 0.0f, 0.0f, 0.0f },
```
fig02, line 3127: outside if possible

src/code/z_collision_check.c
```diff
@@ -3035,16 +3066,18 @@ void func_80062A28(GlobalContext* globalCtx, Vec3f* v) {
...
+    Effect_Add(globalCtx, &pIndex, EFFECT_SPARK, 0, 1, &D_8015CF10);
 }
 
-void func_80062B80(GlobalContext* globalCtx, Vec3f* v) {
+// Spawns water droplets.
+// No actor has a collision type that spawns water droplets.
+void CollisionCheck_SpawnWaterDroplets(GlobalContext* globalCtx, Vec3f* v) {
     static EffectSparkInit D_8015D3D8;
-    s32 sp24;
+    s32 pIndex;
```
fig02, line 3076: same as above, blood prob works fine as a name

src/code/z_collision_check.c
```diff
@@ -3147,17 +3185,17 @@ void func_80062E14(GlobalContext* globalCtx, Vec3f* v, Vec3f* arg2) {
...
     init.lightPoint.y = init.position.y;
     init.lightPoint.z = init.position.z;
 
-    Effect_Add(globalCtx, &sp24, EFFECT_SHIELD_PARTICLE, 0, 1, &init);
-    Audio_PlaySoundGeneral(NA_SE_IT_REFLECTION_WOOD, arg2, 4, &D_801333E0, &D_801333E0, &D_801333E8);
+    Effect_Add(globalCtx, &pIndex, EFFECT_SHIELD_PARTICLE, 0, 1, &init);
+    Audio_PlaySoundGeneral(NA_SE_IT_REFLECTION_WOOD, actorPos, 4, &D_801333E0, &D_801333E0, &D_801333E8);
 }
 
 #define SQXZ(vec) (SQ(vec.x) + SQ(vec.z))
```
fig02, line 3201: perhaps something for math.h?

src/overlays/actors/ovl_En_Reeba/z_en_reeba.c
```diff
@@ -49,8 +49,13 @@ const ActorInit En_Reeba_InitVars = {
 };
 
 static ColliderCylinderInit sCylinderInit = {
-    { COLTYPE_UNK5, 0x11, 0x09, 0x39, 0x10, COLSHAPE_CYLINDER },
-    { 0x00, { 0xFFCFFFFF, 0x08, 0x08 }, { 0xFFCFFFFF, 0x00, 0x00 }, 0x01, 0x05, 0x01 },
+    { COLTYPE_HIT5, AT_ON | AT_ENEMY, AC_ON | AC_PLAYER, OC_ON | OC_ALL, OT_TYPE1, COLSHAPE_CYLINDER },
+    { ELEMTYPE_UNK0,
```
fig02, line 53: im kinda feeling like parens around the flags with | will make it easier to see which belong to which, but want to know what others think

src/code/z_collision_check.c
fig02: also can you convert the colors here to decimal

src/code/z_collision_check.c
petrie911: I apparently hadn't merged this since the math3d cyl vs line seg update. this is in math.h now.

src/code/z_collision_check.c
petrie911: decimal yes, outside no

src/code/z_collision_check.c
fig02: oh i meant the D_ symbol specifically, the temp is the effect index so bloodIndex or something makes sense for that

petrie911: I'm leaving all the bss symbols as D_ for now because of the current bss ordering issues in the file. Since they're all static, it'll be easy to fix them later.

fig02: ah thats fair, np

fig02 (approved):
Im good with these changes, though there are a few things that we need to confirm are good with other people. The names of the Mass defines for example were a hot debate topic that we never reached a conclusion on. But I think I am ok with `HEAVY` and `IMMOBILE` that you have here

include/z64collision_check.h
```diff
@@ -9,245 +9,391 @@
...
+#define OCELEM_UNK6 (1 << 6) // Apparently unused
+#define OCELEM_UNK7 (1 << 7) // Apparently unused
+
+#define OCLINE_NONE 0 // Did not have an OcLine collision
+#define OCLINE_HIT (1 << 0) // Had an OcLine collision
+
+#define DMG_ENTRY(damage, effect) ((damage) | ((effect) << 4))
+
+// These flags are not to be used in code until we figure out how we want to format them. They are only here for reference 
+#define DAMAGE_DEKU_NUT     (1 << 0x00)
```
Roman971, line 356: I suggest shortening the prefix to `DMG`, and reordering common terms to be `ARROW_FIRE` instead of `FIRE_ARROW` for instance (like we did with item enums and such).

I also suggest using decimal instead of hex for the shifts, like usual flags, but that's not very important.

src/code/z_collision_check.c
```diff
@@ -1,13 +1,53 @@
...
+    /* 1 */ u8 effect;
+} HitInfo; // size = 0x2
+
+typedef enum {
+    /* 0 */ NO_BLOOD,
+    /* 1 */ BLUE_BLOOD,
+    /* 2 */ GREEN_BLOOD,
+    /* 3 */ WATER_BURST,
+    /* 4 */ RED_BLOOD,
+    /* 5 */ RED_BLOOD2
```
Roman971, line 22: I suggest using blood as a prefix for this enum, to make things a bit cleaner and more consistent

src/code/z_actor.c
```diff
@@ -3198,8 +3201,8 @@ void func_80033480(GlobalContext* globalCtx, Vec3f* arg1, f32 arg2, s32 arg3, s1
 }
 
 Actor* Actor_GetCollidedExplosive(GlobalContext* globalCtx, Collider* collider) {
-    if ((collider->acFlags & 0x2) && (collider->ac->type == ACTORTYPE_EXPLOSIVES)) {
-        collider->acFlags &= ~0x2;
+    if ((collider->acFlags & AC_HIT) && (collider->ac->type == ACTORTYPE_EXPLOSIVES)) {
+        collider->acFlags &= ~AC_HIT;
```
Roman971, line 3205: I mentioned this before and explained why in discord, but I still really think we shouldn't introduce flags used in code everywhere until we're done or almost done decompiling

src/code/z_collision_check.c
```diff
@@ -72,58 +115,63 @@ s32 Collider_DestroyBase(GlobalContext* globalCtx, Collider* collider) {
...
+    collider->ocType = src->ocType;
     collider->shape = src->shape;
     return 1;
 }
 
-void Collider_BaseSetAT(GlobalContext* globalCtx, Collider* collider) {
+void Collider_ResetATBase(GlobalContext* globalCtx, Collider* collider) {
     collider->at = NULL;
-    collider->atFlags &= ~0x6;
+    collider->atFlags &= ~AT_HIT & ~AT_BOUNCED;
```
Roman971, line 158: Probably relevant in other places, and I don't know if we have any instances already in the repo, but this would typically be written as `~(a | b)` instead
```suggestion
    collider->atFlags &= ~(AT_HIT | AT_BOUNCED);
```

src/code/z_collision_check.c
```diff
@@ -1378,505 +1559,535 @@ void func_8005E10C(GlobalContext* globalCtx, Collider* collider, Vec3f* pos) {
...
+/**
+ * Plays sound effects and displays hitmarks for solid-type AC colliders (METAL, WOOD, HARD, and TREE)
+ */
+void CollisionCheck_HitSolid(GlobalContext* globalCtx, ColliderInfo* info, Collider* collider, Vec3f* hitPos) {
     s32 flags;
 
-    flags = colliderBody->toucherFlags & 0x18;
-    if (flags == 0 && collider->type != COLTYPE_METAL_SHIELD) {
-        EffectSsHitMark_SpawnFixedScale(globalCtx, 0, arg3);
+    flags = info->toucherFlags & (3 << 3);
```
Roman971, line 1596: I think `(3 << 3)` doesn't make much sense for this, it's probably supposed to be flags 8 and 0x10 `|`'d together

src/code/z_collision_check.c
```diff
@@ -2722,244 +2990,286 @@ void func_80061EFC(CollisionCheckInfo* info, DamageTable* damageTable, Collision
...
+    element->dim.plane.normal.z = nz;
+    element->dim.plane.originDist = originDist;
 }
 
-// Set ColliderTrisItem at index
-void func_8006285C(GlobalContext* globalCtx, ColliderTris* collider, s32 index, ColliderTrisItemDimInit* init) {
-    ColliderTrisItem* item = &collider->list[index];
-    Collider_SetTrisItemDim(globalCtx, &item->dim, init);
+/**
+ * Sets the specified ColliderTrisElement's dim usig the values in src
```
Roman971, line 3267:
```suggestion
 * Sets the specified ColliderTrisElement's dim using the values in src
```

src/code/z_collision_check.c
```diff
@@ -3086,90 +3407,105 @@ void func_80062B80(GlobalContext* globalCtx, Vec3f* v) {
...
         35.0f,
         30.0f,
         8,
-        { 0, 0, 0, 0x00, 0x80, 0xFF, 0x00, 300 },
-        1,
+        { 0, 0, 0, { 0, 128, 255 }, false, 300 },
+        true,
     };
-    s32 sp24;
+    s32 shield;
```
Roman971, line 3433: Small naming suggestion (also relevant in the other shield particle spawn)
```suggestion
    s32 effectIndex;
```

src/code/z_collision_check.c
```diff
@@ -3086,90 +3407,105 @@ void func_80062B80(GlobalContext* globalCtx, Vec3f* v) {
...
-    Effect_Add(globalCtx, &sp24, EFFECT_SHIELD_PARTICLE, 0, 1, &init);
+    Effect_Add(globalCtx, &shield, EFFECT_SHIELD_PARTICLE, 0, 1, &sInitMetal);
 }
 
-void func_80062D60(GlobalContext* globalCtx, Vec3f* v) {
-    func_80062CD4(globalCtx, v);
+/**
+ * Spawns streaks of light and makes a metallic sound
+ */
+void CollisionCheck_ShieldParticlesMetal(GlobalContext* globalCtx, Vec3f* v) {
```
Roman971, line 3448: It would probably be a good idea to always include "spawn" in the name, otherwise its purpose could be slightly confusing.

src/code/z_collision_check.c
```diff
@@ -3086,90 +3407,105 @@ void func_80062B80(GlobalContext* globalCtx, Vec3f* v) {
...
     Audio_PlaySoundGeneral(NA_SE_IT_SHIELD_REFLECT_SW, &D_801333D4, 4, &D_801333E0, &D_801333E0, &D_801333E8);
 }
 
-void func_80062DAC(GlobalContext* globalCtx, Vec3f* v, Vec3f* arg2) {
-    func_80062CD4(globalCtx, v);
-    Audio_PlaySoundGeneral(NA_SE_IT_SHIELD_REFLECT_SW, arg2, 4, &D_801333E0, &D_801333E0, &D_801333E8);
+/**
+ * Spawns streaks of light and makes a metallic sound at the specified position
+ */
+void CollisionCheck_ShieldParticlesMetalSound(GlobalContext* globalCtx, Vec3f* v, Vec3f* actorPos) {
```
Roman971, line 3456: I would omit the "actor" in the pos argument, since it could be used with any generic position as far as I can tell

src/overlays/actors/ovl_Bg_Bdan_Switch/z_bg_bdan_switch.c
```diff
@@ -59,17 +59,31 @@ extern CollisionHeader D_06005CF8;
...
+        COLTYPE_NONE,
+        AT_OFF,
+        AC_ON | AC_PLAYER,
+        OC_ON | OC_ALL,
+        OT_TYPE2,
+        COLSHAPE_JNTSPH,
+    },
     1,
-    sJntSphItemsInit,
+    sJntSphElementsInit,
```
Roman971, line 86: It doesn't have to happen in this set of changes, but I think it could be good to design macros for collider inits, so we can reorder arguments as we want, and most importantly avoid duplicated information such as the `COLSHAPE` and the `ARRAY_COUNT` in colliders that can have multiple elements (ie. the 1 right before the pointer to the init array in this example)

src/overlays/actors/ovl_Bg_Bdan_Switch/z_bg_bdan_switch.c
```diff
@@ -59,17 +59,31 @@ extern CollisionHeader D_06005CF8;
 extern Gfx D_060061A0[];
 extern Gfx D_06005A20[];
 
-static ColliderJntSphItemInit sJntSphItemsInit[] = {
+static ColliderJntSphElementInit sJntSphElementsInit[] = {
```
Roman971, line 62: While on the topic of collider inits, I want to suggest that we make defines for the elements count of each collider. It would be in the header of the actor so it can be used in the struct for the size of the elements array, as well as the size of the init array in the C file. This would ensure they are always the same size, mostly for modding purposes.

This also doesn't have to happen in this set of changes tho, just bringing it up so we can start thinking about it.

Roman971 (commented):
Finally had time to write down my thoughts on this, so here they are. Sorry for the very long delay

include/z64actor.h
```diff
@@ -9,6 +9,8 @@
 #define ACTOR_NUMBER_MAX 200
 #define INVISIBLE_ACTOR_MAX 20
 #define AM_FIELD_SIZE 0x27A0
+#define MASS_IMMOBILE 0xFF // Cannot be pushed in OC collisions
```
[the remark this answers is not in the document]
fig02: `MASS_IMMOVABLE` sounds fine to me
unrelated, the comments for both say "pushed in OC collisions", but I think "pushed by" makes more sense

src/code/z_actor.c
```diff
@@ -1464,7 +1467,7 @@ s32 func_8002F0C8(Actor* actor, Player* player, s32 flag) {
             dist = actor->xyzDistToLinkSq;
         }
 
-        return !func_8002F090(actor, D_80115FF8[actor->unk_1F].unk_4 * dist);
+        return !func_8002F090(actor, D_80115FF8[actor->unk_1F].leashMod * dist);
```
fig02, line 1470: does mod make sense for a name here? seems more like a scale factor to me

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -1934,14 +1934,14 @@ s32 func_80833CDC(GlobalContext* globalCtx, s32 index) {
 }
 
 void func_80833DF8(Player* this, GlobalContext* globalCtx) {
-    s32 maskActionParam;
+    s32 ocFlagsctionParam;
```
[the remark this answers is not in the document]
petrie911: yeah.

include/z64.h
```diff
@@ -663,15 +663,15 @@ typedef struct {
...
-    /* 0x0CC */ s32 colAcCount;
-    /* 0x0D0 */ Collider* colAc[COLLISION_CHECK_AC_MAX];
-    /* 0x1C0 */ s32 colOcCount;
-    /* 0x1C4 */ Collider* colOc[COLLISION_CHECK_OC_MAX];
-    /* 0x28C */ s32 colOcLineCount;
-    /* 0x290 */ OcLine* colOcLine[COLLISION_CHECK_OC_LINE_MAX];
+    /* 0x004 */ Collider* colAT[COLLISION_CHECK_AT_MAX];
+    /* 0x0CC */ s32 colACcount;
+    /* 0x0D0 */ Collider* colAC[COLLISION_CHECK_AC_MAX];
+    /* 0x1C0 */ s32 colOCcount;
```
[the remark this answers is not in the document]
petrie911: AT, AC, and OC are consistently used capitalized elsewhere though.

include/z64actor.h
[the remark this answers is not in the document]
petrie911: This seems fine.

include/z64.h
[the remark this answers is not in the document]
petrie911: AT, AC, and OC are consistently capitalized. I'd prefer to keep that here.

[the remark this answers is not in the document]
fig02: think he means count being lowercase

include/z64collision_check.h
petrie911: Would that also be `DMG_KOKIRI_SLASH -> DMG_SLASH_KOKIRI`?

src/code/z_collision_check.c
petrie911: alternatively, it's `TOUCH_SFX_NONE`

src/code/z_actor.c
petrie911: This should probably go in your actor PR rather than here.

fig02: yea it probably makes more sense there, but since its already done here might as well. it doesnt affect external/actor code.
do you agree that `leashScale` or factor or something makes more sense?

src/code/z_collision_check.c
petrie911: `shieldIndex` would be consistent with the current blood spawners. Though I suppose they could all just be `effectIndex`

include/z64collision_check.h
```diff
@@ -9,245 +9,391 @@
...
+#define OCELEM_UNK3 (1 << 3) // Unknown purpose. Used by Dead Hand element 0 and Dodongo element 5
+#define OCELEM_UNK4 (1 << 4) // Apparently unused
+#define OCELEM_UNK5 (1 << 5) // Apparently unused
+#define OCELEM_UNK6 (1 << 6) // Apparently unused
+#define OCELEM_UNK7 (1 << 7) // Apparently unused
+
+#define OCLINE_NONE 0 // Did not have an OcLine collision
+#define OCLINE_HIT (1 << 0) // Had an OcLine collision
+
+#define DMG_ENTRY(damage, effect) ((damage) | ((effect) << 4))
```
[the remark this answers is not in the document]
petrie911: Effect first will make the damage table very hard to read due to the formatter. Consider this excerpt from `EnRr`:
```
    /* Shadow arrow  */ DMG_ENTRY(RR_DMG_SHDW_ARROW, 0),
    /* Spirit arrow  */ DMG_ENTRY(RR_DMG_SPRT_ARROW, 0),
    /* Fire magic    */ DMG_ENTRY(RR_DMG_FIRE, 0),
    /* Ice magic     */ DMG_ENTRY(RR_DMG_ICE, 0),
    /* Light magic   */ DMG_ENTRY(RR_DMG_LIGHT_MAGIC, 0),
    /* Shield        */ DMG_ENTRY(RR_DMG_NONE, 0),
    /* Mirror Ray    */ DMG_ENTRY(RR_DMG_NONE, 0),
```
vs
```
    /* Shadow arrow  */ DMG_ENTRY(15, RR_DMG_SHDW_ARROW),
    /* Spirit arrow  */ DMG_ENTRY(15, RR_DMG_SPRT_ARROW),
    /* Fire magic    */ DMG_ENTRY(4, RR_DMG_FIRE),
    /* Ice magic     */ DMG_ENTRY(3, RR_DMG_ICE),
    /* Light magic   */ DMG_ENTRY(10, RR_DMG_LIGHT_MAGIC),
    /* Shield        */ DMG_ENTRY(0, RR_DMG_NONE),
    /* Mirror Ray    */ DMG_ENTRY(0, RR_DMG_NONE),
```

src/overlays/actors/ovl_Bg_Bdan_Switch/z_bg_bdan_switch.c
petrie911: I've been considering this as well. Definitely something for later, but it will be a good thing to have.

include/z64collision_check.h
Roman971: I think so yeah

src/code/z_collision_check.c
Roman971: I would make them all `effectIndex` yeah

src/code/z_actor.c
petrie911: `leashScale` seems fine.

include/z64collision_check.h
```diff
@@ -9,245 +9,391 @@
...
+#define AT_OFF 0 // Cannot have AT collisions when set as AT
+#define AT_ON (1 << 0) // Can have AT collisions when set as AT
+#define AT_HIT (1 << 1) // Had an AT collision
+#define AT_BOUNCED (1 << 2) // Had an AT collision with an AC_HARD collider
+#define AT_PLAYER (1 << 3) // Has player-aligned damage
+#define AT_ENEMY (1 << 4) // Has enemy-aligned damage
+#define AT_OTHER (1 << 5) // Has non-aligned damage
+#define AT_SELF (1 << 6) // Can have AT collisions with colliders attached to the same actor
+#define AT_UNK7 (1 << 7) // Apparently unused
+#define AT_ALL (AT_PLAYER | AT_ENEMY | AT_OTHER) // Has all three damage alignments
```
Roman971, line 285: `AT_ALL` (and its `AC`/`OC` equivalents) sounds confusing since it's only certain flags. When seeing this name, I would personally expect it to be the equivalent of every flag toggled.

Maybe we can prefix the 3 relevant flags and the combined flag with the same thing? Perhaps something like `AT_TYPE_PLAYER` and such, with the combined flag as `AT_TYPE_ALL`?

include/z64collision_check.h
```diff
@@ -9,245 +9,391 @@
...
+    ELEMTYPE_UNK2,
+    ELEMTYPE_UNK3,
+    ELEMTYPE_UNK4,
+    ELEMTYPE_UNK5,
+    ELEMTYPE_UNK6,
+    ELEMTYPE_UNK7
+} ElementType;
+
+
+#define AT_OFF 0 // Cannot have AT collisions when set as AT
```
Roman971, line 276: There are usually no defines for "no flags set", however this case is a bit particular so it's probably fine. But I think they should all be called `_NONE` or something, like you did for OCLINE, to really emphasize on the fact that they mean "nothing enabled" (and not just the opposite of the `_ON` flag for instance).

petrie911: AC_OFF exists as a separate flag because of Bongo and Fishing (thanks Morita), which have colliders that start off.

include/z64collision_check.h
petrie911: `TYPE` prefix should work.

petrie911: Although `OC_TYPE_PLAYER` is what `OT_PLAYER` is short for.

include/z64collision_check.h
fig02: im confused here actually, isnt it supposed to be the opposite of the `on` flag? calling it `off` seems fine to me

fig02: oh no i understand. the whole field is 0 so its supposed to be none as in "no flags"

fig02 (approved):

include/z64collision_check.h
```diff
@@ -9,245 +9,382 @@
...
+#define OC1_TYPE_2 (1 << 5) // Can have OC collisions with OC type 2
+#define OC1_TYPE_ALL (OC1_TYPE_PLAYER | OC1_TYPE_1 | OC1_TYPE_2) // Can have collisions with all three OC types
+
+#define OC2_NONE 0 // No flags set. Has no OC type
+#define OC2_HIT_PLAYER (1 << 0) // Had an OC collision with OC type player
+#define OC2_UNK1 (1 << 1) // Prevents OC collisions with OC2_UNK2. Some horses and toki_sword have it.
+#define OC2_UNK2 (1 << 2) // Prevents OC collisions with OC2_UNK1. Nothing has it.
+#define OC2_TYPE_PLAYER (1 << 3) // Has OC type player
+#define OC2_TYPE_1 (1 << 4) // Has OC type 1
+#define OC2_TYPE_2 (1 << 5) // Has OC type 2
```
Roman971, line 311: I think we also wanted to do this:
```suggestion
#define OC2_TYPE_PLAYER OC1_TYPE_PLAYER // Has OC type player
#define OC2_TYPE_1 OC1_TYPE_1 // Has OC type 1
#define OC2_TYPE_2 OC1_TYPE_2 // Has OC type 2
```

include/z64collision_check.h
```diff
@@ -9,245 +9,382 @@
...
-} ColliderQuadInit_Set3; // size = 0x50
+} ColliderQuadInitType1; // size = 0x50
 
 typedef struct {
     /* 0x00 */ Linef line;
-    /* 0x18 */ u16 unk_18;
+    /* 0x18 */ u16 ocFlags;
 } OcLine; // size = 0x1C
 
+typedef enum ColliderType {
```
Roman971, line 233:
```suggestion
typedef enum {
```

include/z64collision_check.h
```diff
@@ -9,245 +9,382 @@
...
+    /* 7  */ COLTYPE_HIT7, // Red blood, white hitmark
+    /* 8  */ COLTYPE_HIT8, // Blue blood, red hitmark
+    /* 9  */ COLTYPE_METAL,
+    /* 10 */ COLTYPE_NONE,
+    /* 11 */ COLTYPE_WOOD,
+    /* 12 */ COLTYPE_HARD,
+    /* 13 */ COLTYPE_TREE
+} ColliderType;
+
+typedef enum ColliderShape {
```
Roman971, line 250:
```suggestion
typedef enum {
```

include/z64collision_check.h
```diff
@@ -9,245 +9,382 @@
...
+    /* 4 */ COLSHAPE_INVALID
+} ColliderShape;
+
+/**
+ * Affects the sound Link's sword makes when hitting it, hookability,
+ * and possibly other things. It's definitely not flags, as all checks
+ * are == or !=. Will probably need more actors decomped to truly 
+ * understand what this is.
+ */
+typedef enum ElementType {
```
Roman971, line 264:
```suggestion
typedef enum {
```

src/code/z_en_a_keep.c
```diff
@@ -34,7 +34,14 @@ const ActorInit En_A_Obj_InitVars = {
...
+    {
+        COLTYPE_NONE,
+        0x00,
+        0x39,
+        0x39,
+        0x20,
+        COLSHAPE_CYLINDER,
+    },
     { 0x02, { 0x00000000, 0x00, 0x00 }, { 0xFFCFFFFF, 0x00, 0x00 }, 0x00, 0x01, 0x01 },
     { 25, 60, 0, { 0, 0, 0 } },
```
Roman971, line 46: I believe this collider needs to be updated?

Roman971 (commented):
Noticed a few issues, but otherwise it looks good to me

src/code/z_en_a_keep.c
petrie911: Should be OK now.

fig02:
Rest of the changes made.
Also added the mass defines into code, they were only in Info Inits before

Roman971 (approved):

PR #472, opened 2020-11-09
En_Rr

petrie911:
Nice 0x2000 array there.


src/overlays/actors/ovl_En_Rr/z_en_rr.c
```diff
@@ -1,15 +1,75 @@
...
 
 #define THIS ((EnRr*)thisx)
 
+#define RR_MESSAGE_SHIELD (1 << 0)
+#define RR_MESSAGE_TUNIC (1 << 1)
+#define RR_MOUTH 4
+#define RR_BASE 0
+
+typedef enum { REACH_NONE, REACH_EXTEND, REACH_STOP, REACH_OPEN, REACH_GAPE, REACH_CLOSE } EnRrReachState;
+
```
fig02, line 20: the formatter doing this to the enum is truly a shame lol. and trailing commas on enums induce warnings so, welp

src/overlays/actors/ovl_En_Rr/z_en_rr.c
```diff
@@ -1,15 +1,75 @@
...
+} EnRrDropType;
+
 void EnRr_Init(Actor* thisx, GlobalContext* globalCtx);
 void EnRr_Destroy(Actor* thisx, GlobalContext* globalCtx);
 void EnRr_Update(Actor* thisx, GlobalContext* globalCtx);
 void EnRr_Draw(Actor* thisx, GlobalContext* globalCtx);
 
-/*
+// void EnRr_SetSpeed(EnRr* this, f32 speed);
+
```
fig02, line 48: assuming the commented prototypes arent needed, in which case remove

src/overlays/actors/ovl_En_Rr/z_en_rr.c
```diff
@@ -21,51 +81,772 @@ const ActorInit En_Rr_InitVars = {
...
+                hitPos.x = this->collider1.body.bumper.unk_06.x;
+                hitPos.y = this->collider1.body.bumper.unk_06.y;
+                hitPos.z = this->collider1.body.bumper.unk_06.z;
+                func_8005DFAC(globalCtx, NULL, &hitPos);
+            }
+            switch (this->actor.colChkInfo.damageEffect) {
+                case RR_DAMAGE_LIGHT_ARROW: // Light Arrow
+                    dropType++;             // purple rupee
+                case RR_DAMAGE_UNK_ARROW_1: // Unk Arrow 1
+                    dropType++;             // flexible
```
fig02, line 389: comments not needed with the defines

src/overlays/actors/ovl_En_Rr/z_en_rr.c
```diff
@@ -21,51 +81,772 @@ const ActorInit En_Rr_InitVars = {
...
+    Matrix_Push();
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Rr/func_80AE98F8.s")
+    Matrix_Scale((1.0f + this->bodySegs[RR_BASE].scaleMod.x) * this->bodySegs[RR_BASE].scale.x,
+                 (1.0f + this->bodySegs[RR_BASE].scaleMod.y) * this->bodySegs[RR_BASE].scale.y,
+                 (1.0f + this->bodySegs[RR_BASE].scaleMod.z) * this->bodySegs[RR_BASE].scale.z, 1);
+    gSPMatrix(POLY_XLU_DISP++, Matrix_NewMtx(globalCtx->state.gfxCtx, "../z_en_rr.c", 1501),
+              G_MTX_NOPUSH | G_MTX_LOAD | G_MTX_MODELVIEW);
+    Matrix_Pull();
+    VEC_SET(zeroVec, 0.0f, 0.0f, 0.0f);
```
fig02, line 807: had to be sameline to match?

src/overlays/actors/ovl_En_Rr/z_en_rr.c
```diff
@@ -21,51 +81,772 @@ const ActorInit En_Rr_InitVars = {
...
+    this->effectPos[0] = this->actor.posRot.pos;
+    Matrix_MultVec3f(&zeroVec, &this->mouthPos);
+    gSPDisplayList(POLY_XLU_DISP++, D_06000470);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Rr/func_80AE9D1C.s")
+    CLOSE_DISPS(globalCtx->state.gfxCtx, "../z_en_rr.c", 1551);
+    if (this->effectTimer != 0) {
+        static Vec3f effectOffsets[] = {
+            { 25.0f, 0.0f, 0.0f },
+            { -25.0f, 0.0f, 0.0f },
```
fig02, line 829: have to be in function to match?

src/overlays/actors/ovl_En_Rr/z_en_rr.c
petrie911: Probably not. Should `VEC_SET` only be used if needed?

fig02: for now yeah, we have very few uses of it and it was only created for sameline purposes.
imo the macro probably didnt even exist and it was just written on the same line sometimes. but we can decide if we want to use it more later

src/overlays/actors/ovl_En_Rr/z_en_rr.c
petrie911: No, but it's where it makes the most sense.

src/overlays/actors/ovl_En_Rr/z_en_rr.c
petrie911: Can get it to stack properly by giving the enum a tag.

src/overlays/actors/ovl_En_Rr/z_en_rr.c
fig02: i see. i prefer not to put longer pieces of data inside of functions when possible. this one is kinda on the cusp of where id draw the line.
Id probably put it right above the function if it were me, but its probably fine either way

petrie911: I've moved it outside, but wait on merging this. Taking one last look through the code makes me think I might have misnamed a couple fields.

petrie911: OK, fixed names.

fig02 (approved):

src/overlays/actors/ovl_En_Rr/z_en_rr.c
```diff
@@ -21,51 +83,775 @@ const ActorInit En_Rr_InitVars = {
...
+                case RR_DAMAGE_NORMAL:
+                    // ouch
+                    osSyncPrintf(VT_FGCOL(RED) "いてっ( %d : LIFE %d : DAMAGE %d : %x )！！" VT_RST "\n",
+                                 this->frameCount, this->actor.colChkInfo.health, this->actor.colChkInfo.damage,
+                                 this->actor.colChkInfo.damageEffect);
+                    this->stopScroll = false;
+                    Actor_ApplyDamage(&this->actor);
+                    this->invincibilityTimer = 40;
+                    func_8003426C(&this->actor, 0x4000, 0xFF, 0x2000, this->invincibilityTimer);
+                    if (this->hasPlayer) {
```
NEstelami, line 405: For non-booleans we typically do if (this->hasPlayer != 0).

src/overlays/actors/ovl_En_Rr/z_en_rr.c
```diff
@@ -21,51 +83,775 @@ const ActorInit En_Rr_InitVars = {
...
+    }
+
+    Actor_SetHeight(&this->actor, 30.0f);
+    EnRr_UpdateBodySegments(this, globalCtx);
+    if (!this->isDead && ((this->actor.dmgEffectTimer == 0) || !(this->actor.dmgEffectParams & 0x4000))) {
+        EnRr_CollisionCheck(this, globalCtx);
+    }
+
+    this->actionFunc(this, globalCtx);
+    if (this->hasPlayer == 0x3F80) { // checks if hasPlayer is 1.0f
```
NEstelami, line 740: What is this exactly? Some kind of truncated float?

NEstelami (changes requested):

src/overlays/actors/ovl_En_Rr/z_en_rr.c
petrie911: I think it's checking for if `1.0f` has been mistakenly written to that address instead of `1`. I'm not sure how such a thing could happen, but `hasPlayer` is definitely a boolean.

NEstelami: Interesting. Odd that they used an s16. Is it possible to match by doing something like `if (this->hasPlayer == 1.0f)`? Also, maybe put a comment explaining what you just explained to me, like, `// @bug The programmer who worked on this accidentally used a floating point 1.0 instead of a 1 here`, making it clear it's a bug.

petrie911: I don't think it's a bug. Given that if the check succeeds it asserts and crashes the game, I think checking if `1.0f` has been written there instead of 1 is an intentional error check.

NEstelami: Gotcha.

NEstelami (approved):

src/overlays/actors/ovl_En_Rr/z_en_rr.h
```diff
@@ -6,9 +6,57 @@
 
 struct EnRr;
 
+typedef void (*EnRrActionFunc)(struct EnRr*, GlobalContext*);
+
+typedef struct EnRrBodySegment {
```
Roman971, line 11:
```suggestion
typedef struct {
```

src/overlays/actors/ovl_En_Rr/z_en_rr.c
```diff
@@ -1,15 +1,77 @@
...
 #define FLAGS 0x00000435
 
 #define THIS ((EnRr*)thisx)
 
+#define RR_MESSAGE_SHIELD (1 << 0)
+#define RR_MESSAGE_TUNIC (1 << 1)
+#define RR_MOUTH 4
+#define RR_BASE 0
+
+typedef enum EnRrReachState {
```
Roman971, line 19: Use value comments (as hex) instead of duplicating the name to the `enum`, imo.
That will ensure proper vertical formatting, and it's helpful anyway so I think it should be the default for all enums. Also relevant to the other enums.
```suggestion
typedef enum {
```

src/overlays/actors/ovl_En_Rr/z_en_rr.c
```diff
@@ -1,15 +1,77 @@
...
+    RR_DAMAGE_STUN = 1,
+    RR_DAMAGE_FIRE,
+    RR_DAMAGE_ICE,
+    RR_DAMAGE_LIGHT_MAGIC,
+    RR_DAMAGE_LIGHT_ARROW = 11,
+    RR_DAMAGE_UNK_ARROW_1,
+    RR_DAMAGE_UNK_ARROW_2,
+    RR_DAMAGE_UNK_ARROW_3,
+    RR_DAMAGE_NORMAL
+} EnRrDamageEffect;
```
Roman971, line 38: Imo this general enum doesn't belong in the file, and should be kept for a later set of documentation changes where the enum is applied to the whole codebase.

src/overlays/actors/ovl_En_Rr/z_en_rr.c
```diff
@@ -1,15 +1,77 @@
...
+
+typedef enum {
+    RR_DROP_RANDOM_RUPEE,
+    RR_DROP_MAGIC,
+    RR_DROP_ARROW,
+    RR_DROP_FLEXIBLE,
+    RR_DROP_PURPLE_RUPEE,
+    RR_DROP_RED_RUPEE
```
Roman971, line 46: I suggest keeping these names consistent with the EnItem00 enum names.
For example:
```suggestion
    RR_DROP_RUPEE_RED
```

src/overlays/actors/ovl_En_Rr/z_en_rr.c
```diff
@@ -1,15 +1,77 @@
...
+} EnRrDropType;
+
 void EnRr_Init(Actor* thisx, GlobalContext* globalCtx);
 void EnRr_Destroy(Actor* thisx, GlobalContext* globalCtx);
 void EnRr_Update(Actor* thisx, GlobalContext* globalCtx);
 void EnRr_Draw(Actor* thisx, GlobalContext* globalCtx);
 
-/*
+void EnRr_SetupReach(EnRr* this);
```
Roman971, line 54: Most of these prototypes seem unnecessary so they should be excluded imo.

src/overlays/actors/ovl_En_Rr/z_en_rr.c
```diff
@@ -21,51 +83,775 @@ const ActorInit En_Rr_InitVars = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Rr/func_80AE8C44.s")
+    this->actor.flags |= 1;
+    this->hasPlayer = false;
+    this->ocTimer = 110;
+    this->segMoveRate = 0.0f;
+    this->segPhaseVelTarget = 2500.0f;
+    this->wobbleSizeTarget = 2048.0f;
+    tunic = PLAYER_TUNIC_KOKIRI;
+    shield = PLAYER_SHIELD_NONE;
```
Roman971, line 249: Imo, PLAYER_TUNIC_* and PLAYER_SHIELD_* enums shouldn't be used when dealing with equipment, because it will get pretty confusing. These enums are specifically intended for use when dealing with player fields/functions.

src/overlays/actors/ovl_En_Rr/z_en_rr.c
```diff
@@ -21,51 +83,775 @@ const ActorInit En_Rr_InitVars = {
...
+    gSPSegment(POLY_XLU_DISP++, 0x08,
+               Gfx_TwoTexScroll(globalCtx->state.gfxCtx, 0, (this->scrollTimer * 0) & 0x7F,
+                                (this->scrollTimer * 0) & 0x3F, 32, 16, 1, (this->scrollTimer * 0) & 0x3F,
+                                (this->scrollTimer * -6) & 0x7F, 32, 16));
+    Matrix_Push();
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Rr/func_80AE98F8.s")
+    Matrix_Scale((1.0f + this->bodySegs[RR_BASE].scaleMod.x) * this->bodySegs[RR_BASE].scale.x,
+                 (1.0f + this->bodySegs[RR_BASE].scaleMod.y) * this->bodySegs[RR_BASE].scale.y,
+                 (1.0f + this->bodySegs[RR_BASE].scaleMod.z) * this->bodySegs[RR_BASE].scale.z, 1);
```
Roman971, line 812: Relevant in other places
```suggestion
                 (1.0f + this->bodySegs[RR_BASE].scaleMod.z) * this->bodySegs[RR_BASE].scale.z, MTXMODE_APPLY);
```

src/overlays/actors/ovl_En_Rr/z_en_rr.c
Roman971: 1.0f is 0x3F800000, not 0x3F80. So this comment is technically incorrect

src/overlays/actors/ovl_En_Rr/z_en_rr.c
petrie911: Damage effect indices aren't general. Each actor assigns what they do individually.

src/overlays/actors/ovl_En_Rr/z_en_rr.h
petrie911: Why do we prefer anonymous structs?

Roman971: As long as we can, we use types for structs, not the struct itself (ie. we use them as `EnRrBodySegment`, not `struct EnRrBodySegment`). In this case the struct name isn't needed so we can just omit it.

src/overlays/actors/ovl_En_Rr/z_en_rr.c
Roman971: I understand that actors define their own damage tables, but wouldn't the damage effects themselves be a universal enum?

I checked other decompiled actors and at least it looks like 1 is always used for stun and 2 is always used for fire/burn, so I imagine that's not a coincidence. Effect 0 seems to be consistently "none" as well btw.

petrie911: Leever uses 11, 12, and 13 for stun, 14 for normal damage, and 3 for ice damage. Redead uses 14 for burning, 1 and 13 for stun, and possibly intended 6 for ice. Guay and Octorok use 0 for normal damage. Shell blade uses 1 for hookshot and 13-15 for various other kinds of damage. I haven't looked through Iron Knuckle or Gossip Stone, but their damage effects are all over the place.

Roman971:
I left a comment in Jack's review, in case you missed it

src/overlays/actors/ovl_En_Rr/z_en_rr.c
petrie911: Now technically correct. (The best kind of correct?)

Roman971 (approved):

PR #478, opened 2020-11-11
Updated Texture Asset Handling

NEstelami:
- Texture Compilation: Each texture is now compiled from a `.png` file to a `.inc.c` file by ZAP. These get #include'd into a main `.c` file for a given object. Anytime a png file is modified, the corresponding `.inc.c` and main `.c` file will be recompiled.
- The `.gitignore` has been updated to handle custom assets. Any asset with the filename suffix `_custom` will be kept in the repository.
- Object_Vase and Pu_Box have been decompiled and the corresponding code files have been modified to use the appropriate symbols. Gameplay_Keep and Object_SK2 have also been mostly decompiled, however currently no code makes reference to them.
- A few structs were changed in order to comply with MZX's changes to ZAP. When MZX merges in his PR, it will include more of these changes.
- The `scenes/` folder has been moved to the `assets/` folder. This was discussed previously in the Discord, and while there was some disagreement on the decision, the overall consensus was in favor of it. If further discussion is needed on this, let's discuss it in the contributors channel.


src/overlays/actors/ovl_En_Vase/z_en_vase.c
```diff
@@ -5,6 +5,7 @@
  */
 
 #include "z_en_vase.h"
+#include <../assets/objects/object_vase/object_vase.h>
```
fig02, line 8: with our somewhat recent libultra cleanup pr we dont use <> for includes anymore, only quotes.
We also try to avoid relative paths when possible

src/overlays/actors/ovl_En_Vase/z_en_vase.c
```diff
@@ -40,5 +39,5 @@ void EnVase_Destroy(Actor* thisx, GlobalContext* globalCtx) {
 }
 
 void EnVase_Draw(Actor* thisx, GlobalContext* globalCtx) {
-    Gfx_DrawDListOpa(globalCtx, &D_06000000);
+    Gfx_DrawDListOpa(globalCtx, &objVaseMesh);
```
fig02, line 42: not sure on calling these meshes, a mesh is only one part of the overall display list for an actor right? this is relevant for the names for the other display lists like in stalfos as well

src/overlays/actors/ovl_En_Pu_box/z_en_pu_box.c
```diff
@@ -59,7 +59,7 @@ void EnPubox_Init(Actor* thisx, GlobalContext* globalCtx) {
     this->dyna.unk_15C = 0;
     thisx->unk_1F = 1;
     thisx->gravity = -2.0f;
-    DynaPolyInfo_Alloc(&D_060006D0, &local_c);
+    DynaPolyInfo_Alloc(&puBoxDynaPolyInfo, &local_c);
```
fig02, line 62: pretty sure the terminology we are going with for these is "collision header", but if you want to wait until bgcheck is merged to update those names thats fine too

fig02 (commented):
Other than these small things, Im not sure if there was anything else we needed discussion about. I feel like there was something related to the xml format in general but Im not sure. @Roman971 maybe this was something you had brought up?

src/overlays/actors/ovl_En_Vase/z_en_vase.c
NEstelami: This is part of why I wanted to settle on naming schemes some time ago. I'm open to changing what I have here, but I feel we need some standard otherwise there will be inconsistencies.

fig02: yea thats fair. I just think a safe bet until we have a more defined standard is to just call them dlists since thats what we refer to them as everywhere else

assets/xml/objects/object_sk2.xml
```diff
@@ -0,0 +1,55 @@
+<Root>
+    <File Name="object_sk2">
+        <DList Name="skeltonMeshHead" Address="0x53D0"/>
+        <DList Name="skeltonMeshThirdVertebrae" Address="0x5720"/>
+        <DList Name="skeltonMeshRightShoulderBone" Address="0x57B8"/>
```
fig02, line 5: are these names prefixed with skeleton as in sk2? Id do stalfos just cause that can get confused very quickly with the skeleton from skelanime.
Im sure these names are far from final and not the point of the PR, I just dont want to cause confusion as we start to think of standards for naming this stuff.

Also `skelton_mesh_fourth_vertebrae` below doesnt follow the same camel casing as the rest of the entries

NEstelami: A bunch of these names were either borrowed from the 3DS version or the Cloudmodding wiki.

NEstelami: And yeah, the names are by no means final.

include/z64cutscene_commands.h
```diff
@@ -437,4 +437,4 @@
  */
 #define CS_END() 0xFFFFFFFF, 0x00000000
 
-#endif
+#endif
```
ethteck, line 440: Here as well (and other scripts below)

ethteck (approved):

fixbaserom.py
```diff
@@ -70,4 +70,4 @@ def get_str_hash(byte_array):
 with open("baserom.z64", mode="wb") as file:
     file.write(bytes(strippedContent))
 
-print("Done!")
+print("Done!")
```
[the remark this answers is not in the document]
NEstelami: I have no idea. Every time I try to fix that it changes again.

fig02 (approved):
Im ok with these changes, but I do think its a good time to start using `number` in the spec now.

src/overlays/actors/ovl_En_Vase/z_en_vase.c
Zelllll: Do we really need to focus on naming right now anyway?

NEstelami:
> Im ok with these changes, but I do think its a good time to start using `number` in the spec now.

Do you want that to be accounted for in this PR?

Roman971:
Could you give an example of extracting and including a texture inlined inside a code segment (eg. the texture in `z_magic_dark_gfx.c`)? It doesn't necessarily have to be part of those changes, tho I think it could be good to have an example already done for future reference.

And I agree it would be a good time to start using `number`, ideally in this PR.

mzxrules:
I don't really agree that `number` should be part of the spec format, especially since you can just implement it's functionality in a macro

fig02:
roman has let me know privately that he approves, is unable to access github atm.
will be merging shortly

PR #485, opened 2020-11-14
En_Ssh (House of Skulltula spiders)

petrie911:
The Spider House Rules


src/overlays/actors/ovl_En_Ssh/z_en_ssh.c
```diff
@@ -4,12 +4,54 @@
 
 #define THIS ((EnSsh*)thisx)
 
+#define STATE_STUNNED (1 << 0)
+#define STATE_GROUND_START (1 << 2)
+#define STATE_ATTACKED (1 << 3)
+#define STATE_SPIN (1 << 4)
+
+typedef enum EnSshAnimation {
```
Roman971, line 12:
```suggestion
typedef enum {
```

src/overlays/actors/ovl_En_Ssh/z_en_ssh.c
```diff
@@ -4,12 +4,54 @@
...
+    ANIM_UP = 1,
+    ANIM_WAIT,
+    ANIM_LAND,
+    ANIM_DROP
+} EnSshAnimation;
```
Roman971, line 17: I believe there are missing values in this enum since the arrays in `EnSsh_SetAnimation` have 7 entries.
Also I suggest having a prefix specific to the actor because defines starting with `ANIM_` are very likely to conflict with the rest of the codebase in the future.

src/overlays/actors/ovl_En_Ssh/z_en_ssh.c
```diff
@@ -4,12 +4,54 @@
 
 #define THIS ((EnSsh*)thisx)
 
+#define STATE_STUNNED (1 << 0)
```
Roman971, line 7: For a similar reason, I suggest using a more specific prefix here because `STATE_` might conflict with the rest of the codebase in the future.

src/overlays/actors/ovl_En_Ssh/z_en_ssh.c
```diff
@@ -21,95 +63,798 @@ const ActorInit En_Ssh_InitVars = {
...
+}
+
+void EnSsh_SpawnShockwave(EnSsh* this, GlobalContext* globalCtx) {
+    Vec3f zeroVec = { 0.0f, 0.0f, 0.0f };
+    Vec3f pos;
+
+    pos.x = this->actor.posRot.pos.x;
+    pos.y = this->actor.groundY;
+    pos.z = this->actor.posRot.pos.z;
+    EffectSsBlast_SpawnWhiteCustomScale(globalCtx, &pos, &zeroVec, &zeroVec, 0x64, 0xDC, 8);
```
Roman971, line 105:
```suggestion
    EffectSsBlast_SpawnWhiteCustomScale(globalCtx, &pos, &zeroVec, &zeroVec, 100, 220, 8);
```

src/overlays/actors/ovl_En_Ssh/z_en_ssh.c
```diff
@@ -21,95 +63,798 @@ const ActorInit En_Ssh_InitVars = {
...
+    }
+    if (this->swayTimer != 0) {
+        return true;
+    }
+    if (this->animTimer != 0) {
+        return true;
+    }
+
+    if (this->actor.xzDistFromLink > 160.0f) {
+        // player is more than 160 xz units from the skulltulla
```
Roman971, line 344: These comments are unnecessary imo, since they don't add anything and the conditions are already very easy to understand.

src/overlays/actors/ovl_En_Ssh/z_en_ssh.c
```diff
@@ -21,95 +63,798 @@ const ActorInit En_Ssh_InitVars = {
...
+        Matrix_MultVec3f(&swayVecBase, &swayVec);
+        Matrix_Pull();
+        this->actor.shape.rot.z = -(swayAngle * 2);
+        this->actor.posRot.pos.x = swayVec.x;
+        this->actor.posRot.pos.z = swayVec.z;
+    }
+}
+
+void EnSsh_CheckBodyStickHit(EnSsh* this, GlobalContext* globalCtx) {
+    ColliderBody* body = &this->colCylinder[0].body;
```
Roman971, line 412: Is this temp necessary? The other cylinders are accessed directly so I feel like this might be an optimization (a branch head merge). The player temp might also need to be inlined in order for it to work tho, but that's fine.

src/overlays/actors/ovl_En_Ssh/z_en_ssh.c
```diff
@@ -21,95 +63,798 @@ const ActorInit En_Ssh_InitVars = {
...
+    if (this->stateFlags & STATE_SPIN) {
+        if (this->spinTimer == 0) {
+            this->stateFlags &= ~STATE_SPIN;
+            if (this->actor.params != ENSSH_FATHER) {
+                EnSsh_SetColliderScale(this, 0.5f, 1.0f);
+            } else {
+                EnSsh_SetColliderScale(this, 0.75f, 1.0f);
+            }
+        }
+    } else { // !(this->unk_5CA & STATE_SPIN)
```
Roman971, line 758:
```suggestion
    } else {
```

src/overlays/actors/ovl_En_Ssh/z_en_ssh.c
```diff
@@ -21,95 +63,798 @@ const ActorInit En_Ssh_InitVars = {
...
+    s32 pad;
+    EnSsh* this = THIS;
+
+    frameCount = SkelAnime_GetFrameCount(&D_06000304.genericHeader);
+    if (this->actor.params == ENSSH_FATHER) {
+        if (gSaveContext.inventory.gsTokens >= 100) {
+            Actor_Kill(&this->actor);
+            return;
+        }
+    } else if (gSaveContext.inventory.gsTokens >= (this->actor.params * 10)) {
```
Roman971, line 592: The `EnSshParam` enum should probably not be a thing based on this line. I suggest just using numbers instead, where 0 is the only special value (but doesn't need any name imo).

src/overlays/actors/ovl_En_Ssh/z_en_ssh.c
petrie911: Changing it doesn't work.

src/overlays/actors/ovl_En_Ssh/z_en_ssh.c
petrie911: There's a lot of extra junk in this code because it was largely copied from `EnSt`. I assume the extra animation entries are from that.

src/overlays/actors/ovl_En_Ssh/z_en_ssh.c
petrie911: I'm keeping the special value for 0.

src/overlays/actors/ovl_En_Ssh/z_en_ssh.c
Roman971: I think the enum should be completed if possible, even if they aren't used. You might be able to figure out what they are based on animations and the rest of the data, and otherwise I suggest having unknowns in there.

petrie911: So, I tried playing the animations. 0 and 6 are unused animations that EnSt doesn't use either. Hard to describe, but looks like an animation for being knocked backwards. 5 is just a slightly slower version of 4. EnSt has the same animation parameters but doesn't use them either.

Roman971 (approved):

PR #488, opened 2020-11-15
ovl_Bg_Ganon_Otyuka

Sirius902:


NEstelami (approved):

src/overlays/actors/ovl_Bg_Ganon_Otyuka/z_bg_ganon_otyuka.h
```diff
@@ -4,14 +4,34 @@
 #include "ultra64.h"
 #include "global.h"
 
+#define SIDE_EAST (1 << 0)
+#define SIDE_WEST (1 << 1)
+#define SIDE_SOUTH (1 << 2)
+#define SIDE_NORTH (1 << 3)
+#define SIDE_ALL (SIDE_EAST | SIDE_WEST | SIDE_SOUTH | SIDE_NORTH)
+
+typedef enum { FLASH_NONE, FLASH_GROW, FLASH_SHRINK } FlashState;
```
Roman971, line 13: Value comments are useful and ensure proper vertical formatting as a bonus, so I suggest using them here.
```suggestion
typedef enum {
    /* 0x00 */ FLASH_NONE,
    /* 0x01 */ FLASH_GROW,
    /* 0x02 */ FLASH_SHRINK
} FlashState;
```

src/overlays/actors/ovl_Bg_Ganon_Otyuka/z_bg_ganon_otyuka.h
```diff
@@ -4,14 +4,34 @@
 #include "ultra64.h"
 #include "global.h"
 
+#define SIDE_EAST (1 << 0)
+#define SIDE_WEST (1 << 1)
+#define SIDE_SOUTH (1 << 2)
+#define SIDE_NORTH (1 << 3)
+#define SIDE_ALL (SIDE_EAST | SIDE_WEST | SIDE_SOUTH | SIDE_NORTH)
```
Roman971, line 11: Probably a good idea to add a prefix specific to the actor for those, to avoid any conflict with the rest of the codebase, especially since they are exposed in the header.

Also if they are only meant for use inside the actor itself, they shouldn't be exposed in the header. Same goes for the FlashState enum.

src/overlays/actors/ovl_Bg_Ganon_Otyuka/z_bg_ganon_otyuka_gfx.c
```diff
@@ -0,0 +1,209 @@
...
+    0x62496247, 0x6A496249, 0x7B0B5209, 0x52874A49, 0x42495209, 0x19434207, 0x52876207, 0x52495A49, 0x62095A09,
+    0x5A095289, 0x62475207, 0x4A874A89, 0x52495287, 0x52495205, 0x42075A05, 0x62476247, 0x5A095249, 0x52496249,
+    0x62495249, 0x5A874A07, 0x49C74207, 0x19453945, 0x41C739C7, 0x420741C9, 0x41C74209, 0x42054187, 0x3A0541C5,
+    0x3A0931C7, 0x39C541C7, 0x49C73147, 0x3A0541C5, 0x49C549C7, 0x398941C7, 0x420749C7, 0x49C739C7, 0x420731C7,
+    0x320741C7, 0x08C51903, 0x19431143, 0x21451903, 0x21432105, 0x19432105, 0x19051905, 0x11031905, 0x21451907,
+    0x11072105, 0x19031943, 0x190518C5, 0x214518C5, 0x21432145, 0x11032103, 0x19051905, 0x19031903,
+};
+
+// probably unused?
+static UNK_TYPE D_808772E0[] = { 0x00000000, 0x00000000 };
```
Roman971, line 66: Is this needed? I'm thinking it might just be padding.

src/overlays/actors/ovl_Bg_Ganon_Otyuka/z_bg_ganon_otyuka.c
```diff
@@ -31,18 +32,360 @@ const ActorInit Bg_Ganon_Otyuka_InitVars = {
...
+    Actor* actor;
+    BgGanonOtyuka* platform;
+    BossGanon* ganondorf;
+    f32 spBC = -30.0f;
+
+    OPEN_DISPS(globalCtx->state.gfxCtx, "../z_bg_ganon_otyuka.c", 702);
+
+    actor = globalCtx->actorCtx.actorList[ACTORTYPE_BOSS].first;
+    while (actor != NULL) {
+        if ((actor->id == ACTOR_BOSS_GANON)) {
```
Roman971, line 286:
```suggestion
        if (actor->id == ACTOR_BOSS_GANON) {
```

src/overlays/actors/ovl_Bg_Ganon_Otyuka/z_bg_ganon_otyuka.c
```diff
@@ -31,18 +32,360 @@ const ActorInit Bg_Ganon_Otyuka_InitVars = {
...
+
+void BgGanonOtyuka_DoNothing(Actor* thisx, GlobalContext* globalCtx) {
+}
+
+void BgGanonOtyuka_Update(Actor* thisx, GlobalContext* globalCtx) {
+    BgGanonOtyuka* this = THIS;
+
+    this->actionFunc(this, globalCtx);
+    this->flashTimer++;
+    DECR(this->dropTimer);
```
Roman971, line 268:
```suggestion
    if (this->dropTimer != 0) {
        this->dropTimer--;
    }
```

src/overlays/actors/ovl_Bg_Ganon_Otyuka/z_bg_ganon_otyuka.c
```diff
@@ -31,18 +32,360 @@ const ActorInit Bg_Ganon_Otyuka_InitVars = {
...
+
+static UNK_TYPE sSurfaceTypeList[] = {
+    0x00000000,
+    0x000007C0,
+    0x00000000,
+    0x000007C2,
+};
+
+static CollisionPoly sPolyList[] = {
+    { { 0x00, 0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x02 }, { 32767, 0, 0 }, -60 },
```
Roman971, line 74: Writing data to a filler (`char unk_00[0x8]`) is a bit awkward, but I guess we don't know what type these are just yet? (Mostly a comment for other people looking at this PR)

src/overlays/actors/ovl_Bg_Ganon_Otyuka/z_bg_ganon_otyuka_gfx.c
Sirius902: It is needed, but it works either as an array of one zero or two zeros. Removing it completely unfortunately doesn't match.

src/overlays/actors/ovl_Bg_Ganon_Otyuka/z_bg_ganon_otyuka.c
Sirius902: I'm pretty sure this data structure has a proper layout in bgcheck, but that hasn't been merged yet.

src/overlays/actors/ovl_Bg_Ganon_Otyuka/z_bg_ganon_otyuka.h
Sirius902: I know that `SIDE_ALL` is used in Ganondorf when he makes platforms fall though I'm not sure whether or not flash state is modified outside the actor. I'd assume not, so I could move flash state to the c file if that'd be better.

src/overlays/actors/ovl_Bg_Ganon_Otyuka/z_bg_ganon_otyuka.c
Sirius902: Question, how come here we wouldn't use `DECR` here? Is it more for decrementing timers in if statements?

Roman971: Yeah it's intended for use inline in conditions. We avoid using it otherwise because some decrements towards 0 are done with `> 0` checks and such, so we can't use it consistently.

src/overlays/actors/ovl_Bg_Ganon_Otyuka/z_bg_ganon_otyuka_gfx.c
Roman971: By any chance, does it match if the texture above is an array of `u64`?

src/overlays/actors/ovl_Bg_Ganon_Otyuka/z_bg_ganon_otyuka.h
Roman971: Alright, that seems good to me.

src/overlays/actors/ovl_Bg_Ganon_Otyuka/z_bg_ganon_otyuka.c
fig02: id bring in the relevant struct from the bg check pr in this case. if nothing else is using it it shouldnt cause any conflicts

Roman971: I believe a lot of things were changed in the struct (including the name?), so that's probably not a good idea.

src/overlays/actors/ovl_Bg_Ganon_Otyuka/z_bg_ganon_otyuka_gfx.c
Sirius902: Nope, doesn't seem to change a thing.

Roman971 (approved):

fig02 (approved):

PR #493, opened 2020-11-20
En_Attack_Niw, En_Insect, z_skin_matrix, and z_skelanime matched

petrie911:
last functions matched for each file. Also discovered while doing this that a lot of actors don't have `static`, so I cleaned up what I could find.

EDIT: also a function in z_skin_matrix


fig02 (approved):

src/code/z_skelanime.c
```diff
@@ -752,6 +750,8 @@ s32 func_800A29BC(s32 arg0, s32 arg1, Vec3s* arg2) {
     arg2++;
     temp_a3 += 6;
 
+    if(t & 1){}
```
Roman971, line 753: Make sure to run the formatter again

Roman971 (approved):

PR #497, opened 2020-11-22
SkelAnime "Skeleton" Documentation Pass

petrie911:
Updated SkelAnime with the new skeleton structs and cleaned up its type system. No more dereferencing segmented pointers to get genericHeader.


include/z64animation.h
```diff
@@ -5,57 +5,120 @@
...
 typedef struct {
-    /* 0x000 */ Vec3s translation;      // Translation relative to parent limb.  root limb is a tranlation for entire model.
-    /* 0x006 */ u8 firstChildIndex;     // The first child's index into the limb table.
-    /* 0x007 */ u8 nextLimbIndex;       // The parent limb's next limb index into the limb table.
-    /* 0x008 */ Gfx* displayLists[1];   // Display lists for the limb. Index 0 is the normal display list, index 1 is the far model display list.
-} SkelLimbEntry; // Size = 0xC or 0x10
+    /* 0x00 */ Vec3s translation;      // Translation relative to parent limb.  root limb is a tranlation for entire model.
+    /* 0x06 */ u8 firstChildIndex;     // The first child's index into the limb table.
+    /* 0x07 */ u8 nextLimbIndex;       // The parent limb's next limb index into the limb table.    
+} SkelLimb; // Size = 0x8
```
fig02, line 24: Petrie and I discussed this in discord but want to put it here and get other opinions:
We were discussing the possibility of calling this `Bone` and switching the terminology from limb to bone in general.

This would mean that when something takes a "skeleton" its really just a pointer to the root bone which connects to everything else.
The term bone is apparently the more common terminology for this in video games, and I think it sounds good for the most part. I do think that saying for example "drawing a limb" makes more sense than "drawing a bone", since the display list for a given limb/bone draws the whole limb of an actor, and thinking about that aspect as a bone sounds a bit weird.
But, I really do like the continuity of the whole thing being called the skeleton and the parts being called a bone

relevant discussion is here: https://discordapp.com/channels/688807550715560050/688851317593997489/780217593407471636

fig02 (approved):
Assuming we have no more plans for changes in this pass, Im ok with all of this.
This will break alot of open PRs but I will fix them myself if need be

src/code/z_scene_table.c
```diff
@@ -2348,9 +2348,8 @@ void func_8009FE58(GlobalContext* globalCtx) {
         temp = 0.020000001f;
         func_800AA76C(&globalCtx->view, Math_Coss(D_8012A39C) * (0.00009587531f * temp),
                       Math_Sins(D_8012A39C) * (0.00009587531f * temp), Math_Sins(D_8012A3A0) * (0.00009587531f * temp));
-        func_800AA78C(&globalCtx->view, 1.0f + (Math_Sins(D_8012A3A0) * (0.79999995f * temp)),
-                      1.0f + (Math_Coss(D_8012A3A0) * (0.39999998f * temp)),
-                      1.0f + (Math_Coss(D_8012A39C) * (1 * temp)));
+        func_800AA78C(&globalCtx->view, 1 + (Math_Sins(D_8012A3A0) * (0.79999995f * temp)),
+                      1 + (Math_Coss(D_8012A3A0) * (0.39999998f * temp)), 1 + (Math_Coss(D_8012A39C) * (1 * temp)));
```
Roman971, line 2352: This is a non matching and these changes don't seem to have anything to do with skelanime. Are they intended?
And in any case it would probably be better to keep them separate.

src/code/z_scene_table.c
```diff
@@ -2377,7 +2376,7 @@ void func_8009FE58(GlobalContext* globalCtx) {
     }
 
     if (globalCtx->roomCtx.curRoom.num == 2) {
-        Matrix_Scale(1, sinf(D_8012A398) * 0.8f, 1, MTXMODE_NEW);
+        Matrix_Scale(1.0f, sinf(D_8012A398) * 0.8f, 1.0f, MTXMODE_NEW);
```
Roman971, line 2379: Same here

include/functions.h
```diff
@@ -1300,64 +1300,69 @@ s32 Scene_ExecuteCommands(GlobalContext* globalCtx, SceneCmd* sceneCmd);
...
-Gfx* SkelAnime_Draw2(GlobalContext* globalCtx, Skeleton* skeleton, Vec3s* limbDrawTable,
-                     OverrideLimbDraw2 overrideLimbDraw, PostLimbDraw2 postLimbDraw, Actor* actor, Gfx* gfx);
-Gfx* SkelAnime_DrawSV2(GlobalContext* globalCtx, Skeleton* skeleton, Vec3s* limbDrawTable, s32 dListCount,
-                       OverrideLimbDraw2 overrideLimbDraw, PostLimbDraw2 postLimbDraw, Actor* actor, Gfx* gfx);
+void SkelAnime_DrawLod(GlobalContext* globalCtx, void** skeleton, Vec3s* limbDrawTable,
+                       OverrideLimbDrawOpa overrideLimbDraw, PostLimbDrawOpa postLimbDraw, void* data, s32 dListIndex);
+void SkelAnime_DrawFlexLod(GlobalContext* globalCtx, void** skeleton, Vec3s* limbDrawTable, s32 dListCount,
+                         OverrideLimbDrawOpa overrideLimbDraw, PostLimbDrawOpa postLimbDraw, void* data, s32 dListIndex);
+void SkelAnime_DrawOpa(GlobalContext* globalCtx, void** skeleton, Vec3s* limbDrawTable,
+                    OverrideLimbDrawOpa overrideLimbDraw, PostLimbDrawOpa postLimbDraw, void* data);
```
Roman971, line 1308: `data` seems a bit unclear since this is often a pointer to an actor (among other things). We usually call these generic arguments `arg`, any thoughts on using that here?

petrie911: arg seems like something of unknown purpose, but in this case the purpose is known--it's some form of data the actor needs to perform OverrideLimbDraw and PostLimbDraw. This is usually the actor itself, but some, like Boss_Sst, pass other things.

src/code/z_scene_table.c
petrie911: Ah, I was working on matching that function and must have accidentally done it on this branch. I'll revert.

include/functions.h
fig02: I didnt know we used arg for generic stuff like that personally. I agree that it does make it sound like its unk or something.
Ill think over other alternatives in the mean time.

What does it get used for in bongo petrie?

petrie911: the afterimages of his hands are drawn by skelanime but pass it their own struct instead of Bongo's

Roman971: I don't agree with an actor instance being "data", for example. And I feel like leaving it as "unknown" is fine since that's essentially what it is, it could literally be anything as far as the skelanime system is concerned. (To be clear, I'm only talking about how the skelanime system calls it, not how it would be called in override/postlimbdraw implementations)

fig02: yeah i guess it is literally an unknown arg lol, just not from a decomp sense. im fine with it.
and yeah the implementations will have more appropriate names

Roman971 (approved):

PR #500, opened 2020-11-25
Add remaining undefined syms

petrie911:
Adds the vast majority of remaining undefined syms to actor files as externs and their corresponding values to undefined_syms.txt. I can't guarantee I got everything (e.g., I know Boss_Sst is missing one), but the misses should be rare.

All current PRs have been excluded, so there shouldn't be any conflicts in the .c files.


src/overlays/actors/ovl_Bg_Gnd_Iceblock/z_bg_gnd_iceblock.c
```diff
@@ -20,6 +20,9 @@ void func_8087A184(BgGndIceblock* this, GlobalContext* globalCtx);
 void func_8087A248(BgGndIceblock* this, GlobalContext* globalCtx);
 void func_8087A300(BgGndIceblock* this, GlobalContext* globalCtx);
 void func_8087A39C(BgGndIceblock* this, GlobalContext* globalCtx);
+extern UNK_TYPE D_06004420;
```
Roman971, line 23:
```suggestion
void func_8087A39C(BgGndIceblock* this, GlobalContext* globalCtx);

extern UNK_TYPE D_06004420;
```

Roman971 (approved):
Seems good to me (other than this very minor thing I noticed while skimming through).

Tho I'm still not sure why we separate undefined symbols based on the files that reference them tbh. It's a trend that was kind of randomly started a long time ago and everyone followed it, but I don't think it serves any purpose.
If anything, I'd say they should be grouped by which segments/files they belong to, not by what references them. But we can also not separate them at all since undefined symbols are temporary anyway.

fig02 (approved):

PR #502, opened 2020-11-27
Fix 'make' failing due to missing 'md5sum' command

mnzlmstr:
MacOS's built in md5 command has a different name. installing md5sum

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ For macOS, use homebrew to install the following dependencies:
 * python3
 ```bash
 brew update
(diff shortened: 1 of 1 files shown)


Roman971 (approved):

Roman971:
Looks good to me but I think @ethteck should confirm it since he's the one who setup MacOS originally

ethteck:
Thanks, yeah LGTM. I think this won't let the MacOS build fully pass still since this md5sum acts differently, but at least it can run.

PR #506, opened 2020-12-01
z_bg_bombwall OK

mnzlmstr:


src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Bombwall/func_8086EE40.s")
-
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Bombwall/func_8086EE94.s")
-
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Bombwall/BgBombwall_Update.s")
+static InitChainEntry sInitChain[] = {
+    ICHAIN_F32(uncullZoneForward, 1800, ICHAIN_CONTINUE), // unk f4
+    ICHAIN_F32(uncullZoneScale, 300, ICHAIN_CONTINUE),    // unk f8
+    ICHAIN_F32(uncullZoneDownward, 1000, ICHAIN_STOP),    // unk fc
+};
```
fig02, line 70: you can remove the comments

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Bg_Bombwall/BgBombwall_Draw.s")
+Vec3s D_8086F010[] = { { 0x0028, 0x0055, 0x0015 }, { 0xFFD5, 0x006B, 0x000E }, { 0xFFFF, 0x008E, 0x000E },
+                       { 0xFFE5, 0x002C, 0x001B }, { 0x001C, 0x0018, 0x0014 }, { 0xFFD9, 0x0036, 0x0015 },
+                       { 0x0031, 0x0032, 0x0014 } };
+
```
fig02, line 74: you can turn these into decimal. note that the larger numbers are actually neagtive.
also, the formatter will not indent awkwardly like that if you put a trailing comma on the last entry in the list

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+    colHeader = NULL;
+    DynaPolyInfo_SetActorMove(&this->dyna, 0);
+    DynaPolyInfo_Alloc(&D_050041B0, &colHeader);
+    this->dyna.dynaPolyId =
+        DynaPolyInfo_RegisterActor(globalCtx, &globalCtx->colCtx.dyna, &this->dyna.actor, colHeader);
+
+    if (this->dyna.dynaPolyId == 0x32) {
+        // Warning : move BG login failed
+        osSyncPrintf((const char*)"Warning : move BG 登録失敗(%s %d)(arg_data 0x%04x)\n", "../z_bg_bombwall.c", 0xF3,
+                     this->dyna.actor.params);
```
fig02, line 89: you can remove (const char*)
also the 0xF3 is a line number so it can be decimal

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+
+void BgBombwall_Init(Actor* thisx, GlobalContext* globalCtx) {
+
+    s32 i;
+    s32 j;
+    Vec3f vecs[3];
+    Vec3f sp80;
+
+    s32 pad;
+    BgBombwall* this = THIS;
```
fig02, line 107: you can remove all of the line breaks here and have it as one long list that sits at the top

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+
+    f32 sin = Math_Sins(this->dyna.actor.shape.rot.y);
+    f32 cos = Math_Coss(this->dyna.actor.shape.rot.y);
+
+    Actor_ProcessInitChain(&this->dyna.actor, sInitChain);
+    Actor_SetScale(thisx, 0.1f);
+
+    if (Flags_GetSwitch(globalCtx, this->dyna.actor.params & 0x3F) != 0) {
+        func_8086EE94(this, globalCtx);
```
fig02, line 116: flags can be treated as booleans, so you cont need != 0

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+    } else {
+
+        BgBombwall_InitDynapoly(this, globalCtx);
+        this->unk_2A2 |= 2;
+        Collider_InitTris(globalCtx, &this->collider);
+        Collider_SetTris(globalCtx, &this->collider, this, &sTrisInit, &(this->colliderItems[0]));
```
fig02, line 122: you want to pass the address of the array, not just the first item. while the mean the same thing, the types will line up
so just `this->colliderItems` is fine, you dont need & for functions or arrays

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+
+                vecs[j].x += this->dyna.actor.posRot.pos.x;
+                vecs[j].y += this->dyna.actor.posRot.pos.y;
+                vecs[j].z += this->dyna.actor.posRot.pos.z;
+            }
+            func_800627A0(&this->collider, i, &vecs[0], &vecs[1], &vecs[2]);
+        }
+        this->unk_2A2 |= 1;
+
+        BgBombwall_setFuncPointers(this, globalCtx);
```
fig02, line 142: the s in Set needs to be uppercase
Id probably leave this unnamed until we know what both function pointers are honestly, leaving it as `func_` is preferred over a generic name like set function pointers. it will get a proper name later

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+    }
+    osSyncPrintf("(field keep 汎用爆弾壁)(arg_data 0x%04x)(angY %d)\n", this->dyna.actor.params,
+                 this->dyna.actor.shape.rot.y);
+}
+
+void BgBombwall_DestroyInternal(BgBombwall* this, GlobalContext* globalCtx) {
+    if (this->unk_2A2 & 2) {
+        DynaPolyInfo_Free(globalCtx, &globalCtx->colCtx.dyna, this->dyna.dynaPolyId);
+        this->unk_2A2 &= 0xfffd;
```
fig02, line 151: you can throw these very large flags into a calculator to get the `~` value for them
this is `~2` and below is `~1`

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+    }
+    if (this->unk_2A2 & 1) {
+        Collider_DestroyTris(globalCtx, &this->collider);
+        this->unk_2A2 &= 0xfffe;
+    }
+}
+
+void BgBombwall_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+    BgBombwall_DestroyInternal((BgBombwall*)thisx, globalCtx);
```
fig02, line 160: you should be able to make a temp for this using the THIS macro and pass that instead of casting

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+void func_8086EB5C(BgBombwall* this, GlobalContext* globalCtx) {
+    s16 rand;
+    s16 rand2;
+    Vec3f sp88;
+    s32 i;
+    f32 sin = Math_Sins(this->dyna.actor.shape.rot.y);
+    f32 cos = Math_Coss(this->dyna.actor.shape.rot.y);
+    Vec3f* pos = &this->dyna.actor.posRot.pos;
+    f32 temp;
+    float new_var;
```
fig02, line 172: f32 instead of float

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+
+    sp88.x = pos->x;
+    new_var = pos->y + 90.0f;
+    sp88.y = pos->y + 90.0f;
+    sp88.z = pos->z + 15.0f;
+    func_80033480(globalCtx, &sp88, 40.0f, 4, 0xA, 0x32, 1);
+}
+
+void BgBombwall_setFuncPointers(BgBombwall* this, GlobalContext* globalCtx) {
+    this->funcPtr = &D_05003FC0;
```
fig02, line 192: adding on to my previous comment, the first one here is not a function pointer. it is a display list. it gets passed to `Gfx_DrawDListOpa` to actually draw the model for this actor

you can name this field `dList` and type the externs for the D_05 symbols as Gfx[];

If any of this is confusing feel free to ask me questions

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+    this->actionFunc = BgBombwall_actionFunc;
+}
```
fig02, line 194: similarly i would leave this unnamed instead of giving it a generic name. we'll want to name this after the action when we know what it does

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+
+void BgBombwall_actionFunc(BgBombwall* this, GlobalContext* globalCtx) {
+    u8 temp_v0;
+
+    temp_v0 = this->collider.base.acFlags;
```
fig02, line 199: this temp here is probably not needed. mips2c likes to make alot of temps and you should always try to remove them when possible

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+    if ((temp_v0 & 2) != 0) {
+        this->collider.base.acFlags = temp_v0 & 0xFFFD;
```
fig02, line 201: since this is a flag check you can remove the != 0
additionally after you remove the temp you should be able to write this as 
`this->collider.base.acFlags &= ~2;`

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+        this = this;
+        func_8086EDFC(this, globalCtx);
+        Flags_SetSwitch(globalCtx, this->dyna.actor.params & 0x3F);
+        return;
+    }
+    if (this->dyna.actor.xzDistFromLink < 600.0f) {
```
fig02, line 207: instead of a return here you can tunrn this into an else

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+        CollisionCheck_SetAC(globalCtx, &globalCtx->colChkCtx, (Collider*)&this->collider);
+    }
```
fig02, line 209: instead of casting to Collider you can pass `.base`

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+}
+
+void func_8086EDFC(BgBombwall* this, GlobalContext* globalCtx) {
+    this->funcPtr = &D_05003FC0;
+    this->unk_2A0 = (u16)1;
```
fig02, line 214: you probably dont need this u16 cast here

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+    func_8086EB5C(this, globalCtx);
+    this->actionFunc = (void (*)(struct BgBombwall*, GlobalContext*)) & func_8086EE40;
+}
```
fig02, line 217: the function prototype is something mips2c added, you can remove it. along with the &symbol

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+    temp_v0 = this->unk_2A0;
+    if ((s32)temp_v0 > 0) {
+        this->unk_2A0 = (s16)(temp_v0 - 1);
+        return;
+    }
+    func_8086EE94(this, globalCtx);
+    if ((((s32)this->dyna.actor.params >> 0xF) & 1) != 0) {
+        func_80078884((u16)0x4802U);
+    }
+}
```
fig02, line 231: a few things here
-this temp is probably not real
-you can do -- on the variable directly
-the early return is more likely an else
-the cast on params is probably not needed
-the arg for that last call is a sound effect id, which you can find in sfx.h

also, you can name 2A0 `timer` if you want
```suggestion
void func_8086EE40(BgBombwall* this, GlobalContext* globalCtx) {
    if (this->unk_2A0 > 0) {
        this->unk_2A0--;
    } else {
        func_8086EE94(this, globalCtx);
        if (((this->dyna.actor.params >> 0xF) & 1) != 0) {
            func_80078884(NA_SE_SY_CORRECT_CHIME);
        }
    }
}
```

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+
+void func_8086EE94(BgBombwall* this, GlobalContext* globalCtx) {
+    this->funcPtr = &D_05004088;
+    BgBombwall_DestroyInternal(this, globalCtx);
+    this->actionFunc = NULL;
+}
+
+void BgBombwall_Update(Actor* thisx, GlobalContext* globalCtx) {
+
```
fig02, line 240: id remove the line break at the top here

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+    BgBombwall* this = THIS;
+
+    if (this->actionFunc != 0) {
+        this->actionFunc(this, globalCtx);
+    }
```
fig02, line 245: for function pointers youll want to check against NULL

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -36,29 +61,192 @@ const ActorInit Bg_Bombwall_InitVars = {
...
+}
+
+void BgBombwall_Draw(Actor* thisx, GlobalContext* globalCtx) {
+
+    BgBombwall* this = THIS;
```
fig02, line 250: id remove the line break at the top, and instead put the line break after the variable declarations at the top.
ans as mentioned before, the variable should be names `this->dList`, its not a function pointer

src/overlays/actors/ovl_Bg_Bombwall/z_bg_bombwall.c
```diff
@@ -15,16 +15,41 @@ void BgBombwall_Destroy(Actor* thisx, GlobalContext* globalCtx);
 void BgBombwall_Update(Actor* thisx, GlobalContext* globalCtx);
 void BgBombwall_Draw(Actor* thisx, GlobalContext* globalCtx);
 
-void func_8086E7D0(BgBombwall* this, GlobalContext* globalCtx);
-void func_8086EAC0(BgBombwall* this, GlobalContext* globalCtx);
+// void BgBombwall_InitDynapoly(BgBombwall* this, GlobalContext* globalCtx);
+void BgBombwall_DestroyInternal(BgBombwall* this, GlobalContext* globalCtx);
```
fig02, line 19: you can remove this one thats commented out. and if anymore of these are not required then you can remove them as well. we only forward declare when necessary

fig02 (commented):
nice job on your first actor. there are quite a few suggestions here, please ask me questions if anything doesnt make sense

the biggest thing here is that there are alot more steps that can be taken to clean up mips2c's output. always be wary of the temps that it creates and the casts that it does.

fig02 (approved):
review changes implemented

Roman971 (approved):

PR #507, opened 2020-12-02
message_data_static files OK

Thar0:
Much of the design revolves around keeping all the information relevant to the same message in the same place, such as the text id and the contents of the message itself in all 3 localisations.

Do the encoded text headers need to be explicitly cleaned? I don't believe so since they are re-encoded for each build but we may not want that in the first place, and only re-encode them if the encoded file is not present?

Some not-so-great things:
 - Enum values don't work for text control codes such as `ITEM_ICON`, as you can only concatenate string literals with other string literals. This seems like a hard limitation on what can be done with the preprocessor and the fact that rendering these into C isn't ideal to begin with, but I'd love to hear what others think.
 - `message_data_fmt.h` is preprocessor hell to avoid duplicating the magic values for the control codes. Macros exist to create both proper hex values and string literals out of them, once again motivated by only being able to concat strings with other strings, but there might be a simpler way to achieve this?
 - The message entry tables in z_message_PAL aren't being put in rodata currently, so they are left commented out in `z_message_PAL.c`. As a result, the message_data_static files are not shiftable but will be as soon as those tables are included.


fig02:
So we've concluded in discord that we will be extracting.
Have we agreed on waiting to actually include text in the repo? If so we should be able to close this until we revisit

fig02:
Closing this as we have agreed in discord that text will be extracted, and it will be done at a later time

PR #512, opened 2020-12-06
En_Stream (vortex) 7/8 matching, 1 equivalent

tom-overton:
This MR matches 7 of the 8 functions in En_Stream, with the remaining function being equivalent (the only difference is regalloc). I tried really hard to match func_80B0B81C, but I spent a month on-and-off on it and couldn't ever get the registers to match up.


src/overlays/actors/ovl_En_Stream/z_en_stream.c
```diff
@@ -29,19 +31,117 @@ const ActorInit En_Stream_InitVars = {
...
+    f32 yDistWithOffset;
+    s32 pad30;
+    s32 pad2C;
+
+    if (func_80B0B81C(&this->actor.posRot, &player->actor.posRot, &posRotDifference, this->actor.scale.y) != 0) {
+        xzDist = sqrtf(SQ(posRotDifference.x) + SQ(posRotDifference.z));
+        yDistWithOffset = player->actor.posRot.pos.y - (this->actor.posRot.pos.y - 90.0f);
+        player->windDirection = Math_atan2f(-posRotDifference.x, -posRotDifference.z) * 10430.378f;
+        if (3.0f < xzDist) {
+            Math_SmoothScaleMaxMinF(&player->windSpeed, 3.0f, 0.5f, xzDist, 0.0f);
```
fig02, line 99: you may be able to flip both of these comparisons to `var > value` which is a bit easier to read

fig02 (commented):
nice, only one small thing

src/overlays/actors/ovl_En_Stream/z_en_stream.c
tom-overton: Yep, that works. Thanks for the suggestion, it definitely is easier to read

fig02 (approved):

spec
```diff
@@ -3022,7 +3022,6 @@ endseg
 beginseg
     name "ovl_En_Stream"
     include "build/src/overlays/actors/ovl_En_Stream/z_en_stream.o"
-    include "build/data/overlays/actors/z_en_stream.data.o"
     include "build/data/overlays/actors/z_en_stream.reloc.o"
```
Roman971, line 3025: The generated reloc should be used in `NON_MATCHING` since this overlay has a non matching. You can find examples of that in this file if needed.

src/overlays/actors/ovl_En_Stream/z_en_stream.c
```diff
@@ -29,19 +31,117 @@ const ActorInit En_Stream_InitVars = {
...
+
+void EnStream_Init(Actor* thisx, GlobalContext* globalCtx) {
+    EnStream* this = THIS;
+
+    this->unk_150 = thisx->params & 0xFF;
+    Actor_ProcessInitChain(thisx, sInitChain);
+    if ((this->unk_150 != 0) && (this->unk_150 == 1)) {
+        thisx->scale.y = 0.01f;
+    }
+    EnStream_SetupAction(this, &EnStream_WaitForPlayer);
```
Roman971, line 51: We omit the `&` for function and array pointers
```suggestion
    EnStream_SetupAction(this, EnStream_WaitForPlayer);
```

src/overlays/actors/ovl_En_Stream/z_en_stream.c
```diff
@@ -29,19 +31,117 @@ const ActorInit En_Stream_InitVars = {
...
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Stream/EnStream_Destroy.s")
+void EnStream_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+}
 
+#ifdef NONMATCHING
```
Roman971, line 57: We use `NON_MATCHING` for this. It's important so building with `NON_MATCHING` defined catches the function.
```suggestion
#ifdef NON_MATCHING
```

src/overlays/actors/ovl_En_Stream/z_en_stream.c
```diff
@@ -29,19 +31,117 @@ const ActorInit En_Stream_InitVars = {
...
+            Math_SmoothScaleMaxMinF(&player->actor.posRot.pos.z, this->actor.posRot.pos.z, 0.5f, 3.0f, 0.0f);
+        }
+        if (yDistWithOffset > 0.0f) {
+            Math_SmoothScaleMaxMinF(&player->actor.velocity.y, -3.0f, 0.7f, yDistWithOffset, 0.0f);
+            if (posRotDifference.y < -70.0f) {
+                player->stateFlags2 |= 0x80000000;
+            }
+        }
+    } else {
+        EnStream_SetupAction(this, &EnStream_WaitForPlayer);
```
Roman971, line 112:
```suggestion
        EnStream_SetupAction(this, EnStream_WaitForPlayer);
```

src/overlays/actors/ovl_En_Stream/z_en_stream.c
```diff
@@ -29,19 +31,117 @@ const ActorInit En_Stream_InitVars = {
...
+    }
+}
+
+void EnStream_WaitForPlayer(EnStream* this, GlobalContext* globalCtx) {
+    Player* player = PLAYER;
+    s16 pad;
+    Vec3f temp;
+
+    if (func_80B0B81C(&this->actor.posRot, &player->actor.posRot, &temp, this->actor.scale.y) != 0) {
+        EnStream_SetupAction(this, &EnStream_SuckPlayer);
```
Roman971, line 122:
```suggestion
        EnStream_SetupAction(this, EnStream_SuckPlayer);
```

src/overlays/actors/ovl_En_Stream/z_en_stream.c
```diff
@@ -29,19 +31,117 @@ const ActorInit En_Stream_InitVars = {
...
+    EnStream_SetupAction(this, &EnStream_WaitForPlayer);
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Stream/EnStream_Destroy.s")
+void EnStream_Destroy(Actor* thisx, GlobalContext* globalCtx) {
+}
 
+#ifdef NONMATCHING
+// regalloc differences, checks if the player is range of the vortex
+s32 func_80B0B81C(PosRot* vortexPosRot, PosRot* playerPosRot, Vec3f* posRotDifference, f32 vortexYScale) {
```
Roman971, line 59: This function probably takes pointers to `Vec3f`, not `PosRot`

src/overlays/actors/ovl_En_Stream/z_en_stream.c
```diff
@@ -14,10 +14,12 @@ void EnStream_Init(Actor* thisx, GlobalContext* globalCtx);
 void EnStream_Destroy(Actor* thisx, GlobalContext* globalCtx);
 void EnStream_Update(Actor* thisx, GlobalContext* globalCtx);
 void EnStream_Draw(Actor* thisx, GlobalContext* globalCtx);
+void EnStream_WaitForPlayer(EnStream* this, GlobalContext* globalCtx);
+void EnStream_SuckPlayer(EnStream* this, GlobalContext* globalCtx);
+s32 func_80B0B81C(PosRot* vortexPosRot, PosRot* playerPosRot, Vec3f* v, f32 vortexYScale);
 
 extern UNK_TYPE D_06000950;
```
Roman971, line 21: Based on the usage of this variable in the draw function, we can update the type to a display list (ie. a `Gfx[]`)
```suggestion
extern Gfx D_06000950[];
```

src/overlays/actors/ovl_En_Stream/z_en_stream.c
```diff
@@ -29,19 +31,117 @@ const ActorInit En_Stream_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Stream/EnStream_Draw.s")
+    OPEN_DISPS(globalCtx->state.gfxCtx, "../z_en_stream.c", 295);
+    func_80093D84(globalCtx->state.gfxCtx);
+    gSPMatrix(POLY_XLU_DISP++, Matrix_NewMtx(globalCtx->state.gfxCtx, "../z_en_stream.c", 299),
+              G_MTX_NOPUSH | G_MTX_LOAD | G_MTX_MODELVIEW);
+    multipliedFrames = frames * 20;
+    gSPSegment(POLY_XLU_DISP++, 0x08,
+               Gfx_TwoTexScroll(globalCtx->state.gfxCtx, 0, frames * 30, -multipliedFrames, 0x40, 0x40, 1,
+                                multipliedFrames, -multipliedFrames, 0x40, 0x40));
+    gSPDisplayList(POLY_XLU_DISP++, &D_06000950);
```
Roman971, line 145: After treating D_06000950 as an array (see the previous comment), the `&` should be removed

src/overlays/actors/ovl_En_Stream/z_en_stream.c
```diff
@@ -14,10 +14,12 @@ void EnStream_Init(Actor* thisx, GlobalContext* globalCtx);
 void EnStream_Destroy(Actor* thisx, GlobalContext* globalCtx);
 void EnStream_Update(Actor* thisx, GlobalContext* globalCtx);
 void EnStream_Draw(Actor* thisx, GlobalContext* globalCtx);
+void EnStream_WaitForPlayer(EnStream* this, GlobalContext* globalCtx);
+void EnStream_SuckPlayer(EnStream* this, GlobalContext* globalCtx);
+s32 func_80B0B81C(PosRot* vortexPosRot, PosRot* playerPosRot, Vec3f* v, f32 vortexYScale);
```
Roman971, line 19: We generally don't forward declare functions unless necessary, so you can omit these ones since they aren't used before their declaration.
```suggestion
```
```suggestion
```

src/overlays/actors/ovl_En_Stream/z_en_stream.h
```diff
@@ -6,9 +6,13 @@
 
 struct EnStream;
 
+typedef void (*EnStreamActionFunc)(struct EnStream*, struct GlobalContext*);
```
Roman971, line 9:
```suggestion
typedef void (*EnStreamActionFunc)(struct EnStream*, GlobalContext*);
```

Roman971 (commented):
Nicely done. Just a few things

src/overlays/actors/ovl_En_Stream/z_en_stream.c
tom-overton: I can definitely remove the EnStream_SuckPlayer forward declaration, but removing the func_80B0B81C declaration causes functions that call it to no longer match.

Roman971: Oh it's a non matching, then I would leave the prototype in the #else next to the GLOBAL_ASM pragma since it's only relevant while it's a non matching

Roman971 (approved):

PR #514, opened 2020-12-06
New tool: sfxconvert

petrie911:
New tool to convert sfx ids so you don't have to. Run it on your file and it will replace all arguments of functions listed in functions.h as taking SFX ids with the appropriate macro. You can also have it write to a different file instead of your current one if for some reason you want that.

Also fixed all files that had unconverted SFX ids while I was at it.


fig02 (approved):

tools/sfxconvert.py
```diff
@@ -0,0 +1,164 @@
...
+        return -3
+    index = data.find(func)
+    argnum = AudioFunctions.get(func,-1)
+    if(argnum == -1 or index == -1):
+        print('Function lookup error at line', i, 'in', func)
+        return -1
+    args = data[index:].replace('(',',').replace(')',',').split(',')
+    sfxId = args[argnum + 1].strip()
+    if(sfxId.count('0x') == 0):
+        return 0
```
Roman971, line 84: If I'm reading this correctly, it doesn't handle decimal values? I'd probably suggest handling them, but it may not be worth so I'll leave it up to you

petrie911: Eh... the purpose of that check is to make sure I'm not going to try to replace an SFX id that's already been processed. mips2c outputs hex by default, so I don't think needing decimal is going to be a common use case. Also the program description specifies the values should be in hex.

fig02: you can check for a string to see if you should be replacing

Roman971: Oh I see. But there are actually a bunch of sfx defines with 0x in them, so the current check is probably not enough

petrie911: Whoa you're right. I didn't know about those. This would cause a fatal error when it tries to feed that into fix_sfx_flag. I'll need to rethink that check

petrie911: OK, new error detection. Checks for the string `NA_SE` to consider an argument already converted and will now work on purely decimal SFX ids.

fig02 (approved):

tools/sfxconvert.py
```diff
@@ -0,0 +1,176 @@
...
+parser = argparse.ArgumentParser(description='Convert hex SFX ids to macros')
+parser.add_argument('file', help="source file to be processed")
+parser.add_argument('repo', help="directory of the decomp repo")
+parser.add_argument('-o', metavar='outfile',dest='outfile',help='file to write to instead of original')
+parser.add_argument('-v', action='store_true',help='show what changes are made')
+
+# if __name__ == "__main__":
+#     args = parser.parse_args()
+#     set_verbose(args.v)
+#     fix_sfx(args.file, args.repo, outfile=args.outfile)
```
Roman971, line 174: Intended to have the main execution commented out?

petrie911: Oops, forgot to undo that after testing.

Roman971 (approved):

PR #517, opened 2020-12-07
Docs for the Anime part of SkelAnime

petrie911:
Fixed a lot of names to be both shorter and more correct. Remaining functions named, AnimationContext system more documented, legacy AnimationHeader documented. A full understanding of how it works probably requires a deep dive into z_player.


fig02:
A few things:

-imo we should have `Get` in the name of all of these getter functions, for things like last frame, length, limb count etc

-Im thinking we dont really need to say `animationSeg` and can just say `animation`, the fact that its a segmented address isnt too important in the grand scheme of things. and you can theoretically pass a vitural address to any function that takes that as an arg, as S2V on a virtual address just returns itself.

and lastly imo we dont really gain much from shortening table to tbl for joint and morph table, it doesnt shorten it by all that much and its more natural to just spell it correctly

Ill review in more detail, but wanted to get my initial thoughts down first

fig02:
some additional things looking at the header
-I change my mind on the typedef thing if you want to change it, it doesnt serve much purpose since its not used anywhere else
-I see you ran the formatter in the header, Id take away the alignment on the comments though and have it just come right after the field
-maybe add an L to `prevTrans` and `baseTrans`? `Transl` will better help someone know what that means instead of guessing between translate/transfer/trans-whatever

fig02 (approved):
Im ok with these changes, other than the fact that we need to get things going on the Zap2 side as well

include/z64animation.h
```diff
@@ -29,15 +44,15 @@ typedef struct {
     /* 0x06 */ u8 child;
     /* 0x07 */ u8 sibling;
     /* 0x08 */ Gfx* dLists[2]; // Near and far
-} LodLimb; // size = 0x10
+} LodLimb;                     // size = 0x10
```
fig02, line 47: the formatter moved all of the sizes comments all the way to the side, can you put them back

Roman971: (this is one reason we only run the formatter on .c files, an unfortunate limitation)

Roman971 (commented):
It would be good to fix all multi line comments in here to start with `/**`, and I personally don't think we should note unused functions in comments (at least not if they are named and perfectly functional), but otherwise this seems good to me.

fig02:
The unused comments in here are worth mentioning i think, especially cause some of them are for the legacy animation format. but we could maybe do it in a way that makes more sense for that specific case

fig02:
ZAPD has been updated to pull from its new home in zeldaret, and should update with petries animation changes. Will see what jenkins says

Roman971 (approved):

PR #520, opened 2020-12-09
Match gfxprint.c

Synray:


src/code/gfxprint.c
```diff
@@ -220,9 +215,9 @@ void GfxPrint_SetBasePosPx(GfxPrint* this, s32 x, s32 y) {
     this->baseY = y << 2;
 }
 
-#ifdef NON_MATCHING
-// regalloc and ordering differences
 void GfxPrint_PrintCharImpl(GfxPrint* this, u8 c) {
+    u32 tile = (c & 0xff) * 2;
```
Roman971, line 219:
```suggestion
    u32 tile = (c & 0xFF) * 2;
```

src/code/gfxprint.c
```diff
@@ -132,15 +132,13 @@ u8 sGfxPrintFontData[(16 * 256) / 2] = {
 };
 
 #define gDPSetPrimColorMod(pkt, m, l, rgba)                                                    \
-    _DW({                                                                                      \
```
fig02, line 135: since we know now that this macro is only relevant in this file, im wondering if we want to give it a more custom name that doesnt suggest it was actually an addition to gbi. This is more of a question for everyone tho and doesnt need to be answered right now

fig02 (approved):

Roman971 (approved):

PR #521, opened 2020-12-09
z_demo_tre_lgt OK

shawlucas:
i would like to document this more but i just don't have much time right now.


shawlucas:
oh yeah there was also a misspelling in z64animation.h that i fixed (skelCuve to skelCurve)

src/overlays/actors/ovl_Demo_Tre_Lgt/z_demo_tre_lgt.c
```diff
@@ -23,21 +35,136 @@ const ActorInit Demo_Tre_Lgt_InitVars = {
...
+void DemoTreLgt_Init(Actor* thisx, GlobalContext* globalCtx) {
+    DemoTreLgt* this = THIS;
+
+    if (!SkelCurve_Init(globalCtx, &this->skelCurve, &D_06005EB8, sTransformUpdIdx[0])) {
+        osSyncPrintf("Demo_Tre_Lgt_Actor_ct();コンストラクト失敗\n");
+    }
+
+    // defined in .rodata but not used
+    "1";
+    "../z_demo_tre_lgt.c";
```
Roman971, line 55: I suggest putting an assert here since that's very likely what it was.
It would go something like this (UNK_LINE already exists):
```suggestion
    // This assert is optimized out but it exists due to its presence in rodata
    if (0) {
        __assert("1", "../z_demo_tre_lgt.c", UNK_LINE);
    }
```

src/overlays/actors/ovl_Demo_Tre_Lgt/z_demo_tre_lgt.c
```diff
@@ -23,21 +35,136 @@ const ActorInit Demo_Tre_Lgt_InitVars = {
...
+    }
+}
+
+void func_809937B4(DemoTreLgt* this, GlobalContext* globalCtx, f32 currentFrame) {
+    SkelAnimeCurve* skelCurve = &this->skelCurve;
+    s32 pad;
+
+    this->action = DEMO_TRE_LGT_ACTION_ANIMATE;
+
+    { s32 pad; }
```
Roman971, line 87: Is a separate pad in a block really needed here? Ideally it should just be `s32 pad[2]` at the top of the function, if it can be.

Alternatively, it might be worth looking into using temps for `sDemoTreLgtInfo.frameInfo[gSaveContext.linkAge]` and/or using `((void)0, gSaveContext.linkAge)`

src/overlays/actors/ovl_Demo_Tre_Lgt/z_demo_tre_lgt.h
```diff
@@ -6,11 +6,33 @@
...
+typedef struct {
+    /* 0x00 */ f32 startFrame;
+    /* 0x04 */ f32 endFrame;
+    /* 0x08 */ f32 unk_08;
+    /* 0x0C */ f32 unk_0C;
+} FrameInfo; // size = 0x10
+
+typedef struct {
+    /* 0x00 */ FrameInfo frameInfo[2];
+} DemoTreLgtInfo; // size = 0x20
```
Roman971, line 20: Any reason not to make` sDemoTreLgtInfo` an array of `FrameInfo` instead of making an additional struct for it? Also, `FrameInfo` should probably be moved to the C file, or renamed to something more specific to avoid conflicts.

src/overlays/actors/ovl_Demo_Tre_Lgt/z_demo_tre_lgt.h
```diff
@@ -6,11 +6,33 @@
...
+    /* 0x014C */ SkelAnimeCurve skelCurve;
+    /* 0x016C */ s32 action;
+    /* 0x0170 */ u32 unk_170; // some sort of alpha
+    /* 0x0174 */ u32 unk_174; // another sort of alpha
+    /* 0x0178 */ u8 status;
 } DemoTreLgt; // size = 0x017C
 
+typedef enum {
+    /* 0x00 */ DEMO_TRE_LGT_ACTION_WAIT, // wait until animation is needed
+    /* 0x01 */ DEMO_TRE_LGT_ACTION_ANIMATE,
```
Roman971, line 33: Trailing commas in enum trigger a warning
```suggestion
    /* 0x01 */ DEMO_TRE_LGT_ACTION_ANIMATE
```

src/overlays/actors/ovl_Demo_Tre_Lgt/z_demo_tre_lgt.h
shawlucas: It wouldn't match unless I made them separate structs

src/overlays/actors/ovl_Demo_Tre_Lgt/z_demo_tre_lgt.c
shawlucas: Everything gets messed up when using temps or the gSaveContext thing. the only thing that works is the pad, and yeah pad[2] works

src/overlays/actors/ovl_Demo_Tre_Lgt/z_demo_tre_lgt.h
Roman971: That shouldn't ever matter afaik

shawlucas: you were right, idk why i thought it only matched like that. maybe i was trying to match something and changed that but also changed something else and it made me think it was that

Roman971 (approved):

src/overlays/actors/ovl_Demo_Tre_Lgt/z_demo_tre_lgt.c
```diff
@@ -23,21 +40,135 @@ const ActorInit Demo_Tre_Lgt_InitVars = {
...
+
+    if (currentFrame < sDemoTreLgtInfo[(0, gSaveContext.linkAge)].endFrame) {
+        this->unk_170 = 255;
+    } else {
+        if (currentFrame <= (sDemoTreLgtInfo[(0, gSaveContext.linkAge)].endFrame +
+                             sDemoTreLgtInfo[(0, gSaveContext.linkAge)].unk_08)) {
+            this->unk_170 = ((((sDemoTreLgtInfo[(0, gSaveContext.linkAge)].endFrame - currentFrame) /
+                               sDemoTreLgtInfo[(0, gSaveContext.linkAge)].unk_08) *
+                              255.0f) +
+                             255.0f);
```
fig02, line 111: has the formatter been run? im surprised it ended up looking like this

fig02 (approved):

PR #526, opened 2020-12-10
Add .PHONY and 'test' target to Makefile

Synray:


fig02:
what is this for?
People use a bunch of emulators for testing such as pj64 and stuff, so this seems more of a like a personal thing on your end. but i may be misunderstanding

Synray:
it's just a shortcut to run the rom, so maybe 'test' isn't the best name. I just copied it from sm64's makefile, and tbh I always change EMULATOR to retroarch anyways, but I didn't know if mupen was more common or something.
I think it makes testing the rom a lot less tedious, but maybe that's not the case with p64, idk

.PHONY forces make to run `make clean` or `make all` etc. even if a file with the same name is already in the directory. So if someone creates a file named 'all' or 'clean' for some reason, they won't get confused by make refusing to rebuild.
Even though that situation is unlikely, it's still a good idea to have .PHONY because it improves performance a bit (although probably not by that much) since things like implicit rule search get skipped (where make will look for files named 'all.c', 'all.cpp', 'all.h', 'all.p', etc.).

fig02 (approved):

Roman971 (approved):

PR #531, opened 2020-12-14
Fix colliderinit.py to run from project root

Thar0:
filemap.py was already fixed in #235  to work from anywhere when the cutscene disassembler script was added so the fix was simple. Downside is it no longer works when cd'd to but I don't really see that as an issue since running it from root should be preferred anyway.

### DIFF
--- a/tools/csdis.py
+++ b/tools/csdis.py
@@ -2,8 +2,7 @@
 
 from overlayhelpers import filemap
 
-import argparse
-import struct
+import argparse, os, struct
 
 """
 Enumerations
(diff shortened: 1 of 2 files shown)


ethteck:
If you look at some of the scripts in the top-level of tools, like disassemble_data.py, it actually works relative to the script itself. I'm personally fine with this change because I feel scripts should be run from the root dir, not from their own dir. 

That being said, if you use the approach I just mentioned, you can make it so these can be run from anywhere.

Thar0:
Thanks for the suggestion! I did this for both colliderinit and csdis so they now both run from anywhere.

ethteck (approved):

Roman971 (approved):

PR #538, opened 2020-12-17
Remove inline THIS-> from actors 

petrie911:
Removes all instances of inlined `THIS->`, which turned out to all be unnecessary except for ObjTimeblock. I put the explicit cast in there for consistency.


src/overlays/actors/ovl_En_Rd/z_en_rd.c
```diff
@@ -278,7 +278,7 @@ void func_80AE2C1C(EnRd* this, GlobalContext* globalCtx) {
     Color_RGBA8 sp40 = D_80AE4924;
     Color_RGBA8 sp3C = D_80AE4928;
     Player* player = PLAYER;
-    s32 pad;
+    Player* player2;
```
fig02, line 281: did this end up getting used? should prob stay as pad

petrie911: oh whoops, was trying to remove a PLAYER, turned out to be not happening.

fig02 (approved):

Roman971 (approved):

PR #541, opened 2020-12-17
additional matching functions in z_room, z_camera, and En_Zl3

petrie911:
Matched a function in each of these and made improvements to the matchings in z_room. I also have duplicates of roman's tubo_trap and po sisters. Merge that one first.


Roman971 (approved):

src/overlays/actors/ovl_En_Zl3/z_en_zl3.c
```diff
@@ -234,20 +234,19 @@ void func_80B53974(EnZl3* this, u8 arg1) {
     this->unk_3C8 = arg1;
 }
 
-#ifdef NON_MATCHING
```
ethteck, line 237: nice! 🎊

include/z64math.h
```diff
@@ -95,7 +95,7 @@ typedef struct {
 /**
  * Trig macros
 */
-#define DEGF_TO_BINANG(degreesf) (s16)(degreesf * 182.04167f + 0.5f)
+#define DEGF_TO_BINANG(degreesf) (s16)(degreesf * 182.04167f + .5f)
```
fig02, line 98: im worried that this specific fix is only relevant for camera. if it starts breaking stuff outside of camera then it should probably be a local macro

petrie911: I believe Krim made it for camera. I don't know if anything else even uses that.

fig02: yea he did, but it was put in z64math with the intention of being used everywhere i guess. just something to keep in mind

fig02 (approved):

include/z64math.h
petrie911: speaking of macros in math, BINANG_SUB should probably be SUB16

PR #542, opened 2020-12-18
Consistent naming for Math_ functions

petrie911:
`func_800CA720` and `func_800CA774` are sin and cos respectively, taking a float radian argument. They're the equivalent of `Math_Sins` and `Math_Coss` in that they're wrappers for `sins` and `coss`, so they should probably be `Math_Sinf` and `Math_Cosf`. This will make any actors that use them a lot less opaque.

I also named the rest of the functions in sys_math, though that's not super important because they're unused. Nevertheless, they are `Math_Factorialf` (factorial of float argument), `Math_Factoriali` (factorial of int argument), and `Math_Powf` (float raised to an int power).

EDIT: In addition to the above, it seems the Math_ library functions are inconsistently named. This PR creates a more uniform naming system and updates some names of common functions to be more descriptive of what they do.


fig02 (approved):
think the `i` can probably be omitted from the name for factorial. just specifying `f` for the float variant is probably enough

petrie911:
Agreed on Factoriali, looks better and makes more sense than Factorialf. Also unclear why Factorialf exists since it truncates the float anwyays. Factorial is just superior.

fig02:
we should also make the f uppercase like we do for functions like `Math_SmoothScaleMaxMinF`

petrie911:
Should the same be done with Math_Sinf as Math_SinF? Or, for that matter, Math_Coss to Math_CosS?

fig02:
oh yea didnt realize those are inconsistent. i say yes for uppercase personally since it follows our camel case scheme

petrie911:
The only downside of this is that Math_SinS and Math_CosS are very, *very* heavily used. This is an extensive change and will break all open PRs.

fig02 (approved):

Roman971 (approved):

fig02:
can merge after conflicts resolved

PR #545, opened 2020-12-18
ovl_En_In OK

Thar0:
Intended to do a bit more docs than this but a lot of it relies on currently undocumented systems in z_actor


src/overlays/actors/ovl_En_In/z_en_in.c
```diff
@@ -33,75 +36,938 @@ const ActorInit En_In_InitVars = {
...
+
+u16 func_80A79010(GlobalContext* globalCtx) {
+    Player* player = PLAYER;
+    u16 temp_v0 = Text_GetFaceReaction(globalCtx, 25);
+
+    if (temp_v0 != 0) {
+        return temp_v0;
+    }
+    if (gSaveContext.eventChkInf[1] & 0x100) {
+        if (gSaveContext.nightFlag == 0) {
```
fig02, line 102: !night

src/overlays/actors/ovl_En_In/z_en_in.c
```diff
@@ -33,75 +36,938 @@ const ActorInit En_In_InitVars = {
...
+            return 0x205F;
+        } else {
+            return 0x2057;
+        }
+    }
+    if (gSaveContext.nightFlag == 1) {
+        return 0x204E;
```
fig02, line 109: if night

src/overlays/actors/ovl_En_In/z_en_in.c
```diff
@@ -33,75 +36,938 @@ const ActorInit En_In_InitVars = {
...
+            } else {
+                func_8010B720(globalCtx, this->actor.textId = 0x2039);
+                gSaveContext.eventInf[0] &= ~0xF;
+                gSaveContext.eventInf[0] &= ~0x20;
+                gSaveContext.eventInf[0] &= ~0x40;
+                this->actionFunc = func_80A7A4C8;
+            }
+            break;
+    }
+    if (!gSaveContext.rupees) {}
```
fig02, line 233: void0 might help get rid of this?

src/overlays/actors/ovl_En_In/z_en_in.c
```diff
@@ -33,75 +36,938 @@ const ActorInit En_In_InitVars = {
...
+}
+
+s32 D_80A7B998 = 0;
+
+void EnIn_Init(Actor* thisx, GlobalContext* globalCtx) {
+    EnIn* this = THIS;
+    RespawnData* respawn = &gSaveContext.respawn[RESPAWN_MODE_DOWN];
+    Vec3f respawnPos;
+
+    this->requiredObject = Object_GetIndex(&globalCtx->objectCtx, 0xC0);
```
fig02, line 456: requiredObjectBankIdx or something to make it clear its not an object id

src/overlays/actors/ovl_En_In/z_en_in.c
Thar0: This doesn't match since it's a direct comparison to 1 as opposed to != 0

src/overlays/actors/ovl_En_In/z_en_in.c
Thar0: Didn't seem to help in this case unfortunately.

fig02 (approved):

src/code/z_horse.c
```diff
@@ -127,7 +127,7 @@ void func_8006D0EC(GlobalContext* globalCtx, Player* player) {
             }
         }
     } else if (!Flags_GetEventChkInf(0x18)) {
-        if ((DREG(1) == 0) && (globalCtx->sceneNum == SCENE_SOUKO) && (gSaveContext.nightFlag != 0)) {
+        if ((DREG(1) == 0) && (globalCtx->sceneNum == SCENE_SOUKO) && (gSaveContext.nightFlag)) {
```
Roman971, line 130: `nightFlag` isn't really a boolean (like `linkAge`), so I think we should leave these checks as `== 0`/`!= 0`/`== 1` for now. Eventually, we will probably have `IS_DAY`/`IS_NIGHT` macros to handle it (including the `== 1` checks), so we don't need to worry about treating it like a boolean.

src/overlays/actors/ovl_En_In/z_en_in.h
```diff
@@ -6,9 +6,37 @@
...
+    /* 0x01EE */ s16 unk_1EE;
+    /* 0x01F0 */ s16 camId;
+    /* 0x01F2 */ s16 activeCamId;
+    /* 0x01F4 */ char unk_1F4[0x4];
+    /* 0x01F8 */ s16 unk_1F8;
+    /* 0x01FA */ s16 unk_1FA;
+    /* 0x01FC */ s16 unk_1FC;
+    /* 0x01FE */ Vec3s limbDrawTable[20];
+    /* 0x0276 */ Vec3s transitionDrawTable[20];
+    /* 0x0276 */ char unk_2EE[0x2];
```
Roman971, line 31: probably padding
```suggestion
```

src/overlays/actors/ovl_En_In/z_en_in.c
```diff
@@ -33,75 +36,938 @@ const ActorInit En_In_InitVars = {
...
+}
+
+s32 D_80A7B998 = 0;
+
+void EnIn_Init(Actor* thisx, GlobalContext* globalCtx) {
+    EnIn* this = THIS;
+    RespawnData* respawn = &gSaveContext.respawn[RESPAWN_MODE_DOWN];
+    Vec3f respawnPos;
+
+    this->objectIdx = Object_GetIndex(&globalCtx->objectCtx, 0xC0);
```
Roman971, line 456: Object enum can be used here, and I suggest renaming `objectIdx` to something a bit clearer, like `ingoObjBankIndex`
```suggestion
    this->objectIdx = Object_GetIndex(&globalCtx->objectCtx, OBJECT_IN);
```

src/overlays/actors/ovl_En_In/z_en_in.c
```diff
@@ -33,75 +36,938 @@ const ActorInit En_In_InitVars = {
...
+                    case 5:
+                    case 6:
+                        if (this->actor.params == 4) {
+                            sp3C = 1;
+                        }
+                        break;
+                }
+                if (sp3C != 1) {
+                    Actor_Kill(&this->actor);
+                } else {
```
Roman971, line 543: early return instead of else here, imo

src/overlays/actors/ovl_En_In/z_en_in.c
```diff
@@ -33,75 +36,938 @@ const ActorInit En_In_InitVars = {
...
+            func_80A79690(&this->skelAnime, this, globalCtx);
+        }
+        func_8002E4B4(globalCtx, &this->actor, 0.0f, 0.0f, 0.0f, 4);
+    }
+    func_80A79A2C(this);
+    this->actionFunc(this, globalCtx);
+    if (this->actionFunc != func_80A7A304) {
+        func_80A79AB4(this, globalCtx);
+        if (gSaveContext.timer2Value < 6 && gSaveContext.timer2State != 0 && this->unk_308.unk_00 == 0) {
+            if (func_8002F194(&this->actor, globalCtx)) {}
```
Roman971, line 900: Does this have to be in a `if` condition?

src/code/z_horse.c
fig02: in kankyo where its set at least, it is def just true or false for currently being night. though they do check explicitly for "true" alot 

will it be possible to have a macro that uses both methods of checking?

Roman971: `IS_DAY` and `IS_NIGHT` can handle everything, just like `LINK_IS_CHILD` and `LINK_IS_ADULT` will once they are adjusted.

Right now, my idea is to use `LINK_IS_CHILD` for `linkAge == 0` and `LINK_IS_ADULT` for `linkAge == 1`. Then we can use `!LINK_IS_CHILD` for `linkAge != 0`, and `!LINK_IS_ADULT` for `linkAge != 1`. With the same pattern for day/night stuff ofc

fig02: ah yea that makes sense, alright.
though thats kind of unfortunate for the age stuff. `!LINK_IS_CHILD` is way more common than `LINK_IS_ADULT` and the macro saying adult is a bit easier on the eyes. but thats just how it is i guess

Roman971: I just realized I inverted child and adult in this explanation, but you get the idea. I agree it's unfortunate, but chances are they were just inconsistent about it and I don't think there is much we can do about that.

Roman971: To some extent, it makes sense to check for "not being adult" more often, if we consider that adult is the "normal" age (or the main one)

src/overlays/actors/ovl_En_In/z_en_in.c
Thar0: It does unfortunately, in this case it really does generate an extra branch (likely)

fig02 (approved):

Roman971 (approved):

PR #548, opened 2020-12-19
Decompile z_skin.c

Thar0:
SEGMENTED_TO_VIRTUAL: The File
1 nonmatching, regalloc and stack differences. Looks equivalent.


src/code/z_skin.c
```diff
@@ -1,21 +1,222 @@
...
+            gSPPopMatrix(POLY_OPA_DISP++, G_MTX_MODELVIEW);
+            gDPPipeSync(POLY_OPA_DISP++);
+        }
+    }
+    CLOSE_DISPS(gfxCtx, "../z_skin.c", 433);
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_skin/func_800A5E28.s")
+void func_800A60D8(Actor* actor, GlobalContext* globalCtx, PSkinAwb* skin, SkinCallback callback, SkinCallback2 arg4,
+                   s32 arg5, s32 arg6, s32 arg7) {
```
fig02, line 158: these callbacks look to be the equivalent of override and post limb draw. not documenting now is fine but just wanna point that out

fig02 (approved):

src/code/z_skin.c
```diff
@@ -1,21 +1,222 @@
 #include "global.h"
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_skin/func_800A57C0.s")
+MtxF D_801600C0[60];
+char pad[0x10];
```
Roman971, line 4: `pad` is too generic for a global variable. I suggest leaving it as a D_ name for now, since that's what usually do for unknowns

fig02 (approved):

Roman971 (approved):

PR #550, opened 2020-12-20
Obj_Mure2 (one non-matching)

Zelllll:
A few things to note:
- A lot of these matches are very fake, although I tried for hours to improve them but I couldn't.
- Some documentation may not be entirely correct or be the proper format.
- The non-matching is very close, but changing minor things makes the function explode.

I would greatly appreciate help in fixing these things before it gets merged.


Zelllll:
This should compile correctly now.

src/overlays/actors/ovl_Obj_Mure2/z_obj_mure2.c
```diff
@@ -25,33 +41,198 @@ const ActorInit Obj_Mure2_InitVars = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Obj_Mure2/func_80B9A158.s")
+static f32 sDistSquared1[] = { SQ(1600.0f), SQ(1600.0f), SQ(1600.0f) };
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Obj_Mure2/func_80B9A260.s")
+static f32 sDistSquared2[] = { SQ(1705.0f), SQ(1705.0f), SQ(1705.0f) };
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Obj_Mure2/func_80B9A2B0.s")
+#ifdef NON_MATCHING
+static s16 D_80B9A818[] = { 9, 12, 8 };
```
fig02, line 50: if your intention is to put this inside the func when its matched, then you can put it inside now

Zelllll: I can't since its used between multiple functions.

Zelllll:
Alright I merged in the new math stuff and fixed a fake match

fig02 (approved):

Roman971 (approved):

PR #552, opened 2020-12-21
Cleanup z_lifemeter

Zelllll:
- Fixed data issue. Before none of the data was merged into the file.
 - Added defines for the heart colors, this lets users easily change the heart colors without manually going through the functions and doing so.
 - Added some documentation surrounding the large amount of unused data in this file, which it turns out used to have a purpose based on pre-release footage of the game.
 - Documented more of InterfaceContext in z64.h, namely the heart colors.
 - Improved the only non-matching in z_lifemeter, although unfortunately I was not able to match it, if any brave souls want to take a shot at it though, be my guest!


src/code/z_lifemeter.c
```diff
@@ -1,20 +1,113 @@
...
+ * In the final game these environments only have a timer and do not damage you continuously.
+ ***/
+
+#define HEART_PRIM_BURN_PRIM_R 255
+#define HEART_PRIM_BURN_PRIM_G 190
+#define HEART_PRIM_BURN_PRIM_B 0
+
+#define HEART_PRIM_BURN_ENV_R 255
+#define HEART_PRIM_BURN_ENV_G 0
+#define HEART_PRIM_BURN_ENV_B 0
```
fig02, line 39: these have an extra "PRIM_ after heart

Zelllll: fixed

Zelllll:
I've decided that I am not going to continue trying to match the last non-matching, I've tried for 3+ days and have not been able to, so this PR should be able to be reviewed now.

Zelllll:
This should compile correctly now.

include/functions.h
```diff
@@ -990,8 +990,8 @@ void Color_RGBA8_Copy(Color_RGBA8* dst, Color_RGBA8* src);
 void func_80078884(u16 sfxId);
 void func_800788CC(u16 sfxId);
 void func_80078914(Vec3f* arg0, u16 sfxId);
-void Health_InitData(GlobalContext* globalCtx);
-void Health_UpdateData(GlobalContext* globalCtx);
+void Health_Init(GlobalContext* globalCtx);
+void Health_Update(GlobalContext* globalCtx);
```
fig02, line 994: we may want to consider calling this HealthMeter or something, since it really is just the graphical display

src/code/z_lifemeter.c
```diff
@@ -1,59 +1,139 @@
...
+// Far from matching, but is equivalent. The for loop needs to become unrolled somehow in order to match.
+void Health_Update(GlobalContext* globalCtx) {
+    static s16 sHeartPrimColors[3][3] = {
+        { HEART_PRIM_R, HEART_PRIM_G, HEART_PRIM_B },
+        { HEART_BURN_PRIM_R, HEART_BURN_PRIM_G, HEART_BURN_PRIM_B },    // unused
+        { HEART_DROWN_PRIM_R, HEART_DROWN_PRIM_G, HEART_DROWN_PRIM_B }, // unused
+    };
+    static s16 sHeartEnvColors[3][3] = {
+        { HEART_ENV_R, HEART_ENV_G, HEART_ENV_B },
+        { HEART_PRIM_BURN_ENV_R, HEART_PRIM_BURN_ENV_G },            // unused
```
fig02, line 96: did making this in function improve matching? otherwise it should prob be outside

Zelllll:
fixed all of figs comments

fig02 (approved):

include/functions.h
```diff
@@ -990,8 +990,8 @@ void Color_RGBA8_Copy(Color_RGBA8* dst, Color_RGBA8* src);
 void func_80078884(u16 sfxId);
 void func_800788CC(u16 sfxId);
 void func_80078914(Vec3f* arg0, u16 sfxId);
-void Health_InitData(GlobalContext* globalCtx);
-void Health_UpdateData(GlobalContext* globalCtx);
+void HealthMeter_Init(GlobalContext* globalCtx);
+void HealthMeter_Update(GlobalContext* globalCtx);
 void Health_Draw(GlobalContext* globalCtx);
 void Health_HandleCriticalAlarm(GlobalContext* globalCtx);
 u32 Health_IsCritical(void);
```
Roman971, line 997: Might as well rename the whole system to HealthMeter_ for consistency, if we think that's better (at the very least for the Draw function)

src/code/z_lifemeter.c
```diff
@@ -1,59 +1,147 @@
...
-s16 D_8015FDD0[6];
-s16 D_8015FDE0[6];
-
-void Health_InitData(GlobalContext* globalCtx) {
+/***
+ * These are the colors for the hearts in the interface. The prim color is the red color of the heart
+ * for the base hearts, while the prim color for the double defense hearts is the white outline. The
+ * env color for the base hearts is the purple-ish outline, while the env color for the double defense
+ * hearts is the red color of the hearts.
+ ***/
```
Roman971, line 8: It's fairly minor, but the style for comment blocks usually starts with `/**` and ends with `*/`

src/code/z_lifemeter.c
```diff
@@ -172,18 +249,24 @@ s32 func_80078E84(GlobalContext* globalCtx) {
...
             globalCtx->damagePlayer(globalCtx, -(gSaveContext.health + 1));
             return 1;
         }
     }
-
     return 0;
 }
 
 void Health_Draw(GlobalContext* globalCtx) {
+    static u8* sHeartTextureList[] = {
```
Roman971, line 262: For these arrays, I suggest Textures (plural) instead of TextureList, for consistency with the rest of the codebase.
And imo they would be better left outside the function, especially since they will take a lot more space once they have proper symbols with names.
```suggestion
    static u8* sHeartTextures[] = {
```

include/z64.h
```diff
@@ -399,10 +399,14 @@ typedef struct {
...
-    /* 0x020E */ s16    unk_20E[6];
-    /* 0x021A */ s16    unk_21A[6];
+    /* 0x0202 */ s16    beatingHeartPrim[3];
+    /* 0x0208 */ s16    beatingHeartEnv[3];
+    /* 0x020E */ s16    healthPrimR[2];
+    /* 0x0212 */ s16    healthPrimG[2];
+    /* 0x0216 */ s16    healthPrimB[2];
+    /* 0x021A */ s16    healthEnvR[2];
+    /* 0x021E */ s16    healthEnvG[2];
+    /* 0x0222 */ s16    healthEnvB[2];
```
Roman971, line 409: Thoughts on using the term heart here, since that's what you went with for all other colors? (as far as I can tell)

src/code/z_lifemeter.c
```diff
@@ -1,59 +1,147 @@
...
+#define HEART_ENV_G 40
+#define HEART_ENV_B 60
+
+#define HEART_DOUBLE_DEFENSE_PRIM_R 255
+#define HEART_DOUBLE_DEFENSE_PRIM_G 255
+#define HEART_DOUBLE_DEFENSE_PRIM_B 255
+
+#define HEART_DOUBLE_DEFENSE_ENV_R 200
+#define HEART_DOUBLE_DEFENSE_ENV_G 0
+#define HEART_DOUBLE_DEFENSE_ENV_B 0
```
Roman971, line 24: I suggest shortening DOUBLE_DEFENSE to DD here

src/code/z_lifemeter.c
```diff
@@ -1,59 +1,147 @@
...
+ * In the final game these environments only have a timer and do not damage you continuously.
+ ***/
+
+#define HEART_BURN_PRIM_R 255
+#define HEART_BURN_PRIM_G 190
+#define HEART_BURN_PRIM_B 0
+
+#define HEART_PRIM_BURN_ENV_R 255
+#define HEART_PRIM_BURN_ENV_G 0
+#define HEART_PRIM_BURN_ENV_B 0
```
Roman971, line 39: I think this is what you meant?
```suggestion
#define HEART_BURN_ENV_R 255
#define HEART_BURN_ENV_G 0
#define HEART_BURN_ENV_B 0
```

src/code/z_lifemeter.c
```diff
@@ -172,18 +249,24 @@ s32 func_80078E84(GlobalContext* globalCtx) {
...
-
     return 0;
 }
 
 void Health_Draw(GlobalContext* globalCtx) {
+    static u8* sHeartTextureList[] = {
+        0x02000400, 0x02000100, 0x02000100, 0x02000100, 0x02000100, 0x02000100, 0x02000200, 0x02000200,
+        0x02000200, 0x02000200, 0x02000200, 0x02000300, 0x02000300, 0x02000300, 0x02000300, 0x02000300,
+    };
+    static u8* sDoubleDefenseHeartTextureList[] = {
```
Roman971, line 266: This DoubleDefense can probably also be shortened to DD. And ideally the order of DD and Heart should be consistent throughout the file (mostly referring to color and factor arrays which are currently HeartDD, so this one should probably be HeartDD as well)

src/code/z_lifemeter.c
```diff
@@ -1,59 +1,147 @@
...
+s16 sHeartDDEnvFactors[3][3] = {
+    { 0, 0, 0 },
+    { 55, 0, 0 },     // unused
+    { -200, 0, 255 }, // unused
+};
+
+s16 sHeartPrimCurrent1[3];
+s16 sHeartEnvCurrent1[3];
+s16 sHeartPrimCurrent2[2][3];
+s16 sHeartEnvCurrent2[2][3];
```
Roman971, line 100: I believe these 4 arrays are specifically for DD hearts, as opposed to the colors in interfaceCtx which are normal hearts. Also I think Current2 is what you called beatingHeart in interfaceCtx, so it would be a good idea to name them consistently.

include/z64.h
Zelllll: Ended up switching a lot of things over to health so it can probably stay as is

Zelllll:
resolved roman's feedback

include/z64.h
Roman971: That's also fine I guess, but now it's still not consistent with "beating heart" names, so I'm not sure

Zelllll: yeah a bit unsure of what to name that, I was thinking trailingHeart but that also has "heart" in it. maybe trailingHealth?

fig02: if hearts instead of health makes more sense, it might be better to go in that direction for everything

Roman971: I also think it could make sense to go with hearts for all these color names

Zelllll: alright when i get some time ill make that change then

Zelllll:
Alright I finally should have resolved all of your feedbacks, health is now hearts everywhere.

fig02 (approved):

Roman971 (approved):

PR #560, opened 2020-12-23
Match 4 functions in z_player.c

Synray:
Only Player_Init left


src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -9500,40 +9510,35 @@ void func_80847BA0(GlobalContext* globalCtx, Player* this) {
 
         D_8085360C = ABS(sp9A);
 
-        temp1 = D_8085360C * 0.00008f;
-        if (!(this->actor.bgCheckFlags & 1) || (temp1 >= 1.0f)) {
+        spB0 = D_8085360C * 0.00008f;
+        if (!(this->actor.bgCheckFlags & 1) || (spB0) >= 1.0f) {
             this->unk_880 = R_RUN_SPEED_LIMIT / 100.0f;
```
fig02, line 9515: can remove parens on spB0

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -9864,9 +9859,8 @@ Vec3f D_80854814 = { 0.0f, 0.0f, 200.0f };
 f32 D_80854820[] = { 2.0f, 4.0f, 7.0f };
 f32 D_8085482C[] = { 0.5f, 1.0f, 3.0f };
 
-#ifdef NON_MATCHING
-// stack alloc differences
-void Player_UpdateCommon(Player* this, GlobalContext* globalCtx, Input* input) {
+void Player_UpdateCommon(Player* this, GlobalContext* gc, Input* input) {
+    GlobalContext* globalCtx = gc;
```
fig02, line 9863: can you name `gc` as `globalCtx2`. we already know why they had a temp for it in some cases so it will be changed later, but globalCtx2 will help us find and replace later

also, do you need to use the arg version for the state frames access? it makes sense if so, since that var is part of the base gamestate. just want to make sure its required

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -12588,51 +12582,44 @@ void func_8084FF7C(Player* this) {
     }
 }
 
-#ifdef NON_MATCHING
-// regalloc differences
 void func_8085002C(Player* this) {
-    s32 pad;
+    s16 v0 = D_80858AC8.unk_08, v1 = D_80858AC8.unk_06;
```
fig02, line 12586: for now we are avoiding sameline declared incase we run into a situation where we need to flip stack order in the future. unless its required here?

fig02 (commented):
nice, a couple things

Roman971:
I feel like most of these changes shouldn't be necessary to match, it would be nice to isolate only the stuff that required matching (ideally without moving the stack layout to the top of the function)

src/overlays/actors/ovl_player_actor/z_player.c
Synray: yeah `globalCtx2` needs to be used to access gameplayFrames, or else stack gets changed

fig02 (approved):

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -6072,21 +6070,21 @@ s32 func_8083EC18(Player* this, GlobalContext* globalCtx, u32 arg2) {
...
                 func_80041E4C(&globalCtx->colCtx, this->actor.wallPoly, this->actor.wallPolySource)) {
-                f32 phi_f20;
+                s32 i;
                 CollisionPoly* sp84 = this->actor.wallPoly;
                 f32 sp80;
                 f32 sp7C;
                 f32 phi_f12;
                 f32 phi_f14;
+                Vec3f sp50[3];
+                f32 phi_f20;
```
Roman971, line 6080: I imagine these stack changes aren't necessary, could they be omitted if possible?

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -9476,7 +9468,7 @@ void func_80847BA0(GlobalContext* globalCtx, Player* this) {
         CollisionPoly* spA0;
         u32 sp9C;
         s16 sp9A;
-        f32 temp1;
+        u32 pad;
```
Roman971, line 9471: We generally type pads as s32, but it's pretty minor

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -10109,8 +10103,9 @@ void Player_UpdateCommon(Player* this, GlobalContext* globalCtx, Input* input) {
 
         if ((globalCtx->csCtx.state != 0) && (this->csMode != 6) && !(this->stateFlags1 & 0x800000) &&
             !(this->stateFlags2 & 0x80) && (this->actor.type == ACTORTYPE_PLAYER)) {
+            CsCmdActorAction *linkAction = globalCtx->csCtx.linkAction;
```
Roman971, line 10106: I believe this temp is the only change necessary to match this function, unless I'm missing something. 
Also I suggest naming the temp something more descriptive like `linkActionCsCmd`, but that doesn't really matter. 
And make sure to run the formatter after all changes so things like the pointer star style here gets fixed for you.

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -12588,51 +12580,49 @@ void func_8084FF7C(Player* this) {
...
+    D_80858AC8.unk_08 -= v0 >> 3;
     D_80858AC8.unk_06 += -D_80858AC8.unk_00 >> 2;
     D_80858AC8.unk_08 += -D_80858AC8.unk_02 >> 2;
 
     sp26 = this->actor.posRot.rot.y - this->actor.shape.rot.y;
 
-    sp28 = this->actor.speedXZ * -200.0f * Math_Coss(sp26) * (Math_Rand_CenteredFloat(2.0f) + 10.0f);
-    sp2A = this->actor.speedXZ * 100.0f * Math_Sins(sp26) * (Math_Rand_CenteredFloat(2.0f) + 10.0f);
+    sp28 = (s32)(this->actor.speedXZ * -200.0f * Math_Coss(sp26) * (Math_Rand_CenteredFloat(2.0f) + 10.0f)) & 0XFFFF;
+    sp2A = (s32)(this->actor.speedXZ *  100.0f * Math_Sins(sp26) * (Math_Rand_CenteredFloat(2.0f) + 10.0f)) & 0XFFFF;
```
Roman971, line 12598: I believe the changes in these 2 lines are the only thing necessary to match this one.
Also we use 0xFFFF, not 0XFFFF (not sure if the formatter would fix that on its own)

src/overlays/actors/ovl_player_actor/z_player.c
Roman971: I think there are still some changes in this function that could be omitted (eg. the globalCtx2 stuff and other temp changes)

Synray: you're right, the temp changes got it closer to matching, but I guess the CsCmdActorAction temp fixed everything at once, even making everything else unnecessary.

Roman971 (approved):
Seems good to me, but I think you added back the asm for some functions during a merge, so the Jenkins check is failing.

Synray:
rip. ok, that seems to have worked

PR #562, opened 2020-12-24
Bg_Hidan_Kowarerukabe OK

Zelllll:
Fire Temple Breakable Walls and Floors, all functions OK and documented


AdamKiddle (changes requested):
Is it weird that I've kinda been looking forward to seeing this this actor decomped for quite a while? Maybe I'm going crazy...

Thanks for releasing this one! I left some feedback on stuff like compiler warnings and style choices.

I also show you how to re-write these double loops in a cleaner way (these general rotation operations happen quite often across the codebase and they're often implemented in the same way each time)

Zelllll:
Thanks a lot for all the suggestions! Looks like one of your suggestions is causing the PR to not build anymore though, I'll look into it later.

fig02 (approved):

fig02:
need to merge master and update math functions

Zelllll:
Alright all the math stuff is merged in and conflicts are resolved, should be good to merge

Roman971 (approved):

PR #566, opened 2020-12-27
[WIP] GCC Support

ProjectRevoTPP:
Here it is, a COMPILER flag you can switch to use gcc on. You will need gcc-mips-linux-gnu package for starters.

Second, there are a ton of overlays not working. Code and boot seems to be fine. Mkldscript had to get a fix for gcc to work properly. I was encountering constant crashes and black screens with overlays so its just not going to be for now in this PR. I need someone to help troubleshoot overlays, but since we dont yet have full gcc support this PR is marked as WIP.


ProjectRevoTPP:
This will be reopened when I have a plan for verifyable GCC support in the future.

PR #579, opened 2020-12-30
Fixed descriptions of Zelda's Lullaby, Song of Time, and Scarecrow's Song effect actors

silv3rwing07:
There were incorrect descriptions of Oceff_Wipe and Oceff_Wipe4 which had been copied from old, wrong info which had been duplicated around the web. Fixed the info here and got someone to fix it on the CloudModding page.


fig02:
Thanks for this.
Generally we try to keep the descriptions short and sweet, and not get into any details about param values and such. 
We can make an enum value for params that details that information instead. If you dont want to do that in this pr its fine.

Id go with ` Zelda's Lullaby and Song of Time Ocarina Effect`
and 
`Scarecrow's Song and an unused Ocarina Effect`

and then in the code `(if params == OCEFF_WIPE4_UNUSED)` etc

fig02:
Hey just a reminder about these changes, they are really minor so should be an easy fix

silv3rwing07:
Fixed, thanks for the reminder!

src/overlays/actors/ovl_Oceff_Wipe/z_oceff_wipe.c
```diff
@@ -111,7 +111,7 @@ void OceffWipe_Draw(Actor* thisx, GlobalContext* globalCtx) {
     gSPMatrix(POLY_XLU_DISP++, Matrix_NewMtx(globalCtx->state.gfxCtx, "../z_oceff_wipe.c", 375),
               G_MTX_NOPUSH | G_MTX_LOAD | G_MTX_MODELVIEW);
 
-    if (this->actor.params) {
+    if (this->actor.params) { //Params nonzero == Song of Time
```
fig02, line 114: id say instead of the comment we can just do `if (params != OCEFF_WIPE_ZELDAS)`
this probably shouldnt have been written as a boolean comparison anyway, but it was an error during decopmilation. we can change it now

fig02: Additionally, you can remove the underscrore in the name for both enums
`OceffWipe_Type` -> `OceffWipeType`

then this is good to go

fig02 (approved):
changes implemented

Roman971 (approved):

PR #586, opened 2020-12-31
DemoExt OK and documented

EllipticEllipsis:
Despite appearances, this actor does not suck.

All functions matched, all variables and functions named, enums made for actions and draws. Some names can be improved. Data could be inlined at the expense of putting the InitVars at the bottom, but with the benefit of fewer predeclarations being required.


fig02:
init vars at the bottom is fine, and is done in a few places already. its likely what the original file did

src/overlays/actors/ovl_Demo_Ext/z_demo_ext.c
```diff
@@ -1,17 +1,233 @@
...
+
+void DemoExt_PlayVortexSFX(DemoExt* this) {
+    if (this->alphaTimer <= (kREG(35) + 40.0f) - 15.0f) {
+        Audio_PlaySoundGeneral(NA_SE_EV_FANTOM_WARP_L - SFX_FLAG, &this->actor.projectedPos, 4, &D_801333E0,
+                               &D_801333E0, &D_801333E8);
+    }
+}
+
+CsCmdActorAction* DemoExt_GetNpcCsAction(GlobalContext* globalCtx, s32 npcActionIndex) {
+    if (globalCtx->csCtx.state != 0) {
```
fig02, line 57: can name it GetNpcAction if you like, with a name like that its implied its for a cutscene
and I believe theres an enum for cutscene state, but I dont think we use it everywhere yet. if not then ignore

src/overlays/actors/ovl_Demo_Ext/z_demo_ext.c
```diff
@@ -1,17 +1,233 @@
...
+
+void DemoExt_SetupWait(DemoExt* this) {
+    this->action = EXT_WAIT;
+    this->drawMode = EXT_DRAW_NOTHING;
+}
+
+void DemoExt_SetupMaintainVortex(DemoExt* this, GlobalContext* globalCtx) {
+    CsCmdActorAction* npcAction;
+
+    npcAction = DemoExt_GetNpcCsAction(globalCtx, 5);
```
fig02, line 71: can do this all on one line if you like

src/overlays/actors/ovl_Demo_Ext/z_demo_ext.c
```diff
@@ -1,17 +1,233 @@
...
+
+void DemoExt_DispellVortex(DemoExt* this, GlobalContext* globalCtx) {
+    DemoExt_PlayVortexSFX(this);
+    DemoExt_SetScrollAndRotation(this);
+    DemoExt_SetColorsAndScales(this);
+    DemoExt_FinishClosing(this);
+}
+
+void DemoExt_Update(Actor* thisx, GlobalContext* globalCtx) {
+    static DemoExtActionFunc sActionFuncs[] = { DemoExt_Wait, DemoExt_MaintainVortex, DemoExt_DispellVortex };
```
fig02, line 169: we generally keep arrays of function pointers outside just cause in general they can get quite long, even though that obv isnt the case here.
Id also add a comma so it lists them vertically, that way its easier for a modder to add more functions if they ever felt the need to

src/overlays/actors/ovl_Demo_Ext/z_demo_ext.c
```diff
@@ -1,17 +1,233 @@
...
+    DemoExt* this = THIS;
+
+    if ((this->action < 0) || (this->action >= 3) || sActionFuncs[this->action] == NULL) {
+        // Main mode is abnormal!
```
fig02, line 173: can use action enum

src/overlays/actors/ovl_Demo_Ext/z_demo_ext.c
```diff
@@ -1,17 +1,233 @@
...
+    }
+}
+
+void DemoExt_DrawNothing(Actor* thisx, GlobalContext* globalCtx) {
+}
+
+void DemoExt_DrawVortex(Actor* thisx, GlobalContext* globalCtx) {
+    DemoExt* this = THIS;
+
+    Mtx* mtx;
```
fig02, line 186: no break

src/overlays/actors/ovl_Demo_Ext/z_demo_ext.c
```diff
@@ -1,17 +1,233 @@
...
+    GraphicsContext* gfxCtx;
+    s16* curScroll;
+    Vec3f* scales;
+
+    scales = &this->scales;
+    gfxCtx = globalCtx->state.gfxCtx;
+    mtx = Graph_Alloc(gfxCtx, 0x40);
+
```
fig02, line 194: sizeof(Mtx)

src/overlays/actors/ovl_Demo_Ext/z_demo_ext.c
```diff
@@ -1,17 +1,233 @@
...
+    gSPMatrix(POLY_XLU_DISP++, mtx, G_MTX_PUSH | G_MTX_LOAD | G_MTX_MODELVIEW);
+    gSPDisplayList(POLY_XLU_DISP++, D_0600FAA0);
+    gSPPopMatrix(POLY_XLU_DISP++, G_MTX_MODELVIEW);
+
+    CLOSE_DISPS(gfxCtx, "../z_demo_ext.c", 512);
+}
+
+void DemoExt_Draw(Actor* thisx, GlobalContext* globalCtx) {
+    static DemoExtDrawFunc sDrawFuncs[] = { DemoExt_DrawNothing, DemoExt_DrawVortex };
```
fig02, line 220: same comment here about array of func pointers

src/overlays/actors/ovl_Demo_Ext/z_demo_ext.c
```diff
@@ -1,17 +1,233 @@
...
+    DemoExt* this = THIS;
+
+    if ((this->drawMode < 0) || (this->drawMode >= 2) || sDrawFuncs[this->drawMode] == NULL) {
+        // Draw mode is abnormal!
```
fig02, line 224: draw enum

src/overlays/actors/ovl_Demo_Ext/z_demo_ext.c
```diff
@@ -22,40 +238,4 @@ const ActorInit Demo_Ext_InitVars = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Ext/func_809778AC.s")
-
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Ext/DemoExt_Update.s")
-
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Ext/func_80977944.s")
-
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Ext/func_80977950.s")
-
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Demo_Ext/DemoExt_Draw.s")
+};
```
fig02, line 241: make sure to run formatter, itll insert this newline for you

src/overlays/actors/ovl_Demo_Ext/z_demo_ext.c
EllipticEllipsis: I know this one is easy, but is there a way to tell when to use sizeof in general?

src/overlays/actors/ovl_Demo_Ext/z_demo_ext.c
EllipticEllipsis: Well, this actor does only appear in a cutscene. Is that not quite what you had in mind?

fig02: NpcAction is already a structure specifically for cutscenes, so I was noting that having `Cs` in the name isnt really neeeded

src/overlays/actors/ovl_Demo_Ext/z_demo_ext.c
fig02: generally if an allocation is occurring, its probably not a hardcoded size (though it does happen). you may see a larger number that is really `sizeof(thing) * amount`

src/overlays/actors/ovl_Demo_Ext/z_demo_ext.c
EllipticEllipsis: Ah, okay

fig02 (approved):

src/overlays/actors/ovl_Demo_Ext/z_demo_ext.h
```diff
@@ -6,9 +6,21 @@
...
+    /* 0x014C */ s32 action;
+    /* 0x014C */ s32 drawMode;
+    /* 0x0154 */ s32 previousCsAction;
+    /* 0x015E */ s16 scrollIncr[4];
+    /* 0x0160 */ s16 curScroll[4];
+    /* 0x0168 */ s16 rotationPitch;
+    /* 0x016C */ f32 alphaTimer;
+    /* 0x0170 */ s32 primAlpha;
+    /* 0x0174 */ s32 envAlpha;
+    /* 0x0178 */ Vec3f scales;
```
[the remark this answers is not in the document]
EllipticEllipsis: Ah, hadn't realised that was the convention. I'm used to scale being, well, a scalar

Roman971 (approved):

PR #587, opened 2020-12-31
Use subrepo instead of submodule for ZAPD (and some misc changes)

fig02:
- ZAPD code is now included in the repo. Check pinned message in `#oot-decomp-tools` for steps on how to update (only relevant for people pushing ZAPD updates to our upstream).
- Makefile updated to remove submodule init and changes ZAP2 to ZAPD
- `Hierarchy` -> `Skeleton`
- Leading zero removed from segment numbers in the spec, this causes issues according to zel
- Remove THANKS.md (the involved parties have agreed)


fig02:
gonna do this pr from scratch after zapd's git history is cleared

PR #593, opened 2021-01-02
Gerudo3 OK and documented

EllipticEllipsis:
This seems to be a gutted version of Ge2, that uses GeldB's object.

All functions named, most variables named (left two, one of which appears to be unused by the actor except as a required function argument, the other some vestigial flags), enums made for limbs.

Not surprisingly, there's not a lot to this NPC actor, since it appears once and does one thing.


petrie911 (commented):
Only a few minor comments. Also be on the lookout for the object_geldb merge to get rid of the segment addresses.

src/overlays/actors/ovl_En_Ge3/z_en_ge3.c
```diff
@@ -30,37 +57,250 @@ const ActorInit En_Ge3_InitVars = {
...
+    { 20, 50, 0, { 0, 0, 0 } },
+};
+
+extern FlexSkeletonHeader D_0600A458; // Same as GeldB
+extern AnimationHeader D_0600B07C; // Idle with right hand on hip and left over mouth
+
+void EnGe3_ChangeAction(EnGe3* this, s32 i) {
+    static EnGe3ActionFunc sActionFunc[] = { EnGe3_WaitLookAtPlayer };
+    static AnimationHeader* sAnimationHeader[] = { 0x0600B07C }; // Idle with right hand on hip and left over mouth
+    static u8 sAnimationMode[] = { ANIMMODE_LOOP };
```
Roman971, line 73: Since they are arrays, these should probably be named plural even tho they only have 1 entry in the base game.
And you can go with a simpler `sAnimations` if you want.
```suggestion
    static EnGe3ActionFunc sActionFuncs[] = { EnGe3_WaitLookAtPlayer };
    static AnimationHeader* sAnimations[] = { 0x0600B07C }; // Idle with right hand on hip and left over mouth
    static u8 sAnimationModes[] = { ANIMMODE_LOOP };
```

src/overlays/actors/ovl_En_Ge3/z_en_ge3.c
```diff
@@ -30,37 +57,250 @@ const ActorInit En_Ge3_InitVars = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Ge3/func_80A34E58.s")
+    if (limbIndex == GELDB_LIMB_HEAD) {
+        Matrix_MultVec3f(&D_80A351C8, &this->actor.posRot2.pos);
+    }
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Ge3/func_80A35004.s")
+void EnGe3_Draw(Actor* thisx, GlobalContext* globalCtx2) {
+    static u64* sEyeTextures[] = {
```
Roman971, line 291: I suggest leaving this array right above the function, unless it actually has to be in-function static to match. It looks okay right now, but it could become pretty verbose with the full symbol names. It's also nice to leave it available to other functions, if you wanted to add other draw modes while modding for instance.

In general I recommend to avoid in-function static arrays of functions/pointers for these reasons.

Roman971 (commented):
Well done with this actor, just a few suggestions

src/overlays/actors/ovl_En_Ge3/z_en_ge3.c
```diff
@@ -30,37 +57,250 @@ const ActorInit En_Ge3_InitVars = {
...
 };
-*/
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Ge3/func_80A34620.s")
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Ge3/EnGe3_Init.s")
+static ColliderCylinderInit sCylinderInit = {
+    { COLTYPE_UNK10, 0x00, 0x09, 0x39, 0x10, COLSHAPE_CYLINDER },
+    { 0x00, { 0x00000000, 0x00, 0x00 }, { 0x00000722, 0x00, 0x00 }, 0x00, 0x01, 0x01 },
+    { 20, 50, 0, { 0, 0, 0 } },
+};
```
fig02, line 65: in top of the other comments mentioned here, id merge master and get this updated

fig02 (approved):
updated and review changes made

fig02:
moved the limb enum to geldb
will wait for this to get merged before fixing gerudo 1 and 2

src/overlays/actors/ovl_En_GeldB/z_en_geldb.h
```diff
@@ -6,14 +6,42 @@
...
+    /* 16 */ GELDB_LIMB_SWORD_L,
+    /* 17 */ GELDB_LIMB_THIGH_L,
+    /* 18 */ GELDB_LIMB_SHIN_L,
+    /* 19 */ GELDB_LIMB_FOOT_L,
+    /* 20 */ GELDB_LIMB_THIGH_R,
+    /* 21 */ GELDB_LIMB_SHIN_R,
+    /* 22 */ GELDB_LIMB_FOOT_R,
+    /* 23 */ GELDB_LIMB_WAIST,
+    /* 24 */ GELDB_LIMB_MAX
+} EnGeldBLimbs;
```
Roman971, line 35:
```suggestion
} EnGeldBLimb;
```

src/overlays/actors/ovl_En_GeldB/z_en_geldb.h
```diff
@@ -6,14 +6,42 @@
...
 typedef void (*EnGeldBActionFunc)(struct EnGeldB*, GlobalContext*);
 
 typedef struct EnGeldB {
     /* 0x0000 */ Actor actor;
     /* 0x014C */ Vec3s bodyPartsPos[10];
     /* 0x0188 */ SkelAnime skelAnime;
-    /* 0x01CC */ Vec3s jointTable[24];
-    /* 0x025C */ Vec3s morphTable[24];
+    /* 0x01CC */ Vec3s jointTable[GELDB_LIMB_MAX];
+    /* 0x025C */ Vec3s morphTable[GELDB_LIMB_MAX];
```
Roman971, line 44: Make sure to use `GELDB_LIMB_MAX` in the skelanime init call in geldb (I might just be missing it tho)

src/overlays/actors/ovl_En_GeldB/z_en_geldb.h
```diff
@@ -6,14 +6,42 @@
...
+    /* 13 */ GELDB_LIMB_FOREARM_L,
+    /* 14 */ GELDB_LIMB_WRIST_L,
+    /* 15 */ GELDB_LIMB_HAND_L,
+    /* 16 */ GELDB_LIMB_SWORD_L,
+    /* 17 */ GELDB_LIMB_THIGH_L,
+    /* 18 */ GELDB_LIMB_SHIN_L,
+    /* 19 */ GELDB_LIMB_FOOT_L,
+    /* 20 */ GELDB_LIMB_THIGH_R,
+    /* 21 */ GELDB_LIMB_SHIN_R,
+    /* 22 */ GELDB_LIMB_FOOT_R,
```
Roman971, line 32: Pretty minor but in player I went with `_R_FOOT` and such instead of `_FOOT_R`, and we should probably be consistent in the naming for these

fig02: Any preference? I think direction on the end makes sense cause its more of like a single letter suffix. but i dont mind too much either way

Roman971: They seem grouped by direction so I think treating the direction as a prefix makes slightly more sense, but that's about it

fig02: fair enough, will change

fig02:
changes made

Roman971 (approved):

PR #596, opened 2021-01-03
Automatic object xml creation

petrie911:
yo dawg, I heard you like automatic creation of resources

This tool will take your overlay file, analyze it for symbols, and produce an xml for the object with automatically generated names. It also has options to replace all symbols in your source with those names and to update spec to use the new object file. It doesn't really support textures, though, since it has no way to know their size. You'll have to fix those manually.

Also fixed a couple things I ran across while testing this. Notably, removed GeldB's relocs.


mzxrules:
there's an haheniron in this pr

Zelllll:
> there's an z_bg_jya_haheniron in this pr

yeah he made some adjustments to that actor, probably should mention it in the pr though so people know

fig02 (approved):

Roman971 (approved):

PR #606, opened 2021-01-06
MacOS ZAPD fixes

ethteck:
makes the build not error
Also updates ZAPD to fix another bug with MacOS


Roman971:
I don't think this is a warning/error we want to suppress so ideally we should look for a way to avoid it entirely, if it's possible.
I looked at `Camera_CheckWater` and it looks we might be able to avoid the warning there by using elses instead of early returns, but I don't know if there are any other occurrences in the repo.

Roman971:
Also worth noting the plan is to have gcc warnings elevated as errors eventually, so whether this shows up as an error or warning won't make a difference by then. And we will have to look for ways to avoid as many warnings as we can, so looking at this one right now will save time later.

However it might be worth elevating warnings like this one that show up as errors on certain systems, so we don't accidentally break the support for these systems in the meantime. I think we can do this with `-Werror=return-type`

ethteck:
Makes sense. I'll give actually fixing the issue a shot

ethteck:
I was able to do the else like you said, and luckily this is the only function in the entire game where this was a problem.

I'm not sure how important the bug comment is at this point. Technically the function can finish without returing a value (when the end is reached), but this is warned about by GCC. I left the comment anyway, but let me know your thoughts.

fig02 (approved):

Roman971 (approved):

PR #621, opened 2021-01-11
EnTr, 1 nonmatching, mostly documented

EllipticEllipsis:
Eyhehehehehe...

Twinrova in the two cutscenes outside the boss room. All but one evil float function (which appears to literally be for floating) decompiled. Action functions and variables named, enums for Koume and Kotake. Other names will require understanding the cutscene cue system better than I do at present.

Some rather odd temp choices throughout, which probably is why I can't match the last function.


src/overlays/actors/ovl_En_Tr/z_en_tr.c
```diff
@@ -35,45 +35,522 @@ const ActorInit En_Tr_InitVars = {
...
+    if (1) {}
+
+    if ((globalCtx->csCtx.state == 0) || (globalCtx->csCtx.npcActions[this->actionIndex] == 0)) {
+        this->actor.shape.shadowDrawFunc = NULL;
+        return;
+    }
+
+    this->actor.shape.shadowDrawFunc = ActorShadow_DrawFunc_Circle;
+    OPEN_DISPS(globalCtx->state.gfxCtx, "../z_en_tr.c", 840);
+    func_800943C8(globalCtx->state.gfxCtx);
```
fig02, line 460: i would prob do an else here myself, but theres merit for both

fig02 (approved):

Roman971 (approved):

PR #622, opened 2021-01-11
z_lifemeter OK

Synray:
There were a few versions that matched, so the duplicate `ddfactor`
might be fake. The `if (interfaceCtx)` at the start seems to be required.

I think `sHeartsPrimFactors[0]` vs `sHeartsDDPrimFactors[ddtype]` is
real, because the asm loads `sHeartsDDPrimFactors` for each access,
while it only loads `sHeartsPrimFactors` once. It could have something
to do with `reduceixa` in uopt.


fig02:
nice
only small change id make is the dd temp vars can be made camel case, ``ddType` etc

fig02 (approved):

Roman971 (approved):

PR #631, opened 2021-01-14
DoorKiller OK

AdamKiddle:
I also made some changes in player and EnDoor (added comments to explain more about what the doors are doing, an enum for the different door types and an enum for the different door animation styles).

Sadly in player function `func_80839800` it casts all door actors to the EnDoor type in order to access two variables at 0x190 and 0x191. This is ugly and will cause some issues if ever anyone tries to move the variables around in DoorKiller's struct, but unfortunately that just seems to be how they implemented it.


src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -32,6 +32,13 @@ typedef struct {
 #define CHEST_ANIM_SHORT 0
 #define CHEST_ANIM_LONG 1
 
+typedef enum {
+    /* 0x00 */ HANDLE_ON_LEFT_ADULT,
+    /* 0x01 */ HANDLE_ON_LEFT_CHILD,
+    /* 0x02 */ HANDLE_ON_RIGHT_ADULT,
+    /* 0x03 */ HANDLE_ON_RIGHT_CHILD
+} DoorAnimStyle;
```
fig02, line 40: im unsure if animation style is the term we wanna go with here. can you go into detail on what that means?
we'd also want this enum to have a prefix reflecting what it is, whenever we agree on a term for it

AdamKiddle: specifically for doors with handles, link has 4 different animations he can do depending on which side the door handle is on and if he's adult or child. This is used to index which animation to play (it's common across EnDoor and DoorKiller)

I didn't like the word "style" either but I wasn't able to come up with anything better. Maybe doorHandleAnimation?

fig02: knobDoorAnim, knobbedDoorAnim, something along those lines could work, or handle if you like that better. def think dropping the word style and calling it an animation on its own is best either way

AdamKiddle: Was trying to avoid the word "Knobbed" cause it's a vulgar term in British slang lol

Would you call the enum entries then KNOBBED_DOOR_ANIM_LEFT_KNOB_ADULT etc? Seems a bit long

fig02: KNOB_ANIM_ADULT_R, L for left
prob can assume that knob will imply its for a door. just throwing out random ideas tho maybe others have suggestions

src/overlays/actors/ovl_Door_Killer/z_door_killer.c
```diff
@@ -1,16 +1,35 @@
+/*
+ * File: z_door_killer.c
+ * Overlay: ovl_Door_Killer
+ * Description: Fake doors which attack player
+ *
+ * Associated switch flags: (params >> 8) & 0x3F
+ * ((params >> 8) & 0x3F) == 0x3F means no switch flags are checked / set
+ */
```
fig02, line 8: you can put this near the relevant enums in the header (if there are any) since thats where people will go to find this information. at least until macros for it are created

src/overlays/actors/ovl_Door_Killer/z_door_killer.c
```diff
@@ -67,41 +86,438 @@ static ColliderJntSphInit D_80996000 = {
...
+    }
+    func_8002D7EC(&this->actor);
+}
+
+s32 DoorKiller_IsHit(Actor* thisx, GlobalContext* globalCtx) {
+    DoorKiller* this = THIS;
+    if ((this->colliderCylinder.base.acFlags & 2) && (this->colliderCylinder.info.acHitInfo != NULL)) {
+        return 1;
+    }
+    return 0;
```
fig02, line 234: can use true and false as returns if you like

src/overlays/actors/ovl_Door_Killer/z_door_killer.c
```diff
@@ -67,41 +86,438 @@ static ColliderJntSphInit D_80996000 = {
...
+void DoorKiller_Wait(DoorKiller* this, GlobalContext* globalCtx) {
+    Player* player;
+    Vec3f playerPosRelToDoor;
+    s16 angleToFacingPlayer;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Door_Killer/func_80995318.s")
+    player = PLAYER;
+    func_8002DBD0(&this->actor, &playerPosRelToDoor, &player->actor.world.pos);
+    // playerIsOpening is set from the player overlay when the player opens the door
+    if (this->playerIsOpening != 0) {
```
fig02, line 418: "from the player overlay" can be changed to "by player", the type of file he is isnt really relevant here imo
and if playerIsOpening  is a bool you can remove != 0 (and use true and false where relevant)

fig02 (commented):
other than the door style thing we talked about, just a couple of small things

src/overlays/actors/ovl_Door_Killer/z_door_killer.c
AdamKiddle: I'll move this to the header but I'm not sure I know how to write this kind of check in a macro without it being considered hiding too much info.

Would this be ok?
```
#define DOOR_KILLER_SWITCH_FLAG(params) (params >> 8) & 0x3F
...
if (DOOR_KILLER_SWITCH_FLAG(this->actor.params) != 0x3F) {
    Flags_SetSwitch(globalCtx, DOOR_KILLER_SWITCH_FLAG(this->actor.params))
}
```

fig02: oh no no I said *until* macros are created. we havent worked out exactly how we want to write our macros that unpack complex params just yet. So just including a comment in the header that explains how they are packed is fine for now

Zelllll (approved):
Looks good to me.

src/overlays/actors/ovl_En_Arrow/z_en_arrow.c
```diff
@@ -250,7 +250,6 @@ void EnArrow_Fly(EnArrow* this, GlobalContext* globalCtx) {
     Actor* hitActor;
     Vec3f sp60;
     Vec3f sp54;
-
```
Roman971, line 253: Were all these line breaks removed on purpose? I think they improve readability, so we should probably keep them

src/overlays/actors/ovl_Door_Killer/z_door_killer.h
```diff
@@ -4,11 +4,35 @@
 #include "ultra64.h"
 #include "global.h"
 
+/*
+ * Associated switch flag: (params >> 8) & 0x3F
+ * ((params >> 8) & 0x3F) == 0x3F means no switch flag is checked / set
+*/
+
+typedef struct{
```
Roman971, line 12:
```suggestion
typedef struct {
```

src/overlays/actors/ovl_Door_Killer/z_door_killer.h
```diff
@@ -4,11 +4,35 @@
...
+    /* 0x00 */ s16 objectId;
+    /* 0x04 */ Gfx* texture;
```
Roman971, line 14: A texture can't be of type `Gfx*` so there is something wrong here.
If it can be any pointer type, I suggest using `void*`.

src/overlays/actors/ovl_Door_Killer/z_door_killer.h
```diff
@@ -4,11 +4,35 @@
...
+
 typedef struct DoorKiller {
     /* 0x0000 */ Actor actor;
-    /* 0x014C */ char unk_14C[0x138];
+    /* 0x014C */ SkelAnime skelAnime;
+    /* 0x0190 */ u8 animStyle; // Must be at same offset as animStyle in EnDoor due to the cast in func_80839800
+    /* 0x0191 */ u8 playerIsOpening; // Must be at same offset as playerIsOpening in EnDoor due to the cast in func_80839800
+    /* 0x0192 */ Vec3s jointTable[9];
+    /* 0x01C8 */ ColliderCylinder colliderCylinder;
+    /* 0x0214 */ Gfx* texture;
```
Roman971, line 28: Same issue here

src/overlays/actors/ovl_Door_Killer/z_door_killer.c
```diff
@@ -1,17 +1,33 @@
+/*
+ * File: z_door_killer.c
+ * Overlay: ovl_Door_Killer
+ * Description: Fake doors which attack player
+ */
```
Roman971, line 5:
```suggestion
 */

```

Roman971 (commented):
Sorry for the late review. Just a few things

src/overlays/actors/ovl_Door_Killer/z_door_killer.c
```diff
@@ -68,41 +84,438 @@ static ColliderJntSphInit D_80996000 = {
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Door_Killer/func_80995318.s")
+    player = PLAYER;
+    func_8002DBD0(&this->actor, &playerPosRelToDoor, &player->actor.world.pos);
+    // playerIsOpening is set by player
+    if (this->playerIsOpening != 0) {
+        this->actionFunc = DoorKiller_WaitBeforeWobble;
+        this->timer = 10;
+        this->playerIsOpening = 0;
+        return;
```
petrie911, line 420: This early return and others like it should be `else` blocks.

src/overlays/actors/ovl_Door_Killer/z_door_killer.c
```diff
@@ -68,41 +84,438 @@ static ColliderJntSphInit D_80996000 = {
...
+        this->hasHitPlayerOrGround &= ~1;
+        return;
+    }
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Door_Killer/func_809951C4.s")
+    rotation = Math_SinS(this->timer * 0x2000) * this->timer * 100.0f;
+    for (i = 2; i < 9; i++) {
+        this->jointTable[i].y = rotation;
+    }
+    rotation = (u16)(s16)(-Math_CosS(this->timer * 0x1000) * 1000.0f) + 1000;
```
petrie911, line 390: Can the `s16` cast be `s32`?

src/overlays/actors/ovl_Door_Killer/z_door_killer.c
```diff
@@ -68,41 +84,438 @@ static ColliderJntSphInit D_80996000 = {
...
+        this->timer--;
+    } else {
+        this->timer = 16;
+        this->actionFunc = DoorKiller_Wobble;
+    }
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Door_Killer/func_809952B8.s")
+void DoorKiller_Wait(DoorKiller* this, GlobalContext* globalCtx) {
+    Player* player;
```
petrie911, line 409:
```suggestion
    Player* player = PLAYER;
```
should be fine to put this in the declaration.

src/overlays/actors/ovl_Door_Killer/z_door_killer.c
```diff
@@ -68,41 +84,438 @@ static ColliderJntSphInit D_80996000 = {
...
+        this->jointTable[i].z = -rotation;
+    }
+
+    if (this->timer < 8) {
+        rotation = Math_SinS(this->timer * 0x2000) * this->timer * 100.0f;
+        for (i = 2; i < 9; i++) {
+            this->jointTable[i].y = rotation;
+        }
+    }
+    return;
```
petrie911, line 289: this return is unnecessary.

src/overlays/actors/ovl_Door_Killer/z_door_killer.c
```diff
@@ -68,41 +84,438 @@ static ColliderJntSphInit D_80996000 = {
...
+        this->timer--;
+    } else {
+        Actor_Kill(&this->actor);
+    }
+    func_8002D7EC(&this->actor);
+}
+
+s32 DoorKiller_IsHit(Actor* thisx, GlobalContext* globalCtx) {
+    DoorKiller* this = THIS;
+    if ((this->colliderCylinder.base.acFlags & 2) && (this->colliderCylinder.info.acHitInfo != NULL)) {
```
petrie911, line 229:
```suggestion
    if ((this->colliderCylinder.base.acFlags & AC_HIT) && (this->colliderCylinder.info.acHitInfo != NULL)) {
```

src/overlays/actors/ovl_Door_Killer/z_door_killer.c
```diff
@@ -68,41 +84,438 @@ static ColliderJntSphInit D_80996000 = {
...
+    this->texture = sDoorTextures[this->textureEntryIndex].texture;
+
+    ActorShape_Init(&this->actor.shape, 0.0f, NULL, 0.0f);
+    Actor_SetScale(&this->actor, 0.01f);
+    this->timer = 0;
+    this->hasHitPlayerOrGround = 0;
+    this->animStyle = 0;
+    this->playerIsOpening = 0;
+
+    switch ((u8)(this->actor.params & 0xFF)) {
```
petrie911, line 122: is the (u8) cast needed to match?

src/overlays/actors/ovl_Door_Killer/z_door_killer.c
```diff
@@ -68,41 +84,438 @@ static ColliderJntSphInit D_80996000 = {
...
+void DoorKiller_Wait(DoorKiller* this, GlobalContext* globalCtx) {
+    Player* player;
+    Vec3f playerPosRelToDoor;
+    s16 angleToFacingPlayer;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Door_Killer/func_80995318.s")
+    player = PLAYER;
+    func_8002DBD0(&this->actor, &playerPosRelToDoor, &player->actor.world.pos);
+    // playerIsOpening is set by player
+    if (this->playerIsOpening != 0) {
```
petrie911, line 416:
```suggestion
    if (this->playerIsOpening) {
```
this is a boolean, right?

src/overlays/actors/ovl_Door_Killer/z_door_killer.c
```diff
@@ -68,41 +84,438 @@ static ColliderJntSphInit D_80996000 = {
...
+    DoorKiller* this = THIS;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Door_Killer/func_80995A84.s")
+    this->actionFunc(this, globalCtx);
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Door_Killer/func_80995CDC.s")
+void DoorKiller_SetTexture(Actor* thisx, GlobalContext* globalCtx) {
+    DoorKiller* this = THIS;
+    s32 doorTexture = this->texture;
```
petrie911, line 496:
```suggestion
    u64* doorTexture = this->texture;
```
textures all have the type `u64*`

src/overlays/actors/ovl_Door_Killer/z_door_killer.c
```diff
@@ -68,41 +84,438 @@ static ColliderJntSphInit D_80996000 = {
...
+    CLOSE_DISPS(globalCtx->state.gfxCtx, "../z_door_killer.c", 885);
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Door_Killer/DoorKiller_Update.s")
+void DoorKiller_DrawDoor(Actor* thisx, GlobalContext* globalCtx) {
+    DoorKiller* this = THIS;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Door_Killer/func_80995E40.s")
+    func_800943C8(globalCtx->state.gfxCtx);
+    DoorKiller_SetTexture(thisx, globalCtx);
```
petrie911, line 507: `thisx` needed for matching?

src/overlays/actors/ovl_Door_Killer/z_door_killer.c
fig02: yes

src/overlays/actors/ovl_Door_Killer/z_door_killer.c
fig02: he uses them for action changes so it seems fine to me

fig02 (approved):
review changes made

Roman971 (approved):
We should probably default to `void*` for all textures when a specific pointer type isn't needed, but that's something we can fix globally in a later pass

PR #633, opened 2021-01-15
TwoHeadArena Cleanup

AdamKiddle:
Lmk if you don't want me pushing small individual cleanups like this, I don't want to waste reviewers time


Roman971 (approved):
Doing isolated changes like this is great, actually

fig02 (approved):

PR #648, opened 2021-01-19
En_Arrow OK

fig02:


src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -1968,9 +1969,9 @@ s32 func_8083442C(Player* this, GlobalContext* globalCtx) {
                 magicArrowType = arrowType - 3;
 
                 if (this->unk_860 >= 0) {
-                    if ((magicArrowType >= 0) && (magicArrowType < 3) &&
+                    if ((magicArrowType >= ARROW_NORMAL_LIT) && (magicArrowType <= ARROW_NORMAL) &&
```
Roman971, line 1972: magicArrowType is arrowType - 3 so these enum values seem incorrect. Also I think this check is supposed to convert magic arrows to normal arrows if you're out of magic

fig02: ah, yea thats true. do we want to make an enum specifically for magic arrow types or something? or do the arrow type enum - 3? lol

Roman971: To avoid adding yet another enum that depends on the others, I would do `enum - the enum value for 3`, with `magicArrowType = arrowType - the enum value for 3` so they don't go out of sync

fig02: so
`magicArrowType = arrowType - ARROW_FIRE;`
then
`if ((magicArrowType >= ARROW_FIRE - ARROW_FIRE) && (magicArrowType <= ARROW_LIGHT - ARROW_FIRE)`

seems hella confusing to read at first glance

Roman971: Either that or keep them as 0 < x <= 2 imo, which is also fine tbh

Roman971: But I would definitely say to keep the `magicArrowType = arrowType  - ARROW_FIRE`

fig02: if doing the weird subtraction means the code wont break when changing the enum then i guess it would be better to go with that

Roman971: Now that I think about it, the rest of this code will break if you added/removed a magic arrow type without making some changes (eg. adding/removing entries in `sMagicArrowCosts`), so I think the `0 < x <= 2` is the right way to handle the check since it's really straightforward. This solution would also not break if you changed the rest of the arrow type enum.

fig02: yea makes sense. pushed changes

Roman971 (approved):

Thar0 (approved):

Zelllll (approved):
Looks good

PR #656, opened 2021-01-24
Fix most actor compiler warnings

EllipticEllipsis:
Fixes as many compiler warnings as possible in actors, including Player. This mainly amounts to typing stuff correctly, using correct pointers, adding a few functions to `functions.h`, etc. 

Several actors have issues that I have not been able to fix, or already have fixes due in other PRs (notably #595 ). These include

- Obj_Elevator
- Demo_Go
- En_Test
- En_Viewer
- Bg_Mori_Rakkatenjo
- En_Guest
- Effect_Ss_Solder_Srch_Ball
- Bg_Spot09_Obj
- Bg_Ydan_Hasi
- Anything with a cutscene data file that requires `EARLY` (9 files)
- Anything that produces `Error 1 (ignored)` (21 files)

Fixes to the bottom of the GetItem table in Player are bit of a hack, but do not produce warnings. Filling the othrer fields with `0`s is also an option, but I certainly don't think they should just be left as `{ ITEM_NONE }`.

Also did some cleanup on some actors: bringing `thisx` use up to current convention, translating some of the printfs, adding descriptive headers, naming obvious variables like timers, etc. This has not been done consistently or to be much in the way of documentation, but just because I happened to be in the file; the sheer number of actors I have edited in the past few days inevitably means I have not been particularly consistent or comprehensive in these ancillary aspects.

Not addressed: errors arising from code, gamestates, and basically anything that is not directly related to actors. I have edited `z_actor.c`, but only to correct things that wouldn't compile otherwise. I leave these areas to someone who is more familiar with code decomp than me.

As a measure of progress, this reduces the length of the output from `make` by slightly over 3000 lines, and reduces the number of warnings by approximately 1/3 (1556 to 1039).


src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
```diff
@@ -41,7 +41,7 @@ typedef enum {
     STATE_GO_BOTTOM,
     STATE_GO_MIDDLE_FROM_BOTTOM,
     STATE_GO_MIDDLE_FROM_TOP,
-    STATE_GO_TOP,
+    STATE_GO_TOP
```
mzxrules, line 44: The trailing comment is necessary to prevent clang from formatting this enumeration like so: 

`typedef enum { STATE_GO_BOTTOM, STATE_GO_MIDDLE_FROM_BOTTOM, STATE_GO_MIDDLE_FROM_TOP, STATE_GO_TOP } BgDdanJdState;`

src/overlays/actors/ovl_Bg_Hidan_Kowarerukabe/z_bg_hidan_kowarerukabe.c
```diff
@@ -201,6 +201,7 @@ void func_8088A67C(BgHidanKowarerukabe* this, GlobalContext* globalCtx) {
     Vec3f velocity;
     Vec3f pos;
     s16 arg5;
+    // thisx is not required for matching in this function, but is in BgHidanKowarerukabe_FloorBreak, so done for consistency
```
Roman971, line 204: Imo we don't need these comments. Using `thisx` is technically allowed to simplify accesses, so it should just be treated as a normal thing imo.

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -607,8 +607,10 @@ GetItemEntry sGetItemTable[] = {
     GET_ITEM(ITEM_NUT_UPGRADE_30, OBJECT_GI_NUTS, 0x11, 0xA7, 0x80, CHEST_ANIM_SHORT),
     GET_ITEM(ITEM_NUT_UPGRADE_40, OBJECT_GI_NUTS, 0x11, 0xA8, 0x80, CHEST_ANIM_SHORT),
     GET_ITEM(ITEM_BULLET_BAG_50, OBJECT_GI_DEKUPOUCH, 0x72, 0x6C, 0x80, CHEST_ANIM_LONG),
-    { ITEM_NONE },
-    { ITEM_NONE },
+    GET_ITEM(ITEM_NONE, OBJECT_UNSET_0, -1, 0x00, 0x00, CHEST_ANIM_SHORT),
+    GET_ITEM(ITEM_NONE, OBJECT_UNSET_0, -1, 0x00, 0x00, CHEST_ANIM_SHORT),
+    // { ITEM_NONE, 0, 0, 0, 0 },
+    // { ITEM_NONE, 0, 0, 0, 0 },
```
Roman971, line 613: `{ ITEM_NONE },` is the best option here imo, but that's assuming we ignore the warning (which we arguably should). The other decent options are to have a `GET_ITEM_NONE` macro for this purpose or to fill the remaining fields with 0, but I really don't think using the normal `GET_ITEM` macro here is good.

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -10692,7 +10694,7 @@ void func_8084B78C(Player* this, GlobalContext* globalCtx) {
 
 void func_8084B840(GlobalContext* globalCtx, Player* this, f32 arg2) {
     if (this->actor.wallBgId != BGCHECK_SCENE) {
-        DynaPolyActor* dynaActor = DynaPoly_GetActor(&globalCtx->colCtx, this->actor.wallBgId);
+        DynaPolyActor* dynaActor = (DynaPolyActor*)DynaPoly_GetActor(&globalCtx->colCtx, this->actor.wallBgId);
```
Roman971, line 10697: It's not a change from this PR, but why does `DynaPoly_GetActor` no longer return a `DynaPolyActor*`? Considering what the function does, I would expect it to return a `DynaPolyActor*`, not an `Actor*`, and it would keep things much easier.

src/overlays/actors/ovl_En_Zl1/z_en_zl1.c
```diff
@@ -58,12 +58,12 @@ static ColliderCylinderInit sCylinderInit = {
...
-static UNK_PTR D_80B4E61C[] = { 0x06007208, 0x06009848, 0x06009C48, 0x06009848 };
-static UNK_PTR D_80B4E62C[] = { 0x06007608 };
+static u64* D_80B4E61C[] = { 0x06007208, 0x06009848, 0x06009C48, 0x06009848 };
+static u64* D_80B4E62C[] = { 0x06007608 };
 
 extern AnimationHeader D_06000438;
-extern UNK_TYPE D_06008848;
-extern UNK_TYPE D_06008C48;
+extern u64* D_06008848;
+extern u64* D_06008C48;
```
Roman971, line 66:
```suggestion
extern u64 D_06008848[];
extern u64 D_06008C48[];
```

src/overlays/actors/ovl_Demo_Go/z_demo_go.c
```diff
@@ -52,10 +52,10 @@ const ActorInit Demo_Go_InitVars = {
 
 extern AnimationHeader D_060029A8;
 extern AnimationHeader D_06004930;
-extern UNK_TYPE D_0600E680;
+extern u64* D_0600E680;
```
Roman971, line 55:
```suggestion
extern u64 D_0600E680[];
```

src/overlays/actors/ovl_Bg_Jya_Bombiwa/z_bg_jya_bombiwa.c
```diff
@@ -1,3 +1,8 @@
+/*
+ * File: z_bg_jya_bombiwa.c
+ * Overlay: ovl_Bg_Jya_Bombiwa
+ * Description: Spirit Temple top room bombable wall
+ */
```
Roman971, line 5: Pretty minor but there should be a break between the file comment and the includes, ideally
```suggestion
 */

```

src/overlays/actors/ovl_player_actor/z_player.c
fig02: yea this is something that will be sorted with bgcheck docs

src/overlays/actors/ovl_Bg_Ddan_Jd/z_bg_ddan_jd.c
fig02: trailing comma actually induces a warning on IDO
putting offset comments (which we should have anyway) will prevent this formatting

src/overlays/actors/ovl_Bg_Hidan_Kowarerukabe/z_bg_hidan_kowarerukabe.c
EllipticEllipsis: Okay, so leave as thisx, remove comment?

fig02: yeah

src/overlays/actors/ovl_player_actor/z_player.c
EllipticEllipsis: I don't see the point of having a warning on something that can be resolved simply by padding to agree with the array definition. I don't mind how this is done, but I don't think a warning should be preferred to such an easy fix.

src/overlays/actors/ovl_player_actor/z_player.c
EllipticEllipsis: Okay, what's the best thing to do in the meantime, to be consistent? This function is used in 12 files, 7 of which fall under the purview of this PR, so it would be easy enough to change all of them.

fig02: dont cast and leave the warnings until we sort out the functions return type proper with bgcheck docs

src/overlays/actors/ovl_player_actor/z_player.c
fig02: a `GET_ITEM_NONE` macro sounds like a good compromise here. agree with roman though that filling in the normal macro does not work. `OBJECT_UNSET_0` and `CHEST_ANIM_SHORT` dont make sense in this context

EllipticEllipsis: So something like `#define GET_ITEM_NONE { ITEM_NONE, 0, 0, 0, 0 }` after the `CHEST_ANIM_LONG` one at the top?

fig02 (approved):
looks good to me

i didnt check but im sure the recent merges of the last few days prob had some warnings. If you want to merge master and get those cleaned up too (or maybe you already have) then let us know

EllipticEllipsis:
All up to date with latest merges now. Only extra issues introduced are in `ovl_En_Fr`, but I'm pretty sure those are required to match.

src/overlays/actors/ovl_Bg_Jya_Bombiwa/z_bg_jya_bombiwa.c
```diff
@@ -76,30 +82,32 @@ void BgJyaBombiwa_SetupDynaPoly(BgJyaBombiwa* this, GlobalContext* globalCtx, vo
...
+
+    if ((this->dyna.actor.params & 0x3F) != 0x29) {
         osSyncPrintf(VT_COL(YELLOW, BLACK));
 
         // Warning: Switch Number changed (%s %d)(SW %d
         osSyncPrintf("Ｗａｒｎｉｎｇ : Switch Number が変更された(%s %d)(SW %d)\n", "../z_bg_jya_bombiwa.c", 218,
-                     thisx->params & 0x3F);
-        osSyncPrintf(VT_SGR());
+                     this->dyna.actor.params & 0x3F);
+        osSyncPrintf(VT_SGR(""));
```
Roman971, line 100:
```suggestion
        osSyncPrintf(VT_RST);
```

Roman971 (approved):

PR #658, opened 2021-01-24
Some static texture files "decompiled"

shawlucas:
~~UPDATE:~~
~~PLEASE DONT MERGE THIS YET. I NOTICED A FEW PROBLEMS THAT I NEED TO FIX~~

One thing i need to fix is the hardcoded pointers. Haven't replaced those all yet.

So I "decompiled":
do_action_static
item_name_static
map_grand_static
map_name_static
message_static
message_texture_static
nes_font_static
map_i_static

1/26/21:
Added nintendo_rogo_static
Added title_static

1/28/21:
Added segment= to the texture files

I tried to give appropriate names to every texture. In some files (specifically item_name_static), names got a bit confusing because there are duplicates of a lot of unused japanese textures.

I was just doing this for fun on my free time, but got pretty far into it so I thought I would share and see if its any use


assets/xml/textures/nes_font_static.xml
```diff
@@ -0,0 +1,143 @@
...
+        <Texture Name="gMsgCharYenSign" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharRightSquareBracket" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharCaret" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharUnderscore" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharGraveAccent" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharLowercase_a" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharLowercase_b" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharLowercase_c" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharLowercase_d" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharLowercase_e" Format="i4" Width="16" Height="16"/>
```
fig02, line 72: I see why you used an underscore, but im wondering if theres a way to do it without breaking our naming style. Other than that these seem fine to me

fig02: oh actually, all of these need an OutName so that the pngs have snake_case names without all the extra variable name stuff. can look at gameplay keep as an example

assets/xml/textures/do_action_static.xml
```diff
@@ -0,0 +1,91 @@
...
+        <Texture Name="gGrabActionLabelTexGER" OutName="grab_action_label_ger" Format="ia4" Width="48" Height="16"/>
+        <Texture Name="gStopActionLabelTexGER" OutName="stop_action_label_ger" Format="ia4" Width="48" Height="16"/>
+        <Texture Name="gPutAwayActionLabelTexGER" OutName="put_away_action_label_ger" Format="ia4" Width="48" Height="16"/>
+        <Texture Name="gReelActionLabelTexGER" OutName="reel_action_label_ger" Format="ia4" Width="48" Height="16"/>
+        <Texture Name="g1ActionLabelTexGER" OutName="1_action_label_ger" Format="ia4" Width="48" Height="16"/>
+        <Texture Name="g2ActionLabelTexGER" OutName="2_action_label_ger" Format="ia4" Width="48" Height="16"/>
+        <Texture Name="g3ActionLabelTexGER" OutName="3_action_label_ger" Format="ia4" Width="48" Height="16"/>
+        <Texture Name="g4ActionLabelTexGER" OutName="4_action_label_ger" Format="ia4" Width="48" Height="16"/>
+        <Texture Name="g5ActionLabelTexGER" OutName="5_action_label_ger" Format="ia4" Width="48" Height="16"/>
+        <Texture Name="g6ActionLabelTexGER" OutName="6_action_label_ger" Format="ia4" Width="48" Height="16"/>
```
fig02, line 58: due to the way the indexing of these textures work, the translations are *guaranteed* to be in the same order as their english counterparts. so you should be able to name the german and french ones now. would just use the english word with the language suffix at the end

fig02: oh i see, you have some unnamed in english as well. Any reason why, id expect you to be able to read it in a texture editor

fig02: are these the diving numbers actually?
eitherway i think we can call all of these like `gDoAction<Name><Lang>Tex`, since we call these "do actions" in code
and to make this more clear we can call the diving numbers Num1 or something

fig02 (approved):

spec
```diff
@@ -188,37 +188,43 @@ endseg
...
+    include "build/assets/textures/map_name_static/map_name_static.o"
+    number 11
 endseg
 
 beginseg
     name "do_action_static"
     romalign 0x1000
-    include "build/baserom/do_action_static.o"
+    include "build/assets/textures/do_action_static/do_action_static.o"
+    number 07
```
Roman971, line 206: I don't think we include a leading zero in those anymore, so it should be cleaned up ideally (feel free to remove them from other numbers if there are any)
```suggestion
    number 7
```

Roman971 (approved):

PR #665, opened 2021-01-27
gameplay_dangeon_keep

fig02:


Zelllll:
Did you replace every hardcoded pointer? I'd kind of expect there to be more 🤔

fig02:
@Zelllll dungeon keep isnt all that common. I replaced all that I could find.
keep in mind that 05 can also mean field keep depending on the context, so not all of those are gone yet

fig02:
changes made

Roman971 (approved):

PR #666, opened 2021-01-27
Update regconvert.py

AngheloAlf:
- `regconvert.py` expects a `gGameInfo->data[]` index instead of a raw offset of `gGameInfo`.
- The argument is assume to be a decimal number instead of an hexadecimal.
  - `mips_to_c` outputs the index in decimal (at least for me), so i thought this should be changed accordingly.
  - If the argument starts with `0x` or  `0X`, then the number is treated as an hexadecimal.
  - Use the flag `--hex` to make the script interpret the argument as an hexadecimal even if it doesn't starts with `0x`.
- The flag `--offset` will make the script expect a raw offset.
  - Mix it with `--hex` to have the old behavior.
- Use the flag `--file` with a path to a file to make the script change every occurrence of `gGameInfo->data[number]` to their corresponding `REG` macro in that file.
  - It can be mixed with the flag `--stdout` to print the processed to the terminal instead of overwritting the file.
- The script now checks if the index is in the valid range instead of throwing an exception.

Feedback is welcome.


fig02 (approved):

tools/regconvert.py
```diff
@@ -11,13 +12,105 @@ def get_reg_macro(offset):
     reg = (offset - DATA_OFFSET) // 2
     group = reg // GROUP_SIZE
     reg_in_group = reg % GROUP_SIZE
-    return "%cREG(%d)\n" % (REGISTER_NAMES[group], reg_in_group)
+    return "%cREG(%d)" % (REGISTER_NAMES[group], reg_in_group)
+
+def parse_number(string, force_hex):
+    if string.lower().startswith("0x") or force_hex:
+        return int(string, 16)
+    return int(string, 10)
```
ethteck, line 20: If the hex string doesn't start with 0x and the user doesn't specify the hex arg, it'd be nice to still support automatically parsing a string like `A234`. You can change the second arg of `int()` to 0 and simplify it like this:

```
if force_hex:
    return int(string, 16)
return int(string, 0)
```

tools/regconvert.py
```diff
@@ -11,13 +12,105 @@ def get_reg_macro(offset):
...
+    parsed_contents += file_contents
+    return parsed_contents
+
+def write_file(filename, contents):
+    dst_file = open(filename, "w")
+    dst_file.write(contents)
+    dst_file.close()
+
+def check_valid_index(index):
+    if index < 0x0:
```
ethteck, line 55: I think it would be clearer to just say `return index >= 0 and index <= x` for both of these, but not a big deal

ethteck: actually, I think you can do `0 <= index <= x`

ethteck: and at that point, you can probably just not make these functions. but up to you

tools/regconvert.py
```diff
@@ -11,13 +12,105 @@ def get_reg_macro(offset):
...
+        parsed_contents += get_reg_macro(offset)
+
+        file_contents = file_contents[end:]
+        match = pattern.search(file_contents)
+    
+    parsed_contents += file_contents
+    return parsed_contents
+
+def write_file(filename, contents):
+    dst_file = open(filename, "w")
```
ethteck, line 50: I think with file stuff it's cleaner to wrap the file opening in the with block so it gets automatically closed at the end of the block:

```
with open(filename, "w") as f:
    f.write(contents)
```

ethteck (commented):
Nice changes! 

I commented some minor style things that aren't that important and a hopeful improvement about making parsing more flexible.

tools/regconvert.py
AngheloAlf: I made this change, but I kept the functions because i think it is clearer this way what those numbers means.

AngheloAlf:
@ethteck I made the changes you requested :D

ethteck:
thanks! :)

ethteck (approved):

fig02:
wait actually, does mips2c properly output an array access for this now? Ive always known it to show just an offset, but if that changed recently and I havent seen it yet, then makes sense

ethteck:
yeah it did change in a recent m2c commit

Roman971 (approved):

PR #667, opened 2021-01-27
Fixed improper usage of color combiner macros, replaced hardcoded gSPMatrix values with gbi macros

Mr-Wiseguy:
Not every color combiner mux value can be used for each color combiner input, so this PR fixes that by making separate lookup tables for each CC input. This also replaces hardcoded matrix operations in gSPMatrix with gbi macros, such as G_MTX_LOAD, G_MTX_MODELVIEW, etc.


Mr-Wiseguy:
PR being moved to ZAPD repo.

PR #668, opened 2021-01-27
All* Get Item objects decompiled

Thar0:
Also names functions in `z_draw.c` and creates an enum for the draw ids.

\* `object_st` was not decompiled as while it has a get item model for skulltula tokens, it also has a lot of other things that seem more appropriate to be done after `En_Sw`, the last of 3 actors that depend on this object, is decompiled.

Naming suggestions are most welcome.


extract_single_asset.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/python3
```
Roman971, line 1: Was this file intended to be comitted?

include/z64item.h
```diff
@@ -377,6 +377,127 @@ typedef enum {
...
+    /* 0x6D */ DI_RUPEE_BLUE,
+    /* 0x6E */ DI_RUPEE_RED,
+    /* 0x6F */ DI_BIG_POE,
+    /* 0x70 */ DI_RUPEE_PURPLE,
+    /* 0x71 */ DI_RUPEE_GOLD,
+    /* 0x72 */ DI_BULLET_BAG_50,
+    /* 0x73 */ DI_SWORD_KOKIRI,
+    /* 0x74 */ DI_SKULL_TOKEN_2,
+    /* 0x75 */ DI_MAX
+} GetItemDrawID;
```
Roman971, line 499: I think it could be better to go with `GID_` for the prefix in this enum, for "Get Item Draw"

src/overlays/actors/ovl_En_Ex_Item/z_en_ex_item.h
```diff
@@ -15,7 +15,7 @@ typedef struct EnExItem {
     /* 0x0150 */ s16 getItemObjId;
     /* 0x0152 */ s16 type;
     /* 0x0152 */ s16 unusedParam;
-    /* 0x0156 */ s16 drawItemId;
+    /* 0x0156 */ s16 drawId;
```
Roman971, line 18: Imo, `drawId` is too general outside of the context of get items, so stuff like actor struct fields should probably include `getItem` (or `gi` if we want to keep it short)

src/overlays/actors/ovl_En_GirlA/z_en_girla.c
```diff
@@ -100,7 +100,7 @@ static u16 D_80A3C668[5] = { 0x70B6, 0x70B5, 0x70B4, 0x70B7, 0x70BB };
 
 typedef struct {
     /* 0x00 */ s16 objID;
-    /* 0x02 */ s16 unk_02;
+    /* 0x02 */ s16 drawId;
```
Roman971, line 103: Same thing in this actor imo

include/z64item.h
```diff
@@ -377,6 +377,127 @@ typedef enum {
...
 
+typedef enum {
+    /* 0x00 */ DI_BOTTLE,
+    /* 0x01 */ DI_KEY_SMALL,
+    /* 0x02 */ DI_SONG_MINUET,      // unused
+    /* 0x03 */ DI_SONG_BOLERO,      // unused
+    /* 0x04 */ DI_SONG_SERENADE,    // unused
+    /* 0x05 */ DI_SONG_REQUIEM,     // unused
+    /* 0x06 */ DI_SONG_NOCTURNE,    // unused
+    /* 0x07 */ DI_SONG_PRELUDE,     // unused
```
Roman971, line 388: I personally think the unused comments are unnecessary, and they could be misunderstood as "not implemented"

Roman971 (commented):
Haven't reviewed everything yet, but here are my suggestions for now

extract_single_asset.py
Thar0: I thought it might be useful for others doing object decomp for it to be in the repo so kept it in the commit but I would be happy to exclude it.

fig02: ideally it would be an option passed to the current assets script. we've all just been too lazy to do it lol

Thar0: I've made it an option in extract_assets instead, hopefully that's fine.

fig02 (approved):

Roman971 (approved):

PR #671, opened 2021-01-31
Various README updates

Roman971:
- Changed the shiftability warning to a warning about the codebase being subject to change at any time.
- Moved the Cygwin instructions down to be next to Docker instructions, since it's also an alternative (tho I moved it before Docker in terms of the alternatives we suggest, since it's probably easier to setup and use for most users)
- Improved the wording of some sections
- Adjusted some text to be bold (so it stands out more)

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,9 +3,9 @@
 ```diff
 - WARNING! -
 
-The ROM this repository builds cannot be 'shifted', primarily due to some hardcoded pointers which have yet
-to be dumped. Thus this repository is currently in an experimental and research phase and cannot yet be used
-traditionally as a source code base for general changes.
+This repository is a work in progress, and while it can be used to make certain changes, it's still
+constantly evolving. If you use it for modding purposes in its current state, please be aware that
+the codebase can drastically change at any time.
 ```
 
 This is a WIP decompilation of The Legend of Zelda: Ocarina of Time.
@@ -16,12 +16,11 @@ It builds the following ROM:
 
 * zelda_ocarina_mq_dbg.z64 `md5: f0b7f35375f9cc8ca1b2d59d78e35405`
 
-**Note:** This repository does not include any of the assets necessary to build the ROM.
-A prior copy of the game is required to extract the needed assets.
+**Note: This repository does not include any of the assets necessary to build the ROM. A prior copy of the game is required to extract the needed assets.**
 
-Website: <https://zelda64.dev/>
+**Website:** <https://zelda64.dev>
 
-Discord: <https://discord.zelda64.dev>
+**Discord:** <https://discord.zelda64.dev>
 
 ## Installation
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -3,9 +3,9 @@
 ```diff
 - WARNING! -
 
-The ROM this repository builds cannot be 'shifted', primarily due to some hardcoded pointers which have yet
-to be dumped. Thus this repository is currently in an experimental and research phase and cannot yet be used
-traditionally as a source code base for general changes.
+This repository is a work in progress, and while it can be used to make certain changes, it's still
```
Zelllll, line 6: I'm not entirely sure if changing this is the best idea right now, considering how many things are not actually shiftable, like any struct still being used in assembly, hardcoded pointers in objects that haven't been decompiled, etc.

louist103, line 6: Id probably mention something about partial shiftability. IE any C file can be shifted but anything in ASM, Structs, and non-decompiled objects can not.

fig02: This is probably a good idea yeah, we dont need to list specific examples but just let people be aware of some things that may not be shiftable

Roman971: I feel like it should be obvious that not everything can be done, and it will remain the case for a very long time, but I guess we can leave a small mention about it for now

README.md
```diff
@@ -149,13 +136,36 @@ md5sum: WARNING: 1 computed checksum did NOT match
...
+Then you can continue from step 3 of the Linux instructions.
+
+Note that, before building anything, you will need to run the following commands to fix line endings:
+
+```bash
+dos2unix fixle.sh
+./fixle.sh
+```
+
 ### Docker
```
jpburnett, line 162: I think there was a discussion to remove docker because no one was using or docker and it would quickly become outdated due to lack of support for it, so this entire section could just be removed?

fig02: we ended up deciding to keep it around at least while it still works. but if it becomes a pain to maintain then it can be removed later

fig02 (approved):

fig02:
looking at the shields on your github page, it seems the progress one says "resource not found" or something. not sure if thats an issue that would be fixed when merged in

also, build should say "OK" instead of passing imo Kappa

Roman971:
> looking at the shields on your github page, it seems the progress one says "resource not found" or something. not sure if thats an issue that would be fixed when merged in

The progress badge needs the new endpoint on the website, which will only be added and updated during the jenkins build once this is pushed to master. It works similarly to the progress csv used to update the website after each master build.

fig02:
FeelsOKMan
good to merge this, if anyone else wants to drop an approval

Thar0 (approved):

Roman971:
FYI, the progress badge should end up looking like this:

Progress

PR #674, opened 2021-01-31
Fix most compiler warnings in the boot and code segments

Thar0:
Reduces build output by a little over 3000 lines, similar to #656 for actors.

A few files in `code` still have some warnings:
 * fault.c : `PadMgr_RequestPadData` is not called properly in one of the functions. As this is a genuine bug no attempt to suppress the warning is made.
 * z_actor.c : Some comparisons do not account for the limited range of data types. I do not know of a fix for this that maintains matching.
 * z_camera.c : Occurrences of `sp50[phi_a1++]` may be undefined. This is in the last nonmatching, so was ignored for now.
 * ~~z_view.c : `if (&D_8012ABF0) {}` is a fake-matching trick in `View_Init`, the use of `&` here causes a warning as the address will always evaluate to `true`. `if (D_8012ABF0 == 0) {}` also works to match and does not warn but I am not sure if it's better, perhaps I could get some thoughts on that?~~

Also addresses some style issues such as pointers not being directly compared with NULL, or variables not being assigned values in-line with their declaration where possible.


fig02:
About the z_view warning, since its a clearly fake matching trick I personally feel going with the option that doesnt induce a warning seems better. hopefully we can remove it at some point either way

fig02 (approved):

fig02 (approved):

src/code/z_play.c
```diff
@@ -55,38 +55,38 @@ void func_800BC5E0(GlobalContext* globalCtx, s32 transitionType) {
...
-        transitionCtx->setEnvColor = TransitionCircle_SetEnvColor;
+        transitionCtx->init = (TransitionInit)TransitionCircle_Init;
+        transitionCtx->destroy = (TransitionDestroy)TransitionCircle_Destroy;
+        transitionCtx->start = (TransitionStart)TransitionCircle_Start;
+        transitionCtx->isDone = (TransitionIsDone)TransitionCircle_IsDone;
+        transitionCtx->draw = (TransitionDraw)TransitionCircle_Draw;
+        transitionCtx->update = (TransitionUpdate)TransitionCircle_Update;
+        transitionCtx->setType = (TransitionSetType)TransitionCircle_SetType;
+        transitionCtx->setColor = (TransitionSetColor)TransitionCircle_SetColor;
+        transitionCtx->setEnvColor = (TransitionSetEnvColor)TransitionCircle_SetEnvColor;
```
Roman971, line 66: Why are these casts needed? Shouldn't the function types be the same as the types in the struct?
In general I'd say we should strongly avoid casting functions, so we should look for an alternative if possible

src/code/z_player_lib.c
```diff
@@ -1339,22 +1339,18 @@ void func_80090D20(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3s*
...
     }
 }
 
 u32 func_80091738(GlobalContext* globalCtx, u8* segment, SkelAnime* skelAnime) {
-    s16 linkObjectId;
-    u32 size;
-    void* ptr;
+    s16 linkObjectId = gLinkObjectIds[(void)0, gSaveContext.linkAge];
+    u32 size = gObjectTable[OBJECT_GAMEPLAY_KEEP].vromEnd - gObjectTable[OBJECT_GAMEPLAY_KEEP].vromStart;
+    void* ptr = segment + 0x3800;
```
Roman971, line 1352: It's not a huge deal, but I personally disagree with combining the declarations in cases like these. Here and in some other places, it makes more sense to leave them separate imo.

src/code/z_player_lib.c
```diff
@@ -1508,8 +1501,8 @@ void func_80091A24(GlobalContext* globalCtx, void* seg04, void* seg06, SkelAnime
 
     gSPSegment(POLY_OPA_DISP++, 0x0C, gCullBackDList);
 
-    func_8008F470(globalCtx, arg3->skeleton, arg3->jointTable, arg3->dListCount, 0, tunic, boots, 0, func_80091880,
-                  NULL, &sp12C);
+    func_8008F470(globalCtx, arg3->skeleton, arg3->jointTable, arg3->dListCount, 0, tunic, boots, 0,
+                  (OverrideLimbDrawOpa)func_80091880, NULL, &sp12C);
```
Roman971, line 1505: The type of func_80091880 should be fixed instead of casting here, imo

src/code/z_player_lib.c
Thar0: Could you explain a little more why you think so? I'm currently under the impression we should be doing this wherever we can, not limited to constants.

Roman971: Further down in the function, `size` and `ptr` are assigned values before the `DmaMgr_SendRequest1` call, so it's more consistent to set their values before the function call in both instances.

And in general, initializing the variable in the declaration doesn't necessarily make sense if the variable is written to later on in the function. (this is why I didn't combine them in functions like `Item_Give`, for instance)

fig02 (approved):

Roman971 (approved):

PR #675, opened 2021-02-01
Minor z_bgcheck documentation fixes

mzxrules:


src/code/z_bgcheck.c
```diff
@@ -2514,9 +2514,9 @@ void BgActor_SetActor(BgActor* bgActor, Actor* actor, CollisionHeader* colHeader
 }
 
 /**
- * Test if the BgActor transform changed
+ * Test if the BgActor transform is the same
  */
-s32 BgActor_HasTransformChanged(BgActor* bgActor) {
+s32 BgActor_IsTransformUnchanged(BgActor* bgActor) {
```
fig02, line 2519: wondering if `_TransformMaintained` reads nicer, as opposed to the negative form of changed. 
both work, just a thought

fig02:
@mzxrules ^

src/code/z_bgcheck.c
mzxrules: idk. I feel like maintain is an unusual word for describing it.
I know one thing petrie suggested was to consider naming it IsUnmoved or something relating to movement, since scaling/rotating/changing positions can be considered as moving, since this is a less mathy way of describing what the system is doing. The downside to this is that the system doesn't directly account for mesh transformations.

fig02: i would agree to making it less mathy if it was used alot externally. but its used in one place and is internal to the bgcheck system. so making it more user friendly on the outside isnt as much of a factor. im fine with whatever you guys agree is better tho

fig02 (approved):

include/z64bgcheck.h
```diff
@@ -11,6 +11,8 @@ struct DynaPolyActor;
 #define COLPOLY_VIA_FLAG_TEST(vIA, flags) ((vIA) & (((flags)&7) << 13))
 #define COLPOLY_VTX_INDEX(vI) ((vI)&0x1FFF)
 
+#define DYNAPOLY_INVALIDATE_LOOKUP 1 << 0
```
Roman971, line 14:
```suggestion
#define DYNAPOLY_INVALIDATE_LOOKUP (1 << 0)
```

Roman971 (approved):

PR #678, opened 2021-02-02
Make func_800C213C match original ROM

teaguecl:
The C code for this function was very close to done, and was
functionally equivalent to the ROM.  This commit uses a few tricks
to convince the compiler to allocate the same registers as the original
ROM.  The two tricks used are:
1. Introduced a new temporary variable "yinc".
2. Added a "do while(0) {}" line in two places.

Both of these ideas came from the decomp-permuter tool.  I took the
do/while idea and guess as to what it might have been in the original
code.  I imagined that the "do while(0) {}" was probably left over from
a standard multi-line macro, which for some reason had an empty body.
I further suggest that maybe the empty body is because it was a feature
compiled out during RELEASE builds - perhaps a debug logging macro.


fig02:
Great job on this match!
Since we have no way of knowing what code was actually there, and even if it was a log message (or what it even said), we usually opt for using `if (1) {}`s or `do{} while(0)`'s. Its the simplest way to make the code match without making incorrect assumptions about what it actually could have been, since we have no way of knowing.
It would be preferable if you used do while 0's instead :D

teaguecl:
Thank you flg02 for the feedback.  Will update this PR with that change.

Roman971:
I'll add that some gbi.h macros don't have a wrapping `do {} while (0)` around them yet, so this might be what's happening here. At some point I plan to look into adding them to every gbi.h macro and see whether it helps or completely ruins functions that use these macros.

Roman971 (approved):

fig02 (approved):

PR #686, opened 2021-02-06
Add a segment number option to xmlcreate.py

louist103:
I dont know how much help this will be but the option is there now.


fig02 (approved):

Roman971:
I could be wrong but I thought this script worked off the object in init vars, which has to be number 6 since that's what the actor engine expects. So I feel like this would never get a use?
At least I think the script should default to number 6 so it doesn't require an extra argument for most if not all cases.

petrie911:
I think I mentioned it, but the script only really works if all segment addresses are part of a single object in segment 6, which is also the object in actor init vars. Any other arrangement makes the parsing too hard. This seemed to encompass the majority of actors so it worked out. It's possible I should write a script to parse an existing xml file and replace all segment addresses with the offsets from that file (so _keep objects can be used with it), but I'm not sure if I'll find the time.

fig02:
based on the above conversation i think this change is probably not worth it, or can at least be done in a different way. gonna close for now

PR #689, opened 2021-02-08
En_Clear_Tag OK and documented

Kenix3:


Zelllll (changes requested):
I left quite a few comments and suggestions, let me know if you have any questions.

include/z64.h
```diff
@@ -909,7 +909,7 @@ typedef struct GlobalContext {
     /* 0x11E04 */ s16* setupExitList;
     /* 0x11E08 */ Path* setupPathList;
     /* 0x11E0C */ ElfMessage* cUpElfMsgs;
-    /* 0x11E10 */ char unk_11E10[0x4];
+    /* 0x11E10 */ void* bossEffects;
```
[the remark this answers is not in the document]
fig02: they can point to various effect structs depending on the actor, so void is fine

though, not sure if we want to name this after bosses. that is the main use, but this pr is an example of something using it thats not a boss. something like `customEffects` or `actorEffects` or something is prob more appropriate

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.h
```diff
@@ -8,9 +8,82 @@ struct EnClearTag;
...
-    /* 0x014C */ char unk_14C[0xB8];
+    /* 0x014C */ u8 shouldExplode;
+    /* 0x014D */ u8 drawMode;
+    /* 0x014E */ u8 state;
+    /* 0x0150 */ s16 work[3];
+    /* 0x0158 */ Vec3f targetPosition;
+    /* 0x0164 */ Vec3f targetDirection;
+    /* 0x0170 */ Vec3f acceleration;
+    /* 0x017C */ u8 timer;
+    /* 0x017D */ u8 shouldShootLaser;
```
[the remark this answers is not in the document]
Kenix3: I think the name I have is more descriptive.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.h
```diff
@@ -8,9 +8,82 @@ struct EnClearTag;
...
+    /* 0x0184 */ s16 crashingTimer;
+    /* 0x0186 */ s16 deathTimer;
+    /* 0x0188 */ Vec3f floorTangent;
+    /* 0x0194 */ ColliderCylinder collider;
+    /* 0x01E0 */ u8 demoMode;
+    /* 0x01E2 */ s16 cameraNumber;
+    /* 0x01E4 */ Vec3f demoCameraAt;
+    /* 0x01F0 */ Vec3f demoCameraEye;
+    /* 0x01FC */ s16 cutsceneTimer;
+    /* 0x01FE */ UNK_TYPE1 unk_1FE[0x06];
```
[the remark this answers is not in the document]
Kenix3: It needs to be there or else the struct size doesn't match. Padding out put the full struct size at 0x200. There is probably 2 bytes of padding an unused var.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -22,9 +33,17 @@ const ActorInit En_Clear_Tag_InitVars = {
     (ActorFunc)EnClearTag_Draw,
 };
 
-static ColliderCylinderInit D_809D5C6C = {
+static u8 isParticlesInitialized = 0;
+
+static Vec3f zeroVector = {
+    0.0f,
+    0.0f,
+    0.0f,
```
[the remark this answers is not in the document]
Kenix3: They always tell me to add trailing comma to structs.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +80,993 @@ static ColliderCylinderInit D_809D5C40 = {
...
+        if (effect->type == CLEAR_TAG_PARTICLE_AVAILABLE) {
+            effect->type = CLEAR_TAG_PARTICLE_DEBRIS;
+
+            effect->position = *position;
+            effect->velocity = *velocity;
+            effect->acceleration = *acceleration;
+
+            effect->scale = scale;
+
+            // Set the Debris particles to spawn in a circle.
```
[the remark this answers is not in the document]
Kenix3: Not a valid change.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +80,993 @@ static ColliderCylinderInit D_809D5C40 = {
...
+
+static EnClearTagEffect clearTagParticlesBuffer[CLEAR_TAG_PARTICLE_MAX_COUNT];
+
+#include "../../../../assets/overlays/ovl_En_Clear_Tag/ovl_En_Clear_Tag.c"
+
+/**
+ * Creates a debris particle.
+ * Debris particles are spawned when the Arwing dies. It spawns fire particles.
+ */
+void EnClearTag_CreateDebrisParticle(GlobalContext* globalCtx, Vec3f* position, Vec3f* velocity, Vec3f* acceleration,
```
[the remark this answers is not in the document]
Kenix3: This is debris, not dust.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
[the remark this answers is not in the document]
fig02: for really long ones yeah, but for short stuff like this that can afford to be on one line, no comma looks nice

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.h
```diff
@@ -8,9 +8,82 @@ struct EnClearTag;
...
+    /* 0x004C */ f32 scale;
+    /* 0x0050 */ f32 maxScale;
+    /* 0x0054 */ f32 rotationY;
+    /* 0x0058 */ f32 rotationX;
+    /* 0x005C */ f32 floorHeight;
+    /* 0x0060 */ Vec3f floorTangent;
+} EnClearTagEffect; // size = 0x6C
+
+typedef enum {
+    /* 0x00 */ CLEAR_TAG_CUTSCENE_ARWING = 0,
```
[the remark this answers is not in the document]
Kenix3: Keeping cleartag for now since it's the file name and more consistent. We can replace CLEAR_TAG with Arwing later in all uses.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -22,9 +33,17 @@ const ActorInit En_Clear_Tag_InitVars = {
     (ActorFunc)EnClearTag_Draw,
 };
 
-static ColliderCylinderInit D_809D5C6C = {
+static u8 isParticlesInitialized = 0;
```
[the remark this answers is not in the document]
Kenix3: Rather not use C99 features in my C89

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +80,993 @@ static ColliderCylinderInit D_809D5C40 = {
...
     { 15, 30, 10, { 0, 0, 0 } },
 };
-*/
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Clear_Tag/func_809D35B0.s")
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Clear_Tag/func_809D36A8.s")
+static u32 D_809D5C98 = 0; // unused
+static u32 D_809D5C9C = 0; // unused
+
+static EnClearTagEffect clearTagParticlesBuffer[CLEAR_TAG_PARTICLE_MAX_COUNT];
```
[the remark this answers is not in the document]
Kenix3: There's no info here. It's just a buffer that the actor uses to store particles.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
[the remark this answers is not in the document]
fig02: there are defines for true and false and they are used all around the codebase

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +80,993 @@ static ColliderCylinderInit D_809D5C40 = {
...
+                                     f32 scale, f32 floorHeight) {
+    s16 i;
+    s16 seed;
+    EnClearTagEffect* effect;
+    GlobalContext* globalCtx2;
+
+    globalCtx2 = globalCtx;
+    effect = (EnClearTagEffect*)globalCtx2->bossEffects;
+
+    // Look for an available particle to allocate a Debris particle to.
```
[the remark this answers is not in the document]
Kenix3: These are particles, and this is debris not dust.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +80,993 @@ static ColliderCylinderInit D_809D5C40 = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Clear_Tag/func_809D36A8.s")
+static u32 D_809D5C98 = 0; // unused
+static u32 D_809D5C9C = 0; // unused
+
+static EnClearTagEffect clearTagParticlesBuffer[CLEAR_TAG_PARTICLE_MAX_COUNT];
+
+#include "../../../../assets/overlays/ovl_En_Clear_Tag/ovl_En_Clear_Tag.c"
+
+/**
+ * Creates a debris particle.
```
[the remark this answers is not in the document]
Kenix3: This is debris, not dust. It is a particle.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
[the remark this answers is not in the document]
Zelllll: Maybe `sArwingEffects` then

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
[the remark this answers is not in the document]
Zelllll: There is no such effect called "debris" though. What makes this different?

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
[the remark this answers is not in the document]
Kenix3: `sClearTagEffects` works for me.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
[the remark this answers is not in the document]
Kenix3: These effects are not part of the effects system. They are all self contained, even the assets needed to display them. It doesn't matter whether or not the effects system has a debris effect.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +80,993 @@ static ColliderCylinderInit D_809D5C40 = {
...
+ * Debris particles are spawned when the Arwing dies. It spawns fire particles.
+ */
+void EnClearTag_CreateDebrisParticle(GlobalContext* globalCtx, Vec3f* position, Vec3f* velocity, Vec3f* acceleration,
+                                     f32 scale, f32 floorHeight) {
+    s16 i;
+    s16 seed;
+    EnClearTagEffect* effect;
+    GlobalContext* globalCtx2;
+
+    globalCtx2 = globalCtx;
```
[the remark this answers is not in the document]
Kenix3: I think it's more consistent within this function to NOT do that since moving the two variables set there to the declarations is not OK.

Kenix3:
Updated based on everything I agree with.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.h
```diff
@@ -8,9 +8,82 @@ struct EnClearTag;
 
 typedef struct EnClearTag {
     /* 0x0000 */ Actor actor;
-    /* 0x014C */ char unk_14C[0xB8];
+    /* 0x014C */ u8 shouldExplode;
```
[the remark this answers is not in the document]
Kenix3: I think my name is more descriptive and fits with the usage of `if (shouldExplode)` better.

fig02:
sorry, my latest armos merge caused a conflict. `func_8003426C` is now `Actor_SetColorFilter`

assets/xml/overlays/ovl_En_Clear_Tag.xml
```diff
@@ -0,0 +1,14 @@
+<Root>
+    <File Name="ovl_En_Clear_Tag" BaseAddress="0x809D35B0" RangeStart="0x26F0" RangeEnd="0x89F0" Segment="128">
+        <DList Name="arwingDL" Offset="0x26F0"/>
+        <DList Name="laserDL" Offset="0x6388"/>
+        <DList Name="arwingBackfireDL" Offset="0x6688"/>
```
fig02, line 5: I think zel noted already, but these need to follow our naming convention.
`gArwing<NameOfThing><Suffix>`
siffixes are DL, Anim, Col etc
can use ArwingEff for the effects maybe

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,1001 @@ static ColliderCylinderInit D_809D5C40 = {
...
-*/
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Clear_Tag/func_809D35B0.s")
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Clear_Tag/func_809D36A8.s")
+static u32 D_809D5C98 = 0; // unused
+static u32 D_809D5C9C = 0; // unused
+
+static EnClearTagEffect sClearTagEffects[CLEAR_TAG_EFFECT_MAX_COUNT];
+
+#include "../../../../assets/overlays/ovl_En_Clear_Tag/ovl_En_Clear_Tag.c"
```
fig02, line 85: we dont use relative paths anymore in the project

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,1001 @@ static ColliderCylinderInit D_809D5C40 = {
...
+            effect->rotationY = Rand_ZeroFloat(M_PI * 2);
+            effect->rotationX = Rand_ZeroFloat(M_PI * 2);
+
+            effect->timer = effect->bounces = 0;
+
+            effect->floorHeight = floorHeight;
+
+            seed = (s32)Rand_ZeroFloat(10.0f);
+            effect->seed = seed;
+
```
fig02, line 122: idk if seed is the right word here. thats usually used as part of the random number generation process.
since its a general purpose random number id prob just call it `rand`

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,1001 @@ static ColliderCylinderInit D_809D5C40 = {
...
+    // Initialize the Arwing.
+    this->actor.flags |= 1;
+    this->actor.targetMode = 5;
+    Collider_SetCylinder(globalCtx, &this->collider, &this->actor, &sArwingCylinderInit);
+    this->actor.colChkInfo.health = 3;
+
+    // Update the Arwing to play the intro cutscene.
+    if (this->actor.params == CLEAR_TAG_CUTSCENE_ARWING) {
+        this->work[0] = 70;
+        this->work[1] = 250;
```
fig02, line 288: if youre interested in doing more documentation, defines for the `work` indices would make things much more readable.
I also think we should consider a name other than work, but we dont have to figure that out now

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,1001 @@ static ColliderCylinderInit D_809D5C40 = {
...
+    Vec3f cutsceneCameraAtTarget;
+    Vec3f cutsceneCameraEyeTarget;
+    Vec3f crashEffectLocation;
+    Vec3f crashEffectVelocity;
+    Vec3f debrisEffectAcceleration;
+
+    this->timer++;
+
+    if (this->drawMode != CLEAR_TAG_DRAWMODE_EFFECT) {
+        // Decrement the "work" timers.
```
fig02, line 378: imo comment not needed, you can quickly see thats whats happening

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,1001 @@ static ColliderCylinderInit D_809D5C40 = {
...
+        for (i = 0; i < 3; i++) {
+            if (this->work[i] != 0) {
+                this->work[i]--;
+            }
+        }
+
+        // Decrement the cutscene timer.
+        if (this->cutsceneTimer != 0) {
```
fig02, line 386: same here

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,1001 @@ static ColliderCylinderInit D_809D5C40 = {
...
+                        // Set the Arwing to fly to a hardcoded position.
+                        this->targetPosition.x = 0.0f;
+                        this->targetPosition.y = 300.0f;
+                        this->targetPosition.z = 0.0f;
+                        loseTargetLockDistance = 100.0f;
+                    }
+
+                    // If the Arwing is not in cutscene mode update the roll to zero.
+                    if (this->state != CLEAR_TAG_STATE_DEMO) {
+                        Math_ApproachZeroF(&this->roll, 0.1f, 0.2f);
```
fig02, line 467: same here, the code pretty clearly says this

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,1001 @@ static ColliderCylinderInit D_809D5C40 = {
...
+                        this->shouldShootLaser = 0;
+                        Actor_Spawn(&globalCtx2->actorCtx, globalCtx2, ACTOR_EN_CLEAR_TAG, this->actor.world.pos.x,
+                                    this->actor.world.pos.y, this->actor.world.pos.z, this->actor.world.rot.x,
+                                    this->actor.world.rot.y, this->actor.world.rot.z, CLEAR_TAG_STATE_LASER);
+                    }
+                }
+
+            case CLEAR_TAG_STATE_CRASHING:
+                // Decrement the crashing timer.
+                if (this->crashingTimer != 0) {
```
fig02, line 539: same

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,1001 @@ static ColliderCylinderInit D_809D5C40 = {
...
+                    this->crashingTimer--;
+                }
+
+                // Update Arwing collider to better match a ground collision.
+                func_8002D7EC(&this->actor);
```
fig02, line 544: I dont think this comment is accurate to whats going on. what does it mean for it to better match the ground?
setting focus doesnt seem related to that, and setting AC and AT are not related to that either

include/z64.h
```diff
@@ -909,7 +909,7 @@ typedef struct GlobalContext {
     /* 0x11E04 */ s16* setupExitList;
     /* 0x11E08 */ Path* setupPathList;
     /* 0x11E0C */ ElfMessage* cUpElfMsgs;
-    /* 0x11E10 */ char unk_11E10[0x4];
+    /* 0x11E10 */ void* specialEffects;
```
fig02, line 912: "special" is an alright name, though I do think that `customEffects` moreso gets the point across that you can use this if none of the effects supplied to you by the engine are sufficient. It lets you make your own if the provided ones arent what you need

fig02: curious what others think tho

assets/xml/overlays/ovl_En_Clear_Tag.xml
Kenix3: Ah I missed the g part. Thanks!

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
Kenix3: I like this suggestion a lot.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
Kenix3: Hmmm. Yeah I think that's fair.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
Kenix3: I think it would be better to explain why this is important.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
Kenix3: We can't remove the relative path here because include starts in the current directory. We would have to update extraction script to put the C file for overlays in the overlay's folder, rather than assets.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
Kenix3: It is accurate in my opinion. It's updating the collider to update in a way that makes the ground collision look more correct.

You're right, though, that all of it is accomplishing that goal.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,999 @@ static ColliderCylinderInit D_809D5C40 = {
...
+        this->state = CLEAR_TAG_STATE_DEMO;
+        this->actor.world.rot.x = 0x4000;
+        this->cutsceneMode = CLEAR_TAG_DEMOMODE_SETUP;
+        this->cutsceneTimer = defaultCutsceneTimer;
+        this->work[2] = 20;
+    }
+
+    // Initialize all effects to available if effects have not been initialized.
+    if (!sIsEffectsInitialized) {
+        sIsEffectsInitialized = 1;
```
petrie911, line 298:
```suggestion
        sIsEffectsInitialized = true;
```

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,999 @@ static ColliderCylinderInit D_809D5C40 = {
...
+    for (i = 0; i < CLEAR_TAG_EFFECT_MAX_COUNT; i++, effect++) {
+        if (effect->type == CLEAR_TAG_EFFECT_AVAILABLE) {
+            effect->type = CLEAR_TAG_EFFECT_FLASH;
+
+            effect->position = *position;
+            effect->velocity = sZeroVector;
+            effect->acceleration = sZeroVector;
+
+            effect->scale = 0.0f;
+            effect->maxScale = scale + scale;
```
petrie911, line 226:
```suggestion
            effect->maxScale = 2.0f * scale;
```
This is a compiler optimization.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,999 @@ static ColliderCylinderInit D_809D5C40 = {
...
+ * Flash effects are spawned when the Arwing dies.
+ * Flash effects two components: 1) a billboard flash, and 2) a light effect on the ground.
+ */
+void EnClearTag_CreateFlashEffect(GlobalContext* globalCtx, Vec3f* position, f32 scale, f32 floorHeight,
+                                  Vec3f* floorTangent) {
+    s16 i;
+    EnClearTagEffect* effect;
+    GlobalContext* globalCtx2;
+
+    globalCtx2 = globalCtx;
```
petrie911, line 213: This recast should be unnecessary. Does it match without it?

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,999 @@ static ColliderCylinderInit D_809D5C40 = {
...
+            case CLEAR_TAG_STATE_DEMO:
+            case CLEAR_TAG_STATE_TARGET_LOCKED:
+            case CLEAR_TAG_STATE_FLYING:
+                // Check if the Arwing should crash.
+                if (this->collider.base.acFlags & 2 &&
+                    (this->collider.base.acFlags &= ~2, this->crashingTimer = 20,
+                     Actor_SetColorFilter(&this->actor, 0x4000, 255, 0, 5), this->acceleration.x = Rand_CenteredFloat(15.0f),
+                     this->acceleration.y = Rand_CenteredFloat(15.0f), this->acceleration.z = Rand_CenteredFloat(15.0f),
+                     Audio_PlayActorSound2(&this->actor, NA_SE_EN_FANTOM_THUNDER_GND), this->actor.colChkInfo.health--,
+                     (s8)this->actor.colChkInfo.health <= 0)) {
```
petrie911, line 398: OK this can't possibly be correct.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,999 @@ static ColliderCylinderInit D_809D5C40 = {
...
+                                // Set the Arwing to fly to a random position.
+                                this->targetPosition.x = Rand_CenteredFloat(700.0f);
+                                this->targetPosition.y = Rand_ZeroFloat(200.0f) + 150.0f;
+                                this->targetPosition.z = Rand_CenteredFloat(700.0f);
+                            }
+                        }
+
+                        this->targetDirection.z = 0.0f;
+                        this->targetDirection.y = 0.0f;
+                        this->targetDirection.x = 0.0f;
```
petrie911, line 433:
```suggestion
                        this->targetDirection.x = this->targetDirection.y = this->targetDirection.z = 0.0f;
```

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,999 @@ static ColliderCylinderInit D_809D5C40 = {
...
+                    }
+
+                    rotationScale = 10;
+                    xRotationTarget = 2048;
```
petrie911, line 437:
```suggestion
                    xRotationTarget = 0x800;
```
Power of 2 angles should *definitely* be in hex.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,999 @@ static ColliderCylinderInit D_809D5C40 = {
...
+                        this->work[0] = 0;
+                        if (this->state == CLEAR_TAG_STATE_TARGET_LOCKED) {
+                            this->work[1] = ((s16)Rand_ZeroFloat(100.0f)) + 100;
+                        }
+                        this->state = CLEAR_TAG_STATE_FLYING;
+                    }
+
+                    // Calculate the direction for the Arwing to fly and the rotation for the Arwing
+                    // based on the Arwing's direction, and current rotation.
+                    worldRotationTargetY = (s32)(Math_FAtan2F(vectorToTargetX, vectorToTargetZ) * 10430.378f);
```
petrie911, line 487: This float is `(0x8000 / M_PI)`

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,999 @@ static ColliderCylinderInit D_809D5C40 = {
...
+                    worldRotationTargetZ = Math_SmoothStepToS(&thisx->world.rot.y, worldRotationTargetY, rotationScale,
+                                                              this->targetDirection.y, 0);
+                    Math_ApproachF(&this->targetDirection.x, xRotationTarget, 1.0f, 256.0f);
+                    this->targetDirection.y = this->targetDirection.x;
+                    if (ABS(worldRotationTargetZ) < 4096) {
+                        Math_ApproachS(&thisx->world.rot.z, 0, 15, this->targetDirection.z);
+                        Math_ApproachF(&this->targetDirection.z, 1280.0f, 1.0f, 256.0f);
+
+                        // Check if the Arwing should fire it's laser.
+                        if (!(this->timer & 3) && (Rand_ZeroOne() < 0.75f) &&
```
petrie911, line 506: `((this->timer % 4) == 0)` is more logical.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,999 @@ static ColliderCylinderInit D_809D5C40 = {
...
+                    case CLEAR_TAG_DEMOMODE_SETUP:
+                        // Initializes Arwing cutscene camera data.
+                        this->cutsceneMode = CLEAR_TAG_DEMOMODE_PLAY;
+                        func_80064520(globalCtx2, &globalCtx2->csCtx);
+                        this->cameraId = Gameplay_CreateSubCamera(globalCtx2);
+                        Gameplay_ChangeCameraStatus(globalCtx2, 0, 1);
+                        Gameplay_ChangeCameraStatus(globalCtx2, this->cameraId, 7);
+                    case CLEAR_TAG_DEMOMODE_PLAY:
+                        // Update the Arwing cutscene camera to spin around in a circle.
+                        cutsceneTimer = this->timer << 7;
```
petrie911, line 630:
```suggestion
                        cutsceneTimer = this->timer * 128;
```

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,999 @@ static ColliderCylinderInit D_809D5C40 = {
...
+        for (i = 0; i < 15; i++) {
+            crashEffectVelocity.x = sinf(i * 1.65f) * i * 0.3f;
+            crashEffectVelocity.z = cosf(i * 1.65f) * i * 0.3f;
+            crashEffectVelocity.y = Rand_ZeroFloat(6.0f) + 5.0f;
+            crashEffectVelocity.x += Rand_CenteredFloat(0.5f);
+            crashEffectVelocity.z += Rand_CenteredFloat(0.5f);
+
+            debrisEffectAcceleration.x = 0.0f;
+            debrisEffectAcceleration.z = 0.0f;
+            debrisEffectAcceleration.y = -1.0f;
```
petrie911, line 689: Can this be done in `xyz` order?

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
Kenix3: I think leaving it as `work` for now would be preferable. Then we can more easily replace all `work` variable names in the future.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,999 @@ static ColliderCylinderInit D_809D5C40 = {
...
+        // Check if the Arwing should be removed.
+        if (this->drawMode == CLEAR_TAG_DRAWMODE_EFFECT) {
+            if (this->deathTimer == 0) {
+                isArwingAlive = 0;
+            } else {
+                this->deathTimer--;
+                isArwingAlive = this->deathTimer;
+            }
+
+            if (!isArwingAlive) {
```
petrie911, line 707: This looks like `if(DECR(this->deathTimer) == 0) {`

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,999 @@ static ColliderCylinderInit D_809D5C40 = {
...
+            gSPDisplayList(POLY_XLU_DISP++, gLaserDL);
+        } else {
+            // Draw the Arwing itself.
+            func_80093D18(globalCtx2->state.gfxCtx);
+            gDPSetPrimColor(POLY_OPA_DISP++, 0, 0, 255, 255, 255, 255);
+            if (this->crashingTimer != 0) {
+                f32 xRotation;
+                f32 yRotation;
+                f32 scaledCrashingTimer;
+                scaledCrashingTimer = this->crashingTimer * 0.05f;
```
petrie911, line 750:
```suggestion
                f32 scaledCrashingTimer = this->crashingTimer * 0.05f;
                
```

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,999 @@ static ColliderCylinderInit D_809D5C40 = {
...
+                        if (effect->bounces <= 0) {
+                            effect->bounces++;
+                            effect->velocity.y *= -0.5f;
+                            effect->timer = ((s16)Rand_ZeroFloat(20)) + 25;
+                        } else {
+                            // The Debris effect is done bounding. Set it's velocity and acceleration to 0.
+                            effect->velocity.y = 0.0f;
+                            effect->acceleration.y = 0.0f;
+                            effect->velocity.z = 0.0f;
+                            effect->velocity.x = 0.0f;
```
petrie911, line 863: possible chained assigns, see if it can be done in `xyz` order.

petrie911 (commented):
A few style things. A lot of them also apply generally.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
Kenix3: Give me an alternative. ;)

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
Kenix3: Not OK.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
Kenix3: I've tried DECR in the file and it's not OK.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
Kenix3: Yeah, more consistent with how I generally set vars.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
Kenix3: I had already tried this. Not OK.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
Kenix3: It's generally good practice to explicitly cast your void* pointers to the explicit type for readability, is it not?

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
Kenix3: Yeah, I know of the optimization. I missed it here. Thanks!

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
petrie911: but this isn't a void* pointer.

Kenix3: Do all of the special effects have a shared base struct?

petrie911: No I mean `globalCtx2` should be unnecessary.

Kenix3: Yep. It looks like I copy/pasted the boiler plate code for this function. Only the engine functions _should_ need the base game struct.

Good catch.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
mzxrules:
```suggestion

#include "overlays/ovl_En_Clear_Tag/ovl_En_Clear_Tag.c"
```
Does this not work?

Kenix3: Ah I had tried it from the project root, not the source root. This worked.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
fig02: agreed

fig02 (approved):
other than one of petries comments that got left, and discussion about what to name the effects buffer in global context, this is good to go for me

assets/xml/overlays/ovl_En_Clear_Tag.xml
```diff
@@ -0,0 +1,14 @@
+<Root>
+    <File Name="ovl_En_Clear_Tag" BaseAddress="0x809D35B0" RangeStart="0x26F0" RangeEnd="0x89F0" Segment="128">
+        <DList Name="gArwingDL" Offset="0x26F0"/>
+        <DList Name="gLaserDL" Offset="0x6388"/>
+        <DList Name="gArwingBackfireDL" Offset="0x6688"/>
+        <DList Name="gDebrisEffectMaterialDL" Offset="0x6A38"/>
+        <DList Name="gDebrisEffectDL" Offset="0x6AB8"/>
+        <DList Name="gFireEffectMaterialDL" Offset="0x7100"/>
+        <DList Name="gFireEffectDL" Offset="0x71A8"/>
```
Roman971, line 9: Those names seem a bit too generic and they will probably conflict with other things in the game, so I suggest using something more specific to Arwing (probably fine to just include Arwing in all these names tbh)

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.h
```diff
@@ -8,9 +8,89 @@ struct EnClearTag;
...
+    /* 0x03 */ CLEAR_TAG_EFFECT_SMOKE,
+    /* 0x04 */ CLEAR_TAG_EFFECT_FLASH
+} ClearTagEffectType;
+
+typedef enum {
+    /* 0x00 */ CLEAR_TAG_TIMER_ARWING_UPDATE_STATE = 0,
+    /* 0x00 */ CLEAR_TAG_TIMER_LASER_DEATH = 0,
+    /* 0x01 */ CLEAR_TAG_TIMER_ARWING_ENTER_LOCKED_ON = 1,
+    /* 0x02 */ CLEAR_TAG_TIMER_ARWING_UPDATE_BG_INFO = 2
+} ClearTagTimers;
```
Roman971, line 90: I feel like these should probably be defines of some form, because having multiple enum values resolve to the same thing would be probably be very confusing to work with

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
```diff
@@ -61,25 +76,991 @@ static ColliderCylinderInit D_809D5C40 = {
...
+                isMaterialApplied++;
+            }
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Clear_Tag/func_809D38BC.s")
+            // Draw the ground flash effect.
+            gDPSetPrimColor(POLY_XLU_DISP++, 0, 0, 255, 255, 200, (s8)effect->primColor.a);
+            Matrix_Translate(effect->position.x, effect->floorHeight, effect->position.z, MTXMODE_NEW);
+            Matrix_RotateX(effect->floorTangent.x, MTXMODE_APPLY);
+            Matrix_RotateZ(effect->floorTangent.z, MTXMODE_APPLY);
+            Matrix_Scale(effect->scale + effect->scale, 1.0f, effect->scale + effect->scale, MTXMODE_APPLY);
```
Roman971, line 980: I think these should most likely be `2.0f *` (or just `2 *`)
```suggestion
            Matrix_Scale(2.0f * effect->scale, 1.0f, 2.0f * effect->scale, MTXMODE_APPLY);
```

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.h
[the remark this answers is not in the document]
Roman971: `char` arrays are what we normally use for fillers, while `UNK_TYPE1` is mostly intended for a variable of size 1 byte and unknown signedness, so I would just use `char` here

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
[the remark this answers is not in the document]
Roman971: I think what Zel meant was to have `globalCtx2` as the name for the argument, and `globalCtx` in the function, and I agree it would look better since we could continue using the name `globalCtx` as usual

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
Roman971: Using 1 or 2 goto here should work, and it would probably lead to better code (?), but yeah I can't think of any alternative that doesn't involve goto(s)

include/z64.h
Roman971: "special" seems fine, tho it might not be the best. And I don't think it really matters anyway, since at the end of the day it was just used as a generic pointer to hold various things.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
Kenix3: I did however make an enum.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
Kenix3: It doesn't logically need a goto. It's logically this, but it's not OK.
```c
                if (this->collider.base.acFlags & 2) {
                     this->collider.base.acFlags &= ~2;
                     this->crashingTimer = 20;
                     Actor_SetColorFilter(&this->actor, 0x4000, 255, 0, 5);
                     this->acceleration.x = Rand_CenteredFloat(15.0f);
                     this->acceleration.y = Rand_CenteredFloat(15.0f);
                     this->acceleration.z = Rand_CenteredFloat(15.0f);
                     Audio_PlayActorSound2(&this->actor, NA_SE_EN_FANTOM_THUNDER_GND);
                     this->actor.colChkInfo.health--;
                     if ((s8)this->actor.colChkInfo.health <= 0)) {
                         // Do the rest of the things.
                     }
                }
```

assets/xml/overlays/ovl_En_Clear_Tag.xml
Kenix3: Works for me.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.h
Kenix3: I disagree here. This is a grouping of possible values of the work buffer. Enums are meant for situations like this where you have well defined possible states. 

I'd much rather call value 0 "general timer" or something than just make defines.

I did notice that regardless I should have a total work buffer size so that it can work as intended based on the enum growing/shrinking.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
Kenix3: ye good catch here. Thanks!

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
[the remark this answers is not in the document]
Kenix3: I disagree with that change. It's 1) Inconsistent with how other functions are defined. It's always `globalCtx` and not `globalCtx2`, and 2) That inconsistency will make a regex much more difficult when we actually implement passing in the `game` base state struct rather than `game_play` derived state struct.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.h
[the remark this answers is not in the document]
Kenix3: `UNK_TYPE1 is mostly intended for a variable of size 1 byte and unknown signedness`
This description fits here. We don't know what signedness this data originally had.

Unfortunately we don't know the size of this value either, so `char` doesn't technically fit. I think it's better to use some kind of UNK_TYPE to flag to someone reading that we just don't know what it's type originally was.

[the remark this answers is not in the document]
Roman971: We use `char` arrays for fillers in literally every other case, so I don't see why it would be different here...

It's going to be pretty clear to anyone reading the codebase that `char unk_XXX[...]` is just random space we don't know about and isn't used, since that's we do everywhere and other decomp projects do the same or something very similar.

src/overlays/actors/ovl_En_Clear_Tag/z_en_clear_tag.c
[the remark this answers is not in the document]
Roman971: Having `globalCtx2` for the argument instead of the temp actually makes things much easier for us in the future, since we can just replace the argument to something like `gameState` instead of `globalCtx2` when we make the change. And the `globalCtx` temp, which is the one that's actually used in the rest of the function, won't have to be touched or renamed.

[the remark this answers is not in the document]
fig02: heavily agreed here. when the time comes to cast gameState to globalCtx everywhere, renaming the arg is much easier than having to rename the variable used throughout the function

fig02 (approved):
review implemented
theres one small issue in `EnClearTag_DrawEffects` that we'll need to resolve at some point, probably when switching to using GameState for all globalctx2 uses.
```
 // using globalCtx2 here fixes saved reg usage, but will need to be changed when switching to GameState
func_800D1FD4(&globalCtx2->mf_11DA0);
```

Roman971 (approved):

PR #694, opened 2021-02-13
object_zl2, object_zl2_anime1 & object_zl2_anime2 OK

AngheloAlf:
Decompiled the objects of adult Zelda, used by `z_en_eg`, `z_en_zl2` and. `z_en_zl3`. I also did a bit of renaming of those actors.


fig02 (approved):
looks good, but you can remove the gitkeep files. we only need one in the asset folders root. same goes for your other open prs if they have them

AngheloAlf:
Done

assets/xml/objects/object_zl2.xml
```diff
@@ -0,0 +1,42 @@
+<Root>
+    <File Name="object_zl2" Segment="6">
+        <!-- Zelda 2 skeleton -->
+        <Skeleton Name="gZelda2Skel" Type="Flex" LimbType="Standard" Offset="0x10D70"/>
+
+        <!-- Zelda 2 eyes textures -->
+        <Texture Name="gZelda2EyesPalette" OutName="zelda_2_eyes_palette" Format="rgb5a1" Width="16" Height="16" Offset="0x2B00"/>
```
Roman971, line 7: I don't believe we have discussed a suffix for this yet, and "palette" seems a bit long, so we should probably bring it up in discord and see what people think (something like "pal" might be good enough but it's worth a discussion either way)

EllipticEllipsis: Given that we'll need to support PAL/NTSC versions of objects, that's probably the one abbreviation we don't want to use!

fig02: its important to consider, but i think context helps. and fwiw they originally used the pal abbreviation as well (see skyboxes)

fig02: We should get back to this discussion.
I still think `Pal` is fine and doesnt conflict with the region abbreviation at all (both context and capitalization make it clear imo), but if more people want TLUT then we should settle on what the majority wants

fig02: guess this conversation isnt really going to progress more, so TLUT is probably what the majority of people were favoring

Roman971:
I could be wrong, but I think the idea was to have TLUT as the suffix (all uppercase)

fig02:
Im down with either. currently we do things like `Id` instead of `ID` to follow camel case, so idk if one makes more sense over the other
Worth noting the gfx macro does it in all uppercase

AngheloAlf:
I don't like this suffix being `TLUT` in all caps, but I think we should use it like we use `DL` for display lists.

Roman971 (approved):

PR #696, opened 2021-02-13
object_anubice OK

AngheloAlf:


assets/xml/objects/object_anubice.xml
```diff
@@ -0,0 +1,24 @@
+<Root>
+    <File Name="object_anubice" Segment="6">
+        <!-- Unused -->
+        <Animation Name="gAnubiceAnim_000144" Offset="0x144"/>
+
```
Zelllll, line 5:
```suggestion
```
id remove the new line here

Zelllll (commented):
looks mostly good, document if you can

AngheloAlf:
@Zelllll Done!

assets/xml/objects/object_jya_iron.xml
```diff
@@ -0,0 +1,13 @@
+<Root>
+    <File Name="object_jya_iron" Segment="6">
+        <DList Name="gPillarDL" Offset="0x0240"/>
+        <DList Name="gObjectJyaIronDL_000600" Offset="0x0600"/>
+        <DList Name="gObjectJyaIronDL_000880" Offset="0x0880"/>
+        <DList Name="gObjectJyaIronDL_000AE0" Offset="0x0AE0"/>
+        <Collision Name="gPillarCol" Offset="0x0D48"/>
+        <DList Name="gThroneDL" Offset="0x1050"/>
```
Zelllll: Looks good! However I'd suggest maybe specifying IronKnuckle in these, just since there could potentially be another "pillar" in the game that might conflict in some circumstances.

Zelllll (commented):
Looks good! For the jya_iron though I'd suggest being more specific to avoid conflicts. Since for example there could be multiple "pillar" collisions in the game.

assets/xml/objects/object_jya_iron.xml
AngheloAlf: I don't even have that file in this pr :confused:

Zelllll: Oh weird, it seems to show up in the diff, still worth changing though :V

assets/xml/objects/object_anubice.xml
```diff
@@ -0,0 +1,20 @@
...
+        <Animation Name="gAnubiceAttackingAnim" Offset="0x78C"/>
+        <Animation Name="gAnubiceIdleAnim" Offset="0xF74"/>
+        <DList Name="gAnubiceEyesDL" Offset="0x3468"/>
+        <DList Name="gAnubiceFireAttackDL" Offset="0x3510"/>
+
+        <!-- The following are unused -->
+        <Animation Name="gAnubiceLaughingAnim" Offset="0x144"/>
+        <Animation Name="gAnubiceStandUpAnim" Offset="0x930"/>
+        <DList Name="gAnubiceUnusedParticleDL1" Offset="0x3610"/>
+        <DList Name="gAnubiceUnusedParticleDL2" Offset="0x3688"/>
```
fig02, line 15: we want the suffixes to be the last thing, so particle1DL and particle2DL
we can probably name them after their purpose tho. iirc one dlist has the texture and the other has the geometry?

AngheloAlf:
Done :D

fig02 (approved):

Roman971 (approved):

PR #697, opened 2021-02-13
Fix func_800BC5E0 NON_EQUIVALENT (matched)

louist103:
Apparently this function isnt equivalent.
Tested on N64.


fig02:
looking at the diff, this should 100% be equivalent. something else is the issue

fig02 (approved):

fig02:
For context, this non equivalent was not clear from just the asm that it was wrong. 
Louis watched the intro cutscene which uses transition 13, and it went to the default case and crashed instead. So this was actually just missing a case this whole time. Adding in the case fixed the seemingly unrelated instruction ordering difference as well.

Roman971 (approved):

PR #701, opened 2021-02-15
z_play.c OK

Synray:


src/code/z_play.c
```diff
@@ -333,7 +333,7 @@ void Gameplay_Init(GameState* thisx) {
     Matrix_Init(&globalCtx->state);
     globalCtx->state.main = Gameplay_Main;
     globalCtx->state.destroy = Gameplay_Destroy;
-    globalCtx->sceneLoadFlag = -0x14;
+    globalCtx->sceneLoadFlag = ~0x13;
```
fig02, line 336: are the ~0x13 changes for `sceneLoadFlag` needed to match? the game often uses 0x14 and -0x14 as values which tells me that its probably an actual negative value. not sure if our name of calling it a single "flag" is even correct tbh

but if its needed then nvm

Synray: it's not needed, I just thought it was a flag lol. It looks like it's only set to +-20, so should I change it to -20 instead?

fig02: i would say so, since thats what we have it as everywhere else

fig02: oh id also probably leave it in hex like everywhere else and we can change them all at once when it makes sense to do so. though some second thoughts are welcome
@Roman971 ?

Synray: oh rip, ok lol

fig02 (approved):

Roman971 (approved):

PR #703, opened 2021-02-16
Rename Matrix_Pull to Matrix_Pop

Roman971:
`Matrix_pull` is possibly the original name for this function (based on the AC map) but I believe "pop" is the more appropriate term for the action (pulling something out of the stack), so I think we should rename it.

This function is used in quite a few places so it might create some conflicts, but I figure this is a decent time to do it since we're pretty low on PRs atm.


fig02:
Think @AdamKiddle and others were interested in some more in depth Matrix name changes as well. Now is a great time.

I think Matrix_New can probably be improved as well. Last I had it explained to me, the term new seemed a bit confusing to what it actually does. But matrix stuff in general is still a bit over my head

mzxrules:
Looked it up on Wikipedia just to double check. Pull is an appropriate name for the operation, Pop is just the much more widely used name. 

Matrix_NewMtx is pretty complex yea. It creates a Mtx copy of sCurrentMatrix on the heap.

fig02 (approved):

fig02:
since I think we all agree on this name change specifically, we can bring this one in and then discuss our other name proposals further before making any changes

Roman971:
Fixed the conflicts and I agree on doing this rename on its own first, since the other names are still in discussion

PR #704, opened 2021-02-17
"decompile" ovl_map_mark_data, make minor tweaks to z_map_mark.c

mzxrules:
One thing of note is that z_map_mark_data.c ends up being 12k lines long. I can shorten it by ~8k lines by using `{ 0 }`, but fig suggested I don't do that to improve editability. 
I also created a python script that generates z_map_mark_data so that we can dump the vanilla data. it also lets us change the structure easier if we want to make changes like that later down the road.


fig02:
other people still need to weigh in on the `{ 0 }` discussion. context is here: https://discordapp.com/channels/688807550715560050/688851317593997489/810991521267974154

other than that tho this seems fine

fig02 (approved):

Zelllll (approved):

include/z64map_mark.h
```diff
@@ -0,0 +1,27 @@
+#ifndef _Z64MAP_MARK_H_
```
Roman971, line 1: Imo there isn't really a need to do a separate header for just these few things right now.
We'll probably have a dedicated header for this system once we restructure everything, but there is a good chance its scope will be different and that it won't be in `include/`, so I suggest keeping it in `z64.h` for now.

include/z64map_mark.h
```diff
@@ -0,0 +1,27 @@
...
+
+typedef struct {
+    /* 0x00 */ s8 markType; // 0 for the chest icon, 1 for the boss skull icon, -1 for none
+    /* 0x01 */ u8 count; // number of icons to display
+    /* 0x02 */ MapMarkPoint points[12];
+} MapMarkIconData; // size = 0x26
+
+typedef struct {
+    MapMarkIconData icons[3];
+} MapMarkData; // size = 0x72
```
Roman971, line 23: Is there a reason to create a wrapping struct for this? I feel like it adds additional braces and makes things a bit confusing in code for no real gain, but I might be missing something?

Roman971 (commented):
Sorry for the late review, but here are my few comments on this.

Also, after thinking for a bit, I feel like using `{ 0 }` would be completely fine here, and we could probably shorten this data even more tbh (tho that might require disabling some warnings in this kind of files, not sure).

src/overlays/ovl_map_mark_data/z_map_mark_data.c
```diff
@@ -0,0 +1,12703 @@
+#include "global.h"
+
+MapMarkData mapMarkDekuTree[] = {
+    // Deku Tree map 0
```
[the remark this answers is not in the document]
mzxrules: Dungeon maps are not directly indexed by room index; Inside the Deku Tree for example has minimaps 0x00, 0x0B, and 0x0C, which correspond to different floors of room 0.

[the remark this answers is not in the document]
Roman971: I see, I wasn't aware of that. Nvm then I guess

fig02:
Still dont like the use of { 0 } personally, but after thinking about it more its prob fine if the intention is for this to be generated by tools anyway. So I will stop fighting about this FeelsOkayMan

include/z64map_mark.h
mzxrules: if z64 is being split, i see no reason not to split it in pieces as we go if you have a good sense of how to split it.

include/z64map_mark.h
mzxrules: mainly to eliminate this pattern: ```mapMarkData = &sLoadedMarkDataTable[dungeon][interfaceCtx->mapRoomNum][0]```

Roman971: I'm not sure I see the issue with this pattern, or how using a wrapping struct helps

include/z64map_mark.h
fig02: would normally agree to splitting it out for larger systems that get references alot. but romans last point here is the main thing `but there is a good chance its scope will be different and that it won't be in include/, so I suggest keeping it in z64.h for now.`

include/z64map_mark.h
fig02: same, can you give an example

fig02 (approved):

Roman971 (approved):

PR #706, opened 2021-02-17
Minor gfx cleanups in some actors

Roman971:
Nothing major but I noticed some issues with the names and defines used in gfx files, so I figured we could fix them now since they are pretty straightforward.


fig02 (approved):

src/overlays/actors/ovl_Oceff_Wipe/z_oceff_wipe.c
```diff
@@ -124,7 +124,7 @@ void OceffWipe_Draw(Actor* thisx, GlobalContext* globalCtx) {
     gSPDisplayList(POLY_XLU_DISP++, sTextureDL);
     gSPDisplayList(POLY_XLU_DISP++, Gfx_TwoTexScroll(globalCtx->state.gfxCtx, 0, 0 - scroll, scroll * (-2), 32, 32, 1,
                                                      0 - scroll, scroll * (-2), 32, 32));
-    gSPDisplayList(POLY_XLU_DISP++, sFrustrumDl);
+    gSPDisplayList(POLY_XLU_DISP++, sFrustrumDL);
```
EllipticEllipsis, line 127: Can we correct the spelling of "Frustum" in various places in the repo in this PR as well?

Roman971: Good catch, I didn't even notice that 😅

EllipticEllipsis (approved):

PR #707, opened 2021-02-18
ovl_en_yabusame_mark, 1 nonmatching

emilybrooks:


src/overlays/actors/ovl_En_Yabusame_Mark/z_en_yabusame_mark.h
```diff
@@ -6,9 +6,19 @@
...
-    /* 0x014C */ char unk_14C[0xC4];
+    /* 0x014C */ EnYabusameMarkActionFunc actionFunc;
+	/* 0x0150 */ s16 typeIndex;
+	/* 0x0152 */ s16 subTypeIndex;
+	/* 0x0154 */ Vec3f worldPos;
+	/* 0x0160 */ Vec3f vertexA;
+	/* 0x016C */ Vec3f vertexB;
+	/* 0x0178 */ Vec3f vertexC;
+	/* 0x0184 */ Vec3f vertexD;
+	/* 0x0190 */ ColliderQuad collider;
```
Roman971, line 21: I think some tab characters were inserted here, which should be replaced with proper spaces, ideally

src/overlays/actors/ovl_En_Yabusame_Mark/z_en_yabusame_mark.c
```diff
@@ -40,17 +29,221 @@ static ColliderQuadInit D_80B434F0 = {
...
+    FLAGS,
+    OBJECT_GAMEPLAY_KEEP,
+    sizeof(EnYabusameMark),
+    (ActorFunc)EnYabusameMark_Init,
+    (ActorFunc)EnYabusameMark_Destroy,
+    (ActorFunc)EnYabusameMark_Update,
+    NULL,
+};
+
+Vec3f collisionVerticies[] = {
```
Roman971, line 51:
```suggestion
static Vec3f sCollisionVertices[] = {
```

src/overlays/actors/ovl_En_Yabusame_Mark/z_en_yabusame_mark.c
```diff
@@ -40,17 +29,221 @@ static ColliderQuadInit D_80B434F0 = {
...
+
+    Collider_DestroyQuad(globalCtx, &this->collider);
+}
+
+void EnYabusameMark_Init(Actor* thisx, GlobalContext* globalCtx) {
+    EnYabusameMark* this = THIS;
+
+    osSyncPrintf("\n\n");
+    osSyncPrintf(VT_FGCOL(GREEN) "☆☆☆☆☆ やぶさめまと ☆☆☆☆☆ %x\n" VT_RST, this->actor.params);
+    this->actor.flags &= -2;
```
Roman971, line 105:
```suggestion
    this->actor.flags &= ~1;
```

src/overlays/actors/ovl_En_Yabusame_Mark/z_en_yabusame_mark.c
```diff
@@ -40,17 +29,221 @@ static ColliderQuadInit D_80B434F0 = {
...
 #pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Yabusame_Mark/func_80B42F74.s")
+#endif
+
+void EnYabusameMark_Update(Actor* thisx, GlobalContext* globalCtx) {
+    EnYabusameMark* this = THIS;
+    Vec3f* vertexArray;
+    u32 arrayIndex;
+
+    this->actionFunc(this, globalCtx);
+    arrayIndex = this->typeIndex * 4;
```
Roman971, line 223: Is this temp definitely required to match? I feel like it should be possible to omit it.

src/overlays/actors/ovl_En_Yabusame_Mark/z_en_yabusame_mark.c
```diff
@@ -40,17 +29,221 @@ static ColliderQuadInit D_80B434F0 = {
...
+s32 D_80B435F0 = 0x45536000;
+s32 D_80B435F4 = 0x44D8C000;
+s32 D_80B435F8[] = {
+    0xC59A9000, 0x45520000, 0x44D8C000, 0x43F78000, 0x458D2800, 0x44D24000, 0xC4DE6000,
+    0x458D2800, 0x44D24000, 0xC52FD000, 0x458D5000, 0x44D7E000, 0xC50F8000,
+};
+s32 D_80B4362C[] = {
+    0x41A00000, 0x42200000, 0x42700000, 0x44424000, 0x42200000, 0x42A00000,
+    0x42F00000, 0x44424000, 0x42200000, 0x42F00000, 0x43200000, 0x44424000,
+};
```
Roman971, line 88: I assume this data is in a `NON_MATCHING` block because the asm function references fake symbols like `D_80B435F4` and `D_80B435F8`, but you shouldn't actually need to do this. I think you could replace the references to `D_80B435F4` by `D_80B435F0+4` and `D_80B435F8` by `D_80B435F0+8` in the asm, so you no longer need a `NON_MATCHING` block for the data.

spec
```diff
@@ -3179,8 +3179,11 @@ endseg
 beginseg
     name "ovl_En_Yabusame_Mark"
     include "build/src/overlays/actors/ovl_En_Yabusame_Mark/z_en_yabusame_mark.o"
-    include "build/data/overlays/actors/z_en_yabusame_mark.data.o"
+#ifdef NON_MATCHING
+	include "build/src/overlays/actors/ovl_En_Yabusame_Mark/ovl_En_Yabusame_Mark_reloc.o"
```
Roman971, line 3183: I think there is a tab character here as well (see my first comment)

Roman971 (commented):
Well done with this actor. Just a few things

src/overlays/actors/ovl_En_Yabusame_Mark/z_en_yabusame_mark.c
EllipticEllipsis: Could this be a 2d array, or does the codegen not allow that?

emilybrooks: this array access was really picky, I couldn't get the right combination of sll instructions with a 2d array, and the regalloc is wrong without the temp.

fig02: im not sure of the context cause i didnt read the code yet, but smells like addition being done on the index instead of a 2d array

Roman971 (approved):

src/overlays/actors/ovl_En_Yabusame_Mark/z_en_yabusame_mark.c
mzxrules: i suspect that collisionVerticies (should be Vertices btw) might be an array of Quads. I'm not entirely sure why such a type hasn't been created yet.

Roman971: Oh yeah I missed that collisionVerticies wasn't renamed. It should be changed to `sCollisionVertices`

src/overlays/actors/ovl_En_Yabusame_Mark/z_en_yabusame_mark.c
```diff
@@ -40,17 +29,208 @@ static ColliderQuadInit D_80B434F0 = {
...
+    OBJECT_GAMEPLAY_KEEP,
+    sizeof(EnYabusameMark),
+    (ActorFunc)EnYabusameMark_Init,
+    (ActorFunc)EnYabusameMark_Destroy,
+    (ActorFunc)EnYabusameMark_Update,
+    NULL,
+};
 
+static Vec3f collisionVerticies[] = {
+    { 70.0f, 70.0f, 0.0f },      { 70.0f, -70.0f, 0.0f },      { -70.0f, 70.0f, 0.0f },    { -70.0f, -70.0f, 0.0f },
```
fig02, line 52: like roman mentioned in the other comment, this needs an `s` prefix then its good to go

Roman971 (approved):

fig02 (approved):

PR #711, opened 2021-02-27
Add Nix configuration

s5bug:
Adds a `shell.nix` file that Nix users can use for the project's dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,8 @@ sudo apt-get update
 sudo apt-get install git build-essential binutils-mips-linux-gnu python3
 ```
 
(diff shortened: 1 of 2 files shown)


fig02:
Given how niche this is, it may be hard for us to maintain in the future if our build flags or dependencies change (which will definitely happen when we support other oot versions). Since not many people use this, it would be hard to guarantee that we can test it.

What we can do though is pin the nix file in the discord. If down the line someone comes along who'd like this as well, we can point them to that, and they can make whatever updates are needed. Let me know if this works for you

fig02:
If thats cool with you you can drop a message in the #oot-misc channel with the file and ill pin it

s5bug:
I'm willing to maintain it. I'll send the file there for sure.

PR #714, opened 2021-03-06
object_bg and object_mir_ray OK

EllipticEllipsis:
EnBomBowlMan and MirRay's objects decompiled. Everything except vertices named.

I did my best with the names, but it's pretty hard for some multi-purpose textures.

I have not done enums for EnBomBowlMan's limbs since only the head is referenced, but it would be pretty easy to get them from the named ones in the xml. Also did some minor cleanup on stuff that particularly stood out.

Also fixed a couple of warnings I saw in Obj_Tsubo.


fig02 (approved):

Roman971 (approved):

fig02:
ready to merge, just a small undefined syms conflict to resolve
(imo theres no need to even remove stuff from undefined syms, the file is gonna be thrown out all together anyway when we're done with it)

PR #715, opened 2021-03-07
Tutorial docs

EllipticEllipsis:
Finally PRing these. Any comments welcome.

I know a few of the screenshots are out of date. I can redo any where this matters. I think the text is aligned with current practice.

Some of the docs are more skeletal than others; someone with appropriate knowledge may like to put the dlists on the limbs.


docs/tutorial/contents.md
```diff
@@ -0,0 +1,62 @@
...
+		- Initchains
+		- Actors and dynapoly actors
+		- Colliders
+		- Skelanime
+	
+	- Matching
+		- Using diff
+		- control flow (branches) -> instruction ordering -> register allocation -> stack
+
+- [The rest of the function in the actor](other_functions.md)
```
fig02, line 26: function -> functions

docs/tutorial/helper_scripts.md
```diff
@@ -0,0 +1,176 @@
...
+Collider types supported are
+
+- `ColliderJntSphInit`
+- `ColliderCylinderInit`
+- `ColliderTrisInit`
+- `ColliderQuadInit`
+- `ColliderJntSphItemInit`
+- `ColliderTrisItemInit`
+
+and `num` is used only for `ColliderJntSphItemInit`.
```
fig02, line 110: Items has since changed to Elements

docs/tutorial/types_structs_padding.md
```diff
@@ -0,0 +1,148 @@
...
+Here are the usual names and the sizes of some of the most common structs used in actors and their structs:
+| Type                 | Usual name            | Size            |
+| -------------------- | --------------------- | --------------- |
+| `Actor`              | `actor`               | 0x14C           |
+| `DynaPolyActor`      | `dyna`                | 0x164           |
+| `Vec3f`              |                       | 0xC             | 
+| `Vec3s`              |                       | 0x6             | 
+| `SkelAnime`          | `skelAnime`           | 0x44            |
+| `Vec3s[limbCount]`   | `limbDrawTable`       | 0x6 * limbCount |
+| `Vec3s[limbCount]`   | `transitionDrawTable` | 0x6 * limbCount |
```
fig02, line 50: jointTable and morphTable

docs/tutorial/types_structs_padding.md
```diff
@@ -0,0 +1,148 @@
...
+| `ColliderCylinder`   | `collider`            | 0x4C            |
+| `ColliderQuad`       | `collider`            | 0x80            |
+| `ColliderJntSph`     | `collider`            | 0x20            |
+| `ColliderJntSphItem` | `colliderItems[n]`    | 0x40 * n        |
+| `ColliderTris`       | `collider`            | 0x20            |
+| `ColliderTrisItem`   | `colliderItems[n]`    | 0x5C * n        |
```
fig02, line 56: same comment about elements from earlier

fig02 (commented):
just a few things to bring it up to date with the current state of the repo. i wasn't able to read every file closely, but got through a good portion of it. we can make other changes as needed at a later time if anything

docs/tutorial/types_structs_padding.md
EllipticEllipsis: Shall I change the default name too? I see it hasn't been batch-renamed in the repo yet.

docs/tutorial/types_structs_padding.md
EllipticEllipsis: I'll change this in En_Skb too, since that seems to be the only other place the old names are still used.

fig02 (approved):

docs/tutorial/introduction.md
```diff
@@ -0,0 +1,74 @@
...
+    this->onFire = 1;
+    this->actor.naviEnemyId = 0x11;
+}
+```
+
+which is intended to be as close to the original code as possible. We are doing *matching* decomp: in the right context, and with the right compiler settings, the above C compiles into *precisely* the assembly code above, not just equivalent code.
+
+N.B. We are using only publicly available code. In particular, we are not looking at any of the recent Nintendo source code leaks. (These apparently contain very little Ocarina of Time material anyway.)
+
+Progress of the project can be found at [https://zelda64.dev]. The long-term goal of this project is to obtain a complete compilable version of the code for every publicly released version of Ocarina of Time (there are also sister projects for Majora's Mask and other Zelda games). *We are not working on a PC Port, and this project will not be making one*, although the resulting code will be very useful if someone does intend to make such a port.
```
AngheloAlf, line 47: Why is the link to zelda64.dev in brackets?

docs/tutorial/vscode.md
```diff
@@ -0,0 +1,26 @@
...
+- bracket pair colorizer 2
+- better mips support
+
+
+
+You will probably also find it helpful to watch Fig's video, although it was not meant to be an actual introduction. But it does show various useful features of VSCode, quite apart from the decompilation aspect.
+
+Useful stuff to know:
+
+- Ctrl + Alt + Up/Down gives multicursors across consecutive lines. If you want several cursors in a more diverse arrangement, middle clicking works, at least on Windows.
```
AngheloAlf, line 18: For some reason, in Linux this shortcut is `Ctrl` + `Alt` + `Up`/`Down`

docs/tutorial/helper_scripts.md
```diff
@@ -0,0 +1,176 @@
...
+./tools/sfxconvert.py <path_to_file> <path_to_repo>
+```
+
+Optional arguments are `-o output` to output to a different file and `-v` to give verbose output (i.e. tell you what changes it has made).
+
+## vtfmt
+
+This turns the strange strings in the `osSyncPrintf`s into the human-readable equivalent instructions. Copy the contents, including the quotation marks, and run
+```sh
+./tools/vtfmt.py "contents"
```
AngheloAlf, line 125:
```suggestion
./tools/vt_fmt.py "contents"
```

docs/tutorial/helper_scripts.md
```diff
@@ -0,0 +1,176 @@
...
+
+## format
+
+Shell script that does a standardised format to the C code. Can be run on a file, a directory, or the whole codebase. Run this before you submit a PR.
+
+## find_unused_asm
+
+Tracks down any `.s` files no longer used by the project. Does not ignore comments, so you have to actually remove any `#pragma` lines for it to consider the file unused.
+```sh
+./tools/find_unused_asm.py
```
AngheloAlf, line 149:
```suggestion
./tools/find_unused_asm.sh
```

docs/tutorial/helper_scripts.md
```diff
@@ -0,0 +1,176 @@
...
+
+This converts the cutscene data into macros that the cutscene system uses. Cutscenes are generally very long, so I recommend sending the output straight to a file with `>`, rather than trying to copy it all from the terminal. Run
+```sh
+./tools/csdis.py <address>
+```
+on the address from the `D_address` containing the cutscene data.
+
+## regconvert
+
+This converts the direct memory references, of the form `gGameInfo + 0x<offset>`, into the corresponding REG macros defined in [regs.h](../include/regs.h). Run
```
AngheloAlf, line 163: Currently, this script will works with `gGameInfo->data` indexes by default. For offsets, you need to pass the flag `--offset`. It may be worth noting the flag `-file` which replaces automatically each occurrence of that array in a file.

AngheloAlf (changes requested):
Very good job. Just a few small suggestions.

docs/tutorial/beginning_decomp.md
```diff
@@ -0,0 +1,926 @@
...
+
+```C
+Actor_ProcessInitChain(&this->actor, &D_80A88CE0);
+```
+
+which initialises common properties of actor using an InitChain, which is usually somewhere near the top of the data, in this case in the variable `D_80A88CE0`. Although we don't need to do this now since we we will extern the data, we might as well work out what it is now. Fortunately, we have a script to do this.
+
+The InitChain script is also in the tools directory, and is called `ichaindis.py`. Simply passing it the ROM address will spit out the entire contents of the InitChain, in this case:
+
+```
```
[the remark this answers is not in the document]
EllipticEllipsis: I don't think syntax highlighting really matters for any of these. Some of them probably have `sh` that I didn't remove, actually.

docs/tutorial/contents.md
```diff
@@ -0,0 +1,62 @@
+# Getting started
+
+## [Introduction to decomp](introduction.md)
+- What we are doing
+- Structure of the code
+
+## Pre-decompilation
+- Building the repo (follow the instructions in the README.md)
+- Most of us use VSCode. (You can watch Fig's video to get an idea of how this can be used). Some useful information is [here](vscode.md).
```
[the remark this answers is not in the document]
EllipticEllipsis: If @fig02 's happy with that. I've erred on the side of sending people to Discord for things like this.

docs/tutorial/introduction.md
EllipticEllipsis: This is markdown for a bare link without diferent text: https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links

docs/tutorial/vscode.md
EllipticEllipsis: VS Code's not very consistent across OSes, but is that actually different?

AngheloAlf: Yes, it is different.
image

Also, the other shortcuts are the same as you described them.

AngheloAlf (approved):

docs/tutorial/helper_scripts.md
```diff
@@ -0,0 +1,181 @@
...
+## first_diff
+
+Tells you where your built rom first differs from the baserom. It gives you a memory address that you can use to do, e.g. a binary diff, and also tries too find what function or data this address is in. Run with 
+```C
+./first_diff.py
+```
+
+If the rom is shifted, the first problem will be in gDMADataTable. Ignore this and look at the next one for where you actually need to look to see what's happened. The last line makes a guess on this location you need to edit to fix the problem.
+
+## syminfo
```
Roman971, line 76:
```suggestion
## sym_info
```

docs/tutorial/helper_scripts.md
```diff
@@ -0,0 +1,181 @@
...
+
+Gives information about a `D_address` symbol (ROM address, RAM address, file). Run
+```C
+./syminfo.py <D_number>
```
Roman971, line 80:
```suggestion
./sym_info.py <D_number>
```

docs/tutorial/helper_scripts.md
```diff
@@ -0,0 +1,181 @@
...
+
+This is used to convert the data associated to the `D_address` in
+```C
+Actor_ProcessInitChain(&this->actor, &D_address);
+```
+into an InitChain. It lives in the tools directory. Run
+```sh
+./tools/ichaindis.py <path_to_baserom> <D_address>
+```
+and copy the output. (This used to only take the ROM address, which you would need to get from `syminfo.py`. Now you can just give it the RAM address, or even the raw `D_address`.)
```
Roman971, line 93:
```suggestion
and copy the output. (This used to only take the ROM address, which you would need to get from `sym_info.py`. Now you can just give it the RAM address, or even the raw `D_address`.)
```

Roman971 (commented):
Looks good to me besides a small typo.

As a side note, we should probably remove the Getting Started page on the Wiki to avoid any confusion, since it's getting really outdated now.

Roman971 (approved):

PR #716, opened 2021-03-07
Add distclean rule to makefile

AngheloAlf:
Adds a ~`clean_everything`~ `distclean` that aims to clean all generated files not deleted by `clean`.

It does:
- Delete every file that `make clean` deletes.
- Delete every extracted asset in `assets/`. It doesn't deletes the `assets/xml/` folder.
- Delete `baserom/`
- Delete the generated/compiled tools and files in `tools/`. It only invokes the ~`clean_everything`~ `distclean` rule of the tool's makefile.
  - This will also delete every file generated when compiling ZAPD.

I did this because sometimes your local copy can break because of an residual file or something, and just running `make clean && make setup -j && make -j` will not fix the issue, but erasing every generated file may fix it.

Let me know if I forgot any generated file.

~PS: For comedic purposes, the rule `nuke` was also added and it does the same as ~`clean_everything`~ `distclean`, so you can do `make nuke` instead if you prefer.~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -195,12 +195,9 @@ clean:
 	$(RM) -r $(ROM) $(ELF) build
 
 distclean: clean
-	$(RM) -r $(TEXTURE_BIN_DIRS)
+	$(RM) -r $(ASSET_BIN_DIRS)
 	$(RM) -r baserom/
 	$(MAKE) -C tools distclean
-
-nuke: distclean
-	@echo "Nuked."
 
 setup:
 	$(MAKE) -C tools -j
@@ -212,7 +209,7 @@ resources: $(ASSET_FILES_OUT)
 test: $(ROM)
 	$(EMULATOR) $(EMU_FLAGS) $<
 
-.PHONY: all clean setup test distclean nuke
+.PHONY: all clean setup test distclean
 
 #### Various Recipes ####
 
--- a/Makefile
+++ b/Makefile
@@ -194,12 +194,12 @@ build/undefined_syms.txt: undefined_syms.txt
 clean:
 	$(RM) -r $(ROM) $(ELF) build
 
-clean_everything: clean
+distclean: clean
 	$(RM) -r $(TEXTURE_BIN_DIRS)
 	$(RM) -r baserom/
-	$(MAKE) -C tools clean_everything
+	$(MAKE) -C tools distclean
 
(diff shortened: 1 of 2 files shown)


louist103 (approved):

fig02:
imo id name the rule `clean_all` to make it easier to type, and remove "nuke". feels like having an alias for the same thing could cause confusion and doesnt seem necessary

fig02:
ah i see its since been changed. still feel the same about the alias though

Makefile
```diff
@@ -194,6 +194,14 @@ build/undefined_syms.txt: undefined_syms.txt
 clean:
 	$(RM) -r $(ROM) $(ELF) build
 
+distclean: clean
+	$(RM) -r $(TEXTURE_BIN_DIRS)
```
Roman971, line 198: Shouldn't this be `ASSET_BIN_DIRS` instead of `TEXTURE_BIN_DIRS`?

Makefile
```diff
@@ -194,6 +194,14 @@ build/undefined_syms.txt: undefined_syms.txt
...
+	$(RM) -r baserom/
+	$(MAKE) -C tools distclean
+
+nuke: distclean
+	@echo "Nuked."
```
Roman971, line 203: I agree with Fig that this alias seems unnecessary

Makefile
Roman971: Looking at the rest of the makefile a bit closer, it seems like `TEXTURE_BIN_DIRS` is just a duplicate of `ASSET_BIN_DIRS`, so we should probably remove it and use `ASSET_BIN_DIRS` instead.
On a similar note, `TEXTURE_DIRS` seems useless and looks like it could be removed entirely.

Roman971: I can make a follow up PR after we merge this if you prefer (to clean up what I just mentioned)

AngheloAlf: I don't have problems doing that cleanup myself, I can do it if you like.

Makefile
AngheloAlf: Removed.

fig02 (approved):

Roman971 (commented):
The `TEXTURE_DIRS` declaration itself can also be removed from the file, but otherwise this looks good to me

Roman971 (approved):

PR #719, opened 2021-03-15
OnePointDemo OK

petrie911:
One point demo matched and data imported, with preliminary documentation. Added some camera-related macros to improve readability. Also cleaned up a bunch of actors as I came across them while editing OnePointDemo-related things.


src/code/z_onepointdemo.c
```diff
@@ -1,55 +1,1365 @@
...
 }
 
-void func_8007C820(OnePointDemoCamera* onePointCamera, s16 actionParameters, s16 initTimer,
-                   CutsceneCameraPoint* atPoints, CutsceneCameraPoint* eyePoints) {
-    onePointCamera->atPoints = atPoints;
-    onePointCamera->eyePoints = eyePoints;
-    onePointCamera->actionParameters = actionParameters;
-    onePointCamera->initTimer = initTimer;
+// Ends the onepointdemo in camIdx by setting its timer to 0. For attention demos, it is set to 5 instead.
+s16 OnePointDemo_EndDemo(GlobalContext* globalCtx, s16 camIdx) {
```
fig02, line 1217: function level comments can be /** */

src/code/z_onepointdemo.c
```diff
@@ -1,55 +1,1365 @@
...
+        } else if (category == parentCam->target->category) {
+            return true;
+        }
+    }
+    return false;
+}
 
-void func_80080788(UNK_TYPE arg0, UNK_TYPE arg1) {
+// unused, also empty.
+void OnePointDemo_Noop(GlobalContext* globalCtx, UNK_TYPE arg1) {
```
fig02, line 1364: id say personally we can use s32 for that unk arg, we're never gonna find out what it is.
I treat the UNK types as "filler for now and will properly replace later"

fig02 (approved):

fig02:
This was posted about in discord, but just commenting here for documentations sake: 
One change that should be made is using `cutscene` instead of `demo` wherever possible, like the new macro and such

Roman971 (approved):

fig02 (approved):

PR #723, opened 2021-03-17
PreRender decomped and equivalent (one small non-matching)

petrie911:
Finished the decomp of PreRender and cleaned it up a bit. There's still a small nonmatching, but the file is now definitely equivalent.


Zelllll (approved):
Looks good, glad to see this file finally done

include/color.h
```diff
@@ -21,4 +21,22 @@ typedef struct {
...
+    u16 rgba;
+} Color_RGB5A1;
+
+typedef union {
+    struct {
+        u8 a : 3;
+        u8 b : 5;
+    };
+    u8 c;
+} Struct_Cvg;
```
Roman971, line 40: This doesn't seem to be used anywhere?

petrie911: This was part of a matching idea that didn't go anywhere. I'll take it out.

Roman971 (approved):

fig02 (approved):

PR #728, opened 2021-03-20
Fix some fixable warnings

AngheloAlf:
I tried to fix almost all warnings. The remaining may not be fixable at all because that was how the original devs wrote the code. A shame.

~~Since it was a warning, I changed the `#include "..." EARLY` to `#pragma INCLUDE_EARLY("...")`. This way gcc will stop complaining about the `EARLY` token.
This has one downside, now we need to declare the cutscene as an `extern`, otherwise it will not compile. Feel free to suggest a better way to do this.~~ I undid this change for now, I'll try to figure out another way.

~~To silence a warning in `z_eff_ss_solder_srch_ball`, I had to do an ugly cast. Let me know if I should undo it.~~


assets/xml/objects/object_okuta.xml
```diff
@@ -20,7 +20,7 @@
         <Texture Name="gOctorokFadeToBlackTex" OutName="fade_to_black" Format="rgb5a1" Width="8" Height="8" Offset="0x30D0"/>
 
         <!-- Octorok Skeleton -->
-        <Skeleton Name="gOctorokSkel" Offset="0x3660"/>
+        <Skeleton Name="gOctorokSkel" Type="Normal" LimbType="Standard" Offset="0x3660"/>
```
EllipticEllipsis, line 23: Since you've corrected this one, could you correct the Skeleton Types listed as `Standard` to `Normal` as well? There's 6 of them:
```
assets/xml/objects/object_dekunuts.xml
assets/xml/objects/object_dnk.xml
assets/xml/objects/object_dns.xml
assets/xml/objects/object_dodongo.xml
assets/xml/objects/object_firefly.xml
assets/xml/objects/object_hintnuts.xml```

File: src/overlays/effects/ovl_Effect_Ss_Solder_Srch_Ball/z_eff_ss_solder_srch_ball.c [hunk 2]
```diff
@@ -25,7 +25,7 @@ u32 EffectSsSolderSrchBall_Init(GlobalContext* globalCtx, u32 index, EffectSs* t
     this->update = EffectSsSolderSrchBall_Update;
     this->life = 100;
     this->rUnused = initParams->unused;
-    this->actor = initParams->linkDetected; // actor field was incorrectly used as a pointer to something else
+    this->actor = (Actor*)initParams->linkDetected; // actor field was incorrectly used as a pointer to something else
```
EllipticEllipsis, line 28: This should stay imo: it's a genuine, horrible, programming mistake, and shouldn't be swept under the rug.

fig02:
yea, we agreed that the searchball effect warning should stay as is. the casts make it far more confusing to read
nothing we can really do about the dev incorrectly using that field originally

EllipticEllipsis (commented):
Fewer warnings, hooray! Couple of things.

src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -618,7 +618,7 @@ void Select_Init(GameState* thisx) {
     }
     SREG(30) = 1;
 
-    this->staticSegment = GameState_Alloc(this, size, "../z_select.c", 1114);
+    this->staticSegment = GameState_Alloc(thisx, size, "../z_select.c", 1114);
```
fig02, line 621: &this->state imo

src/overlays/gamestates/ovl_title/z_title.c
```diff
@@ -156,7 +156,7 @@ void Title_Main(GameState* thisx) {
 void Title_Destroy(GameState* thisx) {
     TitleContext* this = (TitleContext*)thisx;
 
-    Sram_InitSram(this, &this->sramCtx);
+    Sram_InitSram(thisx, &this->sramCtx);
```
fig02, line 159: same here

assets/xml/objects/object_okuta.xml
AngheloAlf: Why do we have 6 listed as `Standard`? It seems weird to me. I think there should be a reason for this :thinking:

fig02: people are mixing up Standard Limbs and "normal" skeletons. I guess ZAP just defaults to normal for skeleton if the name is unrecognized

fig02: ah this is why lol 
`else if (skelTypeStr != "Normal" && skelTypeStr != "Standard")`
ideally limbs will have a type called Standard, and skeleton is something else. I mentioned in discord possibly just making flex and skin attributes of the skeleton so we arent forced to give the base skeleton a name. still need to think it over some more though

for now I would just make sure all skeletons use Normal instead of Standard

AngheloAlf: I found the culprit. 
```py

def skeleton_to_xml(var_name,offset, type):
    skel_type = "Flex" if "Flex" in type else "Standard"
    limb_type = "Standard"

    return skeleton_xml.format(var_name, offset.lstrip('0'), skel_type, limb_type)
```
It was `tools/xmlcreate.py`.

Roman971 (approved):

PR #739, opened 2021-03-25
code_80110450 (game over) OK*

fig02:
*this file has the same issue with rodata as z_message. There is a hack in place to allow the functions to OK while leaving the rodata var in z_message's rodata.
This will allow us to tackle the rodata meme with both files at once. Tharo is working on z_message and said this is the approach we should probably take


src/code/z_gameover.c
```diff
@@ -0,0 +1,151 @@
...
+        (gameOverCtx->state >= GAMEOVER_REVIVE_RUMBLE && gameOverCtx->state < GAMEOVER_REVIVE_FADE_OUT)) {
+        Kankyo_FadeInGameOverLights(globalCtx);
+    }
+}
+
+// This file has the same issue as z_message where data is going to rodata.
+// This hack allows the function to OK, though rodata still needs to be fixed along with z_message rodata.
+extern const s16 D_80153D80[];
+#define NON_CONST(x, type) ((type*)(&x))
+#define GAME_OVER_TIMER NON_CONST(D_80153D80, s16)[0]
```
Thar0, line 20: Because of how I ended up implementing the fix on the message branch, I believe it would be fine to remove this hack now if it still matches.
Try changing the extern to simply `extern s16 D_80153D80;` and replacing all `GAME_OVER_TIMER` with `D_80153D80` directly as if it were a regular data variable. Hopefully it still matches.
If this works you can move the extern to `variables.h` as well.

fig02: it does match, yeah. Thanks.
think ill leave the extern in the file tho, as the variable isnt global and doesnt need to be accessible elsewhere

Thar0 (approved):

src/code/z_game_over.c
```diff
@@ -0,0 +1,150 @@
...
+
+            for (i = 0; i < ARRAY_COUNT(gSpoilingItems); i++) { // search for spoiling items
+                if (INV_CONTENT(ITEM_POCKET_EGG) == gSpoilingItems[i]) {
+                    INV_CONTENT(gSpoilingItemReverts[i]) =
+                        gSpoilingItemReverts[i]; // if a spoiling item is found, revert it
+                    for (j = 1; j < ARRAY_COUNT(gSaveContext.equips.buttonItems);
+                         j++) { // search c buttons for the found spoiling item
+                        if (gSaveContext.equips.buttonItems[j] == gSpoilingItems[i]) {
+                            gSaveContext.equips.buttonItems[j] =
+                                gSpoilingItemReverts[i]; // if found on a c button, revert it
```
Roman971, line 45: Formatter is breaking up these various lines because of the comments, so I suggest putting them in the lines above

Roman971 (approved):

PR #740, opened 2021-03-25
Update asm_processor and use static variables in En_Torch2

Roman971:
With this new version of `asm_processor`, static symbols are now usable inside `GLOBAL_ASM`, so we can have static variables even if they are used inside non matchings.

Note that this wasn't tested extensively so it might not always work, but at least everything is OK right now, and there are no issues in the case of `En_Torch2`.

### DIFF
--- a/docs/tutorial/data.md
+++ b/docs/tutorial/data.md
@@ -23,8 +23,6 @@ Both approaches have their advantages and disadvantages.
 <!-- Fig shows how to do this in his video.  -->
 This way is good for smaller actors with little data. It is not really suitable for EnJj, for example, because of the enormous section of data labelled as `D_80A88164`.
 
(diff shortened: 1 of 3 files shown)


fig02 (approved):

EllipticEllipsis:
Do we want to update the tutorial to reflect this? I suppose it doesn't really matter for WIP files.

Roman971:
Updating the tutorial would probably be a good idea, if there are mentions of this limitation. I can look into it myself if needed, but if you want to make a PR with an update to the tutorial that's also fine to me.

EllipticEllipsis:
I think https://github.com/zeldaret/oot/blob/master/docs/tutorial/data.md#data-first is the only explicit reference. The tutorial uses `extern` in the main example, so it's not an issue there.

Roman971:
@EllipticEllipsis I think I updated the tutorial accordingly, let me know if this is what you had in mind

EllipticEllipsis:
Yes, if you're satisfied that it'll work in general, this looks fine to me.

EllipticEllipsis (approved):

PR #749, opened 2021-03-30
named members of TitleCardContext

gamestabled:
I hope this doesnt conflict with bosses, but it does suggest changing some numbers from hex to decimal

I opted for the name "intensity" to be consistent with the name "IA8" for the texture pixel format, but it could also be "luminance" or "luminosity", i think


include/z64.h
```diff
@@ -223,14 +223,14 @@ typedef struct {
...
     /* 0x00 */ u8*      texture;
-    /* 0x04 */ s16      unk_4;
-    /* 0x06 */ s16      unk_6;
-    /* 0x08 */ u8       unk_8;
-    /* 0x09 */ u8       unk_9;
-    /* 0x0A */ u8       delayA;
-    /* 0x0B */ u8       delayB;
-    /* 0x0C */ s16      unk_C;
-    /* 0x0E */ s16      unk_E;
+    /* 0x04 */ s16      screenX;
```
Zelllll, line 226: should just be `x` imo

include/z64.h
```diff
@@ -223,14 +223,14 @@ typedef struct {
...
+    /* 0x06 */ s16      screenY;
```
Zelllll, line 227: should just be `y` imo, or you can do `xPos` and `yPos`

Zelllll (commented):
looks good besides screenX and screenY imo, these are mostly the same changes as what i made the other day that were closed, so good job

include/z64.h
gamestabled: didnt even realize that you had done that and had it closed. i did it independently, and def not trying to come off as stealing your effort (i am aware of the situation)

I'm down to change the names of x/y to whatever, is there somewhere else in the code that has already named screen coordinates to be consistent with?

Zelllll: well either way I'm just glad it's being put in, you can look at the other stuff in there closed pr and use it if you want too. Usually everywhere else just simply uses x and y.

include/z64.h
```diff
@@ -223,14 +223,14 @@ typedef struct {
...
-    /* 0x0A */ u8       delayA;
-    /* 0x0B */ u8       delayB;
-    /* 0x0C */ s16      unk_C;
-    /* 0x0E */ s16      unk_E;
+    /* 0x04 */ s16      x;
+    /* 0x06 */ s16      y;
+    /* 0x08 */ u8       width;
+    /* 0x09 */ u8       height;
+    /* 0x0A */ u8       timer;
+    /* 0x0B */ u8       delay;
```
fig02, line 231: mostly ok with this name, just wondering if it might be worth distinguishing between `delay` and `timer`. I assume `delay` is how long it takes to start appearing, and `timer` is how long its visible for.
maybe `duration` is better than timer? or just keep that names and put a comment next to them. 
others should weigh in, cause im not too sure

fig02: duration actually probably implies the whole duration, and not whats left. but yea maybe just some clarifying comments since theyre both technically timers

Roman971: Yeah I don't know about naming it `duration` since it gets decremented over time. Maybe something like `remainingTime` would be better. And I think even `timer` would be less confusing than `duration` tbh.

Roman971: Alternatively, we could go with a naming pattern like `delayTimer` and `durationTimer`, to make it clear that they both decrement over time. And in the function where they get set, we can leave the arguments as `delay`/`duration` since that's still correct at that point.

fig02: yea im cool with `delayTimer` and `durationTimer`

AngheloAlf (approved):
You would need to fix the merge conflict first, besides that looks good to me.

include/z64.h
```diff
@@ -222,15 +222,15 @@ typedef struct {
...
-    /* 0x00 */ void*    texture;
-    /* 0x04 */ s16      unk_4;
-    /* 0x06 */ s16      unk_6;
-    /* 0x08 */ u8       unk_8;
-    /* 0x09 */ u8       unk_9;
-    /* 0x0A */ u8       delayA;
-    /* 0x0B */ u8       delayB;
-    /* 0x0C */ s16      unk_C;
-    /* 0x0E */ s16      unk_E;
+    /* 0x00 */ u8*      texture;
```
Roman971, line 225: Imo this should remain a `void*`, since we don't really care if it's a `u8*`, `u64*` or something else.

In general I think we should use `void*` for pointers to textures whenever we can, so we avoid casts and limit changes if we end up using a different type for these textures later. In fact, we currently use `u64[]` for ZAPD extracted textures, but there are already discussions about changing them to `u8[]` or making it dependent on the type of texture, so it's pretty likely to change in the future.

Roman971 (approved):

fig02 (approved):

PR #753, opened 2021-04-02
Fix and rename nes_font_static

EllipticEllipsis:
Added a missing space texture to correct the bottom few textures. Also renamed most of the characters to their Unicode names ( https://www.unicode.org/Public/UCD/latest/ucd/NamesList.txt ) for consistency, and renamed the button textures to match their names in Nintendo's manuals.


assets/xml/textures/nes_font_static.xml
```diff
@@ -14,130 +14,131 @@
...
+        <Texture Name="gMsgCharButtonZTex" OutName="msg_char_button_z" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharButtonCUpTex" OutName="msg_char_button_c_up" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharButtonCDownTex" OutName="msg_char_button_c_down" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharButtonCLeftTex" OutName="msg_char_button_c_left" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharButtonCRightTex" OutName="msg_char_button_c_right" Format="i4" Width="16" Height="16"/>
         <Texture Name="gMsgCharZTargetSignTex" OutName="msg_char_z_target_sign" Format="i4" Width="16" Height="16"/>
-        <Texture Name="gMsgCharAnalogStickTex" OutName="msg_char_analog_stick" Format="i4" Width="16" Height="16"/>
-        <Texture Name="gMsgCharDPadTex" OutName="msg_char_dpad" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharControlStickTex" OutName="msg_char_control_stick" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharPlusControlPadTex" OutName="msg_char_plus_control_pad" Format="i4" Width="16" Height="16"/>
```
fig02, line 142: we use dpad elsewhere in the codebase (input macros) and its what a majority of people call it, dont think we should change that

Zelllll: agreed

EllipticEllipsis: Okay, for DPad (although worth noting that the in-game text explicitly calls it Control Pad: 
image
 ). `ControlStick` is used in `gameplay_keep` already, and we don't have to worry about regional spelling differences, so I'd rather keep that one.

mdtauk: Instruction Manuals also call it the Control Pad
image

EllipticEllipsis: Indeed, it's been called that by Nintendo from NES: https://www.manualslib.com/manual/1010406/Nintendo-Nintendo-Entertainment-System.html?page=4#manual to Wii U: https://www.nintendo.com/consumer/downloads/manual-nintendo-wiiu-operations-english.pdf .

AngheloAlf: Don't forget the Switch.
https://cdn02.nintendo-europe.com/media/downloads/support_1/nintendoswitch/NSwitch_Information_ProController_EUR.pdf
image

fig02: Yea, elliptic noted in the original comment that its from the manual. No one calls it that though, and will probably just be confused when seeing it lol

mdtauk: What about ``...Control_D_Pad...`` as a compromise?

fig02: no need imo when dpad is already well established

mzxrules: Control Pad is the official name for it though. Doesn't make sense to break from it just because there's a more popular English term.

fig02: it makes perfect sense to, in the same way we dont use official symbol names when we have something thats better/known more commonly

AngheloAlf: What official symbol name we aren't using?

fig02: search `original name:` in the codebase

mzxrules: There's a difference between symbol names in source and officially published terminology that appears in hundreds of different manuals and game dialog.

AngheloAlf: Is it really comparable a "debug variable name" to the name of a button that is it actually in the game and in every game manual?
Variables names aren't something official imo, they aren't supposed to be seen by the end user.

fig02: When 99% of people who see that name are gonna have to look it up to know what it means, it doesnt really matter how official it is or how many manuals it appears in, in my opinion. Have never heard it called Plus Control Pad ever

fig02: we're discussing the name of the variable as its going to be seen in the code. It should be as recognizable as possible so the person reading the code knows instantly what its for

assets/xml/textures/nes_font_static.xml
```diff
@@ -14,130 +14,131 @@
...
-        <Texture Name="gMsgCharCButtonTex" OutName="msg_char_c_button" Format="i4" Width="16" Height="16"/>
-        <Texture Name="gMsgCharLTriggerTex" OutName="msg_char_l_trigger" Format="i4" Width="16" Height="16"/>
-        <Texture Name="gMsgCharRTriggerTex" OutName="msg_char_r_trigger" Format="i4" Width="16" Height="16"/>
-        <Texture Name="gMsgCharZTriggerTex" OutName="msg_char_z_trigger" Format="i4" Width="16" Height="16"/>
-        <Texture Name="gMsgCharCUpButtonTex" OutName="msg_char_c_up_button" Format="i4" Width="16" Height="16"/>
-        <Texture Name="gMsgCharCDownButtonTex" OutName="msg_char_c_down_button" Format="i4" Width="16" Height="16"/>
-        <Texture Name="gMsgCharCLeftButtonTex" OutName="msg_char_c_left_button" Format="i4" Width="16" Height="16"/>
-        <Texture Name="gMsgCharCRightButtonTex" OutName="msg_char_c_right_button" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharTildeTex" OutName="msg_char_tilde" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharNoBreakSpaceTex" OutName="msg_char_no_break_space" Format="i4" Width="16" Height="16"/>
```
AngheloAlf, line 98: I think this character's name is `DEL` [according to unicode](https://www.compart.com/en/unicode/U+007F).
```suggestion
        <Texture Name="gMsgCharDelTex" OutName="msg_char_del" Format="i4" Width="16" Height="16"/>
```

AngheloAlf (changes requested):
That's a lot of renamed characters.
I would like to suggest preffixing the hex value of each character so they would be ordered and easier to find. This way could be more valuable to modders which want to add custom fonts or so.

assets/xml/textures/nes_font_static.xml
mzxrules: Control Pad is hardly super confusing terminology, and D-pad only makes sense in English.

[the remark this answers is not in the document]
mzxrules: Look again. It reads `+ Control Pad` or `+Control Pad` depending on the manual

assets/xml/textures/nes_font_static.xml
```diff
@@ -14,130 +14,131 @@
...
+        <Texture Name="gMsgCharDigitZeroTex" OutName="msg_char_digit_zero" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharDigitOneTex" OutName="msg_char_digit_one" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharDigitTwoTex" OutName="msg_char_digit_two" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharDigitThreeTex" OutName="msg_char_digit_three" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharDigitFourTex" OutName="msg_char_digit_four" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharDigitFiveTex" OutName="msg_char_digit_five" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharDigitSixTex" OutName="msg_char_digit_six" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharDigitSevenTex" OutName="msg_char_digit_seven" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharDigitEightTex" OutName="msg_char_digit_eight" Format="i4" Width="16" Height="16"/>
+        <Texture Name="gMsgCharDigitNineTex" OutName="msg_char_digit_nine" Format="i4" Width="16" Height="16"/>
```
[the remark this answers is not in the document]
mdtauk: Can I ask that some synchronisation be put in place to match the names of the textures, with the names used in code, and the names/ordering of the character width values.

It will make modding easier, but also help make it easier to read the code, and get what matches with what.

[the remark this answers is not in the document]
EllipticEllipsis: Agreed, whichever we use should be consistent between Name and OutName. I assume Unicode has a rule about using only capital letters and spaces in its names.

assets/xml/textures/nes_font_static.xml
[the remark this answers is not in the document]
EllipticEllipsis: Nintendo *always* uses the + sign when referring to it in text, e.g. https://www.zeldadungeon.net/wiki/images/9/98/Ocarina-of-Time-Master-Quest-Manual-08-09.jpg

assets/xml/textures/nes_font_static.xml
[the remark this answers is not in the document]
mdtauk: >A Unicode character is assigned a unique Name (na).[1] The name is composed of uppercase letters A–Z, digits 0–9, - (hyphen-minus) and <space>. Some sequences are excluded: names beginning with a space or hyphen, names ending with a space or hyphen, repeated spaces or hyphens, and space after hyphen are not allowed. The name is guaranteed to be unique within Unicode, and can be used to identify a code point and its character. Ideographic characters, of which there are tens of thousands, are named in the pattern "cjk unified ideograph-hhhh". For example, ``U+4E00 一 CJK UNIFIED IDEOGRAPH-4E00``. Formatting characters are named too: ``U+00A0   NO-BREAK SPACE``.
>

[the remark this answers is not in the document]
EllipticEllipsis: I'm happy to use digits 0-9. But if we use codepoints in the names as well, it won't matter nearly so much.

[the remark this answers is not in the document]
AngheloAlf: Yeah, I came with that idea later. I think I prefer the hex preffix.

assets/xml/textures/nes_font_static.xml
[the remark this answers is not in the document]
mdtauk: Not the ``+`` character, but an icon of the Control Pad.

mdtauk: Texture Name and Internal Name should be kept as close as possible just for sanity's sake

mdtauk:
> That's a lot of renamed characters.
> I would like to suggest preffixing the hex value of each character so they would be ordered and easier to find. This way could be more valuable to modders which want to add custom fonts or so.

When the project gets around to the Japanese version, you will have to deal with over 5,000 font characters.  Maybe keep one eye on how any solutions decided now, will be impacted down the line...

mzxrules:
overall it might be more sane/performant/convenient to just handle those as large sprite sheets over putting in the effort to slice and identify each character in the xml, but we'd need the font code disassembled first to make that call.

mdtauk:
> overall it might be more sane/performant/convenient to just handle those as large sprite sheets over putting in the effort to slice and identify each character in the xml, but we'd need the font code disassembled first to make that call.

I would be in favour of that, from a modding point of view, its easier to create font sheets than it is individual letters/characters.

The font textures are contiguous in the ROM AFAIU but its a tall strip vertically aligned.  So if you do implement a process to slice a large PNG file into individual 16x16 blocks of data, this will be needed.

EllipticEllipsis:
Changes in accordance with review:
- Compromise to `control_pad` for the `+`-shaped texture at the end of the file.
- Add codepoints (improves file manager sorting with proper configuration).
- Rename 0x7F, since we have no indication of what it actually should be.
image

fig02 (approved):

AngheloAlf (approved):

Zelllll (approved):
plz merge 🙃

Roman971 (approved):

PR #755, opened 2021-04-04
`assert` macro OK

AngheloAlf:
I was bored, so I made an assert macro.
Replaced every use of `__assert` with this macro.

If the macro `NDEBUG` is defined, then every assert call will be replaced to `(void) 0`. This could be useful when we add retail roms support.
If the macro `REAL_ASSERT_MACRO` (please suggest a better name) is defined, then the real filename line number and condition expression will be used instead of the matching ones. This could be useful for people modding this game in the future.

In my small testing, using `NDEBUG` leads to random crashes, probably because of shifting. Using `REAL_ASSERT_MACRO` seems to work fine, or at least it doesn't _seem_  to crash.


include/macros.h
```diff
@@ -129,4 +129,12 @@ extern GraphicsContext* __gfxCtx;
 
 #define VTX_T(x,y,z,s,t,cr,cg,cb,a) { { x, y, z }, 0, { s, t }, { cr, cg, cb, a } }
 
+#ifdef NDEBUG
+#define assert(cond, msg, file, line) ((void) 0)
```
Thar0, line 133: All our macros so far are uppercase so this should be too.

AngheloAlf: Done :D

louist103 (approved):

Zelllll (approved):
this looks good, someone should merge it (once the conflicts are fixed)

fig02 (approved):

src/code/z_elf_message.c
```diff
@@ -47,7 +47,7 @@ u32 ElfMessage_CheckCondition(ElfMessage* msg) {
 
     // "Unplanned conditions"
     LOG_STRING("企画外 条件", "../z_elf_message.c", 156);
-    __assert("0", "../z_elf_message.c", 157);
+    ASSERT(false, "0", "../z_elf_message.c", 157);
```
Roman971, line 50: I think this kind of assert just be `ASSERT(0)` like it was originally
```suggestion
    ASSERT(0, "0", "../z_elf_message.c", 157);
```

include/macros.h
```diff
@@ -129,4 +129,12 @@ extern GraphicsContext* __gfxCtx;
 
 #define VTX_T(x,y,z,s,t,cr,cg,cb,a) { { x, y, z }, 0, { s, t }, { cr, cg, cb, a } }
 
+#ifdef NDEBUG
+#define ASSERT(cond, msg, file, line) ((void) 0)
+#elif defined(REAL_ASSERT_MACRO)
+#define ASSERT(cond, msg, file, line) ((cond) ? (void) 0 : __assert(#cond, __FILE__, __LINE__))
+#else
+#define ASSERT(cond, msg, file, line) ((cond) ? (void) 0 : __assert(msg, file, line))
```
Roman971, line 137: Very minor but I think this would be a bit cleaner
```suggestion
#define ASSERT(cond, msg, file, line) ((void)0)
#elif defined(REAL_ASSERT_MACRO)
#define ASSERT(cond, msg, file, line) ((cond) ? ((void)0) : __assert(#cond, __FILE__, __LINE__))
#else
#define ASSERT(cond, msg, file, line) ((cond) ? ((void)0) : __assert(msg, file, line))
```

Roman971 (commented):
The other `ASSERT(0)` should be changed to use `0` instead of `false` imo, but otherwise this looks fine to me

Roman971 (approved):

PR #763, opened 2021-04-06
Match the last function in PreRender.c

Roman971:


louist103 (approved):

src/code/PreRender.c
```diff
@@ -520,13 +515,12 @@ void func_800C2FE4(PreRenderContext* this) {
                     u8* temp_s2 = &buffB[x - 1];
 
                     if (((HREG(80) == 0xF) ? HREG(81) : 0) == 3) {
-                        medR = MEDIAN3(temp_s0[0], temp_s0[1], temp_s0[2]);
-                        medG = MEDIAN3(temp_s1[0], temp_s1[1], temp_s1[2]);
-                        medB = MEDIAN3(temp_s2[0], temp_s2[1], temp_s2[2]);
```
petrie911, line 525: Looks cleaner to me if we leave these temps in. Perhaps they should be declared down here if so.

Roman971: There was a regalloc diff here so the temps had to be removed (at least that's the only clean way I found to fix it)

ethteck (approved):

src/code/PreRender.c
petrie911: oh, that's fine then.

fig02 (approved):

PR #783, opened 2021-04-18
Add `actorfixer.py` ~~and `graphovl.py`~~; and a few improvements to `extract_assets.py`

AngheloAlf:
- Changes in `extract_assets.py`
  - Will check if `ZAPD` crashed/failed, stop the extraction and exit with a non-zero exit value (so `make` will know if something went wrong with this step).
  - Will extract _only_ from XMLs that has been updated since the last extraction, instead of extracting all of them every time.
  - Add a `-f`/`--force` flag to force the extraction of every XML (No last-update checks will be performed). `-s` will always force.
- Add `actorfixer.py`.
- ~~Add `graphovl.py` and do a few changes~~
  - ~~`--loners` flag to include functions that are not called or call any other overlay function.~~
  - ~~Can parse enum and macro values used as actionFunc indexes for array-based actors.~~
  - ~~`-s`/`--style` to select a color style for the output image. The styles are files stored in `tools/graphovl_styles/`. The default is `solarized_dark`.~~


louist103:
I dont know if I agree with actorfixer since whenever there is a large change it will need to be updated anyway but the other tools I think are good to have.

Zelllll:
I'm not entirely sure I agree with only reextrating assets with changed xmls, if you change an asset and want to reextract it if you broke something then you have no option now, why not just simple add an option to extract a single asset?

AngheloAlf:
The option to extract a single asset already exists (`-s`).
If you broke something you can force the re-extraction of everything with `-f`/`--force`.
Extracting assets of changed XMLs would make a bit faster changing between branches or merging master into your branch.

fig02:
In regards to actorfixer @louist103, I think its fine to have in the case someone has a super out of date actor on git that they come back to and need to update. 
Ideally animfixer and math fixer would merge all into one script and we can just upkeep one script intended for mass renaming. Though these days all 3 scripts are becoming less needed, I think most if not all branches have been updated by now. But it wont hurt to keep around for a bit

fig02 (approved):

louist103 (approved):

fig02:
are we waiting on anything for this? I think this is good to go

AngheloAlf:
> are we waiting on anything for this? I think this is good to go

I think we are just waiting for Roman's review, right?

Roman971 (commented):
The extract_assets.py changes and actorfixer.py seem good to me, but imho graphovl doesn't belong in the repo, because it's not needed for decomp purposes and won't be relevant to end users. It also includes custom themes and all, so that seems better maintained separately.
I see it has its own repo now, so my question is do we really need to include it in the decomp repo? (we could keep the .gitignore entry either way, if it's still relevant)

fig02:
its a pretty useful tool for documenting. i had the same impression until trying it out when documenting en_test. was really nice to visualize the control flow

even if its removed at a later time when the repo is ready for more general purpose use, i think its nice to have for now

AngheloAlf:
Yeah, now it has its own repo.
I could either add it as a subrepo or remove it entirely. Whichever you think is the best option.
Personally I would prefer if it were part of the repo because of how useful it is to document and understand the function flow of an actor, as fig said.

Roman971:
It seems pretty useful for sure, but my point is it doesn't have to be in the repo itself.
We don't include other tools like mips2c and permuter since they have their own repos, and they are arguably much more relevant. In general I think we should only include tools that are either needed by users, or specific to oot (ie. not with their own repo). `diff.py` would be the only real exception currently, but it's so important that it makes sense imo.

fig02:
fair enough

AngheloAlf:
I removed graphovl.

Roman971 (approved):

PR #785, opened 2021-04-18
Add clean assets

louist103:
Added `make cleanassets` so that deleting tools isn't needed to clean up assets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -191,14 +191,14 @@ build/undefined_syms.txt: undefined_syms.txt
 clean:
 	$(RM) -r $(ROM) $(ELF) build
 
-cleanassets:
+assetclean:
 	$(RM) -r $(ASSET_BIN_DIRS)
 	$(RM) -r build/assets
 
-cleantools:
+toolclean:
 	$(MAKE) -C tools distclean
 
-distclean: clean cleanassets cleantools
+distclean: clean assetclean toolclean
 	$(RM) -r baserom/
 	
 
--- a/Makefile
+++ b/Makefile
@@ -195,9 +195,12 @@ cleanassets:
 	$(RM) -r $(ASSET_BIN_DIRS)
 	$(RM) -r build/assets
 
-distclean: clean cleanassets
-	$(RM) -r baserom/
+cleantools:
 	$(MAKE) -C tools distclean
+
+distclean: clean cleanassets cleantools
+	$(RM) -r baserom/
+	
 
 setup:
 	$(MAKE) -C tools -j
--- a/Makefile
+++ b/Makefile
@@ -193,6 +193,7 @@ clean:
 
 cleanassets:
 	$(RM) -r $(ASSET_BIN_DIRS)
+	$(RM) -r build/assets
 
 distclean: clean cleanassets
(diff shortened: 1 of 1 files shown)


Zelllll (approved):
Good idea, I think there could be a better name for it though, maybe make clean-all or something?

louist103:
> 
> 
> Good idea, I think there could be a better name for it though, maybe make clean-all or something?

We already have distclean which clean everything (tools, assets and runs `make clean`)

AngheloAlf:
What if `cleanassets` would also delete `build/assets/`? Just to be sure you get rid of every asset-related file.
Also, I don't like the name `cleanassets`, but that's personal taste, but I don't have any better suggestion.

fig02:
I personally dont see cleaning just assets on its own to be needed, but I may be missing some use cases. `distclean` seems fine for this job imo

AngheloAlf:
One use case could be testing changes made to ZAPD. At least I tend to delete that folder multiples times while adding stuff to ZAPD and testing everything is still OK.

fig02:
thats fair. I guess distclean works well for me cause doing setup is pretty fast, but that will vary for everyone

fig02 (approved):

louist103:
> 
> 
> I personally dont see cleaning just assets on its own to be needed, but I may be missing some use cases. `distclean` seems fine for this job imo

The idea is when IDO is built during setup (and also removed with `distclean`) it will take much longer to rebuild all the tools so doing `distclean` will get very time consuming if all you want to do is clean assets.

Roman971 (commented):
Ideally we shouldn't ever need to clean assets directly, but that will require some improvements to the process and tools, so I guess it's fine to have a command like this in the meantime.

Tho I don't think it needs to remove build/assets (seems unnecessary since files leftover in build are irrelevant), and it should probably be called `assetclean` for consistency with `distclean` and such

louist103:
Its a nice thing to have when doing testing with ZAPD or if something goes wrong with asset extraction or something like that.  Mostly for the future when IDO is built it will take much longer to rebuild everything.
I can change the names for consistency.

Roman971 (approved):
Imo the `toolclean` recipe is unnecessary, but this looks fine to me either way

fig02:
Would agree that it doesnt seem necessary

AngheloAlf (changes requested):
Could you add the new rules to `.PHONY`?

AngheloAlf (approved):

AngheloAlf:
What are we waiting for in this PR? It already has 2 leads approvals :thinking:

fig02:
as mentioned above, still dont understand the addition of a recipe that cleans tools only, and dont find it very necessary

louist103:
> 
> 
> as mentioned above, still dont understand the addition of a recipe that cleans tools only, and dont find it very necessary

Its really only there on the off chance it is needed one day.  I don't think it will be used very often though.

fig02:
id add it when and if its needed then, personally

fig02 (approved):

Roman971 (approved):

PR #791, opened 2021-04-24
Improve names of files run through the asm processor

louist103:
Files now have names resembling their actual names.  Example:
Instead of `/tmp/preprocessedypvrmuch.c` files now have names like `/tmp/z_play.cvfbopkja.c`

### DIFF
--- a/tools/asm_processor/build.py
+++ b/tools/asm_processor/build.py
@@ -28,8 +28,7 @@ del compile_args[out_ind]
 
 in_dir = os.path.split(os.path.realpath(in_file))[0]
 opt_flags = [x for x in compile_args if x in ['-g3', '-g', '-O1', '-O2', '-framepointer']]
-
-preprocessed_file = tempfile.NamedTemporaryFile(prefix='preprocessed', suffix='.c', delete=False)
+preprocessed_file = tempfile.NamedTemporaryFile(prefix=os.path.basename(os.path.normpath(in_file)), suffix='.c', delete=False)
 
 try:
     asmproc_flags = opt_flags + [in_file, '--input-enc', 'utf-8', '--output-enc', 'euc-jp']


Zelllll:
Is there any way to simply just use z_play.c?

Roman971:
The asm_processor.py we use is a straight copy from its repo (https://github.com/simonlindholm/asm-processor), so this change should probably be PR'd and merged in its master branch before porting it over.

louist103:
> 
> 
> The asm_processor.py we use is a straight copy from its repo (https://github.com/simonlindholm/asm-processor), so this change should probably be PR'd and merged in its master branch before porting it over.

I didnt actually change anything in the processor it self.   I only changed the `build.py` script.

Roman971:
I don't remember whether that file is custom or comes from the repo since I don't think we ever had to make changes to it.
If it's custom for our repo then it's fine, tho the change is probably relevant to MM as well

louist103:
As far as I can tell it does use `build.py` it uses `compile.sh`.

Thar0:
`build.py` is our own, mm uses a different script called `preprocess.py` which is pretty much the same thing.

AngheloAlf (approved):

fig02:
Seconding zels question, is there anyway to remove the preprocessor jibberish from the name entirely? 
One time zel and I were trying to compare roms of two different builds to find an issue, and the different strings for these names was creating alot of false positives

AngheloAlf:
The easiest way to remove the jibberish would be not use temp files.
It should be easy to change, but we would need to create a folder or something for the temps files. Maybe `build/temp/`?

louist103:
The nice thing about /tmp is that its a ramdisk.  So people with HDDs will have an increase in build time.

AngheloAlf:
Are you sure? I think that statement is not true unless you configured your `/tmp` folder to be used as ramdisk.

fig02:
Not being able to build and have a consistent rom with the same crc output everytime is not desirable imo

louist103:
#793 does the same thing but better.  Closing.

PR #792, opened 2021-04-24
Add scene macros

AngheloAlf:
Since ZAPD will output the cutscenes as an array of macros ([PR 124](https://github.com/zeldaret/ZAPD/pull/124)), so this macros will be needed in the repo.

The idea is to discuss the macros one last time before merging them.

When this get merged, I will PR this to MM too. So it would be nice if MM people join this discussion so we can have the same set of macros here and there.


include/z64scene.h
```diff
@@ -417,4 +419,139 @@ typedef enum {
...
+
+#define SCENECMD_WIND_SETTINGS(xDir, yDir, zDir, strength)                     \
+    { SCENECMD_ID_WIND_SETTINGS, 0, CMD_BBBB(xDir, yDir, zDir, strength) }
+
+#define SCENECMD_ENTRANCE_LIST(entranceList)                    \
+    { SCENECMD_ID_ENTRANCE_LIST, 0, CMD_PTR(entranceList) }
+
+#define SCENECMD_SPECIAL_FILES(elfMessageFile, specialObject)               \
+    { SCENECMD_ID_SPECIAL_FILES, elfMessageFile, CMD_W(specialObject) }
+
```
fig02, line 480: I dont know much about this command, but it seems odd that its expecting something called `elfMessage` specifically. Is the command for navis hints specifically, or is it intended for other files as well?
If the former then we may want a more specific name for the command. if the latter, then the args should be more general

include/z64scene.h
```diff
@@ -417,4 +419,139 @@ typedef enum {
...
+#define SCENECMD_MESH(meshHeader)                    \
+    { SCENECMD_ID_MESH, 0, CMD_PTR(meshHeader) }
+
+#define SCENECMD_OBJECT_LIST(objectNum, objectList)                 \
+    { SCENECMD_ID_OBJECT_LIST, objectNum, CMD_PTR(objectList) }
+
+#define SCENECMD_POS_LIGHT_LIST(unusedLightNum, unusedLightList)             \
+    { SCENECMD_ID_POS_LIGHT_LIST, unusedLightNum, CMD_PTR(unusedLightList) } 
+
```
fig02, line 500: We talked about it in discord but we can change this to just `LIGHT_LIST`, remove the word POS

include/z64scene.h
```diff
@@ -417,4 +419,139 @@ typedef enum {
...
+#define SCENECMD_PATH_LIST(pathList)                    \
+    { SCENECMD_ID_PATH_LIST, 0, CMD_PTR(pathList) }
+
+#define SCENECMD_TRANSI_ACTOR_LIST(transiNum, transiList)             \
+    { SCENECMD_ID_TRANSI_ACTOR_LIST, transiNum, CMD_PTR(transiList) } 
+
+#define SCENECMD_LIGHT_LIST(lightNum, lightList)                 \
+    { SCENECMD_ID_LIGHT_LIST, lightNum, CMD_PTR(lightList) }
+
```
fig02, line 509: Along with the other name changing, this can be `ENV_LIGHT_SETTINGS`

include/z64scene.h
```diff
@@ -417,4 +419,139 @@ typedef enum {
...
+    { SCENECMD_ID_CUTSCENE_LIST, cutsceneCount, CMD_PTR(cutsceneList) }
+
+#define SCENECMD_ALTERNATE_HEADER_LIST(alternateHeaderList)                    \
+    { SCENECMD_ID_ALTERNATE_HEADER_LIST, 0, CMD_PTR(alternateHeaderList) }
+
+// OoT
+#define SCENECMD_MISC_SETTINGS(camMode, worldMapLocation)               \
+    { SCENECMD_ID_MISC_SETTINGS, camMode, CMD_W(worldMapLocation) }
+// MM
+#define SCENECMD_WORLD_MAP_VISITED()                   \
```
fig02, line 541: thinking we should leave out the MM specific commands in OoT. In a modding context if someone would like to back port some of the new commands they can add it themselves. But as far as OoT goes, they arent relevant here

include/z64scene.h
```diff
@@ -417,4 +419,139 @@ typedef enum {
...
+            currRoomUnk2 | _SHIFTL(currRoomUnk5, 8, 1) | _SHIFTL(msgCtxunk12044, 10, 1) |                   \
+                _SHIFTL(enablePosLights, 11, 1) | _SHIFTL(kankyoContextUnkE2, 12, 1)                        \
+    }
+
+#define SCENECMD_UNUSED_9()                   \
+    { SCENECMD_ID_UNUSED_9, 0, CMD_W(0) }
+
+#define SCENECMD_MESH(meshHeader)                    \
+    { SCENECMD_ID_MESH, 0, CMD_PTR(meshHeader) }
+
```
fig02, line 494: While we're talking about names, I've never really liked the term "mesh" for this. Its pretty a pretty loose term and can be referencing just the geometry that makes up the maps, when really this is the whole display list. `DLISTS` or something would make more sense to me, but curious what others think

include/z64scene.h
```diff
@@ -417,4 +419,139 @@ typedef enum {
...
+    { SCENECMD_ID_WIND_SETTINGS, 0, CMD_BBBB(xDir, yDir, zDir, strength) }
+
+#define SCENECMD_ENTRANCE_LIST(entranceList)                    \
+    { SCENECMD_ID_ENTRANCE_LIST, 0, CMD_PTR(entranceList) }
+
+#define SCENECMD_SPECIAL_FILES(elfMessageFile, specialObject)               \
+    { SCENECMD_ID_SPECIAL_FILES, elfMessageFile, CMD_W(specialObject) }
+
+#define SCENECMD_ROOM_BEHAVIOR(currRoomUnk3, currRoomUnk2, currRoomUnk5, msgCtxunk12044, enablePosLights,  \
+                                kankyoContextUnkE2)                                                         \
```
fig02, line 482: this is the MM version of this macro, isnt it

fig02 (commented):
A few comments regarding names.
Also I think I would prefer `SCENE_CMD_` over `SCENECMD_` but curious what others think

Zelllll:
> A few comments regarding names.
> Also I think I would prefer `SCENE_CMD_` over `SCENECMD_` but curious what others think

yeah, I called it `SCENE_CMD_`, i have no idea why it was changed

include/z64scene.h
Zelllll: this is for navi hints specifically

include/z64scene.h
Zelllll: yeah that's the mm version

include/z64scene.h
Zelllll: DLISTS imo is too technical, I'm find with MODEL though.

include/z64scene.h
Zelllll: yeah, those were meant to only in the mm repo

include/z64scene.h
fig02: alright, then id say we can consider a better name than "special files"

include/z64scene.h
AngheloAlf: I thought it may be simpler to just have only one macro that would work for both games, but I can change it back if you prefer.

fig02: It would just lead to confusion imo, cause then you have to know what is actually supported or not. If someone wants to backport mm stuff to oot they can do that in their own projects

include/z64scene.h
mzxrules: extra files? It's kind of weird to name because the two data points are completely irrelevant. specialObject should maybe be keepObjectId or seg5ObjectId

fig02: ah i see, it sets both segment 5 and elf message. for some reason i thought that was done elsewhere.
Yeah naming that is hard. Extra is a little bit better, since I dont think theyre any more special than anything else.
And yea, agreed with `keepObjectId`

include/z64scene.h
fig02: additionally i would take `unused` out of the arg names.
And instead of lightNum it should prob be `numLights`. this goes for a few other things in this file

include/z64scene.h
```diff
@@ -417,4 +419,132 @@ typedef enum {
...
+#define SCENE_CMD_SPECIAL_FILES(elfMessageFile, keepObjectId) \
+    { SCENE_CMD_ID_SPECIAL_FILES, elfMessageFile, CMD_W(keepObjectId) }
+
+#define SCENE_CMD_ROOM_BEHAVIOR(curRoomUnk3, curRoomUnk2, showInvisActors, msgCtxUnkE40C) \
+    { SCENE_CMD_ID_ROOM_BEHAVIOR, curRoomUnk3, \
+        curRoomUnk2 | _SHIFTL(showInvisActors, 8, 1) | _SHIFTL(msgCtxUnkE40C, 10, 1) }
+
+#define SCENE_CMD_UNUSED_09() \
+    { SCENE_CMD_ID_UNUSED_09, 0, CMD_W(0) }
+
```
fig02, line 488: if this goes entirely unused do we need a macro for it? The function that corresponds to this is empty afaik

include/z64scene.h
```diff
@@ -417,4 +419,132 @@ typedef enum {
...
+#define SCENE_CMD_LIGHT_LIST(numLights, lightList) \
+    { SCENE_CMD_ID_POS_LIGHT_LIST, numLights, CMD_PTR(lightList) } 
+
+#define SCENE_CMD_PATH_LIST(pathList) \
+    { SCENE_CMD_ID_PATH_LIST, 0, CMD_PTR(pathList) }
+
+#define SCENE_CMD_TRANSI_ACTOR_LIST(numTransi, transiList) \
+    { SCENE_CMD_ID_TRANSI_ACTOR_LIST, numTransi, CMD_PTR(transiList) } 
+
```
fig02, line 503: imo id just have the full word `transition` for both the macro and the args
`SCENE_CMD_TRANSITION_ACTOR_LIST(numTransitionActors, transitionActorList)`

its a bit longer yea, but it doesnt leave you filling in the blanks on your own, and its not like this is gonna appear everywhere in our code. tools will mostly deal with this stuff

include/z64scene.h
```diff
@@ -417,4 +419,132 @@ typedef enum {
...
+#define SCENE_CMD_ENV_LIGHT_SETTINGS(numLights, lightList) \
+    { SCENE_CMD_ID_ENV_LIGHT_SETTINGS, numLights, CMD_PTR(lightList) }
+
+#define SCENE_CMD_TIME_SETTINGS(hour, min, speed) \
+    { SCENE_CMD_ID_TIME_SETTINGS, 0, CMD_BBBB(hour, min, speed, 0) }
+
+#define SCENE_CMD_SKYBOX_SETTINGS(externalTextureFileID, skyboxId, weather, lightMode) \
+    { SCENE_CMD_ID_SKYBOX_SETTINGS, externalTextureFileID, CMD_BBBB(skyboxId, weather, lightMode, 0) }
```
fig02, line 511: `externalTextureFileID` -> `externalTextureFileId`

include/z64scene.h
```diff
@@ -417,4 +419,132 @@ typedef enum {
...
+
+#define SCENE_CMD_ENTRANCE_LIST(entranceList) \
+    { SCENE_CMD_ID_ENTRANCE_LIST, 0, CMD_PTR(entranceList) }
+
+#define SCENE_CMD_SPECIAL_FILES(elfMessageFile, keepObjectId) \
+    { SCENE_CMD_ID_SPECIAL_FILES, elfMessageFile, CMD_W(keepObjectId) }
+
+#define SCENE_CMD_ROOM_BEHAVIOR(curRoomUnk3, curRoomUnk2, showInvisActors, msgCtxUnkE40C) \
+    { SCENE_CMD_ID_ROOM_BEHAVIOR, curRoomUnk3, \
+        curRoomUnk2 | _SHIFTL(showInvisActors, 8, 1) | _SHIFTL(msgCtxUnkE40C, 10, 1) }
```
AngheloAlf, line 484: About this macro, since `showInvisActors` and `msgCtxUnkE40C` are 1-bit flags, would make more sense if ZAPD outputted `true` and `false` for those?

For example:
```c
static SCmdBase ganontika_room_16Set0000[8] = {
	SCENE_CMD_ECHO_SETTINGS(6),
	SCENE_CMD_ROOM_BEHAVIOR(0x01, 0x00, true, false),
	SCENE_CMD_SKYBOX_DISABLES(0, 1),
	SCENE_CMD_TIME_SETTINGS(255, 255, 0),
	SCENE_CMD_MESH(&ganontika_room_16PolygonType2_0000B0),
	SCENE_CMD_OBJECT_LIST(5, ganontika_room_16ObjectList_000040),
	SCENE_CMD_ACTOR_LIST(6, ganontika_room_16ActorList_00004C),
	SCENE_CMD_END(),
}; 
```
instead of 
```c
static SCmdBase ganontika_room_16Set0000[8] = {
	SCENE_CMD_ECHO_SETTINGS(6),
	SCENE_CMD_ROOM_BEHAVIOR(0x01, 0x00, 1, 0),
	SCENE_CMD_SKYBOX_DISABLES(0, 1),
	SCENE_CMD_TIME_SETTINGS(255, 255, 0),
	SCENE_CMD_MESH(&ganontika_room_16PolygonType2_0000B0),
	SCENE_CMD_OBJECT_LIST(5, ganontika_room_16ObjectList_000040),
	SCENE_CMD_ACTOR_LIST(6, ganontika_room_16ActorList_00004C),
	SCENE_CMD_END(),
}; 
```

include/z64scene.h
```diff
@@ -417,4 +419,132 @@ typedef enum {
...
+    { SCENE_CMD_ID_END, 0, CMD_W(0) }
+
+#define SCENE_CMD_SOUND_SETTINGS(audioSessionId, nighttimeSfx, bgmId) \
+    { SCENE_CMD_ID_SOUND_SETTINGS, audioSessionId, CMD_BBBB(0, 0, nighttimeSfx, bgmId) }
+
+#define SCENE_CMD_ECHO_SETTINGS(echo) \
+    { SCENE_CMD_ID_ECHO_SETTINGS, 0, CMD_BBBB(0, 0, 0, echo) }
+
+#define SCENE_CMD_CUTSCENE_LIST(cutsceneCount, cutsceneList) \
+    { SCENE_CMD_ID_CUTSCENE_LIST, cutsceneCount, CMD_PTR(cutsceneList) }
```
fig02, line 529: possibly `numCutscenes` to keep it consistent with the other macros

include/z64scene.h
```diff
@@ -417,4 +419,132 @@ typedef enum {
...
+    { SCENE_CMD_ID_TEXTURE_ANIM_LIST, 0, CMD_PTR(textureAnimList) }
+
+#define SCENE_CMD_ACTOR_CUTSCENE_LIST(actorCutsceneCount, actorCutsceneList) \
+    { SCENE_CMD_ID_ACTOR_CUTSCENE_LIST, actorCutsceneCount, CMD_PTR(actorCutsceneList) }
+
+#define SCENE_CMD_MINIMAP_INFO(minimapInfo) \
+    { SCENE_CMD_ID_MINIMAP_INFO, 0, CMD_PTR(minimapInfo) }
+
+#define SCENE_CMD_MINIMAP_COMPASS_ICON_INFO(compassIconCount, compassIconInfo) \
+    { SCENE_CMD_ID_MINIMAP_COMPASS_ICON_INFO, compassIconCount, CMD_PTR(compassIconInfo) }
```
fig02, line 547: these are mm specific and should be removed

include/z64scene.h
fig02: for showinvis actors id say yes. not sure about the message context one though, cause im not sure what that variable is for

fig02: the same can apply to skybox disables

fig02: according to tharos message decomp, `msgCtxUnkE40C` is `disableWarpSongs`. can rename the arg and use true/false

include/z64scene.h
AngheloAlf: What about `POLYGON_TYPE`? The `meshHeader` ptr points to a `PolygonType0`, `PolygonType1` or `PolygonType2`

include/z64scene.h
```diff
@@ -417,4 +419,117 @@ typedef enum {
...
+    /* 0x15 */ SCENE_CMD_ID_SOUND_SETTINGS,
+    /* 0x16 */ SCENE_CMD_ID_ECHO_SETTINGS,
+    /* 0x17 */ SCENE_CMD_ID_CUTSCENE_LIST,
+    /* 0x18 */ SCENE_CMD_ID_ALTERNATE_HEADER_LIST,
+    /* 0x19 */ SCENE_CMD_ID_MISC_SETTINGS,
+    /* 0x1A */ SCENE_CMD_ID_TEXTURE_ANIM_LIST,
+    /* 0x1B */ SCENE_CMD_ID_ACTOR_CUTSCENE_LIST,
+    /* 0x1C */ SCENE_CMD_ID_MINIMAP_INFO,
+    /* 0x1D */ SCENE_CMD_ID_UNUSED_1D,
+    /* 0x1E */ SCENE_CMD_ID_MINIMAP_COMPASS_ICON_INFO
```
fig02, line 455: these can go as well

include/z64scene.h
fig02: its more than just that, it points to the display lists for the whole room. I dont think thats too technical, its just describing what it is.

I dont like model for the same reason as mesh, model sounds like just the 3d geometry when its the whole thing, materials and all.

If you dont want to use the word dlists can maybe go with simply GFX. but we use dlist to reference the same thing everywhere else, so idk

fig02 (approved):
Im fine with this as is for the time being, even if we dont come to a conclusion on some of the remaining name changes, like SPECIAL_FILES and MESH_HEADER. 
Those can be discussed in more length and decided on later if need be

include/z64scene.h
```diff
@@ -417,4 +419,112 @@ typedef enum {
...
+
+#define SCENE_CMD_PATH_LIST(pathList) \
+    { SCENE_CMD_ID_PATH_LIST, 0, CMD_PTR(pathList) }
+
+#define SCENE_CMD_TRANSITION_ACTOR_LIST(numTransitionActors, transitionActorList) \
+    { SCENE_CMD_ID_TRANSI_ACTOR_LIST, numTransitionActors, CMD_PTR(transitionActorList) } 
+
+#define SCENE_CMD_ENV_LIGHT_SETTINGS(numLights, lightList) \
+    { SCENE_CMD_ID_ENV_LIGHT_SETTINGS, numLights, CMD_PTR(lightList) }
+
```
fig02, line 498: lightList -> lightSettingsList

include/z64scene.h
```diff
@@ -417,4 +419,112 @@ typedef enum {
...
+#define SCENE_CMD_ENTRANCE_LIST(entranceList) \
+    { SCENE_CMD_ID_ENTRANCE_LIST, 0, CMD_PTR(entranceList) }
+
+#define SCENE_CMD_SPECIAL_FILES(elfMessageFile, keepObjectId) \
+    { SCENE_CMD_ID_SPECIAL_FILES, elfMessageFile, CMD_W(keepObjectId) }
+
+#define SCENE_CMD_ROOM_BEHAVIOR(curRoomUnk3, curRoomUnk2, showInvisActors, msgCtxUnkE40C) \
+    { SCENE_CMD_ID_ROOM_BEHAVIOR, curRoomUnk3, \
+        curRoomUnk2 | _SHIFTL(showInvisActors, 8, 1) | _SHIFTL(msgCtxUnkE40C, 10, 1) }
```
fig02, line 479: `msgCtxUnkE40C` can be `disableWarpSongs`
this is using knowledge from tharos decomp which isnt in the repo yet, but we are very sure thats what it is

fig02 (commented):
sry missed these last time, this should be it from me

include/z64scene.h
AngheloAlf: If that's the case, then using `true`/`false` would make sense, right?
Also, do you know if this applies for MM too?

fig02: yea, thats what i was saying here https://github.com/zeldaret/oot/pull/792#discussion_r619849645
but i guess it got resolved before you saw it

MM doesnt have warp songs, so I'm not sure. it might disable song of soaring, but someone would need to confirm

include/z64scene.h
fig02: the num arg should be changed as well. i edited shortly after so you might have saw it before that, apologies

AngheloAlf: Now should be ok, right?

fig02 (approved):
for real this time FeelsOkMan

include/z64scene.h
AngheloAlf: If that's the case, then `DLISTS` sounds good to me. But I would like to hear if anybody else has something to say about this

include/z64scene.h
Zelllll: Yeah either way I'm not as concerned with MM until it gets done there

include/z64scene.h
Zelllll: Model or mesh is pretty obvious what it does though, even if there is a 2% chance that someone might see it as something else, I think it's more clear to people who don't know all the technical details about the game etc like what a displaylist is

include/z64scene.h
```diff
@@ -417,4 +419,112 @@ typedef enum {
     /* 0x6E */ SCENE_ID_MAX
 } SceneID;
 
+// Scene commands
+
+typedef enum {
+    /* 0x00 */ SCENE_CMD_ID_SPAWN_LIST,
+    /* 0x01 */ SCENE_CMD_ID_ACTOR_LIST,
+    /* 0x02 */ SCENE_CMD_ID_ACTOR_CUTSCENE_CAM_LIST,
```
Zelllll, line 427: this is not used in oot so it should just be UNUSED_02

include/z64scene.h
```diff
@@ -417,4 +419,112 @@ typedef enum {
...
+#define SCENE_CMD_TRANSITION_ACTOR_LIST(numTransitionActors, transitionActorList) \
+    { SCENE_CMD_ID_TRANSI_ACTOR_LIST, numTransitionActors, CMD_PTR(transitionActorList) } 
+
+#define SCENE_CMD_ENV_LIGHT_SETTINGS(numLightSettings, lightSettingsList) \
+    { SCENE_CMD_ID_ENV_LIGHT_SETTINGS, numLightSettings, CMD_PTR(lightSettingsList) }
+
+#define SCENE_CMD_TIME_SETTINGS(hour, min, speed) \
+    { SCENE_CMD_ID_TIME_SETTINGS, 0, CMD_BBBB(hour, min, speed, 0) }
+
+#define SCENE_CMD_SKYBOX_SETTINGS(externalTextureFileId, skyboxId, weather, lightMode) \
```
Zelllll, line 502: this is not correct, the oot macro should be this:

```
#define SCENE_CMD_SKYBOX_SETTINGS(skyboxId, weather, lightMode) \
    { SCENE_CMD_ID_SKYBOX_SETTINGS, 0, CMD_BBBB(skyboxId, weather, lightMode, 0) }```

File: include/z64scene.h [hunk 18]
```diff
@@ -417,4 +419,112 @@ typedef enum {
...
+#define SCENE_CMD_END() \
+    { SCENE_CMD_ID_END, 0, CMD_W(0) }
+
+#define SCENE_CMD_SOUND_SETTINGS(audioSessionId, nighttimeSfx, bgmId) \
+    { SCENE_CMD_ID_SOUND_SETTINGS, audioSessionId, CMD_BBBB(0, 0, nighttimeSfx, bgmId) }
+
+#define SCENE_CMD_ECHO_SETTINGS(echo) \
+    { SCENE_CMD_ID_ECHO_SETTINGS, 0, CMD_BBBB(0, 0, 0, echo) }
+
+#define SCENE_CMD_CUTSCENE_LIST(numCutscene, cutsceneList) \
```
Zelllll, line 520: this is not correct either, the oot macro should be this:

```#define SCENE_CMD_CUTSCENE_DATA(cutsceneData) \
    { SCENE_CMD_ID_CUTSCENE_DATA, 0, CMD_PTR(cutsceneData) }

**Review by Zelllll — CHANGES_REQUESTED** [review]
Please fix the issues I mentioned in a comment earlier before this gets merged.

File: include/z64scene.h (hunk 18 above)
Comment (fig02) [reply include/z64scene.h:520]: In that case, I think we could make that more clear with the name. its really more of the "default" cutscene for the scene, since if you want to play something else youd have to set the csSegment yourself in code somewhere else
`SCENE_CMD_DEFAULT_CUTSCENE` and `defaultCutscene` for the arg

File: include/z64scene.h (hunk 17 above)
Comment (fig02) [reply include/z64scene.h:502]: what you have as `lightMode` is actually a flag for whether the area is indoors or not. Its true that it does affect lighting, but the context its used in kankyo it makes more sense as an indoor/outdoor flag

Id change it do `isIndoors` and use true/false in the output

File: include/z64scene.h [hunk 19]
```diff
@@ -456,9 +456,6 @@ typedef enum {
 #define SCENE_CMD_ACTOR_LIST(numActors, actorList) \
     { SCENE_CMD_ID_ACTOR_LIST, numActors, CMD_PTR(actorList) }
 
-#define SCENE_CMD_CAM_LIST(numCams, camList) \
-    { SCENE_CMD_ID_ACTOR_CUTSCENE_CAM_LIST, numCams, CMD_PTR(camList) }
```
Zelllll: This didn't need to be removed, just replaced with a more accurate macro (see the original scene_header.h that I posted) on the discord

include/z64scene.h
```diff
@@ -499,8 +496,8 @@ typedef enum {
 #define SCENE_CMD_TIME_SETTINGS(hour, min, speed) \
     { SCENE_CMD_ID_TIME_SETTINGS, 0, CMD_BBBB(hour, min, speed, 0) }
 
-#define SCENE_CMD_SKYBOX_SETTINGS(externalTextureFileId, skyboxId, weather, lightMode) \
-    { SCENE_CMD_ID_SKYBOX_SETTINGS, externalTextureFileId, CMD_BBBB(skyboxId, weather, lightMode, 0) }
+#define SCENE_CMD_SKYBOX_SETTINGS(skyboxId, weather, isIndoors) \
```
Zelllll, line 499: This isn't really accurate, "indoors" is just a different lighting mode, hence the original name I gave it

include/z64scene.h
```diff
@@ -517,8 +514,8 @@ typedef enum {
 #define SCENE_CMD_ECHO_SETTINGS(echo) \
     { SCENE_CMD_ID_ECHO_SETTINGS, 0, CMD_BBBB(0, 0, 0, echo) }
 
-#define SCENE_CMD_CUTSCENE_LIST(numCutscene, cutsceneList) \
-    { SCENE_CMD_ID_CUTSCENE_LIST, numCutscene, CMD_PTR(cutsceneList) }
+#define SCENE_CMD_DEFAULT_CUTSCENE(defaultCutscene) \
+    { SCENE_CMD_ID_CUTSCENE_LIST, 0, CMD_PTR(defaultCutscene) }
```
Zelllll, line 518: Not sure if I'd call this "default" cutscene

Zelllll (changes requested):

include/z64scene.h
AngheloAlf: If it is unused, why have a macro for it? Does it even works?

include/z64scene.h
Thar0: There is only one scripted cutscene per scene setup in oot so default should be omitted, as there is only one.

AngheloAlf: So `cutsceneData` should be fine?

include/z64scene.h
Zelllll: It still has functionality so it should be supported even if functionally it does not do much

include/z64scene.h
Zelllll: yes, and the command name should reflect it as well

include/z64scene.h
AngheloAlf: Does make sense for this value for having a different value other than 0/1?

Zelllll: Hmmm the only valid values are 0 and 1 so admittedly I'm not entirely sure, fig worked on decompiling some of this stuff so perhaps he is more knowledgable, but to understanding the "indoor" mode doesn't use directional lights (which is used by the sun and moon) so calling it indoor *technically* makes sense although I don't think it's the best name for this, making it true or false is probably fine but I think we should maybe understand it a little better in order to properly name it.

AngheloAlf: We can always cameback and rename this in the future when we have more knowledge about how this works. So I think this name is fine for now.

include/z64scene.h
AngheloAlf: From what I'm seeing, it only sets a pointer, but that pointer doesn't seem to be used in the whole repo to date.
Also, from reading the comments of that function, we know it was supposed to be a camera list for cutscenes ([here](https://github.com/zeldaret/oot/blob/0b8252cfe99651b3174ac8652ba133781e8e2105/src/code/z_scene.c#L208)), so if we will keep this macro, it should have an appropriate name, and maybe leave a comment saying it is unused and it functionality is non-implemented

Zelllll: Well the thing is that we *don't* know. That information came from it's use case in majora's mask, but there is zero proof of it being intended to do that in oot which is why I don't not suggest naming it as such, especially when it is meant to be used in conjunction with command 1B which does not exist in OoT at all.

AngheloAlf: If that the case, then we should remove that comment and change the name of `SCmdCsCameraList` and `csCameraList` to something more appropriate too.

Zelllll: yes that needs to be changed, which you can do in this pr if you'd like, eventually I'm gonna do some docs on z_scene too

AngheloAlf: Changed.

Zelllll (approved):
Alright we should be good now

include/z64scene.h
fig02: indoors is perfectly fine for this, ive already named it that in my kankyo branch.

fig02: you say its not accurate in the first message, if you have any examples of how its not accurate feel free to explain. but based on the code Ive worked with, its a fine description. Its a true or false flag that says whether it should be using the time of day entries to control lighting (outdoors) or use the light settings (indoors). and its used in a couple more places as well

fig02: > but to understanding the "indoor" mode doesn't use directional lights 
they do use directional lights, but they are not oriented to the position of the sun and moon. the direction and color is indicated by the light settings

Zelllll: I believe its used in the exterior of Ganon's Castle, which is not exactly "out"doors, among other places.

Thar0 (approved):

Roman971 (approved):

PR #793, opened 2021-04-24
Set __FILE__ in asm_processor

zelda2774:
Alternative to #791. (Though that PR may still be useful.)

### DIFF
--- a/src/code/z_collision_check.c
+++ b/src/code/z_collision_check.c
@@ -2254,7 +2254,6 @@ void CollisionCheck_AC_CylVsQuad(GlobalContext* globalCtx, CollisionCheckContext
 
 static s8 sBssDummy0;
 static s8 sBssDummy1;
-static s8 sBssDummy2;
 
 /**
  * AC overlap check. Calculates the center of each collider element and the point of contact.
--- a/src/code/z_parameter.c
+++ b/src/code/z_parameter.c
@@ -3012,11 +3012,10 @@ s16 D_80125B5C[] = { 91, 91 }; // unused
 
 // Due to an unknown reason, bss ordering changes within the 5 static variables in the function below.
 // In order to restore the correct order, we need a specific number of bss variables in the file before that point.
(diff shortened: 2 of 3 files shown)


louist103:
The only issue is that it still has the random characters at the end of the file.  Thats something they wanted to avoid.

zelda2774:
The idea of this is to set `__FILE__` to equal the original filename, before preprocessing. So while the file getting compiled would still have random characters in its name, it wouldn't show up in error messages or in `__FILE__`. 

The CI failure is apparently due to `#line` introducing bss reorderings...

louist103 (approved):

louist103:
After this is merged in I may make the build.py change so the temp files have a recognizable name.  Unless you want to do it your self.

zelda2774:
> After this is merged in I may make the build.py change so the temp files have a recognizable name. Unless you want to do it your self.

Please go ahead if you want, it sounds like a useful change.

AngheloAlf:
> The asm_processor.py we use is a straight copy from its repo (https://github.com/simonlindholm/asm-processor), so this change should probably be PR'd and merged in its master branch before porting it over.
- Roman, 24/04/2021, 6 hours ago

zelda2774:
Done

fig02 (approved):

Thar0 (approved):

Roman971 (approved):
This seems fine to me. I'm worried it will cause issues with bss ordering in the future and might be straight up incompatible with some files, but I guess we'll deal with that if it happens

AngheloAlf:
How can setting a macro can cause bss ordering issues? :thinking:

Roman971:
> How can setting a macro can cause bss ordering issues? 🤔

The #line directive isn't a macro, and it looks like it has the same effect as a .bss variable or external pointer reference, which can cause bss reorderings in large files. Some of these reorderings are necessary so we sometimes need a specific amount of variables/references (see the extra dummy variables in z_parameter.c and z_collision_check.c), which might be impossible to reach with an extra #line directive.

PR #795, opened 2021-04-25
EnFish, OK and documented

EllipticEllipsis:
*Uncorks bottle, turns upside down, drops fish.*

Some fishy stuff in here, but overall not too bad.

All functions matched, and named, most variables named (I left a few of the temps, and the struct variable that's used in bgCheck since I don't know what it's for, although I assume it's to do with having a cylinder). Also renamed its assets in `gameplay_keep`, added the type to Player, and corrected a few SkelAnime functions I spotted.

Not a lot of documentation required for the contents of functions, but I think it's worth considering including something like the graph I [posted in Discord](https://discord.com/channels/688807550715560050/688851317593997489/835653101313720360) since for that type, it's the function flow itself that's complicated, rather than the contents.

(But why does he drop the fish?)


src/overlays/actors/ovl_En_Fish/z_en_fish.c
```diff
@@ -47,73 +63,709 @@ static ColliderJntSphInit D_80A17040 = {
...
+        yaw = Math_Vec3f_Yaw(&this->actor.world.pos, &this->actor.home.pos);
+        Math_StepToAngleS(&this->actor.world.rot.y, yaw, 3000);
+    } else if ((this->actor.child != NULL) && (&this->actor != this->actor.child)) {
+        yaw = Math_Vec3f_Yaw(&this->actor.world.pos, &this->actor.child->world.pos);
+        Math_StepToAngleS(&this->actor.world.rot.y, yaw, 2000);
+    } else if (playerClose) {
+        yaw = this->actor.yawTowardsPlayer + 0x8000;
+        frames = globalCtx->state.frames;
+
+        if (frames & 0x10) {
```
louist103, line 290: Does `%` not work here?

src/overlays/actors/ovl_En_Fish/z_en_fish.c
```diff
@@ -47,73 +63,709 @@ static ColliderJntSphInit D_80A17040 = {
...
+
+void EnFish_Destroy(Actor* thisx, GlobalContext* globalCtx2) {
+    GlobalContext* globalCtx = globalCtx2;
+    EnFish* this = THIS;
+
+    Collider_DestroyJntSph(globalCtx, &this->collider);
+}
+
+void EnFish_SetYOffset(EnFish* this) {
+    this->actor.shape.yOffset += ((Math_SinS(this->slowPhase) * 10.0f) + (Math_SinS(this->fastPhase) * 5.0f));
```
AngheloAlf, line 164:
```suggestion
    this->actor.shape.yOffset += (Math_SinS(this->slowPhase) * 10.0f) + (Math_SinS(this->fastPhase) * 5.0f);
```

src/overlays/actors/ovl_En_Fish/z_en_fish.c
```diff
@@ -47,73 +63,709 @@ static ColliderJntSphInit D_80A17040 = {
...
+    SkelAnime_Update(&this->skelAnime);
+
+    if (this->timer <= 0) {
+        Actor_Kill(&this->actor);
+        return;
+    }
+
+    if (this->timer <= 60) {
+        // Blink when about to disappear
+        if (frames & 4) {
```
louist103, line 465: Same here with the `%`

louist103 (changes requested):
Looks good.  Just one (or two?) things.

src/overlays/actors/ovl_En_Fish/z_en_fish.c
```diff
@@ -47,73 +63,709 @@ static ColliderJntSphInit D_80A17040 = {
...
+    this->timer = Rand_S16Offset(5, 35);
+    this->unk_250 = 0;
+    EnFish_SetInWaterAnimation(this);
+    this->actionFunc = EnFish_Respawning_SlowDown;
+}
+
+void EnFish_Respawning_SlowDown(EnFish* this, GlobalContext* globalCtx) {
+    EnFish_SetYOffset(this);
+    Math_SmoothStepToF(&this->actor.speedXZ, 0.0f, 0.05f, 0.3f, 0.0f);
+    this->skelAnime.playSpeed = CLAMP_MAX((this->actor.speedXZ * 1.4f) + 0.8f, 2.0f);
```
AngheloAlf, line 207:
```suggestion
    this->skelAnime.playSpeed = CLAMP_MAX(this->actor.speedXZ * 1.4f + 0.8f, 2.0f);
```

src/overlays/actors/ovl_En_Fish/z_en_fish.c
```diff
@@ -47,73 +63,709 @@ static ColliderJntSphInit D_80A17040 = {
...
+
+    if ((EnFish_XZDistanceSquared(&this->actor.world.pos, &this->actor.home.pos) > SQ(80.0f)) || (this->timer < 4)) {
+        Math_StepToAngleS(&this->actor.world.rot.y, Math_Vec3f_Yaw(&this->actor.world.pos, &this->actor.home.pos),
+                          3000);
+    } else {
+        if ((this->actor.child != NULL) && (&this->actor != this->actor.child)) {
+            Math_StepToAngleS(&this->actor.world.rot.y,
+                              Math_Vec3f_Yaw(&this->actor.world.pos, &this->actor.child->world.pos), 3000);
+        }
+    }
```
AngheloAlf, line 245: Does this work?
```suggestion
    } else if ((this->actor.child != NULL) && (&this->actor != this->actor.child)) {
        Math_StepToAngleS(&this->actor.world.rot.y,
                          Math_Vec3f_Yaw(&this->actor.world.pos, &this->actor.child->world.pos), 3000);
    }
```

src/overlays/actors/ovl_En_Fish/z_en_fish.c
```diff
@@ -47,73 +63,709 @@ static ColliderJntSphInit D_80A17040 = {
...
+
+    this->actor.shape.rot.y = this->actor.world.rot.y;
+    this->skelAnime.playSpeed = CLAMP_MAX((this->actor.speedXZ * 1.5f) + 0.8f, 4.0f);
```
AngheloAlf, line 248:
```suggestion
    this->skelAnime.playSpeed = CLAMP_MAX(this->actor.speedXZ * 1.5f + 0.8f, 4.0f);
```

src/overlays/actors/ovl_En_Fish/z_en_fish.c
```diff
@@ -47,73 +63,709 @@ static ColliderJntSphInit D_80A17040 = {
...
+            if (frames & 0x20) {
+                yaw -= 0x2000;
+            }
+        }
+        if (globalCtx) {}
+        Math_StepToAngleS(&this->actor.world.rot.y, yaw, 2000);
+    }
+
+    this->actor.shape.rot.y = this->actor.world.rot.y;
+    this->skelAnime.playSpeed = CLAMP_MAX((this->actor.speedXZ * 1.5f) + 0.8f, 4.0f);
```
AngheloAlf, line 304:
```suggestion
    this->skelAnime.playSpeed = CLAMP_MAX(this->actor.speedXZ * 1.5f + 0.8f, 4.0f);
```

src/overlays/actors/ovl_En_Fish/z_en_fish.c
EllipticEllipsis: No, they all have to be `&`, unfortunately.

src/overlays/actors/ovl_En_Fish/z_en_fish.c
```diff
@@ -47,73 +63,709 @@ static ColliderJntSphInit D_80A17040 = {
...
+        this->actor.velocity.y = (Rand_ZeroOne() * 1.2f) + 0.2f;
+        playSound = true;
+    } else {
+        this->actor.velocity.y = 0.0f;
+
+        if (Rand_ZeroOne() < 0.2f) {
+            playSound = true;
+        } else {
+            playSound = false;
+        }
```
AngheloAlf, line 424: I guess this doesn't work, right?
```suggestion
        playSound = Rand_ZeroOne() < 0.2f;
```

src/overlays/actors/ovl_En_Fish/z_en_fish.c
```diff
@@ -47,73 +63,709 @@ static ColliderJntSphInit D_80A17040 = {
...
+    EnFish_SetInWaterAnimation(this);
+    this->actionFunc = EnFish_Unique_SwimIdle;
+}
+
+void EnFish_Unique_SwimIdle(EnFish* this, GlobalContext* globalCtx) {
+    static f32 speedStopping[] = { 0.0f, 0.04f, 0.09f };
+    static f32 speedMoving[] = { 0.5f, 0.1f, 0.15f };
+    f32 playSpeed;
+    u32 frames;
+    f32* speedType;
```
AngheloAlf, line 542: Does this make sense?
```suggestion
    static Vec3f speedStopping = { 0.0f, 0.04f, 0.09f };
    static Vec3f speedMoving = { 0.5f, 0.1f, 0.15f };
    f32 playSpeed;
    u32 frames;
    Vec3f* speedType;
```

src/overlays/actors/ovl_En_Fish/z_en_fish.c
```diff
@@ -47,73 +63,709 @@ static ColliderJntSphInit D_80A17040 = {
...
+    s32 pad2;
+    f32 extraPlaySpeed;
+    s32 pad3;
+
+    frames = globalCtx->gameplayFrames;
```
AngheloAlf, line 547:
```suggestion
    u32 frames = globalCtx->gameplayFrames;
    f32* speedType;
    s32 pad2;
    f32 extraPlaySpeed;
    s32 pad3;
```

src/overlays/actors/ovl_En_Fish/z_en_fish.c
```diff
@@ -47,73 +63,709 @@ static ColliderJntSphInit D_80A17040 = {
...
+    startPos.z = csAction->startPos.z;
+    endPos.x = csAction->endPos.x;
+    endPos.y = csAction->endPos.y;
+    endPos.z = csAction->endPos.z;
+
+    progress = func_8006F93C(csAction->endFrame, csAction->startFrame, globalCtx->csCtx.frames);
+
+    this->actor.world.pos.x = ((endPos.x - startPos.x) * progress) + startPos.x;
+    this->actor.world.pos.y = ((endPos.y - startPos.y) * progress) + startPos.y + D_80A17014;
+    this->actor.world.pos.z = ((endPos.z - startPos.z) * progress) + startPos.z;
```
AngheloAlf, line 675:
```suggestion
    this->actor.world.pos.x = (endPos.x - startPos.x) * progress + startPos.x;
    this->actor.world.pos.y = (endPos.y - startPos.y) * progress + startPos.y + D_80A17014;
    this->actor.world.pos.z = (endPos.z - startPos.z) * progress + startPos.z;
```

src/overlays/actors/ovl_En_Fish/z_en_fish.c
```diff
@@ -47,73 +63,709 @@ static ColliderJntSphInit D_80A17040 = {
...
+
+        if (Actor_HasParent(&this->actor, globalCtx)) {
+            this->actor.parent = NULL;
+
+            if (this->actor.params == FISH_DROPPED) {
+                Actor_Kill(&this->actor);
+                return;
+            }
+
+            EnFish_BeginRespawn(this);
```
AngheloAlf, line 716:
```suggestion
            if (this->actor.params == FISH_DROPPED) {
                Actor_Kill(&this->actor);
            }
            else {
                EnFish_BeginRespawn(this);
            }
```

AngheloAlf (commented):
Good one.
Just a few minor comments.

src/overlays/actors/ovl_En_Fish/z_en_fish.c
EllipticEllipsis: I disagree with this: it's not treated like a Vec3f at all: it's the goal and the step size arguments.

src/overlays/actors/ovl_En_Fish/z_en_fish.c
EllipticEllipsis: We usually do early returns after `Actor_Kill`.

src/overlays/actors/ovl_En_Fish/z_en_fish.c
louist103: does a 2d array work?

src/overlays/actors/ovl_En_Fish/z_en_fish.c
EllipticEllipsis: It needs an extra branch. Can go with `playSound = (Rand_ZeroOne() < 0.2f ? true : false);` if that's better.

src/overlays/actors/ovl_En_Fish/z_en_fish.c
EllipticEllipsis: It does (Yay, another fake symbol!). Sadly `speedType = speedData[this->timer < 12];` doesn't work

EllipticEllipsis: That said, I'm not actually sure it's clearer with a 2D array.

src/overlays/actors/ovl_En_Fish/z_en_fish.c
```diff
@@ -47,73 +63,709 @@ static ColliderJntSphInit D_80A17040 = {
...
+    this->actor.world.pos.z = ((endPos.z - startPos.z) * progress) + startPos.z;
+
+    this->actor.floorHeight = BgCheck_EntityRaycastFloor4(&globalCtx->colCtx, &this->actor.floorPoly, &bgId,
+                                                          &this->actor, &this->actor.world.pos);
+}
+
+// Update functions and Draw
+
+void EnFish_OrdinaryUpdate(EnFish* this, GlobalContext* globalCtx) {
+    if (this->timer > 0) {
```
fig02, line 684: not a fan of "ordinary update", but have no suggestions for now

src/overlays/actors/ovl_En_Fish/z_en_fish.c
fig02: its fine as is without 2d array imo. though id rename the `speedType` var. that makes it sound like the var denotes what the type is, when really its just the array of speed values. I think just `speed` is fine

src/overlays/actors/ovl_En_Fish/z_en_fish.c
```diff
@@ -47,73 +63,709 @@ static ColliderJntSphInit D_80A17040 = {
...
+    this->actor.shape.rot.z = this->actor.world.rot.z;
+    SkelAnime_Update(&this->skelAnime);
+
+    if (this->actor.bgCheckFlags & 1) { // On floor
+        this->timer = 400;
+        EnFish_Dropped_SetupFlopOnGround(this);
+    } else if (this->actor.bgCheckFlags & 0x20) { // In water
+        EnFish_Dropped_SetupSwimAway(this);
+    } else if ((this->timer <= 0) && (this->actor.params == FISH_DROPPED) && (this->actor.floorHeight < -31990.0f)) {
+        osSyncPrintf(VT_COL(YELLOW, BLACK));
```
fig02, line 390: i guess -31990 is `BGCHECK_Y_MIN + 10.0f`

src/overlays/actors/ovl_En_Fish/z_en_fish.c
```diff
@@ -47,73 +63,709 @@ static ColliderJntSphInit D_80A17040 = {
...
+    Vec3f sp1C;
+
+    if (this->actor.xzDistToPlayer < 32.0f) {
+        sp1C.x = (Math_SinS(this->actor.yawTowardsPlayer + 0x8000) * 16.0f) + player->actor.world.pos.x;
+        sp1C.y = player->actor.world.pos.y;
+        sp1C.z = (Math_CosS(this->actor.yawTowardsPlayer + 0x8000) * 16.0f) + player->actor.world.pos.z;
+
+        //! @bug: this check is superfluous: it is automatically satisfied if the coarse check is satisfied. It may have
+        //! been intended to check the actor is in front of Player, but yawTowardsPlayer does not depend on Player's
+        //! world.rot.
```
fig02, line 180: in this error message (and the other places it was copied to) I would just say "world rotation" instead of `world.rot`

src/overlays/actors/ovl_En_Fish/z_en_fish.c
EllipticEllipsis: Yeah, the problem with this function is it's very much "not any of the special updates", since Update itself is basically a wrapper to choose which sub-update to use. And something like UpdateNormal or NormalUpdate sounds like it's to do with normals or something

src/overlays/actors/ovl_En_Fish/z_en_fish.c
EllipticEllipsis: Whoever wrote Fish and Insect seemed to like this number as their min, rather than the actual BGCHECK number.

fig02: yea, id say to still write it in terms of the actual minimum

EllipticEllipsis:
I changed it, and some other actors to use `BGCHECK_Y_MIN` where possible, and fixed a couple more examples I spotted elsewhere. There are a few cases where it actually checks for `+32000.0f`; do we want `BGCHECK_Y_MAX` too?

fig02 (approved):

louist103 (approved):

src/code/z_actor.c
```diff
@@ -1544,6 +1544,7 @@ u32 Actor_HasParent(Actor* actor, GlobalContext* globalCtx) {
     }
 }
 
+// GI_NONE is used for picking things up, GI_MAX for catching in a bottle.
 s32 func_8002F434(Actor* actor, GlobalContext* globalCtx, s32 getItemId, f32 xzRange, f32 yRange) {
```
Roman971, line 1548: Imo this isn't really the right place to put this comment, because it's mostly a hacky way the function was used rather than what it's intended for, especially regarding GI_MAX since any value >= GI_MAX would have the exact same effect.

Instead I suggest including a mention above the calls that use special values like these, to 'justify' their usage.

src/overlays/actors/ovl_En_Fish/z_en_fish.c
```diff
@@ -47,73 +63,705 @@ static ColliderJntSphInit D_80A17040 = {
...
+                                200)) {
+            extraPlaySpeed = 0.5f;
+        }
+    } else if ((this->timer < 4) && !Math_ScaledStepToS(&this->actor.world.rot.y, frames * 0x80, 100)) {
+        extraPlaySpeed = 0.5f;
+    }
+
+    this->actor.shape.rot.y = this->actor.world.rot.y;
+    playSpeed = (this->actor.speedXZ * 1.2f) + 0.2f + extraPlaySpeed;
+    this->skelAnime.playSpeed = CLAMP(playSpeed, 1.5f, 0.5);
```
Roman971, line 575: is no f required here?
```suggestion
    this->skelAnime.playSpeed = CLAMP(playSpeed, 1.5f, 0.5f);
```

Roman971 (commented):
Just a couple things but otherwise this looks good to me.

Also I think this might need one last run of the formatter, tho I could be wrong on that

src/overlays/actors/ovl_En_Fish/z_en_fish.c
EllipticEllipsis: It is.

src/code/z_actor.c
EllipticEllipsis: This comment was intended to be descriptive rather than prescriptive. Looking again, I think the reverse is true, actually: GI_MAX is used consistently for this by the 4 relevant actors. But GI_NONE isn't used consistently: EnKusa doesn't use it, for example, but EnIshi, EnNiw, and ObjTsubo do.

fig02: I agree with roman here, a comment above the call explaining the hacky usage of the GI value seems more appropriate than putting it above the function itself

EllipticEllipsis: Okay, how would you like these comments phrased?
```
// This is used for catching the actor in a bottle
func_8002F434(&this->actor, globalCtx, GI_MAX, 80.0f, 20.0f);
```
and
```
// This is used for lifting the actor
func_8002F434(&this->actor, globalCtx, GI_NONE, 80.0f, 20.0f);
```
?

fig02: `GI_MAX` in this case allows the player to catch the actor in a bottle
`GI_NONE` in this case allows the player to lift the actor

just trying to put emphasis on the gi values being unusual and not used for their actual purpose

fig02 (approved):

Roman971 (approved):

PR #796, opened 2021-04-26
Some actor.c docs (in preparation for en_test)

fig02:
instead of having the en_test pr cluttered with extra stuff, im doing this first separately 
 
`func_8002E084` -> `Actor_YawInRangeWithPlayer`
`func_8002E0D0` -> `Actor_YawInRangeWithActor`
`func_8002E12C` -> `Actor_YawAndDistInRangeWithPlayer`
`func_8002E1A8` -> `Actor_YawAndDistInRangeWithActor`
`func_80033A84` -> `Actor_IsTargeted`
`func_80033AB8` -> `Actor_OtherIsTargeted`
`func_80035650` ->  `Actor_SetDropFlag`
`func_8003573C` ->  `Actor_SetDropFlagJntSph`
`func_80033780` ->  `Actor_GetProjectileActor`
`func_80033260` ->  `Actor_SpawnFloorDust`
`func_80032C7C` -> `Actor_PlayDeathFx`

`actorCtx.unk_00` -> `actorCtx.freezeFlashTimer`


include/functions.h
```diff
@@ -413,9 +413,9 @@ s32 func_8002DF54(GlobalContext* globalCtx, Actor* actor, u8 arg2);
 void func_8002DF90(DynaPolyActor* dynaActor);
 void func_8002DFA4(DynaPolyActor* dynaActor, f32 arg1, s16 arg2);
 s32 func_8002DFC8(Actor* actor, s16 arg1, GlobalContext* globalCtx);
-s32 func_8002E084(Actor* actor, s16 arg1);
-s32 func_8002E12C(Actor* actor, f32 arg1, s16 arg2);
-s32 func_8002E1A8(Actor* actorA, Actor* actorB, f32 arg2, s16 arg3);
+s32 Actor_YawInRangeWithPlayer(Actor* actor, s16 arg1);
+s32 Actor_YawAndDistInRangeWithPlayer(Actor* actor, f32 arg1, s16 arg2);
+s32 Actor_YawAndDistInRangeWithActor(Actor* actorA, Actor* actorB, f32 arg2, s16 arg3);
```
Roman971, line 418: The arguments for these functions need to be updated.

Also I think `functions.h` could use a formatter run after these changes

src/code/z_actor.c
```diff
@@ -3003,8 +3006,12 @@ Actor* Actor_Find(ActorContext* actorCtx, s32 actorId, s32 actorCategory) {
     return NULL;
 }
 
-void func_80032C7C(GlobalContext* globalCtx, Actor* actor) {
-    globalCtx->actorCtx.unk_00 = 5;
+/**
+ * Play the death sound effect and flash the screen white for 4 frames.
+ * While the screen flashes, the game freezes.
+ */
+void Actor_PlayDeathFx(GlobalContext* globalCtx, Actor* actor) {
```
Roman971, line 3013: I feel like the name or documentation for this function should include "enemy" somewhere so it doesn't get confused with the player death. Also I think this would typically be called a "finishing blow" effect, but I don't know if it's worth including that in some of the names or documentation.

src/code/z_actor.c
```diff
@@ -1084,45 +1084,45 @@ s32 func_8002E020(Actor* actorA, Actor* actorB, s16 arg2) {
...
+    if (ABS(yawDiff) < yaw) {
         return true;
     }
 
     return false;
 }
 
-s32 func_8002E0D0(Actor* actorA, Actor* actorB, s16 arg2) {
-    s16 var = Actor_WorldYawTowardActor(actorA, actorB) - actorA->shape.rot.y;
+s32 Actor_YawInRangeWithActor(Actor* actorA, Actor* actorB, s16 yaw) {
```
Roman971, line 1097: I'm not sure about the name for these functions because the "InRange" part makes it sound like it's doing some kind of distance check, but I also can't think of anything better, aside from maybe reordering terms to make it clear it's doing a range check with the yaw, not between the actors or anything like that.

src/code/z_actor.c
fig02: I like finishing blow, yeah.
`Actor_EnemyPlayFinishingBlowFx` is a bit long, but its not like youll see this function a ton.
My instinct almost wants to make the system `Enemy_` but im not sure how many enemy specific functions we'll have lol

src/code/z_actor.c
fig02: do you have a specific example of how to reorder the terms? I kinda think this is clear, since i tried to make it sound like the sentence "the yaw is in range with the actor"
vs the other variation of the function which says "yaw and distance is in range"

but if its not clear to others then we can def think of something else

fig02: just bumping this and the comments above. still not sure on the best route to take

fig02: pls 🙏

Roman971: I don't know what the best route is either, but I think the current names could be confusing and the sentence "the yaw is in range with the actor" doesn't really make sense to me.
To me it looks like these functions are checking that the yaw difference between 2 actors is smaller than a certain value, so maybe we could just call them "Actor_YawCheck" (or "Actor_ActorYawCheck") and "Actor_PlayerYawCheck", with a similar pattern for the yaw + distance checks.
Also we should probably consider func_8002E020 and func_8002E084 while naming these, since they are doing similar types of checks but from the other actor's perspective (?)

fig02: to me the "less than a certain value" is the range check. its looking for the angle difference to be within a certain range.
Since its not looking for an exact value for the difference, but if the actors are facing each other to a certain extent, the check will pass

fig02: I agree that theres probably something better. but something like yaw check is too ambiguous to me

Roman971: My main concern is the term "range" makes it sound like it's doing a distance check to me, especially when saying "in range with the actor/player". I think a different order like "Actor_YawWithActorInRange" would already be less confusing, but it's not ideal either.

fig02: yea dunno, range doesnt mean just distance to me. maybe theres a totally different direction to take this thats better, if anyone has ideas

EllipticEllipsis: Perhaps include the word `Arc`? `YawInArcRange` or similar?

fig02: Possibly? I feel like i would just be more confused when reading the name for the first time, trying to figure out how an arc plays into it. I understand what youre going for though 

"yaw in range" is sufficient to me, because there is a range of angle values that will satisfy the check. range doesnt mean just distance, to me. And when distance is involved, the name specifies that its checking for both the yaw and the distance to be within a specified range.

we use the phrasing "yaw toward player" for that one actor struct variable, so maybe using that kind of phrasing would help? not sure

PR #797, opened 2021-04-26
Document BodyBreak (en_part, body part spawner)

fig02:
This is used when some enemies break apart when they die. 
Slightly weird implementation, but pretty neat.

This is also in preparation for en_test, along with https://github.com/zeldaret/oot/pull/796


src/code/z_actor.c
```diff
@@ -3037,104 +3037,111 @@ s16 func_80032D60(s16* arg0, s16 arg1, s16 arg2, s16 arg3) {
...
-                Lib_MemSet((u8*)arg0->unk_00, sp28, 0);
-                Lib_MemSet((u8*)arg0->unk_0C, sp24, 0);
-                Lib_MemSet((u8*)arg0->unk_04, sp20, 0);
-                arg0->unk_10 = 1;
+void BodyBreak_Alloc(BodyBreak* bodyBreak, s32 count, GlobalContext* globalCtx) {
+    u32 matricesSize;
+    u32 dListsSize;
+    u32 objectIdsSize;
+
+    matricesSize = (count * sizeof(*bodyBreak->matrices)) + sizeof(*bodyBreak->matrices);
```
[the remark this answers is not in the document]
fig02: this works. pushed

include/z64actor.h
```diff
@@ -201,14 +201,17 @@ typedef struct DynaPolyActor {
...
+    /* 0x04 */ s16* objectIds;
+    /* 0x08 */ s16 count;
+    /* 0x0C */ Gfx** dLists;
+    /* 0x10 */ s32 val; // used for various purposes: both a status indicator and counter
+    /* 0x14 */ s32 prevLimbIndex;
+} BodyBreak;
+
+#define BODYBREAK_OBJECT_DEFAULT -1 // use the same object as the actor
+#define BODYBREAK_STATUS_READY -1
+#define BODYBREAK_STATUS_FINISHED 0
```
AngheloAlf, line 214: Shouldn't be better an enum instead?

src/code/z_actor.c
```diff
@@ -3037,104 +3037,111 @@ s16 func_80032D60(s16* arg0, s16 arg1, s16 arg2, s16 arg3) {
...
-                Lib_MemSet((u8*)arg0->unk_00, sp28, 0);
-                Lib_MemSet((u8*)arg0->unk_0C, sp24, 0);
-                Lib_MemSet((u8*)arg0->unk_04, sp20, 0);
-                arg0->unk_10 = 1;
+void BodyBreak_Alloc(BodyBreak* bodyBreak, s32 count, GlobalContext* globalCtx) {
+    u32 matricesSize;
+    u32 dListsSize;
+    u32 objectIdsSize;
+
+    matricesSize = (count + 1) * sizeof(*bodyBreak->matrices);
```
AngheloAlf, line 3045:
```suggestion
    u32 matricesSize = (count + 1) * sizeof(*bodyBreak->matrices);
    u32 dListsSize;
    u32 objectIdsSize;

```

src/overlays/actors/ovl_En_Skb/z_en_skb.c
```diff
@@ -409,14 +409,14 @@ void func_80AFD7B4(EnSkb* this, GlobalContext* globalCtx) {
...
     EffectSsDeadSound_SpawnStationary(globalCtx, &this->actor.projectedPos, NA_SE_EN_STALKID_DEAD, 1, 1, 0x28);
     EnSkb_SetupAction(this, func_80AFD880);
 }
 
 void func_80AFD880(EnSkb* this, GlobalContext* globalCtx) {
-    if (func_8003305C(&this->actor, &this->unk_28C, globalCtx, 1) != 0) {
+    if (BodyBreak_SpawnParts(&this->actor, &this->bodyBreak, globalCtx, 1)) {
         if (this->actor.scale.x == 0.01f) {
             Item_DropCollectibleRandom(globalCtx, &this->actor, &this->actor.world.pos, 0x10);
         } else {
```
AngheloAlf, line 422: Github doesn't let me take the range of lines I want to make a suggestion, so I'm adding a screenshot.
Could you merge the `else` and the `if`? Does it match?
image

src/code/z_actor.c
fig02: I dont agree with this personally, the other sizes cant be declared with their var so it makes it less consistent.
The declarations could be moved to their relevant blocks, but id say to wait until retail decomp where the scope of the declaration will prob matter more

include/z64actor.h
fig02: the first is unrelated to the other two. and the status only has two values (which already dont start nicely at zero). 
And as a modder its not like youd want to add more really, so I dont think an enum is much of an improvement personally

src/overlays/actors/ovl_En_Skb/z_en_skb.c
fig02: prob can but its a bit out of scope from what im doing here, this kind of thing will get fixed in other passes of this actor.
Will try it anyway tho

AngheloAlf (approved):

src/code/z_actor.c
fig02: thinking about it the declarations actually cant be moved, cause the vars have to be available in the block below where things get freed. so it should prob say like this

Roman971 (approved):

PR #798, opened 2021-04-27
Improve the format in z_map_mark_data.c and cleanup mapmark.py

Roman971:
* Removed unused entries from the tables in z_map_mark_data.c (see the recent discussion in discord for more detail)
* Shortened map mark defines
* Cleaned up mapmark.py a bit


mzxrules:
seems good to me

fig02 (approved):

PR #799, opened 2021-04-28
General object cleanup

AngheloAlf:
Fixes some typos, warnings and invalid attributes.

This change will be needed by the future attribute's checker of ZAPD ([PR 126](https://github.com/zeldaret/ZAPD/pull/126)).

### DIFF
--- a/assets/xml/objects/object_mm.xml
+++ b/assets/xml/objects/object_mm.xml
@@ -7,7 +7,7 @@
         <Animation Name="gRunningManSprintAnim" Offset="0x0468"/>
         <Animation Name="gRunningManExcitedAnim" Offset="0x73A0"/>
         <Animation Name="gRunningManHappyAnim" Offset="0x8060"/>
-        <Texture Name="gRunningManMouthClosedTex" Outname="running_man_mouth_closed" Format="ci8" Width="32" Height="16" Offset="0xE30"/>
-        <Texture Name="gRunningManMouthOpenTex" Outname="running_man_mouth_closed" Format="ci8" Width="32" Height="16" Offset="0xC30"/>
+        <Texture Name="gRunningManMouthClosedTex" OutName="running_man_mouth_open" Format="ci8" Width="32" Height="16" Offset="0xE30"/>
+        <Texture Name="gRunningManMouthOpenTex" OutName="running_man_mouth_closed" Format="ci8" Width="32" Height="16" Offset="0xC30"/>
     </File>
 </Root>
--- a/assets/xml/objects/object_mm.xml
+++ b/assets/xml/objects/object_mm.xml
@@ -1,13 +1,16 @@
 <Root>
     <File Name="object_mm" Segment="6">
         <Skeleton Name="gRunningManSkel" Type="Flex" LimbType="Standard" Offset="0x5E18"/>
+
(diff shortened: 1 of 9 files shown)


assets/xml/objects/object_mm.xml
```diff
@@ -7,7 +7,7 @@
         <Animation Name="gRunningManSprintAnim" Offset="0x0468"/>
         <Animation Name="gRunningManExcitedAnim" Offset="0x73A0"/>
         <Animation Name="gRunningManHappyAnim" Offset="0x8060"/>
-        <Texture Name="gRunningManMouthClosedTex" Outname="running_man_mouth_closed" Format="ci8" Width="32" Height="16" Offset="0xE30"/>
-        <Texture Name="gRunningManMouthOpenTex" Outname="running_man_mouth_closed" Format="ci8" Width="32" Height="16" Offset="0xC30"/>
+        <Texture Name="gRunningManMouthClosedTex" OutName="running_man_mouth_open" Format="ci8" Width="32" Height="16" Offset="0xE30"/>
```
louist103, line 10: Which one is it? One them is obviously wrong. Might as well fix it here.

louist103 (changes requested):
One small thing then its good to go IMO.

assets/xml/objects/object_mm.xml
AngheloAlf: Fixed.

src/overlays/actors/ovl_En_Mm/z_en_mm.c
```diff
@@ -522,7 +522,7 @@ void EnMm_Update(Actor* thisx, GlobalContext* globalCtx) {
 extern Gfx D_0602CA38[]; // bunny hood dlist from object_link_child. replace with proper symbol later
 
 void EnMm_Draw(Actor* thisx, GlobalContext* globalCtx) {
-    static u64* mouthTextures[] = { gRunningManMouthClosedTex, gRunningManMouthOpenTex };
+    static u64* mouthTextures[] = { gRunningManMouthOpenTex, gRunningManMouthClosedTex };
```
Zelllll, line 525: I suggest making this be good*

AngheloAlf: We have been using `u64*` for textures everywhere in the repo, so I don't see why change only this one to be `void*`.

fig02: using void* when possible is something roman has said we may want to do, but yeah it will all be made consistent at the same time

AngheloAlf: I feel like that change is out of scope for this PR, but I could do that change in a future PR :D

Zelllll: Cool 👍 a void* pr for all textures could be good, I'm tempted to say that we should just simply make a typedef instead though like SM64 which just has Texture, which I believe in sm64 is actually just a u8, which would make it easy to change the type for textures if we decide to though--anyway that's out of the scope of this pr and we can discuss it some other time, good work on this.

Zelllll (approved):
Should be good to merge

louist103 (approved):

Roman971 (approved):

fig02 (approved):

PR #802, opened 2021-05-01
ViMode progress

zbanks:
Hey everyone! I know `z_vimode.c` isn't a very interesting section of code - but I was looking for something to get my feet wet with, and avoid colliding with others' work.

`z_vimode.c` is mostly self-contained, and barely referenced by the rest of the code. It seems like it's main purpose is to manage custom VI modes, which probably were just for debugging and not used in practice?

Let me know if there's anything I missed or should clean up!


AngheloAlf:
This file was reserved by @Random06457 [here](https://trello.com/c/pHH3PWdj/33-zvimodec).
You should check the trello or ask in the discord server before trying to match a random file. If you aren't in the discord server, there is an invitation link in https://zelda64.dev/

zbanks:
Oh I'm sorry, it looked like it was one of the few files that wasn't reserved on the trello board? (It didn't have a WIP/Reserved tag?)

I'll pop in the discord though, thanks.

include/z64.h
```diff
@@ -1653,11 +1653,22 @@ typedef struct {
...
+    /* 0x0068 */ s32 viModeBase; // enum: {0: None, 1, 2, 3: NTSC}
+    /* 0x006c */ s32 viTvType;
+    /* 0x0070 */ u32 unkFlagTop; // bool
+    /* 0x0074 */ u32 unkFlagBottom; // bool
+    /* 0x0078 */ u32 unkFlagLeft; // bool
+    /* 0x007c */ u32 unkFlagRight; // bool
     /* 0x0080 */ u32 viFeatures;
-    /* 0x0084 */ char unk_84[4];
-} unk_80166528;
+    /* 0x0084 */ u32 _unused;
```
AngheloAlf, line 1670: You should leave it as an `unk` unless you are sure it is not used anywhere in the game's code. Also we don't use the `_` prefix for variables.
```suggestion
    /* 0x0084 */ char unk_84[4];
```

src/code/z_vimode.c
```diff
@@ -1,17 +1,214 @@
 #include "global.h"
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_vimode/func_800AC030.s")
+void ViMode_LogPrint(OSViMode* viMode) {
+    OSViMode** viMode_ptr = &viMode;
+#define _LOG(line, fmt_str, val)                 \
+    LogUtils_LogThreadId("../z_vimode.c", line); \
```
AngheloAlf, line 6: We usually don't define macros inside functions.
Also, we already have a `LOG_ADDRESS` macro you could use.

Random06457:
It's okay, I didn't make any progress on this file for the past 3 months anyways. The only thing I did more was starting `func_800AC2F4` (https://github.com/Random06457/oot/blob/z_vimode/src/code/z_vimode.c#L32) which is pretty annoying because of the stupid conditions.

src/code/z_vimode.c
```diff
@@ -1,17 +1,214 @@
 #include "global.h"
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_vimode/func_800AC030.s")
+void ViMode_LogPrint(OSViMode* viMode) {
+    OSViMode** viMode_ptr = &viMode;
+#define _LOG(line, fmt_str, val)                 \
```
Random06457, line 5: We already have the `LOG_ADDRESS` macro for this (defined [here](https://github.com/zeldaret/oot/blob/98aef8988adbcb5960e8001e70141d0d1a9034bd/include/macros.h#L72)).

AngheloAlf (changes requested):
Good job! Just a few comments.

src/code/z_vimode.c
```diff
@@ -1,17 +1,214 @@
...
+    LogUtils_LogThreadId("../z_vimode.c", line); \
+    osSyncPrintf(fmt_str, val)
+    _LOG(87, "osvimodep = %08x\n", (*viMode_ptr));
+    _LOG(88, "osvimodep->comRegs.ctrl = %08x\n", (*viMode_ptr)->comRegs.ctrl);
```
mzxrules, line 9: considering the source is right there, I'm assuming it's possible to this like so:
`osViModePtr->comRegs.ctrl`

src/code/z_vimode.c
```diff
@@ -1,17 +1,214 @@
...
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_vimode/func_800AC9A4.s")
+void ViMode_Load(ViMode* viMode) {
+    GameInfo* _gameInfo = gGameInfo;
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_vimode/func_800ACA28.s")
+    // NB: This condition is always false
+    s16 width = SREG(49);
+    if ((width & ((s16)0xfffc)) == 1) {
```
mzxrules, line 66:
```suggestion
    s16 width = SREG(49);
    // NB: This condition is always false
    if ((width & ((s16)0xfffc)) == 1) {
```

mzxrules: Also, I don't think this is the correct way to flag this, but I also don't know what our standard is.

src/code/z_vimode.c
```diff
@@ -1,17 +1,214 @@
...
+    }
+    if (viFeatures & OS_VI_GAMMA_DITHER_OFF) {
+        ctrl = ctrl & ~OS_VI_GAMMA_DITHER;
+    }
+    if (viFeatures & OS_VI_DIVOT_ON) {
+        ctrl = ctrl | OS_VI_DIVOT;
+    }
+    if (viFeatures & OS_VI_DIVOT_OFF) {
+        ctrl = ctrl & ~OS_VI_DIVOT;
+    }
```
mzxrules, line 120:
```suggestion
    if (viFeatures & OS_VI_GAMMA_ON) {
        ctrl |= OS_VI_GAMMA;
    }
    if (viFeatures & OS_VI_GAMMA_OFF) {
        ctrl &= ~OS_VI_GAMMA;
    }
    if (viFeatures & OS_VI_GAMMA_DITHER_ON) {
        ctrl |= OS_VI_GAMMA_DITHER;
    }
    if (viFeatures & OS_VI_GAMMA_DITHER_OFF) {
        ctrl &= ~OS_VI_GAMMA_DITHER;
    }
    if (viFeatures & OS_VI_DIVOT_ON) {
        ctrl |= OS_VI_DIVOT;
    }
    if (viFeatures & OS_VI_DIVOT_OFF) {
        ctrl &= ~OS_VI_DIVOT;
    }
```

src/code/z_vimode.c
```diff
@@ -1,17 +1,214 @@
...
+    _LOG(103, "osvimodep->fldRegs[1].origin = %08x\n", (*viMode_ptr)->fldRegs[1].origin);
+    _LOG(104, "osvimodep->fldRegs[1].yScale = %08x\n", (*viMode_ptr)->fldRegs[1].yScale);
+    _LOG(105, "osvimodep->fldRegs[1].vIntr = %08x\n", (*viMode_ptr)->fldRegs[1].vIntr);
+#undef _LOG
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_vimode/func_800AC2F4.s")
+// This function configures the custom VI mode (`viMode.customViMode`) based
+// on the other flags in `viMode`.
+#pragma GLOBAL_ASM("asm/non_matchings/code/z_vimode/ViMode_Configure.s")
```
mzxrules, line 32: Curious why this function wasn't attempted. Ideally we like to have all functions at least attempted, and wrap #NON_MATCHING or #NON_EQUIVALENT around the code if it doesn't OK, but it's fine if you'd rather not do it.

src/code/z_vimode.c
```diff
@@ -1,17 +1,214 @@
...
+            }
+            if (viMode->unk64 < -50) {
+                viMode->unk64 = -50;
+            }
+            if (viMode->unk58 >= 17) {
+                viMode->unk58 = 16;
+            }
+            if (viMode->unk60 >= 51) {
+                viMode->unk60 = 50;
+            }
```
mzxrules, line 187:
```suggestion
            if (viMode->unk58 > 16) {
                viMode->unk58 = 16;
            }
            if (viMode->unk60 > 50) {
                viMode->unk60 = 50;
            }
```

src/code/z_vimode.c
zbanks: I can try to massage this again -- I agree the whole `OSViMode** viMode_ptr = &viMode;` is weird C code, but I needed to add it to get the asm to match up.

src/code/z_vimode.c
zbanks: Ah, I'll wrap it in `NON_MATCHING` - I didn't pick up on that convention. Thanks!

src/code/z_vimode.c
```diff
@@ -1,17 +1,214 @@
...
+    viMode->unkFlagLeft = true;
+    viMode->unkFlagBottom = false;
+    viMode->unkFlagTop = true;
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_vimode/func_800ACAF8.s")
+    ViMode_Save(viMode);
+}
+
+void ViMode_Destroy(ViMode* viMode) {
+    (void)viMode;
```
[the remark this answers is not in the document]
zbanks: Sorry, it's not. Habit from `-Wunused`, I can remove it to be more consistent with the rest of the codebase.

include/z64.h
zbanks: Makes sense, I can put it back (as far as I could tell, it was unused)

src/code/z_vimode.c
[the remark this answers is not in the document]
AngheloAlf: We have that warning disabled iirc, so please remove it.

zbanks:
I think I addressed all the comments left, thanks for the feedback.

I managed to make some more progress on `func_800AC2F4`, it's still not matching but I got a bit closer. (I tried to use naming conventions more similar to Random06457's code, thanks for linking to it)

include/ultra64/vi.h
```diff
@@ -92,4 +97,6 @@ typedef struct {
 #define OS_VI_PAL_HPN2      26
 #define OS_VI_PAL_HPF2      27
 
+#define OS_VI_UNK28         28
```
zbanks, line 100: This constant, `OS_VI_UNK28` is the one thing I wasn't sure how best to handle.
It's weird that the other `OS_VI_UNKxxx` are all hex flags; whereas here `28` is a decimal number not 0x28.

src/code/z_vimode.c
zbanks: Do I have to update the spec file? None of the files I deleted were referenced by the spec file. (The only file I modified that is referenced by the spec file is `z_vimode.o` afaict?)

AngheloAlf: Forget about what I said. That recommendation is only for overlays/actors. My bad :sweat_smile:

AngheloAlf (approved):

src/code/z_vimode.c
```diff
@@ -1,17 +1,331 @@
...
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_vimode/func_800AC030.s")
+#define LOG_MODE_FIELD(exp, line) LOG_ADDRESS(#exp, exp, "../z_vimode.c", line)
 
-#pragma GLOBAL_ASM("asm/non_matchings/code/z_vimode/func_800AC2F4.s")
+void ViMode_LogPrint(OSViMode* osvimodep) {
+    LOG_MODE_FIELD(osvimodep, 87);
+    LOG_MODE_FIELD(osvimodep->comRegs.ctrl, 88);
+    LOG_MODE_FIELD(osvimodep->comRegs.width, 89);
+    LOG_MODE_FIELD(osvimodep->comRegs.burst, 90);
```
fig02, line 9: not sure about this custom macro that includes the vimode file string. We dont really do this anywhere else currently, feel like we can just use the LOG_ADDRESS macro as is.
can get some other opinions tho

zbanks: I switched to a plain `LOG_ADDRESS` macro.

fig02 (approved):

src/code/z_vimode.c
```diff
@@ -1,17 +1,329 @@
...
+            default:
+                osSyncPrintf("Custom\n");
+                ViMode_LogPrint(&viMode->customViMode);
+                break;
+        }
+    }
+}
+
+void ViMode_Load(ViMode* viMode) {
+    GameInfo* _gameInfo = gGameInfo;
```
Roman971, line 180: This seems very fake, is there any way to avoid it, like not using a `width` variable?
Or at least just doing a `if (gGameInfo) {}`?

src/code/z_vimode.c
```diff
@@ -1,17 +1,329 @@
...
+
+    s16 width = SREG(49);
+    if ((width & ((s16)0xfffc)) == 1) {
```
Roman971, line 183:
```suggestion
    if ((width & ((s16)0xFFFC)) == 1) {
```

src/code/z_vimode.c
```diff
@@ -1,17 +1,329 @@
...
+        SREG(49) = width + 4;
+    }
+
+    viMode->viModeBase = SREG(48);
+    viMode->viWidth = SREG(49) & (~3);
```
Roman971, line 188:
```suggestion
    viMode->viWidth = SREG(49) & ~3;
```

src/code/z_vimode.c
zbanks: I managed to address all your comments and simplify this function down even more. Thanks!
(It is still a bit weird, because `(x & ~3) == 1` can never be true? it sort of looks like bugged rounding logic?)

Roman971 (approved):

Issue #597, opened 2021-01-03
make setup crashes on low-end pcs


shawlucas:
Make setup crashes on low-end pcs, or significantly slows the pc down because of the recent addition of the -j flag. Maybe we can either remove -j, or make it so it does something like -j(Number of cpu cores) because -j without any number makes it run unlimited which crashes on low-end pcs (both my laptop and my desktop, as they are quite old)

ethteck:
Presumably, make setup should pass the -j value passed in via the command line instead of using whatever it wants.
`make setup` (no -j)
`make setup -j` (pass -j)
`make setup -j4` (pass -j4)

shawlucas:
It also passes -j no matter what. I think that should be changed

fig02:
https://github.com/zeldaret/oot/pull/804

PR #804, opened 2021-05-03
Delete `-j` from `make setup` and document use of `-jN` and `-j` in readme

EllipticEllipsis:
~~I was having serious memory issues building the newest version of ZAPD during `make setup`, which were fixed by replacing `-j` in the Makefile by ``-j`nproc` ``, which uses the number of available threads.~~

Remove `-j` from `make setup` for tools. This addresses #597 ~~: although afaik there is no way to pass in the argument of `j` from the original argument of `make`, this is a far better default that what we currently have, at least for lower-end machines.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,13 @@ md5sum: WARNING: 1 computed checksum did NOT match
 
 This means that the built ROM isn't the same as the base one, so something went wrong or some part of the code doesn't match.
 
+**NOTE:** to speed up the build, you can either:
+* pass `-jN` to `make setup` and `make`, where N is the number of threads to use in the build. The generally-accepted wisdom is to use the number of virtual cores your computer has.
+* pass `-j` to `make setup` and `make`, to use as many threads as possible, but beware that this can use too much memory on lower-end systems.
+
+Both of these have the disadvantage that the ordering of the terminal output is scrambled, so for debugging it is best to stick to one.
+
+
 ### Cygwin
 
 If you want to use Cygwin, you will need to:
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ This means that the built ROM isn't the same as the base one, so something went 
 * pass `-jN` to `make setup` and `make`, where N is the number of threads to use in the build. The generally-accepted wisdom is to use the number of virtual cores your computer has.
 * pass `-j` to `make setup` and `make`, to use as many threads as possible, but beware that this can use too much memory on lower-end systems.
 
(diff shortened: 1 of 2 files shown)


fig02 (approved):

louist103 (approved):

AngheloAlf (approved):

EllipticEllipsis:
Apparently this is not quite as portable as I hoped: Macs at least need a different solution. Please hold off on merging until we find a satisfactory one.

Mastergatto:
I'll repost here what I have wrote in #730:

> Passing -j without a specific number means that there wouldn't be any limit on the number of jobs, maybe it's too much for a Raspberry Pi.
> 
> Instead of doing that, may I suggest you to add MAKEFLAGS += -j $(shell nproc) within the platform detection code?
> 
> e.g.:
> 
>     UNAME_S := $(shell uname -s)
>     ifeq ($(UNAME_S),Linux)
>         DETECTED_OS=linux
>         MAKEFLAGS += -j $(shell nproc)
>     endif
> 
> nproc is an useful utility belonging to GNU coreutils, thus it is commonly found on Linux systems. As for its function, according to its manual, it automatically prints the number of cores available to the current process, this can be used to set the appropriate number of jobs for every system. For example, it would set 4 jobs on a Raspberry Pi that has four cores, but on a desktop with i7 or Ryzen it would set 8 (or more) jobs instead.

Also, there's a mac equivalent of nproc:  `sysctl -n hw.logicalcpu`.

EllipticEllipsis:
Following discussion in Discord, we can scrap the complicated platform-dependent stuff and just instruct the user to pass `-j` or `-jN` to the original setup command instead.

README.md
```diff
@@ -151,6 +151,13 @@ md5sum: WARNING: 1 computed checksum did NOT match
 
 This means that the built ROM isn't the same as the base one, so something went wrong or some part of the code doesn't match.
 
+**NOTE:** to speed up the build, you can either:
+* pass `-jN` to `make setup` and `make`, where N is the number of threads to use in the build. The generally-accepted wisdom is to use the number of virtual cores your computer has.
+* pass `-j` to `make setup` and `make`, to use as many threads as possible, but beware that this can use too much memory on lower-end systems.
+
+Both of these have the disadvantage that the ordering of the terminal output is scrambled, so for debugging it is best to stick to one.
```
AngheloAlf, line 158: > so for debugging it is best to stick to one.
To one... what?

AngheloAlf (approved):

Roman971 (approved):

PR #805, opened 2021-05-03
Add every unreserved object in an automated way

AngheloAlf:
- Adds every unreserved object.
- Updates the spec accordingly.
- Edit `format.sh` script to add newlines to XML files.
- The names are autogenerated.
- It doesn't has non referenced stuff in the objects (See eyes/mouth textures), so this would need to be done manually.
- Doesn't update the pointers in code.
- There are a few problematic files.
  - The "blank" ones. I added them as small blobs to make them match, but this should be discussed in the future.
  - The objects with unaligned textures. See ZAPD issue [125](https://github.com/zeldaret/ZAPD/issues/125). I added them, but commented most of their contents.


fig02 (approved):

EllipticEllipsis (commented):
Very nice. I just have two comments: a rewrite suggestion for part of the tutorial, and a request that the files for the objects with unaligned are commented more consistently, to make it easier to work with them when we decide what to do with them.

AngheloAlf:
I updated the comments in those objects.
What part of the tutorial would you like to change?

docs/tutorial/object_decomp.md
```diff
@@ -21,25 +22,23 @@ Because these are regarded as copyrighted, we do not want them in the repository
...
+Select the XML file of your selected object, which should be in `assets/xml/objects/object_name.xml`.
 
-The ZAPD output will go in the folder `assets/objects/object_name/`: you'll want this folder open later to check the output is correct.
+The ZAPD output will go in the folder `assets/objects/object_name/`. You'll want this folder open later to check the output is correct.
 
+### Examining actor files
+
+Most objects are used by at least one actor. For those used by an actor, we can use the script `tools/xmlcreate.py` on the actor to separate all the blobs of data in the object that we already know about. (While it is possible to do this manually, it is much simpler to run the script first and sort it out afterwards, since it won't miss anything accidentally.)
 
+Many objects were added in an automated way, so most parts of each object are already identified, but still need to name and document them them properly. Also, they usually have some unreferenced blobs by the object itself, most of the time those are referenced by the actor who uses said object.
```
EllipticEllipsis, line 35:
```suggestion
Many objects have been added in an automated way, so most constituent parts of each object are already identified, but will still need to be named and documented properly. Also, these objects usually have some blobs unreferenced by the object's own contents and hence not automatically extracted; most of the time these can be identified by looking at references in the actor which uses said object.
```

fig02:
That part will still be manual. All this does is auto generate xmls for the things that are easily detectable. A person still has to label a few more unaccounted things and replace symbols.

AngheloAlf:
That's what "object decomp" will be from now on. Naming and documenting every object; and replacing the hardcoded pointers in the codebase and undefined syms.

Roman971 (commented):
This seems fine to me overall.

My only concern is generated names like `gobject_BbAnim_00000184` seem wrong, even as defaults. Maybe we should use a camel case version of the object name as the default, or at least remove the g since it doesn't really make sense when followed with a snake case name.

assets/xml/objects/object_ahg.xml
```diff
@@ -0,0 +1,67 @@
+<Root>
+    <File Name="object_ahg" Segment="6">
+        <!-- The textures used by this limbs are unaligned, so this won't compile to a matching ROM. -->
```
Roman971, line 3: Also minor suggestions for these comments:
```suggestion
        <!-- The textures used by these limbs aren't 64-bit aligned, so this won't compile to a matching ROM. -->
```

assets/xml/objects/object_ahg.xml
```diff
@@ -0,0 +1,67 @@
...
+        <Limb Name="gobject_ahgLimb_00000090" LimbType="Standard" Offset="0x90" />
+        <Limb Name="gobject_ahgLimb_0000009C" LimbType="Standard" Offset="0x9C" />
+        <Limb Name="gobject_ahgLimb_000000A8" LimbType="Standard" Offset="0xA8" />
+        -->
+
+        <!-- 
+        <Skeleton Name="gobject_ahgSkel_000000F0" Type="Flex" LimbType="Standard" Offset="0xF0" />
+        -->
+
+        <!-- Unaligned textures. -->
```
Roman971, line 26:
```suggestion
        <!-- 32-bit aligned textures. -->
```

AngheloAlf:
What about dropping the `gobject_` part of the name?
For example: `gobject_BbAnim_00000184` -> `BbAnim_00000184`

Roman971:
> What about dropping the `gobject_` part of the name?
> For example: `gobject_BbAnim_00000184` -> `BbAnim_00000184`

I think it would be better to keep the object_ part but drop the `g`, and maybe add an extra `_` before the `Anim` (or `Tex`, etc).
For example: `object_Bb_Anim_00000184`

This should make pretty clear it's a default name and which part of it is the object's name.

Alternatively we could turn the object_ name into a camel case version while keeping the `g`, so it follows our naming scheme, but I don't know if that's worth doing or if it will turn out good for some object names.

Roman971 (approved):

PR #809, opened 2021-05-10
Port extract_baserom to C

louist103:
Discussion for context:
https://discord.com/channels/688807550715560050/688851451014938642/841030065998397460


Makefile
```diff
@@ -203,10 +203,9 @@ distclean: clean
 	$(MAKE) -C tools distclean
 
 setup:
-	$(MAKE) -C tools
-	python3 fixbaserom.py
-	python3 extract_baserom.py
-	python3 extract_assets.py
+	$(MAKE) -C tools -j
+	mkdir -p baserom
+	tools/extract_baserom
```
AngheloAlf, line 208: What happened to `fixbaserom.py` and `extract_assets.py`?

```suggestion
	$(MAKE) -C tools
	mkdir -p baserom
	tools/extract_baserom
```

tools/extract_baserom.c
```diff
@@ -0,0 +1,1614 @@
...
+
+        if (fread(&virtualEnd, sizeof(unsigned int), 1, rom) != 1)
+        {
+            fprintf(stderr, "Could not read end address for file %s\n", sDmaMgrFileNames[i]);
+            exit(1);
+        }
+
+        virtualEnd = __bswap_32(virtualEnd);
+
+        void *data = malloc(virtualEnd - virtualStart);
```
AngheloAlf, line 1582: You are leaking lots of RAM here. That _may_ be a problem to the guys using WSL, since they always complain about WSL eating all their RAM, but I don't know

AngheloAlf (commented):
So, this is dropping support for future compressed ROMs, right?

tools/extract_baserom.c
louist103: I had a `free(data)` on a the original but I guess it got lost when moving it into the repo.

AngheloAlf: Thinking about it, doing a `malloc` in each iteration of the loop may be slow. You could just do a huge `malloc` before the for loop and re-use that  pointer over and over again. Just do a big enough `enough` malloc to ensure every file will always fit in it.

Makefile
```diff
@@ -204,9 +204,9 @@ distclean: clean
 
 setup:
 	$(MAKE) -C tools
+	mkdir -p baserom
```
AngheloAlf, line 207: Is this line necessary? There's a line with a comment saying something like 
```
# create build directories
```
And it has `baserom` on it.

Makefile
```diff
@@ -204,9 +204,9 @@ distclean: clean
...
 	python3 fixbaserom.py
-	python3 extract_baserom.py
-	python3 extract_assets.py
+	tools/extract_baserom
```
AngheloAlf, line 210: I feel like a dejavu, did you removed `extract_assets.py` again?

tools/extract_baserom.c
```diff
@@ -0,0 +1,1604 @@
...
+
+        if (fread(data, virtualEnd - virtualStart, 1, rom) != 1) {
+            fprintf(stderr, "Could not read rom file %s\n", sDmaMgrFileNames[i]);
+        }
+
+        printf("Extracting file %s	0x%x - 0x%x\n", sDmaMgrFileNames[i], virtualEnd, virtualStart);
+
+        if (fwrite(data, virtualEnd - virtualStart, 1, outFile) != 1) {
+            fprintf(stderr, "Could not write rom file %s\n", sDmaMgrFileNames[i]);
+        }
```
AngheloAlf, line 1598: I was thinking, do we really need to read the file and write it again?
Maybe there's a way to write a portion of the rom to a new file without actually moving it to RAM. I remember reading something about `mmap` was used for stuff like this, but I don't really remember.

Makefile
louist103: That creates `build/*`  This creates `baserom/`  I need to do it here because the C program failed if the folder doesn't exist.

AngheloAlf: What about adding `baserom` to that line?

tools/extract_baserom.c
louist103: From what I just looked up `mmap` is just wrapped by `malloc` and on top of that ts a system function which could make more of a mess.

Makefile
louist103: I could but it seems that its there for making `build`. I'll add it and see if anyone objects.

fig02 (approved):

Roman971:
Could you explain the purpose of the filelist_all_* files?

Also shouldn't extract_baserom.py be removed now since it will be obsolete?

Makefile
```diff
@@ -136,7 +136,7 @@ TEXTURE_FILES_OUT := $(foreach f,$(TEXTURE_FILES_RGBA32:.rgba32.png=.rgba32.inc.
 					 $(foreach f,$(TEXTURE_FILES_JPG:.jpg=.jpg.inc.c),build/$f) \
 
 # create build directories
-$(shell mkdir -p build/baserom $(foreach dir,$(SRC_DIRS) $(ASM_DIRS) $(ASSET_BIN_DIRS),build/$(dir)))
+$(shell mkdir -p build/baserom baserom/ $(foreach dir,$(SRC_DIRS) $(ASM_DIRS) $(ASSET_BIN_DIRS),build/$(dir)))
```
Roman971, line 139: Any chance the program itself could create the `baserom/` folder instead of doing it here? This is supposed to create build directories, and `baserom/` doesn't belong in that category.

louist103:
I ran the numbers for the performance of this and here they are.  fixbaserom really shows a difference when it actually has to byte/word swap.

Fix baserom:
| Cycles | Time | Instructions | | 
| --- | --- | --- | --- |
| 289295586 | 0.131 | 513197361 | C Big |
| 447530378 | 0.226 | 607286694 | PY Big |
| 849692398 | 0.226 | 1820213461 | C Little |
| 2444020985 | 0.884 | 5142172060 | PY Little |
| 1182633799 | 0.323 | 1820213901 | C Byte Swap |
| 3614763020 | 1.303 | 8054267007 | PY Byte Swap |

Extract baserom:
| Cycles | Time | Instructions | | 
| --- | --- | --- | --- |
| 7135630 | 0.061 | 6752230 | C |
| 261451777 | 0.098 | 230069662 | PY |

Something interesting to note is that the C version of Extract baserom is still faster than python despite python being multi threaded and the C version being single threaded.

fig02:
one of the commits says removed file lists, but it looks like theyre still there

louist103:
> one of the commits says removed file lists, but it looks like theyre still there

I meant I removed the ones that will never be used (List all and list objects).  The other lists will be used by other versions later.

Roman971:
Having `extract_baserom.py` rewritten to C seems fine because it's not planned to stay forever, tho it's probably slower overall since users have to compile it, so I question the gains. But I don't think we should have scripts like `fixbaserom.py` in C.
Imo the performance gains we can get on this kind of one-time script would never outweigh the benefits of working with Python, especially with `fixbaserom.py` since it will require many changes in the future (to handle compression, multi version support, etc).

fig02:
I think I agree here, starting to think we should just keep what we have now

louist103:
My thinking is that when we support other versions these tools will need to be updated anyway.  I think there is a python version of extractbaserom somewhere but I haven't timed the difference between mine and theirs.

fig02:
These tools will need to be updated yea. Romans point is that when we do inevitably need to update these tools, the advantages of doing quick and dirty tooling in python (like we already do with many tools) outweighs the likely small performance gains of these things being C. 
I see no need to move away from what we are doing currently, personally

PR #810, opened 2021-05-10
Label a bug in Door_Shutter

fig02:
This one took like 3 days to figure out cause rcp crash with no helpful debug screen :)))

This was found while working on a separate project with custom maps. Its a miracle this bug never shows itself in the original game.
Thanks to tharo for helping me debug this as well, was quite an adventure

Below is debug printf output showing the order of events, if anyone is curious.

```
FAR DOOR INIT START
Next action set: DoorShutter_SetupType
FAR DOOR INIT END

FAR DOOR DRAW START
this->unk_16C: 0
sp70->b: 060067A0 (gohmas door display list, not the right one)
FAR DOOR DRAW END

FAR DOOR UPDATE START
DoorShutter_SetupType is starting
Required object index: 0
Is the object loaded?.. Yes
continuing setup
Starting DoorShutter_SetupDoor
16C set in DoorShutter_SetupDoor: 8
FAR DOOR UPDATE END

FAR DOOR DRAW START
this->unk_16C: 8
sp70->b: 0404B0D0 (gDungeonDoorDL)
FAR DOOR DRAW END
```

### DIFF
--- a/src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
+++ b/src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
@@ -678,7 +678,7 @@ void DoorShutter_Draw(Actor* thisx, GlobalContext* globalCtx) {
     DoorShutter* this = THIS;
 
     //! @bug This actor is not fully initialized until the required object dependency is loaded.
-    //! In  most cases, the check for objBankIndex to equal requiredObjBankIndex prevents the actor
+    //! In most cases, the check for objBankIndex to equal requiredObjBankIndex prevents the actor
     //! from drawing until initialization is complete. However when the required object is 
     //! gameplay_keep, this check will pass no matter what.
     //! This only matters in very specific scenarios, when the door is unculled on the first possible frame after
--- a/src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
+++ b/src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
@@ -677,17 +677,17 @@ s32 func_80997A34(DoorShutter* this, GlobalContext* globalCtx) {
 void DoorShutter_Draw(Actor* thisx, GlobalContext* globalCtx) {
     DoorShutter* this = THIS;
 
-//! @bug This actor is not fully initialized until the required object dependency is loaded.
-//! In most cases, the check for objBankIndex to equal requiredObjBankIndex prevents the actor
-//! from drawing until initialization is complete. However if the required object is the same as the
-//! object dependency listed in init vars (gameplay_keep in this case), the check will pass even though
-//! initialization has not completed. When this happens, it will try to draw the display list of the 
-//! first entry in `D_80998134`, which will likely crash the game.
-//! This only matters in very specific scenarios, when the door is unculled on the first possible frame
-//! after spawning. It will try to draw without having run update yet.
-//!
-//! The best way to fix this issue (and what was done in Majora's Mask) is to null out the draw function in
-//! the init vars for the actor, and only set draw after initialization is complete.
+    //! @bug This actor is not fully initialized until the required object dependency is loaded.
+    //! In most cases, the check for objBankIndex to equal requiredObjBankIndex prevents the actor
+    //! from drawing until initialization is complete. However if the required object is the same as the
+    //! object dependency listed in init vars (gameplay_keep in this case), the check will pass even though
+    //! initialization has not completed. When this happens, it will try to draw the display list of the
+    //! first entry in `D_80998134`, which will likely crash the game.
+    //! This only matters in very specific scenarios, when the door is unculled on the first possible frame
+    //! after spawning. It will try to draw without having run update yet.
+    //!
+    //! The best way to fix this issue (and what was done in Majora's Mask) is to null out the draw function in
+    //! the init vars for the actor, and only set draw after initialization is complete.
 
     if (this->dyna.actor.objBankIndex == this->requiredObjBankIndex &&
         (this->unk_16B == 0 || func_80997A34(this, globalCtx) != 0)) {
--- a/src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
+++ b/src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
@@ -677,20 +677,27 @@ s32 func_80997A34(DoorShutter* this, GlobalContext* globalCtx) {
 void DoorShutter_Draw(Actor* thisx, GlobalContext* globalCtx) {
     DoorShutter* this = THIS;
 
+    //! @bug This actor is not fully initialized until the required object dependency is loaded.
+    //! In  most cases, the check for objBankIndex to equal requiredObjBankIndex prevents the actor  the
+    //! from drawing until initialization is complete. However when the required object is 
+    //! gameplay_keep, this check will pass no matter what.
+    //! This only matters in very specific scenarios, when the door is unculled on the first possible frame after
+    //! spawning. It will try to draw on the first possible frame without having run update yet.
+    //! If the required object is gameplay_keep, the check will pass even though initialization has not completed
+    //! yet, and will try to draw the display list of the first entry in `D_80998134`, which will likely crash the game.
+    //!
+    //! The best way to fix this issue (and what was done in Majora's Mask) is to null out draw in the init vars
+    //! for the actor, and only set draw after initialization is complete.
+
     if (this->dyna.actor.objBankIndex == this->requiredObjBankIndex &&
         (this->unk_16B == 0 || func_80997A34(this, globalCtx) != 0)) {
         s32 pad[2];
         ShutterInfo* sp70 = &D_80998134[this->unk_16C];
 
-        //! @bug In very niche scenarios, depending on the placement of other doors when a new room is loading,
-        //! it is possible that a newly spawned door will try to draw after init, without having run it's first update.
-        //! In this case, `DoorShutter_SetupDoor` will not have run yet, so unk_16C is not initialized.
-        //! When this occurs, it will try to draw the display list of the first entry in `D_80998134`.
-        //! This bug can be avoided either by initialized 16C in init, or by checking 
-        //! to make sure the setup has complete before allowing it to draw.
-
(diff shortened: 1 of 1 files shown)


src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
```diff
@@ -682,6 +682,13 @@ void DoorShutter_Draw(Actor* thisx, GlobalContext* globalCtx) {
         s32 pad[2];
         ShutterInfo* sp70 = &D_80998134[this->unk_16C];
 
+        //! @bug In very niche scenarios, depending on the placement of other doors when a new room is loading,
+        //! it is possible that a newly spawned door will try to draw after init, without having run it's first update.
```
EllipticEllipsis, line 686: Typo
```suggestion
        //! it is possible that a newly spawned door will try to draw after init, without having run its first update.
```

src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
```diff
@@ -682,6 +682,13 @@ void DoorShutter_Draw(Actor* thisx, GlobalContext* globalCtx) {
...
+        //! In this case, `DoorShutter_SetupDoor` will not have run yet, so unk_16C is not initialized.
+        //! When this occurs, it will try to draw the display list of the first entry in `D_80998134`.
+        //! This bug can be avoided either by initialized 16C in init, or by checking
```
EllipticEllipsis, line 689:
```suggestion
        //! This bug can be avoided either by initialized unk_16C in init, or by checking 
```
Recommend using the full var name so a find/replace will catch it easily when it's renamed.

src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
```diff
@@ -682,6 +682,13 @@ void DoorShutter_Draw(Actor* thisx, GlobalContext* globalCtx) {
         s32 pad[2];
         ShutterInfo* sp70 = &D_80998134[this->unk_16C];
 
+        //! @bug In very niche scenarios, depending on the placement of other doors when a new room is loading,
```
EllipticEllipsis, line 685: Is it possible/worth it to be any more specific here, especially if it was so hard to figure out in the first place?

EllipticEllipsis (commented):
Wow, nice work figuring this out. Couple of minor things, and one more major one.

src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
fig02: yea I didnt want the comment to be a novel lmao
I think I got the point across though, if theres a door placed such that it will be drawing the same frame it spawns (a door nearby where you are entering from another door), the issue can occur. Thats what I mean by `depending on the placement of other doors when a new room is loading`

fig02: We could get into more detail about how the doors need to be placed, but really any new project should just patch this asap and not worry about it. Cause it all depends on the placement of thing in the map

Zelllll: Yeah I kind of agree since when you explained this to me earlier I was still a little confused why this happens at all, but I'd just emphasize that the displaylist index not being set is what causes the bug

AngheloAlf: I think the more information we have about a bug, the better, even if everybody will just patch the bug.

EllipticEllipsis: So does it happen when it's placed so that the camera sees it too early (and so demands Draw before Update happens for the first time), is that the point?

fig02: Yes, if it is not culled the same frame it is spawned (init has just run), then it will draw once before update 

How this issue doesn't happen in vanilla, idk. In this rom hack I mentioned, the door is shaped like a U, and the exit is directly in line of sight behind the player during the door closing animation.

EllipticEllipsis: Ah, in that case I think something like "the door is placed so that the camera is on it in the first frame after init" or "placed so that it needs to be drawn too soon" or similar would cover it without going overboard. Just to be more specific about exactly what about the placement causes the issue.

src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
```diff
@@ -682,6 +682,13 @@ void DoorShutter_Draw(Actor* thisx, GlobalContext* globalCtx) {
...
+        //! it is possible that a newly spawned door will try to draw after init, without having run its first update.
+        //! In this case, `DoorShutter_SetupDoor` will not have run yet, so unk_16C is not initialized.
+        //! When this occurs, it will try to draw the display list of the first entry in `D_80998134`.
+        //! This bug can be avoided either by initialized unk_16C in init, or by checking
```
pixel-stuck, line 689:
```suggestion
        //! This bug can be avoided either by initializing unk_16C in init, or by checking 
```
weird tense issue

src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
mzxrules: After spending some time talking to Fig and understanding the glitch more, I think the comment should be something like this, positioned above the IF statement:
 
            This actor isn't fully initialized until the first time Update is called when the actor's dependent
            object is loaded. The check objBankIndex == requiredObjBankIndex stops the actor from drawing until Update 
            completes initialization, except in the case when requiredObjBankIndex is OBJECT_GAMEPLAY_KEEP.
            In this case, unk_16C will be 0, and try to draw the display list of the first entry in `D_80998134`, which can crash 
            the game.

fig02:
finally got around to updating this. Took @mzxrules's suggestion and added a little bit more to it.

EllipticEllipsis (approved):

mzxrules:
The following lines are redundant:  

`However if the required object is the same as the object dependency listed in init vars, this check will pass no matter what.`
`If the required object is the same as the init vars requirement (gameplay_keep in this case), the check will pass even though initialization has not completed.`

Also I feel things flow better written like so: 

    //! @bug This actor is not fully initialized until the required object dependency is loaded.
    //! In most cases, the check for objBankIndex to equal requiredObjBankIndex prevents the actor
    //! from drawing until initialization is complete. However if the required object is the same as the
    //! object dependency listed in init vars (gameplay_keep in this case), the check will pass even though
    //! initialization has not completed. When this happens, it will try to draw the display list of the 
    //! first entry in `D_80998134`, which will likely crash the game.
    //! This only matters in very specific scenarios, when the door is unculled on the first possible frame
    //! after spawning. It will try to draw without having run update yet.
    //!
    //! The best way to fix this issue (and what was done in Majora's Mask) is to null out the draw function in
    //! the init vars for the actor, and only set draw after initialization is complete.

src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
```diff
@@ -677,13 +677,28 @@ s32 func_80997A34(DoorShutter* this, GlobalContext* globalCtx) {
...
+    //! from drawing until initialization is complete. However if the required object is the same as the
+    //! object dependency listed in init vars, this check will pass no matter what.
+    //! This only matters in very specific scenarios, when the door is unculled on the first possible frame
+    //! after spawning. It will try to draw without having run update yet.
+    //! If the required object is the same as the init vars requirement (gameplay_keep in this case), the check will
+    //! pass even though initialization has not completed. When this happens, it will try to draw the display list
+    //! of the first entry in `D_80998134`, which will likely crash the game.
+    //!
+    //! The best way to fix this issue (and what was done in Majora's Mask) is to null out the draw function in
+    //! the init vars for the actor, and only set draw after initialization is complete.
```
fig02, line 691:
```suggestion
//! @bug This actor is not fully initialized until the required object dependency is loaded.
//! In most cases, the check for objBankIndex to equal requiredObjBankIndex prevents the actor
//! from drawing until initialization is complete. However if the required object is the same as the
//! object dependency listed in init vars (gameplay_keep in this case), the check will pass even though
//! initialization has not completed. When this happens, it will try to draw the display list of the 
//! first entry in `D_80998134`, which will likely crash the game.
//! This only matters in very specific scenarios, when the door is unculled on the first possible frame
//! after spawning. It will try to draw without having run update yet.
//!
//! The best way to fix this issue (and what was done in Majora's Mask) is to null out the draw function in
//! the init vars for the actor, and only set draw after initialization is complete.
```

fig02:
thanks, yea that wording is good. changed

mzxrules (approved):

AngheloAlf (approved):

Zelllll (approved):

src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
```diff
@@ -677,13 +677,27 @@ s32 func_80997A34(DoorShutter* this, GlobalContext* globalCtx) {
...
+//! In most cases, the check for objBankIndex to equal requiredObjBankIndex prevents the actor
+//! from drawing until initialization is complete. However if the required object is the same as the
+//! object dependency listed in init vars (gameplay_keep in this case), the check will pass even though
+//! initialization has not completed. When this happens, it will try to draw the display list of the 
+//! first entry in `D_80998134`, which will likely crash the game.
+//! This only matters in very specific scenarios, when the door is unculled on the first possible frame
+//! after spawning. It will try to draw without having run update yet.
+//!
+//! The best way to fix this issue (and what was done in Majora's Mask) is to null out the draw function in
+//! the init vars for the actor, and only set draw after initialization is complete.
```
Roman971, line 690: Shouldn't these comments be indented? And I would be surprised if clang-format leaves it like that

fig02: oh, i used githubs built in commit thing for mzx's suggestion lol. will fix in a moment

Roman971 (approved):

PR #814, opened 2021-05-13
Simplify PNGs rules in the makefile

AngheloAlf:
- Automatically detect the format of a PNG based on its extension.
- Doesn't need to list every possible format.
- Updates `.gitignore` accordingly.
- Add `-eh` to every ZAPD invocation.

Thanks to @Roman971 for the makefile rule.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -242,12 +242,12 @@ assets/%.c: assets/%.xml
 	$(CC) -c $(CFLAGS) $(MIPS_VERSION) $(OPTFLAGS) -o build/$(@:.c=.o) $@
 
 build/%.inc.c: %.png
-	$(ZAPD) btex -eh -tt $(lastword ,$(subst ., ,$(basename $<))) -i $< -o $@
+	$(ZAPD) btex -eh -tt $(subst .,,$(suffix $*)) -i $< -o $@
 
 build/assets/%.bin.inc.c: assets/%.bin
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -254,41 +238,16 @@ build/src/libultra_code_O1/llcvt.o: src/libultra_code_O1/llcvt.c
...
+build/%.inc.c: %.png
+	$(ZAPD) btex -eh -tt $(lastword ,$(subst ., ,$(basename $<))) -i $< -o $@
 
 build/assets/%.bin.inc.c: assets/%.bin
-	$(ZAPD) bblb -i $< -o $@
+	$(ZAPD) bblb -eh -i $< -o $@
 
 build/assets/%.jpg.inc.c: assets/%.jpg
-	$(ZAPD) bren -i $< -o $@ -eh
+	$(ZAPD) bren -eh -i $< -o $@ -eh
```
Roman971, line 251: I'm not sure what `-eh` does, but is it intended to appear twice in the command here?

Roman971:
Seems good to me besides the small thing I mentioned

Makefile
AngheloAlf: `-eh` is to enable ZAPD's error handler. And it should appear twice.

Roman971 (approved):

Zelllll (approved):
looks good

fig02 (approved):

PR #816, opened 2021-05-15
Small bgcheck documentation pass

mzxrules:
* Give a few functions names
* Rename LineTestImpl -> CheckLineImpl to make it consistent with other names. I didn't rename the public wrappers for LineTestImpl because I don't intent do give them final names, and I didn't want to touch a bunch of files.
* Rename SphVsWallImpl -> CheckWallImpl since not all collision tests respect the sphere
* Mark the check in CheckWallImpl that doesn't use checkHeight as a bug.
* Document the GetMinY bug, which does occur in vanilla.


include/functions.h
```diff
@@ -570,19 +570,19 @@ void CollisionPoly_GetVerticesByBgId(CollisionPoly* poly, s32 bgId, CollisionCon
...
-// ? func_8003AEA8(?);
-// ? func_8003B04C(?);
-// ? func_8003B218(?);
-// ? func_8003B3C8(?);
-// ? func_8003BB18(?);
+// ? BgCheck_GetSubdivisionMinBounds(?);
+// ? BgCheck_GetSubdivisionMaxBounds(?);
+// ? BgCheck_GetPolySubdivisionBounds(?);
+// ? BgCheck_PolyIntersectsSubdivision(?);
+// ? BgCheck_InitializeStaticLookup(?);
```
fig02, line 585: for all of these commented prototypes, the should either be added here or deleted if they are internal functions to bgcheck not meant to be called externally

fig02 (approved):

Roman971 (approved):

PR #819, opened 2021-05-20
Removes qemu from the README.md

Kenix3:
This has also been requested by Roman.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ Once mips-linux-binutils is installed you will need to install the following pac
 * dos2unix
 * python3
 
-Then you can continue from step 2 of the Linux instructions.
+Then you can continue from step [step 2](#2-clone-the-repository) of the Linux instructions.
 
 Note that, before building anything, you will need to run the following commands to fix line endings:
 
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ dos2unix fixle.sh
 
 To use Docker, you'll need either Docker Desktop or Docker Toolbox installed and setup based on your system.
 
-You'll also need to prepare a local version of the project with a copied base ROM (see steps 2 and 3 of the Linux instructions).
+You'll also need to prepare a local version of the project with a copied base ROM (see steps [2](#2-clone-the-repository) and [3](#3-prepare-a-base-rom) of the Linux instructions).
 
 #### 2. Create the docker image
 
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ docker run -it --rm --mount type=bind,source="$(pwd)",destination=/oot oot /bin/
 
 #### 4. Setup and Build the ROM
 
-Once inside the container, you can follow steps 4 and 5 of the Linux instructions to setup and build the ROM, or run any other command you need.
+Once inside the container, you can follow steps [4](#4-setup-the-rom-and-build-process) and [5](#5-build-the-rom) of the Linux instructions to setup and build the ROM, or run any other command you need.
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ You'll also need to [build and install mips-linux-binutils](docs/BUILDING_BINUTI
 Going forward in this guide, please use `gmake` whenever you encounter a `make` command.
 The `make` that comes with MacOS behaves differently than GNU make and is incompatible with this project.
 
-You should now be able to continue from step 2 of the Linux instructions.
+You should now be able to continue from [step 2](#2-clone-the-repository) of the Linux instructions.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -88,35 +88,20 @@ sudo apt-get update
...
-
-Open up your .bashrc file (~/.bashrc), scroll to the bottom, and add the following, replacing the paths as necessary:
-
-```text
-export QEMU_IRIX="path/to/your/qemu-irix"
-```
-
-Save and close/reopen your terminal window.
-
-#### 3. Clone the repository
```
AngheloAlf, line 106: What about moving this to a file in `docs/` with instruction on how to set up this instead of removing it completely?

fig02: imo, no need. We kept the old ido method around incase recomp had any issues, but we've been using it for months with no hiccups at all. This method is worse in every way and really should never be needed going forward.
We'll still have the files and now how to use it in the very rare case we do need to use it. But the chance of that happening is really low

Kenix3: I think at the level that QEMU is actually needed (no known cases) this is best stored in Discord.

README.md
```diff
@@ -88,35 +88,20 @@ sudo apt-get update
...
+#### 2. Clone the repository
```
Roman971, line 91: Since this changes the step numbers, I think there are references to steps that need to be updated in the rest of the README (in the Docker instructions iirc but there might be more)

README.md
AngheloAlf: I remember one guy a few months ago needed to use the QEMU solution to properly setup decomp on their machine, so there's 1 known case.

Roman971: I think having a pin with these instructions in Discord should be enough. This way we at least have somewhere we can point to in case the setup is needed.

README.md
Kenix3: Good catch!

README.md
AngheloAlf: Okey.
(Please resolve this conversation)

README.md
```diff
@@ -68,7 +68,7 @@ You'll also need to [build and install mips-linux-binutils](docs/BUILDING_BINUTI
 Going forward in this guide, please use `gmake` whenever you encounter a `make` command.
 The `make` that comes with MacOS behaves differently than GNU make and is incompatible with this project.
 
-You should now be able to continue from step 3 of the Linux instructions.
+You should now be able to continue from step 2 of the Linux instructions.
```
AngheloAlf, line 71:
```suggestion
You should now be able to continue from [step 2](#2-clone-the-repository) of the Linux instructions.
```

README.md
```diff
@@ -172,7 +157,7 @@ Once mips-linux-binutils is installed you will need to install the following pac
 * dos2unix
 * python3
 
-Then you can continue from step 3 of the Linux instructions.
+Then you can continue from step 2 of the Linux instructions.
```
AngheloAlf, line 160:
```suggestion
Then you can continue from step [step 2](#2-clone-the-repository) of the Linux instructions.
```

fig02 (approved):

README.md
```diff
@@ -187,7 +172,7 @@ dos2unix fixle.sh
 
 To use Docker, you'll need either Docker Desktop or Docker Toolbox installed and setup based on your system.
 
-You'll also need to prepare a local version of the project with a copied base ROM (see steps 3 and 4 of the Linux instructions).
+You'll also need to prepare a local version of the project with a copied base ROM (see steps 2 and 3 of the Linux instructions).
```
AngheloAlf, line 175:
```suggestion
You'll also need to prepare a local version of the project with a copied base ROM (see steps [2](#2-clone-the-repository) and [3](#3-prepare-a-base-rom) of the Linux instructions).
```

README.md
```diff
@@ -207,7 +192,7 @@ docker run -it --rm --mount type=bind,source="$(pwd)",destination=/oot oot /bin/
 
 #### 4. Setup and Build the ROM
 
-Once inside the container, you can follow steps 5 and 6 of the Linux instructions to setup and build the ROM, or run any other command you need.
+Once inside the container, you can follow steps 4 and 5 of the Linux instructions to setup and build the ROM, or run any other command you need.
```
AngheloAlf, line 195:
```suggestion
Once inside the container, you can follow steps [4](#4-setup-the-rom-and-build-process) and [5](#5-build-the-rom) of the Linux instructions to setup and build the ROM, or run any other command you need.
```

AngheloAlf (changes requested):

AngheloAlf (approved):

Roman971 (approved):

PR #822, opened 2021-05-22
Interface improvements

Zelllll:
- Improve `Interface_Draw` (permuter claims the score is about half of the old score but probably not the best measurement, either way its improved quite a bit).
- Add some macros for positioning in the interface, and made `z64interface.h`
- Renamed `R_MINIMAP_TOGGLED` to `R_MINIMAP_DISABLED` since it sounds more clear (does toggled mean toggled *on* or *off*?)
- Changed some of the combine modes in `Interface_Draw` to make more sense.
- Made a lot of hex numbers that should be decimal be decimal in `z_construct`.
- `z_parameter` now uses proper symbols for all of its external data.
- Inlined a lot of data inside their corresponding function, and made them `static` when applicable in `z_parameter`.


src/code/z_parameter.c
```diff
@@ -2668,39 +2668,47 @@ void Interface_DrawItemButtons(GlobalContext* globalCtx) {
...
     gDPSetPrimColor(OVERLAY_DISP++, 0, 0, R_C_BTN_COLOR(0), R_C_BTN_COLOR(1), R_C_BTN_COLOR(2),
                     interfaceCtx->cRightAlpha);
-    gSPTextureRectangle(OVERLAY_DISP++, R_ITEM_BTN_X(3) << 2, R_ITEM_BTN_Y(3) << 2,
-                        (R_ITEM_BTN_X(3) + R_ITEM_BTN_WIDTH(3)) << 2, (R_ITEM_BTN_Y(3) + R_ITEM_BTN_WIDTH(3)) << 2,
-                        G_TX_RENDERTILE, 0, 0, R_ITEM_BTN_DD(3) * 2, R_ITEM_BTN_DD(3) * 2);
+    gSPTextureRectangle(
+        OVERLAY_DISP++, R_ITEM_BTN_X(HUD_ITEM_BTN_C_RIGHT) << 2, R_ITEM_BTN_Y(HUD_ITEM_BTN_C_RIGHT) << 2,
+        (R_ITEM_BTN_X(HUD_ITEM_BTN_C_RIGHT) + R_ITEM_BTN_WIDTH(HUD_ITEM_BTN_C_RIGHT)) << 2,
+        (R_ITEM_BTN_Y(HUD_ITEM_BTN_C_RIGHT) + R_ITEM_BTN_WIDTH(HUD_ITEM_BTN_C_RIGHT)) << 2, G_TX_RENDERTILE, 0, 0,
+        R_ITEM_BTN_DD(HUD_ITEM_BTN_C_RIGHT) * 2, R_ITEM_BTN_DD(HUD_ITEM_BTN_C_RIGHT) * 2);
```
louist103, line 2702: Can these all do shifting or does that not work? If they can it should be consistent.

louist103 (approved):
Nice job on this.  Just some small things.

src/code/z_parameter.c
Zelllll: Hm what shift is * 2?

src/code/z_parameter.c
```diff
@@ -2566,7 +2568,7 @@ void Interface_DrawMagicBar(GlobalContext* globalCtx) {
             // Yellow part of the bar being used when casting a spell
```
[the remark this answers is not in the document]
fig02: its not limited to just spells. would just end the sentence at "deducted"

src/code/z_parameter.c
mzxrules: log2(2) is 1

AngheloAlf: > Hm what shift is * 2?

`X << 1`

src/code/z_parameter.c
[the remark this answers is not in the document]
Zelllll: what do you mean? It is only for spells I'm pretty sure and it's kinda useless in general because you only see it for one frame anyways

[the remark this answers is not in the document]
fig02: charging a spin is not a spell

Zelllll:
fixed everyone's comments so far, and made most texture rectangles use proper values for scales

src/code/z_parameter.c
```diff
@@ -3002,7 +3002,7 @@ void func_8008A994(InterfaceContext* interfaceCtx) {
     func_800AB2C4(&interfaceCtx->view);
 }
 
-#ifdef NON_MATCHING
+#if 1
 void Interface_Draw(GlobalContext* globalCtx) {
```
fig02: youll need to revert this back

Zelllll: lmao oops

src/code/code_80097A00.c
```diff
@@ -33,23 +36,137 @@ u16 gUpgradeCapacities[][4] = {
...
-    0x08036000, 0x08037000, 0x08038000, 0x08039000, 0x0803A000, 0x0803B000, 0x0803C000, 0x0803D000, 0x0803E000,
-    0x0803F000, 0x08040000, 0x08041000, 0x08042000, 0x08043000, 0x08044000, 0x08045000, 0x08046000, 0x08047000,
-    0x08048000, 0x08049000, 0x0804A000, 0x0804B000, 0x0804C000, 0x0804D000, 0x0804E000, 0x0804F000, 0x08050000,
-    0x08051000, 0x08052000, 0x08053000, 0x08054000, 0x08055000, 0x08056000, 0x08057000, 0x08058000, 0x08059000,
-    0x08089440, 0x08089440, 0x08089440, 0x08089440, 0x08089440, 0x08089440, 0x08089440, 0x08089440, 0x08089440,
-    0x08089440, 0x08089440, 0x08089440, 0x09000000, 0x09000900, 0x09001200, 0x09001B00, 0x09002400, 0x09002D00,
-    0x09003600, 0x09003F00, 0x09004800, 0x09005100, 0x09005A00, 0x09006300, 0x09006C00, 0x09007500, 0x09007E00,
-    0x09009000, 0x09008700, 0x09007E00, 0x0900A200, 0x0900AB00, 0x0805A000, 0x0805A900, 0x0805B200, 0x02002D40,
-    0x02002A40, 0x02002C40, 0x02002B40, 0x02002940,
+    gDekuStickIconTex,
```
AngheloAlf, line 40: The formatting of this array looks weird. Did you ran `./format.sh` right?

src/code/gfxprint.c
```diff
@@ -222,21 +222,21 @@ void GfxPrint_PrintCharImpl(GfxPrint* this, u8 c) {
 
         if (this->flag & GFXPRINT_FLAG64) {
             gSPTextureRectangle(this->dlist++, (this->posX + 4) << 1, (this->posY + 4) << 1, (this->posX + 4 + 32) << 1,
-                                (this->posY + 4 + 32) << 1, tile, (u16)(c & 4) * 64, (u16)(c >> 3) * 256, 512, 512);
+                                (this->posY + 4 + 32) << 1, tile, (u16)(c & 4) * 64, (u16)(c >> 3) * 256, (1 << 10) / 2, (1 << 10) / 2);
         } else {
             gSPTextureRectangle(this->dlist++, this->posX + 4, this->posY + 4, this->posX + 4 + 32, this->posY + 4 + 32,
-                                tile, (u16)(c & 4) * 64, (u16)(c >> 3) * 256, 1024, 1024);
+                                tile, (u16)(c & 4) * 64, (u16)(c >> 3) * 256, 1 << 10, 1 << 10);
```
AngheloAlf, line 228: Why use `1 << 10` instead of `1024` here?

src/code/z_parameter.c
```diff
@@ -3094,26 +3091,24 @@ void Interface_Draw(GlobalContext* globalCtx) {
...
-                    phi_s2 = 42;
-                    if (interfaceCtx->counterDigits[2] != 0) {
-                        OVERLAY_DISP = Gfx_TextureI8(OVERLAY_DISP, D_02003040[interfaceCtx->counterDigits[2]], 8, 16,
-                                                     phi_s2, 190, 8, 16, 1024, 1024);
-                        phi_s2 = 50;
+                    svar3 = 42;
+
+                    if (interfaceCtx->counterDigits[2]) {
+                        OVERLAY_DISP = Gfx_TextureIA8(
+                            OVERLAY_DISP, ((u8*)gHUDCounterDigit0Tex + (8 * 16 * interfaceCtx->counterDigits[2])), 8,
```
AngheloAlf, line 3098: Changes like this doesn't make sense at all.
The old `D_02003040[]` was an array of textures, so doing this kind of trickery seems bad. We should extract those assets as the proper texture array it is.

src/code/code_80097A00.c
Zelllll: ill try again, but yeah i think so

src/code/gfxprint.c
Zelllll: because its correct: http://n64devkit.square7.ch/n64man/gsp/gSPTextureRectangle.htm

src/code/z_parameter.c
Zelllll: its not an array of textures, and it won't match the same way if you try that.

AngheloAlf: How do you know it isn't?

Zelllll: try changing these to not use a u8 cast and the pointer math and it will not be the same asm. This is more than likely what they wrote, and there isn't really much we can do about it besides making macros to accommodate.

Zelllll: Also if I don't do it like this the entire bss section breaks.

AngheloAlf: I mean, an array of pointers of `u8`
```
u8 D_02003040[10][0x80] = {
    { /* First texture */ },
    { /* Second texture */ },
    { /* Third texture */ },
    // ...
};
```

Zelllll: yes i know, and that affects matching somehow, and even breaks the bss section, which is why there were the random bss dummy variables in this file before.

fig02: So many things affect bss reordering, that we cant really say that doing it this way "fixes it". Even something like changing the amount of includes in the file will change what dummy variables are required to match. The fact that changing this happened to fix bss is a coincidence more than anything

Zelllll: its literally not, again try messing with it before saying im wrong

fig02: I agree with you that in this case it fixes bss, but it will be messed up again in the future when other things change. There are many many variables to bss reordering

Zelllll: well this issue should be able to be resolved then?

fig02: not sure what you mean.
I am just making sure its clear that bss reodering being fixed does not mean it proves that this is the correct way to write this. its going to be messed up again after something else seemingly unrelated changes, cause its very volatile.

If there are other matching issues that this fixes then thats fine, but bss reodering should not be considered as proof

Zelllll: i've done quite a bit to this file and nothing else has messed it up, i could include tons of headers with symbols in them and it does not do anything. the only thing affecting this is it not being a 2d array

Zelllll:
everything brought up here has been fixed for a while now, this seems ready to merge to me honestly

include/z64interface.h
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Button HUD Colors
+ */
+#define A_BUTTON_R 0   // 90 on N64
+#define A_BUTTON_G 200 // 90 on N64
+#define A_BUTTON_B 50  // 255 on N64
+
+#define B_BUTTON_R 255 // 0 on N64
+#define B_BUTTON_G 30  // 150 on N64
+#define B_BUTTON_B 30  // 0 on N64
```
Roman971, line 13: These color values are only ever used to set the corresponding regs which are already named properly, so I don't really see the point in having an extra layer with defines for these.
The other defines in this file seem fine tho, since they are actually used for multiple things.

src/overlays/actors/ovl_kaleido_scope/z_kaleido_equipment.c
```diff
@@ -140,7 +141,7 @@ void KaleidoScope_DrawEquipment(GlobalContext* globalCtx) {
             gDPPipeSync(POLY_OPA_DISP++);
             gSPVertex(POLY_OPA_DISP++, &pauseCtx->equipVtx[j], 4, 0);
 
-            POLY_OPA_DISP = KaleidoScope_QuadTextureIA8(POLY_OPA_DISP, D_02000A00[1], 32, 32, 0);
+            POLY_OPA_DISP = KaleidoScope_QuadTextureIA8(POLY_OPA_DISP, gHUDEmptyMapTex, 32, 32, 0);
```
Roman971, line 144: FYI the name `gHUDEmptyMapTex` is incorrect. It's the square around selected items in the inventory and equipment screens. (also not HUD related)

src/overlays/actors/ovl_kaleido_scope/z_kaleido_item.c
```diff
@@ -49,7 +50,7 @@ void KaleidoScope_DrawAmmoCount(PauseContext* pauseCtx, GraphicsContext* gfxCtx,
     if (i != 0) {
         gSPVertex(POLY_OPA_DISP++, &pauseCtx->itemVtx[(sAmmoVtxOffset[item] + 27) * 4], 4, 0);
 
-        gDPLoadTextureBlock(POLY_OPA_DISP++, D_020035C0[i], G_IM_FMT_IA, G_IM_SIZ_8b, 8, 8, 0,
+        gDPLoadTextureBlock(POLY_OPA_DISP++, ((u8*)gHUDAmmo0Tex + (8 * 8 * i)), G_IM_FMT_IA, G_IM_SIZ_8b, 8, 8, 0,
```
Roman971, line 53: The names for all these HUD symbols used in the pause menu should probably be adjusted

src/code/z_parameter.c
```diff
@@ -2711,7 +2713,7 @@ void Interface_DrawItemButtons(GlobalContext* globalCtx) {
                                    G_TX_NOMIRROR | G_TX_WRAP, G_TX_NOMIRROR | G_TX_WRAP, G_TX_NOMASK, G_TX_NOMASK,
                                    G_TX_NOLOD, G_TX_NOLOD);
 
-            dxdy = 1024.0f / (R_START_LABEL_DD(gSaveContext.language) / 100.0f);
+            dxdy = (f32)(1 << 10) / (R_START_LABEL_DD(gSaveContext.language) / 100.0f);
```
Roman971, line 2716: Is the cast also necessary here?

src/code/z_parameter.c
```diff
@@ -2992,52 +2995,37 @@ void func_8008A994(InterfaceContext* interfaceCtx) {
...
-
-// Due to an unknown reason, bss ordering changes within the 5 static variables in the function below.
-// In order to restore the correct order, we need a specific number of bss variables in the file before that point.
-// For this, we introduce 2 dummy variables which end up in padding at the end of the file's bss, so they don't actually
-// take space.
-s8 sBssDummy1;
-s8 sBssDummy2;
-
 #ifdef NON_MATCHING
-// mostly regalloc, minor ordering and stack usage differences
```
Roman971, line 3024: Any chance the comment could be updated to describe what the remaining issues are that makes it nonmatching?

src/code/gfxprint.c
```diff
@@ -222,21 +222,23 @@ void GfxPrint_PrintCharImpl(GfxPrint* this, u8 c) {
 
         if (this->flag & GFXPRINT_FLAG64) {
             gSPTextureRectangle(this->dlist++, (this->posX + 4) << 1, (this->posY + 4) << 1, (this->posX + 4 + 32) << 1,
-                                (this->posY + 4 + 32) << 1, tile, (u16)(c & 4) * 64, (u16)(c >> 3) * 256, 512, 512);
+                                (this->posY + 4 + 32) << 1, tile, (u16)(c & 4) * 64, (u16)(c >> 3) * 256, (1 << 10) / 2,
+                                (1 << 10) / 2);
```
Roman971, line 226: Not too sure about the few `(1 << 10) / 2` there. I'm not too knowledgeable about this stuff but I'm pretty sure that's not how it's supposed to be (referring specifically to the `/ 2`, not the `(1 << 10)`s)

src/overlays/actors/ovl_kaleido_scope/z_kaleido_equipment.c
Zelllll: Yeah I meant to render that one here

src/overlays/actors/ovl_kaleido_scope/z_kaleido_item.c
Zelllll: How should they change?

Roman971: I think we could just drop the HUD part of it, or use a name that fits both the hud and the pause menu.

Zelllll: gAmmoDigit0Tex ?

Zelllll:
alright all of roman's comments have now been addressed

Roman971 (approved):

fig02 (approved):

PR #825, opened 2021-05-24
Improve z_jpeg documentation

vijfhoek:
As my first contribution I looked at improving the documentation a bit to familiarize myself a little with the toolchain. z_jpeg was my first victim.


data/rsp.rodata.s
```diff
@@ -33,5 +33,5 @@ glabel D_80157580
 glabel D_801579A0
     .incbin "baserom.z64", 0xBCEB40, 0x390
 
-glabel D_80157D30
+glabel njpegUCodeData
```
fig02, line 36: we only really use the `g` prefix for global, and `s` for static. Both of these youve changed here look like they should be g.
rest looks good to me

src/code/z_jpeg.c
```diff
@@ -90,6 +95,9 @@ void Jpeg_CopyToZbuffer(u16* src, u16* zbuffer, s32 x, s32 y) {
     }
 }
 
+/**
+ * Gets an u16 from an (possibly unaligned) address in memory, avoiding unaligned memory accesses.
+ */
 u16 Jpeg_GetU16(u8* ptr) {
     if (((u32)ptr & 1) == 0) { // if the address is aligned to 2
         return *(u16*)ptr;
```
fig02, line 103: if i had to guess, this probably has a more specific purpose than just "get any u16", and can probably be named accordingly. But I havent read the system at all so I could be wrong

vijfhoek: I think the only real reason for the existence of this function is for the ability to read an unaligned u16 without crashing the CPU due to use of the `lhu` instruction (as the already existing comment suggests)

data/rsp.rodata.s
vijfhoek: "njpeg" is a term here, referring to Nintendo's JPEG decoding microcode. What do you think, `gNjpeg` or `gNJpeg`?

fig02: is it an existing convention to call it nJpeg, or is that something you added now on your own?
If the latter, I dont see a need to specify that its nintendos personally. Everything in this codebase already is

vijfhoek: There's a `#define M_NJPEGTASK 4` I based it on (include/ultra64/sptask.h)

src/code/z_jpeg.c
```diff
@@ -13,32 +14,33 @@
 #define MARKER_COM 0xFE
 #define MARKER_EOI 0xD9
 
+/**
+ * Sends the JPEG decoding task to the JPEG microcode and waits for it to finish.
```
Thar0, line 18: This function creates an `OSTask`, which is responsible for sending both the microcode and the decoding task data to the RSP, and passes it on to the scheduler, so perhaps a comment like
`Schedules a JPEG decoding task and waits for it to finish.`
would be more accurate. Perhaps the function name could also better reflect that it's only scheduled at this time also?

data/rsp.rodata.s
fig02: hmm ok. I think we can probably take our liberties with the names here, but it doesnt matter too much to me atm. Can leave the n for now and figure out later if we want to keep 

but yea, `gNJpeg..` would be the way to go if we're keeping it

vijfhoek: I think I've been convinced, for two reasons - 1. as you said, of course it's Nintendo, and 2. "UCode" already makes it pretty clear anyway

src/code/z_jpeg.c
vijfhoek: How about the new name and description?

src/code/z_jpeg.c
fig02: that sounds better yea. though as you noted in your comment, it may not be unaligned. So that can possibly be confusing.
Imo id say leave what you have for now and someone else who reviews this can give their thoughts as well, Im not sure which is better

vijfhoek: I've been skimming through the rest of the source code, it seems like there are more situations like this, where an unaligned 16-bit access is needed. As far as I can see, those are all inlined though.

vijfhoek: >that sounds better yea. though as you noted in your comment, it may not be unaligned. So that can possibly be confusing.

That's true, although I feel the documentation comment clarifies this by saying "possibly unaligned"

vijfhoek: I'll just change it to GetUnalignedU16 with some slightly improved documentation - I'll see what others think of it :slightly_smiling_face:

fig02 (approved):

Thar0 (approved):

Zelllll (approved):

Roman971 (approved):

PR #832, opened 2021-05-29
All of `z_actor` OK

Zelllll:
~~Although all the functions are now matching, one of them is still technically not OK since using it creates bss issues that I am unaware how to resolve.~~
~~I also added `ACTOR_CAT_MAX` since it appears to possibly exist based on the Navi colors.~~

The entire file is now OK.


src/code/z_actor.c
```diff
@@ -62,9 +62,9 @@ void ActorShadow_DrawCircle(Actor* actor, Lights* lights, GlobalContext* globalC
 }
 
 void ActorShadow_DrawWhiteCircle(Actor* actor, Lights* lights, GlobalContext* globalCtx) {
-    static Color_RGBA8 white = { 255, 255, 255, 255 };
+    static Color_RGBA8 whiteCol = { 255, 255, 255, 255 };
```
Roman971, line 65: `whileColor` would be much better than `whiteCol` imo, tho `white` seems good enough to describe it tbh

src/code/z_actor.c
```diff
@@ -232,18 +222,24 @@ typedef struct {
...
+    { { 0, 255, 0, 255 }, { 0, 255, 0, 0 } },         // ACTORCAT_EXPLOSIVE,
+    { { 150, 150, 255, 255 }, { 150, 150, 255, 0 } }, // ACTORCAT_NPC,
+    { { 255, 255, 0, 255 }, { 200, 155, 0, 0 } },     // ACTORCAT_ENEMY,
+    { { 0, 255, 0, 255 }, { 0, 255, 0, 0 } },         // ACTORCAT_PROP,
+    { { 0, 255, 0, 255 }, { 0, 255, 0, 0 } },         // ACTORCAT_ITEMACTION,
+    { { 0, 255, 0, 255 }, { 0, 255, 0, 0 } },         // ACTORCAT_MISC,
+    { { 255, 255, 0, 255 }, { 200, 155, 0, 0 } },     // ACTORCAT_BOSS,
+    { { 0, 255, 0, 255 }, { 0, 255, 0, 0 } },         // ACTORCAT_DOOR,
+    { { 0, 255, 0, 255 }, { 0, 255, 0, 0 } },         // ACTORCAT_CHEST
+    { { 0, 255, 0, 255 }, { 0, 255, 0, 0 } },         // ACTORCAT_MAX
```
Roman971, line 238: Having an entry marked as `ACTORCAT_MAX` doesn't make sense imo because this array would never be accessed with `ACTORCAT_MAX`. It would be better to just mark it as `// unknown extra entry` or something like that.

Also not sure about having a comment for every entry, but they should at least not have commas imo.

src/code/z_actor.c
```diff
@@ -3484,13 +3459,12 @@ typedef struct {
...
-    { 0.644f, 12000.0f, 8000.0f, 1.0f, 0.0f, 0x06001530, 0x06001400 },
-    { 0.64000005f, 8500.0f, 8000.0f, 1.75f, 0.1f, 0x050011F0, 0x05001100 },
-};
-
 void Actor_DrawDoorLock(GlobalContext* globalCtx, s32 frame, s32 type) {
+    static struct_801160DC D_801160DC[] = {
+        { 0.54f, 6000.0f, 5000.0f, 1.0f, 0.0f, 0x050011F0, 0x05001100 },
+        { 0.644f, 12000.0f, 8000.0f, 1.0f, 0.0f, 0x06001530, 0x06001400 },
+        { 0.64000005f, 8500.0f, 8000.0f, 1.75f, 0.1f, 0x050011F0, 0x05001100 },
+    };
```
Roman971, line 3467: Is there a reason to move all these variables inside functions? Imo we should avoid doing that unless it's needed or really makes sense (a lot of the things you moved make more sense outside to me)

src/code/z_actor.c
```diff
@@ -1771,166 +1766,149 @@ void func_8002FA60(GlobalContext* globalCtx) {
...
+                curPos->z = nextPos->z + (dist.z * factor);
+                length *= 0.5f;
+                dx = diff - length;
+                yOffset += sqrtf(SQ(length) - SQ(dx)) * 0.2f;
+                osSyncPrintf("-------- DISPLAY Y=%f\n", yOffset);
+            }
+
+            effectPos.x = curPos->x + Rand_CenteredFloat(6.0F);
+            effectPos.y = curPos->y + 80.0f + (6.0f * Rand_ZeroOne());
+            effectPos.z = curPos->z + Rand_CenteredFloat(6.0F);
```
Roman971, line 1826:
```suggestion
            effectPos.x = curPos->x + Rand_CenteredFloat(6.0f);
            effectPos.y = curPos->y + 80.0f + (6.0f * Rand_ZeroOne());
            effectPos.z = curPos->z + Rand_CenteredFloat(6.0f);
```

src/code/z_actor.c
```diff
@@ -1771,166 +1766,149 @@ void func_8002FA60(GlobalContext* globalCtx) {
...
-                          G_MTX_MODELVIEW | G_MTX_LOAD);
-                gSPDisplayList(POLY_XLU_DISP++, &gEffFlash1DL);
+            gDPPipeSync(POLY_XLU_DISP++);
+            gDPSetPrimColor(POLY_XLU_DISP++, 128, 128, 255, 255, 200, alpha);
+            gDPSetEnvColor(POLY_XLU_DISP++, 100, 200, 0, 255);
 
-                Matrix_Pop();
-                phi_f6 = ~((globalCtx->gameplayFrames * 1200) & 0xFFFF);
-                Matrix_RotateZ((phi_f6 * M_PI) / 32768.0f, MTXMODE_APPLY);
+            Matrix_RotateZ((f32)((globalCtx->gameplayFrames * 1500) & 0xFFFF) * M_PI / 32768.0f, MTXMODE_APPLY);
```
Roman971, line 1892: Is the f32 cast needed here? I feel like it should happen implicitly with the multiplication to M_PI

src/code/z_actor.c
```diff
@@ -1771,166 +1766,149 @@ void func_8002FA60(GlobalContext* globalCtx) {
...
-                gSPDisplayList(POLY_XLU_DISP++, &gEffFlash1DL);
-            }
-            {
-                Vec3f lightPos;
-
-                lightPos.x = gSaveContext.respawn[RESPAWN_MODE_TOP].pos.x;
-                lightPos.y = gSaveContext.respawn[RESPAWN_MODE_TOP].pos.y + spD8;
-                lightPos.z = gSaveContext.respawn[RESPAWN_MODE_TOP].pos.z;
+            Matrix_Pop();
+            Matrix_RotateZ((f32)(~((globalCtx->gameplayFrames * 1200) & 0xFFFF)) * M_PI / 32768.0f, MTXMODE_APPLY);
```
Roman971, line 1898: Same thing here with the f32 cast, also I think there is an extra set of parens which isn't necessary

src/code/z_actor.c
```diff
@@ -96,114 +96,104 @@ void ActorShadow_DrawFoot(GlobalContext* globalCtx, Light* light, MtxF* arg2, s3
...
+        s32 j;
+        s32 numLights = lights->numLights - 2;
+        Light* firstLightPtr = &lights->l.l[0];
+        Vec3f* feetPosPtr = actor->shape.feetPos;
+        f32* floorHeightPtr = floorHeight;
 
         OPEN_DISPS(globalCtx->state.gfxCtx, "../z_actor.c", 1741);
 
-        POLY_OPA_DISP = Gfx_CallSetupDL(POLY_OPA_DISP, 0x2C);
+        POLY_OPA_DISP = Gfx_CallSetupDL(POLY_OPA_DISP, 44);
```
Roman971, line 135: The setup index passed to this function should remain as hex imo, since that's what is used inside the system

src/code/z_actor.c
```diff
@@ -96,114 +96,104 @@ void ActorShadow_DrawFoot(GlobalContext* globalCtx, Light* light, MtxF* arg2, s3
...
+            floorHeightPtr++;
         }
 
         if (!(actor->bgCheckFlags & 1)) {
             actor->shape.feetFloorFlags = 0;
         } else if (actor->shape.feetFloorFlags == 3) {
-            temp_f0 = actor->shape.feetPos[FOOT_LEFT].y - actor->shape.feetPos[FOOT_RIGHT].y;
-            actor->shape.feetFloorFlags = ((spE0[0] + temp_f0) < (spE0[1] - temp_f0)) ? 2 : 1;
+            f32 footDistY = actor->shape.feetPos[FOOT_LEFT].y - actor->shape.feetPos[FOOT_RIGHT].y;
+            if ((floorHeight[0] + footDistY) < (floorHeight[1] - footDistY)) {
```
Roman971, line 192:
```suggestion
            f32 footDistY = actor->shape.feetPos[FOOT_LEFT].y - actor->shape.feetPos[FOOT_RIGHT].y;

            if ((floorHeight[0] + footDistY) < (floorHeight[1] - footDistY)) {
```

src/code/z_actor.c
```diff
@@ -1771,166 +1766,149 @@ void func_8002FA60(GlobalContext* globalCtx) {
...
+        if ((globalCtx->csCtx.state == CS_STATE_IDLE) &&
+            (((void)0, gSaveContext.respawn[RESPAWN_MODE_TOP].entranceIndex) ==
+             ((void)0, gSaveContext.entranceIndex)) &&
+            (((void)0, gSaveContext.respawn[RESPAWN_MODE_TOP].roomIndex) == globalCtx->roomCtx.curRoom.num)) {
+            f32 scale = 0.025f * ratio;
 
-                gDPPipeSync(POLY_XLU_DISP++);
-                gDPSetPrimColor(POLY_XLU_DISP++, 0x80, 0x80, 255, 255, 200, spD0);
-                gDPSetEnvColor(POLY_XLU_DISP++, 100, 200, 0, 255);
+            POLY_XLU_DISP = Gfx_CallSetupDL(POLY_XLU_DISP, 25);
```
Roman971, line 1879: Same thing here about the setup index

src/code/z_actor.c
fig02: yea I already do this in many other places. i dont think specifying a color is necessary when context makes it very clear, and the type

src/code/z_actor.c
fig02: the comments are helpful, I had to add them myself when trying to modify colors for a separate project. agreed on no comma tho

src/code/z_actor.c
fig02: agreed, just cause something is only used in one function doesnt always mean it makes sense as in function static imo

mzxrules:
You should be able to eliminate the NON-MATCHING by hoisting `CollisionPoly* sCurCeilingPoly; s32 sCurCeilingBgId;` to the top of the file;

src/code/z_actor.c
Zelllll: hm why avoid it when it seems to clearly be correct? I've been doing this on every file I work on thus far

src/code/z_actor.c
Zelllll: Ok I can remove the one for max, I thought it was a bit weird too honestly lol

src/code/z_actor.c
Zelllll: I felt a variable named "white" just seemed slightly weird lol, bit I can change it back or do whiteColor

src/code/z_actor.c
Zelllll: Oh might not be, let me see, I think I did just to be safe while matching it or something

src/code/z_actor.c
Zelllll: I thought it was just an index? Can you explain why it should be hex?

src/code/z_actor.c
louist103: Adding on to this not having as much static might make it easier to find the bss issue.

Zelllll:
> You should be able to eliminate the NON-MATCHING by hoisting `CollisionPoly* sCurCeilingPoly; s32 sCurCeilingBgId;` to the top of the file

that worked! thanks!

src/code/z_actor.c
Zelllll: bss is already fixed, and everything in here should be static at the very least unless used globally

src/code/z_actor.c
```diff
@@ -95,123 +98,109 @@ void ActorShadow_DrawFoot(GlobalContext* globalCtx, Light* light, MtxF* arg2, s3
...
+
+    if (distToFloor > 20.0f) {
+        f32 shadowScale = actor->shape.shadowScale;
+        u8 shadowAlpha = actor->shape.shadowAlpha;
+        f32 alphaRatio;
+
         actor->shape.shadowScale *= 0.3f;
-        actor->shape.shadowAlpha *= ((temp_f20 - 20.0f) * 0.02f) > 1.0f ? 1.0f : ((temp_f20 - 20.0f) * 0.02f);
+        alphaRatio = (distToFloor - 20.0f) * 0.02f;
+        actor->shape.shadowAlpha = (f32)actor->shape.shadowAlpha * ((alphaRatio > 1.0f) ? 1.0f : alphaRatio);
```
AngheloAlf, line 111: I guess a CLAMP doesn't work, right?

src/code/z_actor.c
AngheloAlf: Do we even know what is this array for?
I think we should not move those variables into of functions until we figure out their meaning.

src/code/z_actor.c
Zelllll: yes it does, just added it

src/code/z_actor.c
Zelllll: now we do :)

Zelllll:
can we merge this or does anything else need to change..?

louist103 (approved):

PR #833, opened 2021-05-29
En_River_Sound OK

Zelllll:


fig02 (approved):

src/overlays/actors/ovl_En_River_Sound/z_en_river_sound.c
```diff
@@ -57,45 +57,36 @@ void EnRiverSound_Destroy(Actor* thisx, GlobalContext* globalCtx) {
...
+
+    vec[2].x = vec[1].x - vec[0].x;
+    vec[2].y = vec[1].y - vec[0].y;
+    vec[2].z = vec[1].z - vec[0].z;
+
+    temp = ((vec[2].x * vec[0].x) + (vec[2].y * vec[0].y)) + (vec[2].z * vec[0].z);
+
+    if (((((vec[2].x * vec[1].x) + (vec[2].y * vec[1].y)) + (vec[2].z * vec[1].z)) * temp) < 0.0f) {
+        temp = -temp / (SQ(vec[2].x) + SQ(vec[2].y) + SQ(vec[2].z));
+
```
fig02, line 80: glank noted in discord that you can use the DOTXYZ macro for these

Zelllll:
fixed the thing fig mentioned and added some docs to this actor

src/overlays/actors/ovl_En_River_Sound/z_en_river_sound.h
```diff
@@ -8,11 +8,38 @@ struct EnRiverSound;
...
+    /* 14 */ RS_SANDSTORM,
+    /* 15 */ RS_LAKESIDE_LAB_TANK,
+    /* 16 */ RS_CHAMBER_OF_SAGES_1,
+    /* 17 */ RS_CHAMBER_OF_SAGES_2,
+    /* 18 */ RS_RUMBLING,
+    /* 19 */ RS_UNK_19,
+    /* 20 */ RS_TORCH_CRACKLING,
+    /* 21 */ RS_COW_MOOING,
+    /* 247 */ RS_UNK_247 = 247,
+    /* 248 */ RS_MAX
```
Roman971, line 40: It's not a big deal but I think it would be much better if these were hex offsets in the comments and names. We should almost always use hex for this imo, but it's especially important here since they are supposed to fit in 1 byte (see how params are used in the init function)

Zelllll:
fixed all of roman's comments on this one

Roman971 (approved):

PR #834, opened 2021-05-30
z_sram OK

Zelllll:
I matched the last function in this file and made some data in this file `static`. I also had to make `gSramSlotOffsets` no longer be a 2D array because for some reason it did not match otherwise, however I added a comment about the last three saves being for the backups. Luckily the dev strings do show that this wasn't a 2D array anyways, so we aren't losing much.

As for the function itself, I had to do some pretty dumb stuff with the loops, and this function reuses variables in a pretty weird way (and in an even weirder way according to the printf's, like making `j` be the name of the checksum for some reason??), but I tried my best to make the variable names make sense. The loops are also pretty weird which appears to be the main reason the function didn't match before, but they seem to have to be this way in order to match. Also `slotNum == 0` had to be made `!slotNum` in order to match for some reason, I'm guessing IDO treats `== 0` a bit differently?


louist103 (approved):
Nice job.  Looks good to me.

src/code/z_sram.c
```diff
@@ -505,77 +502,69 @@ void Sram_WriteSave(SramContext* sramCtx) {
...
                 // backup save didnt work, make new save
-                // offPtr = &gSramSlotOffsets[1][slotNum];
-                osSyncPrintf("ＥＲＲＯＲ！！！ ＝ %x(%d+3)\n", offset, slotNum);
+                osSyncPrintf("ＥＲＲＯＲ！！！ ＝ %x(%d+3)\n", gSramSlotOffsets[slotNum + 3], slotNum);
                 bzero(&gSaveContext.entranceIndex, sizeof(gSaveContext.entranceIndex));
                 bzero(&gSaveContext.linkAge, sizeof(gSaveContext.linkAge));
                 bzero(&gSaveContext.cutsceneIndex, sizeof(gSaveContext.cutsceneIndex));
-                bzero(&gSaveContext.dayTime, sizeof(gSaveContext.dayTime));
+                //! @bug: The size here was probably not intended to be multiplied by 2
+                bzero(&gSaveContext.dayTime, sizeof(gSaveContext.dayTime) * 2);
```
Roman971, line 562: I think all these should be `sizeof(s32)` since this indicates they clearly didn't use sizeof on the actual variable. It's not really a bug either imo, since it just zeroes out words in the struct and alignment ensures it won't overwrite something unintended.

Zelllll:
fixed roman's comments on this one

Roman971 (approved):

fig02 (approved):

PR #843, opened 2021-06-02
Fix lots of undefined syms

Zelllll:
this is by no means all of them, but figured i should pr my progress now since some of these fixes are high priority


src/overlays/actors/ovl_En_Ma3/z_en_ma3.c
```diff
@@ -79,9 +82,6 @@ static UNK_PTR D_80AA38B0[] = {
     0x06003070,
 };
 
-extern u32 D_06005420;
-extern FlexSkeletonHeader D_06008D90;
-extern AnimationHeader D_060093BC;
 extern AnimationHeader D_06009EE0;
```
EllipticEllipsis, line 85: Missed this one? Looks like there are a couple of animations in the blob in `object_ma2`

src/overlays/actors/ovl_En_Ma3/z_en_ma3.c
```diff
@@ -352,8 +352,8 @@ void EnMa3_PostLimbDraw(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Ve
     if (limbIndex == 18) {
         Matrix_MultVec3f(&vec, &this->actor.focus.pos);
     }
-    if ((limbIndex == 14) && (this->skelAnime.animation == &D_060093BC)) {
-        gSPDisplayList(POLY_OPA_DISP++, &D_06005420);
+    if ((limbIndex == 14) && (this->skelAnime.animation == &object_ma2_Anim_0093BC)) {
+        gSPDisplayList(POLY_OPA_DISP++, &object_ma2_DL_005420);
```
EllipticEllipsis, line 356:
```suggestion
        gSPDisplayList(POLY_OPA_DISP++, object_ma2_DL_005420);
```

EllipticEllipsis (commented):
Mostly looks good so far, only spotted a couple of issues.

src/overlays/actors/ovl_En_Ma3/z_en_ma3.c
Zelllll: yeah they're in a blob, I haven't messed with any xmls yet

src/overlays/actors/ovl_En_Ma2/z_en_ma2.c
```diff
@@ -75,9 +78,6 @@ static UNK_PTR D_80AA28C0[] = {
     0x06003070,
 };
```
EllipticEllipsis, line 79: More segment 6 symbols in here

src/overlays/actors/ovl_En_Ma3/z_en_ma3.c
EllipticEllipsis: There's some segment 6 stuff here too

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -76,17 +77,6 @@ UNK_TYPE D_80B16E88[] = {
     0x060072C0,
```
EllipticEllipsis, line 77: And here

src/overlays/actors/ovl_En_Toryo/z_en_toryo.c
```diff
@@ -99,8 +100,6 @@ static EnToryoAnimation sEnToryoAnimation = { 0x06000E50, 1.0f, 0, 0 };
```
EllipticEllipsis, line 100: And here. There are probably others in some of these files I didn't see in the preview.

src/overlays/actors/ovl_En_Ma3/z_en_ma3.c
Zelllll: this pr wasn't really intended to actually fix symbols, necessarily, it was just supposed to add the auto gen'd object symbols

src/overlays/actors/ovl_Demo_Im/z_demo_im.c
```diff
@@ -107,7 +107,7 @@ extern AnimationHeader D_06000AFC;
 extern Gfx D_0600EDE8[];
 extern AnimationHeader D_060014E4;
 extern AnimationHeader D_06001868;
-extern FlexSkeletonHeader D_0600F788;
+extern FlexSkeletonHeader object_im_Skel_00F788;
```
AngheloAlf, line 110: Shouldn't you be removing this and adding the header?
Also, what about the other symbols?

AngheloAlf (commented):
The replaced symbols feels a bit inconsistent, a bit random because of the actors that you fixed but still have undefined syms on them. I guess it is because the XMLs for those symbols are not in the repo yet.
For future fixes like this, I suggest either choosing actors that have every XML they need already on the repo, or at least list the ones that don't have an XML yet in your PR's post.

assets/xml/objects/object_link_child.xml
```diff
@@ -106,7 +106,7 @@
         <DList Name="gLinkChildBottle2DL" Offset="0x18648"/>
         <DList Name="gLinkChildSlinghotStringDL" Offset="0x221A8"/>
         <DList Name="gLinkChildDekuShieldDL" Offset="0x224F8"/>
-        <Mtx Name="gLinkChildDekuShieldMtx"  Offset="0x22648"/>
+        <!--<Mtx Name="gLinkChildDekuShieldMtx"  Offset="0x22648"/>-->
```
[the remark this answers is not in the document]
Zelllll: that was for something else and wasn't intended to be in here

fig02:
From the start this PR seemed to be a bit half baked. Perhaps we can close this and reopen at a later time when the changes are a bit less sporadic?

PR #847, opened 2021-06-09
Mark Stale Reference Manipulation bug in z_actor.c

louist103:


src/code/z_actor.c
```diff
@@ -2469,13 +2469,17 @@ void Actor_FreezeAllEnemies(GlobalContext* globalCtx, ActorContext* actorCtx, s3
...
 
-void func_80031B14(GlobalContext* globalCtx, ActorContext* actorCtx) {
+void Actor_ChangeRoom(GlobalContext* globalCtx, ActorContext* actorCtx) {
     Actor* actor;
     s32 i;
 
     for (i = 0; i < ARRAY_COUNT(actorCtx->actorLists); i++) {
         actor = actorCtx->actorLists[i].head;
         while (actor != NULL) {
+            //! @bug Stale Reference Manipulation.  Under very specific circumstances an actor can be culled and held by
```
Zelllll, line 2479: I've never heard anyone besides a speedrunner use the term "stale reference manipulation," so I'd suggest using something more technical or just explaining that the pointer is not unset under some circumstances and thus can refer to invalid data.

src/code/z_actor.c
```diff
@@ -2469,13 +2469,17 @@ void Actor_FreezeAllEnemies(GlobalContext* globalCtx, ActorContext* actorCtx, s3
...
+            //! the player.  Because of this its room is never set to -1 causing it to be deallocated during a room
```
Zelllll, line 2480: I don't think we double space after punctuation anywhere else in the repo, should I wouldn't do that in this comment

src/code/z_actor.c
pixel-stuck: technical term here is "use after free." I think it could have both, and for documentation purposes, the community probably knows this term better, and will search for it first.

Zelllll: I think that really only applies to speedrunners which is probably a section of people that will look at this, but definitely not everyone. I had never heard of this term until speedrunners invented it so not really sure what is best here. Use after free sounds much better though.

fig02:
this glitch has many moving parts, and this is only one part of a much larger puzzle. Id feel more comfortable if we wait till later to accurately describe everything that allows this glitch to happen, perhaps after actor.c is fully documented

fig02:
And imo this description is not fully accurate. The issue is more so about the Actor_Kill process being skipped, which allows the actors update function to remain non-null in its instance. This is what keeps link from dropping it. Its not really about the room number thing, although it is a factor for getting the thing to unload in general. Getting the room number to not change to -1 is achieved by just delaying the pickup usually

src/code/z_actor.c
fig02: fwiw I think its fine to at least mention the name given to it by speedrunners. doesnt have to be right at the start, but making it easily searchable would be nice. 

Itd be like if we label the RBA glitch as reverse bottle adventure. Name is kinda silly yes, but we're the ones who found these things and can name them whatever we like :p 
Ofc we can include the technical term for what it actually is, like how RBA is really a buffer overflow. But my point is that the name for a glitch != the technical term for the actual bug

fig02:
going to close this for now and we can come back to this another time

PR #850, opened 2021-06-11
BgGndFiremeiro OK and documented

EllipticEllipsis:
Ever get that sinking feeling? Putting on hover boots is not enough: you need to get off the platform too.

Theoretically simple actor, but surprisingly memetic. I have tried to minimise the memes, but some remain. This actor's so simple that I wonder if some of them are suggestive of other things, for example, the match in `Destroy`.


src/overlays/actors/ovl_Bg_Gnd_Firemeiro/z_bg_gnd_firemeiro.c
```diff
@@ -35,17 +31,122 @@ const ActorInit Bg_Gnd_Firemeiro_InitVars = {
...
+
+void BgGndFiremeiro_Shake(BgGndFiremeiro* this, GlobalContext* globalCtx) {
+    s32 pad;
+    f32 randSign;
+
+    if (func_8004356C(&this->dyna)) { // Player standing on it
+        if (this->timer > 0) {
+            this->timer--;
+
+            randSign = ((this->timer & 1) ? 2.0f : -2.0f);
```
louist103, line 94:
```suggestion
            randSign = (this->timer & 1) ? 2.0f : -2.0f;
```

louist103 (changes requested):
Just one thing then its good IMO.

src/overlays/actors/ovl_Bg_Gnd_Firemeiro/z_bg_gnd_firemeiro.c
EllipticEllipsis: I like my ternaries in brackets for clarity's sake. Can change it if it's contrary to our style.

louist103: I dont think we have a set style but IMO it looks nicer without them

Roman971 (approved):

fig02 (approved):

PR #851, opened 2021-06-12
Decompile `object_link_boy`, fix issues in `object_link_child`, and document `z_player_lib`

Zelllll:


louist103 (changes requested):
Great job on this.  Just some things.

assets/xml/objects/object_link_boy.xml
```diff
@@ -0,0 +1,229 @@
+<Root>
+	<File Name="object_link_boy" Segment="6">
+		<!--Eyes-->
+        <Texture Name="gLinkAdultEyesOpenTex" OutName="link_adult_eyes_open" Format="ci8" TlutOffset="0x5C00" Width="64" Height="32" Offset="0x0"/>
+		<Texture Name="gLinkAdultEyesHalfTex" OutName="link_adult_eyes_half" Format="ci8" TlutOffset="0x5C00" Width="64" Height="32" Offset="0x800"/>
```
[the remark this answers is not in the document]
Zelllll: I think you did yeah, was to be consistent, I can easily just remove it from both though if we want that, seems like a fine change since duplicate names won't matter because they are in a different directory anyways

src/code/z_player_lib.c
```diff
@@ -44,197 +141,249 @@ u8 gPlayerModelTypes[][5] = {
     { 0, 0, 13, 18, 20 }, { 0, 0, 14, 18, 20 }, { 0, 7, 8, 18, 20 },  { 0, 2, 8, 19, 20 },
 };
 
-Gfx* D_80125CE8[] = {
-    0x060226E0, gLinkChildRightHandClosedNearDL,        0x06027690, gLinkChildRightHandClosedFarDL,
-    0x060226E0, gLinkChildRightFistAndDekuShieldNearDL, 0x06027690, gLinkChildRightFistAndDekuShieldFarDL,
-    0x06022970, gLinkChildRightHandClosedNearDL,        0x06027918, gLinkChildRightHandClosedFarDL,
-    0x060241C0, gLinkChildRightHandClosedNearDL,        0x06028B40, gLinkChildRightHandClosedFarDL,
+Gfx* gPlayerRightHandAndShieldDLists[] = {
```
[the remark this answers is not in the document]
Zelllll: They can be but I left them not static since some of them are actually global and I didn't like the inconsistency

src/code/z_player_lib.c
```diff
@@ -613,47 +682,38 @@ s32 func_8008F2F8(GlobalContext* globalCtx) {
...
+    gLinkAdultMouthSmileTex,
 };
 
-void func_8008F470(GlobalContext* globalCtx, void** skeleton, Vec3s* jointTable, s32 dListCount, s32 lod, s32 tunic,
-                   s32 boots, s32 face, OverrideLimbDrawOpa overrideLimbDraw, PostLimbDrawOpa postLimbDraw,
-                   void* data) {
+void Player_DrawImpl(GlobalContext* globalCtx, void** skeleton, Vec3s* jointTable, s32 dListCount, s32 lod, s32 tunic,
+                     s32 boots, s32 face, OverrideLimbDrawOpa overrideLimbDraw, PostLimbDrawOpa postLimbDraw,
+                     void* data) {
+    static Color_RGB8 tunicColors[] = {
```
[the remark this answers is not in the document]
Zelllll: Nah these are normal variables technically, so if they are inlined in a function they shouldn't be labelled like that

assets/xml/objects/object_link_boy.xml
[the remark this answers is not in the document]
louist103: I'd say remove it then.  Since that is the standard.

src/code/z_player_lib.c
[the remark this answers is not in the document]
louist103: I'd say still name them as if they were static if they can be.  I feel that `g` implies it is global not just not static.

src/code/z_player_lib.c
```diff
@@ -1070,138 +1142,99 @@ void Player_DrawGetItemImpl(GlobalContext* globalCtx, Player* this, Vec3f* refPo
 }
 
 void Player_DrawGetItem(GlobalContext* globalCtx, Player* this) {
-    if (!this->giObjectLoading || !osRecvMesg(&this->giObjectLoadQueue, NULL, OS_MESG_NOBLOCK)) {
+    if (!(this->giObjectLoading) || !(osRecvMesg(&this->giObjectLoadQueue, NULL, OS_MESG_NOBLOCK))) {
```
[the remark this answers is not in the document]
Zelllll: nope isn't OK like that

src/code/z_player_lib.c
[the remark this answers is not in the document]
Zelllll: I think its a safe assumption that they are intended to be global but idk

louist103 (approved):

Zelllll:
ok everything that louist brought up should be fixed

src/code/z_player_lib.c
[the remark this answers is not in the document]
fig02: unrelated to the suggestion, alot of these parens added around bool checks are unnecessary. they should only be needed if there are multiple parts, ie
`! (something & something)`
if its just one variable or a single function call, the parens do nothing but add noise

Zelllll:
i removed the uneeded parentheses like fig suggested

Dragorn421:
https://github.com/Zelllll/oot/pull/3

src/code/z_player_lib.c
```diff
@@ -374,7 +443,7 @@ void Player_SetModelGroup(Player* this, s32 modelGroup) {
     Player_SetModels(this, modelGroup);
 }
 
-void func_8008EC70(Player* this) {
+void Player_UnsetItemAction(Player* this) {
     this->itemActionParam = this->heldItemActionParam;
     Player_SetModelGroup(this, Player_ActionToModelGroup(this, this->heldItemActionParam));
     this->unk_6AD = 0;
```
fig02, line 449: This function doesnt really have to do with unsetting anything.
Basic testing and a little bit of reading shows its related specifically to put an old item back in links hand after using a custscene item (ocarina, magic spell, etc). But there are probably more contexts where its used so I wouldnt feel comfortable naming this yet

I am planning a docs pass of the item system/action parameters specifically, so you can leave this unnamed and Ill get to it later with a deeper understanding of the system

src/code/z_player_lib.c
```diff
@@ -501,12 +570,12 @@ s32 Player_ActionToMagicSpell(Player* this, s32 actionParam) {
...
-s32 Player_HoldsHookshot(Player* this) {
+s32 Player_IsHoldingHookshot(Player* this) {
     return (this->heldItemActionParam == PLAYER_AP_HOOKSHOT) || (this->heldItemActionParam == PLAYER_AP_LONGSHOT);
 }
 
-s32 func_8008F128(Player* this) {
-    return Player_HoldsHookshot(this) && (this->heldActor == NULL);
+s32 Player_IsShootingHookshot(Player* this) {
+    return Player_IsHoldingHookshot(this) && (this->heldActor == NULL);
 }
```
fig02, line 579: links heldActor pointer is set to the hookshot even when hes aiming it or just holding it in his hand. So this name is not accurate.
We will need to distinguish the difference between this and only have the held item action parameter set to the hookshot. 
Again, I recommend leaving unnamed and ill get to it in my items documentation

src/code/z_player_lib.c
```diff
@@ -710,35 +783,29 @@ void func_8008F470(GlobalContext* globalCtx, void** skeleton, Vec3s* jointTable,
...
-f32 D_80126058[] = { SQ(13.04f), SQ(6.95f) };
-f32 D_80126060[] = { 10.019104f, -19.925102f };
-f32 D_80126068[] = { 5.0f, 3.0f };
+static Vec3f sZeroVec = { 0.0f, 0.0f, 0.0f };
 
-Vec3f D_80126070 = { 0.0f, -300.0f, 0.0f };
-
-void func_8008F87C(GlobalContext* globalCtx, Player* this, SkelAnime* skelAnime, Vec3f* pos, Vec3s* rot,
+void Player_FeetIK(GlobalContext* globalCtx, Player* this, SkelAnime* skelAnime, Vec3f* pos, Vec3s* rot,
                    s32 thighLimbIndex, s32 shinLimbIndex, s32 footLimbIndex) {
```
fig02, line 789: what is IK? I see this term used alot in modding contexts, but since i dont really do modeling stuff I dont know

src/code/z_player_lib.c
```diff
@@ -1076,132 +1148,93 @@ void Player_DrawGetItem(GlobalContext* globalCtx, Player* this) {
     }
 }
 
-void func_80090A28(Player* this, Vec3f* vecs) {
-    D_8012608C.x = D_80126080.x;
+void Player_GetSwordTip(Player* this, Vec3f* swordTip) {
+    sSwordTipPos2.x = sSwordTipPos1.x;
```
fig02, line 1152: this is also used in the context of sticks as well. I am 100% ok with grouping sticks in with swords, cause it is basically a reskinned sword. Just wanted to make sure we are all on board with that

src/code/z_player_lib.c
```diff
@@ -1076,132 +1148,93 @@ void Player_DrawGetItem(GlobalContext* globalCtx, Player* this) {
...
+    Matrix_MultVec3f(&sSwordTipPos1, &swordTip[0]);
+    Matrix_MultVec3f(&sSwordTipPos2, &swordTip[1]);
+    Matrix_MultVec3f(&sSwordTipPos3, &swordTip[2]);
 }
 
-void func_80090AFC(GlobalContext* globalCtx, Player* this, f32 arg2) {
-    static Vec3f D_801260C8 = { -500.0f, -100.0f, 0.0f };
-    CollisionPoly* sp9C;
+void Player_DrawHookshotPointer(GlobalContext* globalCtx, Player* this, f32 arg2) {
+    static Vec3f posOffset = { -500.0f, -100.0f, 0.0f };
```
fig02, line 1168: the most common term to refer to this is a "reticle", and should be used imo. Also will remove any confusion a programmer may have with pointer lol
I think the DL name should be changed to that as well

src/code/z_player_lib.c
```diff
@@ -1076,132 +1148,93 @@ void Player_DrawGetItem(GlobalContext* globalCtx, Player* this) {
...
-
-Vec3f D_801261CC = { 630.0f, 100.0f, -30.0f };
-Vec3s D_801261D8 = { 0, 0, 0x7FFF };
-
-Vec3f D_801261E0[] = {
-    { 200.0f, 300.0f, 0.0f },
-    { 200.0f, 200.0f, 0.0f },
-};
+typedef struct {
+    /* 0x00 */ void* dList;
```
fig02, line 1208: dlist can be typed as Gfx*. It was probably void before out of a precautionary measure, but we know for sure it cant be anything else

src/code/z_player_lib.c
```diff
@@ -1540,8 +1615,11 @@ void func_80091A24(GlobalContext* globalCtx, void* seg04, void* seg06, SkelAnime
     CLOSE_DISPS(globalCtx->state.gfxCtx, "../z_player_lib.c", 3288);
 }
 
-void func_8009214C(GlobalContext* globalCtx, u8* segment, SkelAnime* skelAnime, Vec3f* pos, Vec3s* rot, f32 scale,
-                   s32 sword, s32 tunic, s32 shield, s32 boots) {
+/**
+ * Draws the player in the equip screen on the pause menu.
+ */
+void Player_PauseDraw(GlobalContext* globalCtx, u8* segment, SkelAnime* skelAnime, Vec3f* pos, Vec3s* rot, f32 scale,
```
fig02, line 1621: think adding the word `Menu` to the name will make it read a bit nice, `Player_PauseMenuDraw`

src/code/z_player_lib.c
```diff
@@ -1,35 +1,132 @@
...
+            currentBoots = PLAYER_BOOTS_EX_IRON_UNDERWATER;
+        }
+        REG(27) = 500;
+        REG(48) = 100;
+    }
+
+    entry = &sMovementInfo[currentBoots];
+    REG(19) = entry->walkSpeedMax;
+    REG(30) = entry->sideWalkAnimSpeedBase;
+    REG(32) = entry->sideWalkAnimSpeed;
```
fig02, line 67: what does base mean in this context? how does it differ from the normal speed?

fig02 (commented):
apologies for the late review on this. As we talked about already, we were prioritizing other decomp prs first

A few things to discuss here

src/code/z_player_lib.c
fig02: (this is bow string info, i forgot to include that context in the comment)

src/code/z_player_lib.c
Dragorn421: https://en.wikipedia.org/wiki/Inverse_kinematics

ELI5: find rotations that look nice for bones between feet and body given only the position of the feet and the position of the body

fig02: i feel like theres probably a name we can use that would be easier to understand without needing that context, but i dont have any suggestions for now

fig02: A brief description about what IK is and what this is doing to achieve that would be an option as well.
I just want to make sure this is actually IK. I cant follow any of that math thats going on, did you verify that what its achieving is accurately described as IK?

src/code/z_player_lib.c
Dragorn421: I'm very not familiar with the player code but doesn't it also get used with hammer (and maybe fishing rod?)

fig02: it will be used by anything that sets sword state to non zero.
so fishing rod no, hammer yes

if we want to address that term and make it more broad, `swordState` and related vars will need to be addressed as well lol

fig02: but as far as players internal systems are concerned, hammer is just a shorter stubbier sword with slightly different properties. so i personally dont have an issue with it lol. but can understand if others do

Dragorn421: what about `meleeWeapon` (I was looking for a synonym of "close combat weapon" and wikipedia saved the day https://en.wikipedia.org/wiki/Melee_weapon )

Zelllll: not a fan of the vagueness that comes from that name, sword is fine imo

src/code/z_player_lib.c
Zelllll: I did this like 4-5 months ago so don't remember tbh, this stuff was difficult to name but I think the "base" speed was kind of like the minimum speed which they gets multiplied

fig02: it might be worth to revisit and confirm your reasoning, and potentially rename if it seems fitting. base doesnt really make sense to me at first glance without reading some more code 
i understand it being hard to name tho yeah, it seems like it

src/code/z_player_lib.c
Zelllll: yeah I called it IK because it handles moving links feet to be in the right spot on stairs etc

src/code/z_player_lib.c
Zelllll: I would say this falls in line with the "player boots," which are not actually always boots (link has different "boots" if he's inside so he walks slower), but I was still told to call it boots, so sword here should be fine too I think since that's the main purpose of it, but it gets reused for other things

src/code/z_player_lib.c
fig02: i guess one thing that could help, would it make more sense to say this updating the feet IK? instead of just *being* IK
the term `FeetIK` on its own doesnt really help tell me what its achieving, even after learning what IK is

src/code/z_player_lib.c
Dragorn421: Well yeah the point is to be more vague so it is also accurate for non-swords melee weapons like stick and hammer

But it is also precise enough (and actually not vague) because swords, stick and hammer are all the melee weapon available to Link.

src/code/z_player_lib.c
```diff
@@ -1304,35 +1351,45 @@ void func_80090D20(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3s*
...
             Matrix_Get(&this->shieldMf);
-            func_80090604(globalCtx, this, &this->shieldQuad, D_80126154);
+            Player_SetShieldCollision(globalCtx, this, &this->shieldQuad, shieldQuadVertices);
         }
 
         if (this->actor.scale.y >= 0.0f) {
             if ((this->heldItemActionParam == PLAYER_AP_HOOKSHOT) ||
                 (this->heldItemActionParam == PLAYER_AP_LONGSHOT)) {
-                Matrix_MultVec3f(&D_80126184, &this->unk_3C8);
+                static Vec3f hookshotHandPos = { 100.0f, 1500.0f, 0.0f };
```
Dragorn421, line 1374: this name `hookshotHandPos` doesn't really provide any information, it's obvious this is hand (`PLAYER_LIMB_R_HAND`) and hookshot (`PLAYER_AP_HOOKSHOT`) related

I would revert to `D_80126184` until more documentation (eg on `unk_3C8`)

src/code/z_player_lib.c
```diff
@@ -1076,132 +1148,93 @@ void Player_DrawGetItem(GlobalContext* globalCtx, Player* this) {
...
+    Matrix_MultVec3f(&sSwordTipPos1, &swordTip[0]);
+    Matrix_MultVec3f(&sSwordTipPos2, &swordTip[1]);
+    Matrix_MultVec3f(&sSwordTipPos3, &swordTip[2]);
 }
 
-void func_80090AFC(GlobalContext* globalCtx, Player* this, f32 arg2) {
-    static Vec3f D_801260C8 = { -500.0f, -100.0f, 0.0f };
-    CollisionPoly* sp9C;
+void Player_DrawHookshotReticle(GlobalContext* globalCtx, Player* this, f32 arg2) {
+    static Vec3f posOffset = { -500.0f, -100.0f, 0.0f };
```
Dragorn421, line 1168: why `posOffset`? looks more like `posModel` (model space, kinda) to me

src/code/z_player_lib.c
```diff
@@ -1076,132 +1148,93 @@ void Player_DrawGetItem(GlobalContext* globalCtx, Player* this) {
...
     s32 bgId;
-    Vec3f sp8C;
-    Vec3f sp80;
-    Vec3f sp74;
-    Vec3f sp68;
-    f32 sp64;
-    f32 sp60;
+    Vec3f posA;
+    Vec3f posB;
+    Vec3f aimProjectedPos; // position that the hookshot is aimed towards
```
Dragorn421, line 1173: I think "projected" gives the wrong idea (at first I thought it had to do with the projection matrix)

I think just `aimPos` is fine

src/code/z_player_lib.c
```diff
@@ -1215,45 +1248,55 @@ void func_80090D20(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3s*
...
                     hookedActor->shape.rot = hookedActor->world.rot;
                 } else if (this->stateFlags1 & 0x800) {
-                    Vec3s spB8;
-
-                    Matrix_Get(&sp14C);
-                    Matrix_MtxFToYXZRotS(&sp14C, &spB8, 0);
+                    Vec3s leftHandAngle;
 
+                    Matrix_Get(&matrix);
+                    Matrix_MtxFToYXZRotS(&matrix, &leftHandAngle, 0);
```
Dragorn421, line 1297:
```suggestion
                    Matrix_MtxFToYXZRotS(&matrix, &leftHandRotWorld, 0);
```

src/code/z_player_lib.c
```diff
@@ -1269,19 +1312,23 @@ void func_80090D20(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3s*
...
+            Matrix_Translate(bowStringInfoPtr->pos.x, bowStringInfoPtr->pos.y, bowStringInfoPtr->pos.z, MTXMODE_APPLY);
 
             if ((this->stateFlags1 & 0x200) && (this->unk_860 >= 0) && (this->unk_834 <= 10)) {
-                Vec3f sp90;
+                Vec3f bowPos;
                 f32 distXYZ;
 
-                Matrix_MultVec3f(&D_8012602C, &sp90);
-                distXYZ = Math_Vec3f_DistXYZ(D_80160000, &sp90);
+                Matrix_MultVec3f(&sZeroVec, &bowPos);
```
Dragorn421, line 1330:
```suggestion
                Matrix_MultVec3f(&sZeroVec, &bowPosWorld);
```

src/code/z_player_lib.c
```diff
@@ -810,20 +878,21 @@ void func_8008F87C(GlobalContext* globalCtx, Player* this, SkelAnime* skelAnime,
...
+                                void* thisx) {
     Player* this = (Player*)thisx;
 
     if (limbIndex == PLAYER_LIMB_ROOT) {
-        D_80160014 = this->leftHandType;
-        D_80160018 = this->rightHandType;
-        D_80160000 = &this->swordInfo[2].base;
+        sLeftHandType = this->leftHandType;
+        sRightHandType = this->rightHandType;
+        sPartsPosPtr = &this->swordInfo[2].base;
```
Dragorn421, line 895:
```suggestion
        sCurLimbPartPos = this->bodyPartsPos - 1;
```
with that name, I'm trying to reflect that `sCurLimbPartPos` is set in `OverrideLimbDraw` and used in `PostLimbDraw` when drawing each limb

Dragorn421 (changes requested):
If you want I can take care of this by PRing to your branch

Sorry for merging #897 before, unlike what you wished, but in case you didn't see I looked at what conflicted with this branch and the answer was "nothing but undefined_syms.txt" so it felt fine to go for it

src/code/z_player_lib.c
Dragorn421: I or someone else will open a separate PR for sword -> MeleeWeapon (after more discussion eventually) since such a change (that you aren't eager to do anyway 😅) would require touching more than this PR touches, and this PR is big enough already.

Zelllll:
Although there are things that I still plan to change here, any of the current feedback on this pr can be handled by maintainers. I'm not going to spend that much time on it after you all decided to ban me.

src/code/z_player_lib.c
Zelllll: I've now documented this function, and renamed it to `Player_SnapFootToFloor` to be more general.

src/code/z_player_lib.c
Zelllll: why would i revert it? its the position of the hookshot itself (i.e. in hand), while the hook position is the position of the tip of the hookshot (the spike)

Zelllll:
As far as I know I've now addressed most of the feedback given.

Zelllll:
I have taken the time out of my life to update this pr and address all the feedback that was given, yet like usual I've been met with no response at all. This has been open for 4 months now...

PR #857, opened 2021-06-23
`z_actor` OK and lots of documentation

Zelllll:


src/code/z_actor.c
```diff
@@ -222,186 +212,180 @@ void Actor_SetFeetPos(Actor* actor, s32 limbIndex, s32 leftFootIndex, Vec3f* lef
...
-        func_8002BE64(targetCtx, i, 0.0f, 0.0f, 0.0f);
-        entry->color.r = naviColor->inner.r;
-        entry->color.g = naviColor->inner.g;
-        entry->color.b = naviColor->inner.b;
+    entry = &targetCtx->targetInfo[0];
+    for (i = 0; i < ARRAY_COUNT(targetCtx->targetInfo); i++) {
+        Target_SetPos(targetCtx, i, 0.0f, 0.0f, 0.0f);
+        entry->color.r = targetColor->r;
+        entry->color.g = targetColor->g;
+        entry->color.b = targetColor->b;
```
AngheloAlf, line 260: What is this `for` loop doing?
By looking at the old code it seems it was iterating over the "inner" colors only
```c
    // left column
    { { 0, 255, 0, 255 }, { 0, 255, 0, 0 } },         // ACTORCAT_SWITCH,
    { { 0, 255, 0, 255 }, { 0, 255, 0, 0 } },         // ACTORCAT_BG,
    { { 255, 255, 255, 255 }, { 0, 0, 255, 0 } },     // ACTORCAT_PLAYER,
```
But looking at this code, it would imply it is somehow getting the first 3 entries of that data, like this
```c
    { { 0, 255, 0, 255 }, { 0, 255, 0, 0 } },         // ACTORCAT_SWITCH,
    { { 0, 255, 0, 255 }, 
```

So, removing the `NaviColor` struct just made this piece of code a bit less readable and confusing imo.

src/code/z_actor.c
```diff
@@ -222,186 +212,180 @@ void Actor_SetFeetPos(Actor* actor, s32 limbIndex, s32 leftFootIndex, Vec3f* lef
...
+    targetCtx->targetInfo[targetIndex].pos.y = y;
+    targetCtx->targetInfo[targetIndex].pos.z = z;
+    targetCtx->targetInfo[targetIndex].radius = targetCtx->targetRadius;
+}
 
-// unused
-Gfx D_80115FF0[] = {
-    gsSPEndDisplayList(),
+// Used for both the Z Target and Navi when targetting
+static Color_RGBA8 sNaviTargetColorList[][2] = {
```
AngheloAlf, line 228: We try to avoid naming stuff in the codebase based on character names (for example Player vs Link). So we should avoid using `Navi` too.
```suggestion
static Color_RGBA8 sFairyColorList[][2] = {
```

src/code/z_actor.c
```diff
@@ -419,101 +403,95 @@ void func_8002C124(TargetContext* targetCtx, GlobalContext* globalCtx) {
     CLOSE_DISPS(globalCtx->state.gfxCtx, "../z_actor.c", 2158);
 }
 
-void func_8002C7BC(TargetContext* targetCtx, Player* player, Actor* actorArg, GlobalContext* globalCtx) {
+void Target_Update(TargetContext* targetCtx, Player* player, Actor* actorArg, GlobalContext* globalCtx) {
     s32 pad;
-    Actor* unkActor;
+    Actor* naviTarget = NULL; // Actor that Navi is currently targetting
```
AngheloAlf, line 408:
```suggestion
    Actor* targetedActor = NULL;
```

AngheloAlf (changes requested):
It's a bit late. I'll continue reviewing this another day

src/code/z_actor.c
Zelllll: Calling it fairy is a slight issue because other fairies technically exist in the game (i.e. the ones belonging to the Kokiri children)

src/code/z_actor.c
Zelllll: This is actually specifically for the actor that Navi is focusing on, but not nessescarily what is being targeted (lock on)

src/code/z_actor.c
Zelllll: The struct removal was done for a few reasons. For one I don't really like having tons of structs for things that were more than likely 2D arrays (and match as such), but also for the fact that there was a lot of debate over whether it should be called Navi color or whatever so just removing the struct entirely fixes that. Either way though I can add comments to make it clearer, but a struct is unnecessary imo.

src/code/z_actor.c
AngheloAlf: Then the word `target` shouldn't be used if it isn't necessarily related to the target system.

Zelllll: how is it not related to the target system? 🤔 Navi is very intertwined with it, but anyways whenever Navi hovers over something the arrow also appears above it, but that does not mean it's locked into (targeted)

src/code/z_actor.c
AngheloAlf: The fact that it can match using something else doesn't imply we should do it.
The old struct made very clear the fact what each entry is supposed to be (inner and outer colors), and not having them just obfuscates that information for no good reason, and not even leaving a comment about how that data should be structured.

> for things that were more than likely 2D arrays 

Is there any code that access this data as a 2D array or seems to imply it? 
By what I'm seeing, not even you accessed it as a 2D array, otherwise you would have made something like this instead.
```c
        entry->color.r = targetColor[0].r;
        entry->color.g = targetColor[0].g;
        entry->color.b = targetColor[0].b;
```

Zelllll: What does "accessing it like a 2D array" even mean?

AngheloAlf: You are the guy who keep saying those were likely 2D arrays, so you should be the one who knows more about it.

For example, a for loop
```c
for (i = 0; i < 2; i++) { 
    something = targetColor[i].r;
}
```

Zelllll: If you just have an array of 2 colors, I think a 2d array is logical. Also doing it this way makes it so I don't have to have a "Navi color" pointer in the functions for drawing the target triangles, rather just a ColorRGB8 pointer instead, which seems more logical than having a struct.

Zelllll:
Update:
- Some review comments are addressed and some documentation has been improved to be more accurate
- actorfixer.py is updated to reflect most of the changes in this pr

src/code/z_actor.c
Zelllll: Renaming this to anything besides Navi is outside the scope of this pr I think since it's called Navi everywhere else in the repo, so I think I'll have to resolve this.

src/code/z_actor.c
```diff
@@ -1473,14 +1454,14 @@ typedef struct {
...
-TargetRangeParams D_80115FF8[] = {
+static TargetRangeParams sTargetRanges[] = {
     TARGET_RANGE(70, 140),   TARGET_RANGE(170, 255),    TARGET_RANGE(280, 5600),      TARGET_RANGE(350, 525),
     TARGET_RANGE(700, 1050), TARGET_RANGE(1000, 1500),  TARGET_RANGE(100, 105.36842), TARGET_RANGE(140, 163.33333),
     TARGET_RANGE(240, 576),  TARGET_RANGE(280, 280000),
 };
 
-u32 func_8002F090(Actor* actor, f32 arg1) {
-    return arg1 < D_80115FF8[actor->targetMode].rangeSq;
+u32 func_8002F090(Actor* actor, f32 range) {
```
AngheloAlf, line 1463: Seems weird that this function returns an unsigned while it is doing a boolean operation. Would this match with an `s32`?

src/code/z_actor.c
```diff
@@ -1532,41 +1514,44 @@ s32 func_8002F1C4(Actor* actor, GlobalContext* globalCtx, f32 arg2, f32 arg3, u3
...
-    f32 var1 = 50.0f + actor->colChkInfo.cylRadius;
+s32 Actor_RequestToTalk(Actor* actor, GlobalContext* globalCtx) {
+    f32 radius = 50.0f + actor->colChkInfo.cylRadius;
 
-    return func_8002F2CC(actor, globalCtx, var1);
+    return Actor_RequestToTalkInRange(actor, globalCtx, radius);
 }
 
-u32 func_8002F334(Actor* actor, GlobalContext* globalCtx) {
+u32 Actor_IsDoneTalking(Actor* actor, GlobalContext* globalCtx) {
```
AngheloAlf, line 1531: Wording is weird here. Maybe `Actor_HasFinishedTalking`?
Also, the same question about `u32` vs `s32` here.

src/code/z_actor.c
```diff
@@ -1766,13 +1753,13 @@ s32 func_8002F9EC(GlobalContext* globalCtx, Actor* actor, CollisionPoly* poly, s
...
-s32 D_8015BC14;
-f32 D_8015BC18;
+// Local data used for Farore's Wind light
+static LightInfo D_8015BC00;
+static LightNode* D_8015BC10;
+static s32 D_8015BC14;
+static f32 D_8015BC18;
 
-void func_8002FA60(GlobalContext* globalCtx) {
+void Actor_InitFaroresWindPointer(GlobalContext* globalCtx) {
```
AngheloAlf, line 1762: I'm not a fan of the pointer suffix here. What do you mean with this? Like, a C pointer or something in-game?

src/code/z_actor.c
```diff
@@ -2472,7 +2454,10 @@ void func_800315AC(GlobalContext* globalCtx, ActorContext* actorCtx) {
     CLOSE_DISPS(globalCtx->state.gfxCtx, "../z_actor.c", 6563);
 }
 
-void func_80031A28(GlobalContext* globalCtx, ActorContext* actorCtx) {
+/**
+ * Kills all actors in ActorContext whose object dependency is not loaded.
+ */
+void Actor_KillIfObjectIsNotLoaded(GlobalContext* globalCtx, ActorContext* actorCtx) {
```
AngheloAlf, line 2460: Wording is weird here, but I don't have any suggestions.

src/code/z_actor.c
```diff
@@ -3230,17 +3223,20 @@ Actor* Actor_GetCollidedExplosive(GlobalContext* globalCtx, Collider* collider)
     return NULL;
 }
 
-Actor* func_80033684(GlobalContext* globalCtx, Actor* explosiveActor) {
-    Actor* actor = globalCtx->actorCtx.actorLists[ACTORCAT_EXPLOSIVE].head;
+/**
+ * Returns a pointer to a nearby exploding actor to an actor, or NULL if none are found.
+ */
+Actor* Actor_FindExplosionNearby(GlobalContext* globalCtx, Actor* actor) {
```
AngheloAlf, line 3229: What about `Actor_FindNearbyExplosion`?

src/code/z_actor.c
```diff
@@ -1514,14 +1495,15 @@ u32 func_8002F194(Actor* actor, GlobalContext* globalCtx) {
     return false;
 }
 
-s32 func_8002F1C4(Actor* actor, GlobalContext* globalCtx, f32 arg2, f32 arg3, u32 exchangeItemId) {
+s32 Actor_RequestToTalkAndExchangeItemInRange(Actor* actor, GlobalContext* globalCtx, f32 radius, f32 height,
```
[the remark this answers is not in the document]
Zelllll: Yeah the reason why it was named this way is because there were similar functions and I wanted to highlight the difference between them. Maybe `Actor_RequestToTalkTradeItemInRange` instead? Still a bit long unfortunately.

src/code/z_actor.c
Zelllll: yeah I had trouble with this one too. I wanted to name it something like `Actor_KillAllObjectsInContextIfTheirObjectDependencyIsNotLoaded` but that doesn't work well for obvious reasons.

src/code/z_actor.c
Zelllll: It means like the literal pointer that shows the player where the spell was casted. Maybe something like `Actor_InitFaroresWindSpellIndicator`, I'm not sure, still doesn't sound great.

tools/actorfixer.py
```diff
@@ -89,6 +89,33 @@
     "func_80033260": "Actor_SpawnFloorDust",
     "func_80032C7C": "Enemy_StartFinishingBlow",
     "actorCtx.unk_00": "actorCtx.freezeFlashTimer",
+    # Added in Zel's z_actor.c docs
```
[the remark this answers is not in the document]
Zelllll: ah yeah i can remove that, i only had that for myself when i was editing the script anyways

src/code/z_actor.c
AngheloAlf: I think that name is better but I agree it doesn't sound great either.

src/code/z_actor.c
```diff
@@ -3652,86 +3645,87 @@ s32 func_800343CC(GlobalContext* globalCtx, Actor* actor, s16* arg2, f32 arg3, c
...
 
-    if (arg8) {
-        Math_SmoothStepToS(&actor->shape.rot.y, sp44, 6, 2000, 1);
+    if (waistCanRotate) {
+        Math_SmoothStepToS(&actor->shape.rot.y, yAngle1, 6, 2000, 1);
     }
 
-    temp1 = CLAMP(sp46, arg4, (s16)(u16)arg3);
-    Math_SmoothStepToS(&arg1->unk_08.x, temp1, 6, 2000, 1);
+    offset = CLAMP(xAngle1, neckAngleMinX, (s16)(u16)neckAngleMaxX);
```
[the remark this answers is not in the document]
Zelllll: strangely it apparently is

src/code/z_actor.c
petrie911: `Actor_KillActorsMissingObjects`?

Zelllll: My only issue with that is the redundancy of the word "actor" twice, but overall I don't mind that name.

louist103: `KillIfObjectMissing`?

Zelllll: maybe `Actor_KillAllIfObjectMissing`?

src/code/z_actor.c
```diff
@@ -1577,6 +1562,7 @@ u32 Actor_HasParent(Actor* actor, GlobalContext* globalCtx) {
     }
 }
 
+// Related to carrying an actor as well as giving the player an item?
 s32 func_8002F434(Actor* actor, GlobalContext* globalCtx, s32 getItemId, f32 xzRange, f32 yRange) {
```
AngheloAlf, line 1566: I think this function is the one used to give the player a GI item.
What do you think of some simple name like `Actor_GetItem`?

Zelllll: I think there is already something like that, but no its not just that, this is also for carrying an item, i think the crate uses it or something really weird like that

AngheloAlf: Doing a quick search, it seems only 3 actors explicitly call this function using `GI_NONE` to be hold by player (ishi, niw and tsubo). Every other hold-able actor calls `func_8002F580` to do that (which internally call this function with `GI_NONE`).

Considering that, I propose the following:
- Rename `func_8002F434` to `Actor_GetItem` and add an appropriate comment about the effects of passing `GI_NONE` and `GI_MAX`.
- Rename `func_8002F554` to `Actor_GetItemNearby`
- Rename `func_8002F580` to `Actor_HoldActor`

Zelllll: What does passing in `GI_MAX` do?

AngheloAlf: It is used to catch the actor in a bottle. It is used by insect, ice_hono, fish and elf.

Zelllll: something like `Actor_Carry` makes more sense imo

AngheloAlf: Yeah, that's sounds good too, I'm not very good naming stuff :sweat_smile:

Zelllll: I was saying Actor_Carry for Actor_GetItem to clarify though. Link technically "carries" the item above his head

AngheloAlf: > Carry
>  to take or support from one place to another; convey; transport
https://www.dictionary.com/browse/carry

Link isn't taking an item from one place to another above his head, so `Actor_Carry` would be a bit of a stretch.

Zelllll: I feel like the reason that this function is so weird though has to be some idea like this though, like the devs somehow thinking that all this functionality is related 🤔

src/code/z_actor.c
Zelllll: Note:
Should be renamed to `Actor_InitFaroresWindMarker`

src/code/z_actor.c
Zelllll: Note:
Should he renamed to `Actor_PickUp` along with a comment explaining it's weird usage

PR #858, opened 2021-06-24
Extract all "place name" title cards

louist103:


Zelllll (approved):
Looks good, but I'd maybe suggest renaming it to `place_titles` though to be short or `place_title_cards`

fig02 (approved):

Zelllll (approved):
Looks good!

assets/xml/textures/place_title_cards.xml
```diff
@@ -0,0 +1,287 @@
+<Root>
+    <File Name="g_pn_01">
+        <Texture Name="gForestTempleTitleCardEngTex" OutName="forest_temple_english" Format="ia8" Width="144" Height="24" Offset="0x0"/>
+        <Texture Name="gForestTempleTitleCardGerTex" OutName="forest_temple_german" Format="ia8" Width="144" Height="24" Offset="0xD80"/>
+        <Texture Name="gForestTempleTitleCardFraTex" OutName="forest_temple_french" Format="ia8" Width="144" Height="24" Offset="0x1B00"/>
```
Roman971, line 5: I think we should try to be consistent when naming for different languages, and it looks like so far we've been using `_eng`/`_ger`/`_fra` for file names + `ENG`/`GER`/`FRA` for symbol names

Zelllll: agreed

Roman971:
You marked the comment as resolved but I don't see any changes, so I think you might have forgotten to push them?

assets/xml/textures/place_title_cards.xml
```diff
@@ -0,0 +1,287 @@
+<Root>
+    <File Name="g_pn_01">
```
AngheloAlf, line 2: Could you add a segment number?
I ask because one of the things that was planned to add to ZAPD's error handler was checking if the segment number was missing in a XML.
I understand that a segment number is not needed here and it probably doesn't even make sense, but it would simplify ZAPD's side to not have special cases like this.

fig02:
Not sure where we are regarding this `number 0` conversation, some of it was happening in discord.
My general opinion is that we should not be using number 0 for this purpose I dont think

Roman971:
I mentioned this in discord the other day but imo we shouldn't use number 0 for this type of segments (both in spec and in XMLs). I don't really think it makes sense to require a segment number in XMLs either, since not all assets are contained in numbered segments (eg. assets in `code` or overlays).

This is just my opinion but I think it deserves more discussion, so maybe we could move the number 0 part to a separate PR so we don't hold this one further.

Roman971 (approved):

fig02 (approved):

PR #859, opened 2021-06-27
Fix all xmls to explicitly use offsets

louist103:


Dragorn421 (commented):
This seems to contain #858 , is that intended

fig02 (approved):

AngheloAlf (approved):

Zelllll (approved):
Looks great!

Roman971 (approved):

PR #815, opened 2021-05-14
Decompiled A LOT of objects

shawlucas:
I decompiled a lot of objects. The full list is on trello. object_mag is done too but it wont extracct correctly until there is an update to ZAPD.

I hope my xml files are okay...i have been away for a while and don't know if we have new naming schemes or something

Thanks

PR #860, opened 2021-06-29
Document EnGoroiwa and BgJyaGoroiwa

Dragorn421:
long time no contribute, go ham on me

~~#815 has the Goroiwa object used by both actors, probably merge it first~~


fig02 (approved):
I cant speak for the changes to the mtx functions, since im not knowledgeable with that stuff. Will defer to others on that
But the rest looks good to me

src/overlays/actors/ovl_En_Goroiwa/z_en_goroiwa.h
```diff
@@ -13,18 +13,18 @@ typedef struct EnGoroiwa {
...
-    /* 0x01C8 */ s16 collisionTimer;
+    /* 0x01C6 */ s16 bounceCount;
+    /* 0x01C8 */ s16 collisionDisabledTimer;
     /* 0x01CA */ s16 endWaypoint;
     /* 0x01CC */ s16 currentWaypoint;
     /* 0x01CE */ s16 nextWaypoint;
     /* 0x01D0 */ s16 pathDirection;
     /* 0x01D2 */ u8 isInKokiri;
-    /* 0x01D3 */ u8 unk_1D3;
+    /* 0x01D3 */ u8 flags;
```
Roman971, line 27: We don't necessarily have to change it here but imo we should avoid using names of fields that are already in the base actor struct. In this case it might be better to have it as stateFlags or something else if it makes sense.

src/code/sys_matrix.c
```diff
@@ -739,7 +739,11 @@ void func_800D1FD4(MtxF* mf) {
     cmf->zz = mf->zz * temp3;
 }
 
-void func_800D20CC(MtxF* mf, Vec3s* vec, s32 flag) {
+/**
+ * Gets the rotation the specified matrix represents, using Euler ZXY (most likely).
+ * The flag value doesn't matter for a rotation matrix. Not 0 does extra calculation.
+ */
+void Matrix_MtxFToRotS(MtxF* mf, Vec3s* rotDest, s32 flag) {
```
Roman971, line 746: Do we know the difference between this function and func_800D2264? I don't know how I feel about naming one and not the other since they are both used and seem to be very similar.

Also I just noticed one uses an early return where the other uses an else, which should probably be fixed while we're at it.

Dragorn421: hm. looking at usage the difference may be Euler ZXY vs Euler ZYX, I'll see if I can confirm that

src/overlays/actors/ovl_En_Goroiwa/z_en_goroiwa.c
```diff
@@ -417,93 +426,96 @@ s32 func_80A4CB78(EnGoroiwa* this, GlobalContext* globalCtx) {
...
-    temp = &sp80;
-    if (this->unk_1D3 & 8) {
-        func_80A4BF28(this, globalCtx, &sp28);
-        Math3D_Vec3f_Cross(&unitY, &this->actor.velocity, temp);
+    rollAngleDiff *= this->rollRotSpeed;
+    rollAxisPtr = &rollAxis;
+    if (this->flags & ENGOROIWA_RETAIN_ROT_SPEED) {
+        /*
+         * EnGoroiwa_GetPrevWaypointDiff has no side effects and its result goes unused,
+         * the devs probably meant to use the result instead of the actor's velocity in the
```
Dragorn421, line 450: "the devs"

src/code/sys_matrix.c
Dragorn421: I'm confused for several reasons with matrices,

1) in `mf->xy` the x is the column and y the line which is unlike common usage

2) if I'm right about the rotation orders I'm documenting, already documented `RPY` stuff is wrong like in `SkinMatrix_SetRotateRPY`:

```c
/**
 * Produces a rotation matrix = (roll rotation matrix) * (pitch rotation matrix) * (yaw rotation matrix)
 */
void SkinMatrix_SetRotateRPY(MtxF* mf, s16 roll, s16 pitch, s16 yaw)
```
that would mean it uses Euler XYZ except it clearly doesn't, eg if you look at `mf->xy` (column 1, line 2):
in `SkinMatrix_SetRotateRPY`: `xy (2,1) = sin(yaw)*cos(pitch) == sin(z)*cos(y)`
[wikipedia](https://en.wikipedia.org/wiki/Euler_angles#Rotation_matrix) XYZ: `xy (2,1) = cos(x)*sin(z) + cos(z)*sin(x)*sin(y)`
[wikipedia](https://en.wikipedia.org/wiki/Euler_angles#Rotation_matrix) ZYX: `xy (2,1) = cos(y)*sin(z)`

Or maybe I am missing something

Regardless, I changed `Matrix_MtxFToRotS` to `Matrix_MtxFToYXZRotS` (and not ZXY, oops) and `func_800D2264` to `Matrix_MtxFToZYXRotS` (also cleaned up that function to match the YXZ version)

zbanks: I have a few points to add about rotations:

I like to use [this NASA PDF](https://ntrs.nasa.gov/api/citations/19770019231/downloads/19770019231.pdf) as a reference, because it has both matrix -> angles & angle -> matrix for all 12 sets of Euler angles. It might not be the best *descriptive* reference for OOT, because some of the conventions differ. (Also: I'm pretty sure OOT uses a right-hand coordinate system? That would also be important to know...)

Like Dragorn said, the rows/columns are swapped, as if all matrices are transposed. 
I think this explains why `void SkinMatrix_MtxFMtxFMult(MtxF* mfB, MtxF* mfA, MtxF* dest); // dest = mfA * mfB`, seems to have the order of the arguments swapped. Since `A^T B^T = (BA)^T`, if we renamed elements of `MtxF` to be column-major, then we could say `void SkinMatrix_MtxFMtxFMult(MtxF* mfA, MtxF* mfB, MtxF* dest); // dest = mfA * mfB` which seems more likely?

To map the OOT code  to the linked PDF, I used the following conversions:
- `m_12` maps to `mf->yx`, `m_33` maps to `mf->zz`, etc. (rows/cols swapped)
- `tan^-1(y/x)` maps to `Math_FAtan2F(y, x) * (0x8000 / M_PI)`
- This function uses `sqrt(m_31^2 + m_33^2)` instead of `sqrt(1-m_32^2)` (which should be equal for a rotation matrix)
- This function handles [gimbal lock](https://en.wikipedia.org/wiki/Gimbal_lock) when the middle angle is exactly +/- pi/2 (+/- 90 degrees) by putting all the rotation in the first angle (Y) and setting the third angle (Z) to 0, which is a pretty common solution.
    - However - the actual value used here seems suspicious to me? I can't easily prove that it's equivalent to anything simpler or more obviously correct?

I agree this function `Matrix_MtxFToRotS` is for YXZ - but I think it has a bug in `rotDest->y = Math_FAtan2F(mf->zx, mf->zz) * (0x8000 / M_PI);`. I think this should be `rotDest->y = Math_FAtan2F(mf->xz, mf->zz) * (0x8000 / M_PI);` (`zx --> xz`), and I don't think the two terms are equivalent. The gimbal lock case mentioned above may also be incorrect?

For naming, I'd recommend using names like `ZYX` instead of `RPY` because it avoids ambiguity in labeling the yaw/pitch/roll axes, and is explicit about being based on intrinsic rotations (`ZYX` vs. `ZYZ`).

And for those unfamiliar: in general, you can't mix different forms of Euler angles: passing the angles from `Matrix_MtxFToYXZRotS` into `SkinMatrix_SetRotateRPY` (ZYX) doesn't produce the original orientation. 3D rotations are non-commutative, which makes it pretty hard to convert between the different Euler angle forms (it's not as simple as swapping the angles!). 

(I hope chiming in here isn't out of place -- I've had to deal with 3D rotations a lot in the past; I'm happy to contribute/answer questions)

fig02:
Whats the status on the matrix stuff? A good amount of the conversation went over my head to be honest, so its hard for me to keep up lol

Dragorn421:
Roman brought up that the matrix function I named `Matrix_MtxFToRotS` was very similar to another one, so I named both (the difference turned out to be the "rotation order", one is YXZ the other is ZYX (see: Euler angles))

zbanks agrees on the YXZ one being YXZ and I believe I'm not mistaken about the ZYX one so it should be good for this PR.

The extensive discussion about oot's matrix stuff on discord between zbanks and me (and some of the stuff they bring up in their above comment) will lead to me opening another PR about matrices.

TLDR: this PR is good for me (but I'm the author so that weighs peanuts)

fig02:
alrighty, sounds good

fig02 (approved):

Roman971 (approved):

PR #861, opened 2021-07-01
EnMb documented (Moblins)

Dragorn421:
EnMb features a lot of unused code and content.

The actor is "split" in three behaviors:

- "Spear Guard" (variable `-1`): uses a spear, walks around home point, charges player if too close, unused in the final game
- "Club" (variable `0`): uses a club, stands still and smashes its club on the ground when the player approaches (I'm not sure how the particles hitting Link works)
- "Spear Patrol" (variable `0xPP00` PP=pathId): uses a spear, patrols following a path, charges

Spear and Club moblins use a different skeleton but the hierarchy (limbs relations/ordering) is the same

I prefixed action setup functions and action functions according to which type of moblin they handled. It makes for really long function names unfortunately. Some functions handle both Spear Guard and Spear Patrol and use just "Spear" as prefix.

Looking at the commit before "Use actual decomp-style function names" may be more readable. It uses the format `EnMb_Prefix_[Setup]FuncName` instead of `EnMb_[Setup]PrefixFuncName`

<details>
<summary>Here's what the colliders look like on the spear moblin when charging (click to show image)</summary>

</details>


include/functions.h
```diff
@@ -483,15 +483,15 @@ void BodyBreak_Alloc(BodyBreak* bodyBreak, s32 count, GlobalContext* globalCtx);
 void BodyBreak_SetInfo(BodyBreak* bodyBreak, s32 limbIndex, s32 minLimbIndex, s32 maxLimbIndex, u32 count, Gfx** dList,
                        s16 objectId);
 s32 BodyBreak_SpawnParts(Actor* actor, BodyBreak* bodyBreak, GlobalContext* globalCtx, s16 type);
-void Actor_SpawnFloorDust(GlobalContext* globalCtx, Actor* actor, Vec3f* arg2, f32 arg3, s32 arg4, f32 arg5, s16 arg6,
-                          s16 arg7, u8 arg8);
+void Actor_SpawnFloorDustRing(GlobalContext* globalCtx, Actor* actor, Vec3f* posXZ, f32 radius, s32 amountMinusOne, f32 randAccelWeight,
+                              s16 scale, s16 scaleStep, u8 useLighting);
```
EllipticEllipsis, line 487: When renaming functions that are used in undecomped stuff, we now add them to the dictionary in `actorfixer.py` to make updating to the new names easier.

src/overlays/actors/ovl_En_Mb/z_en_mb.c
```diff
@@ -134,170 +155,186 @@ static ColliderQuadInit sQuadInit = {
...
+    Collider_InitTris(globalCtx, &this->frontShielding);
+    Collider_SetTris(globalCtx, &this->frontShielding, &this->actor, &sFrontShieldingInit, this->frontShieldingTris);
+    Collider_InitQuad(globalCtx, &this->attackCollider);
+    Collider_SetQuad(globalCtx, &this->attackCollider, &this->actor, &sAttackColliderInit);
 
     switch (this->actor.params) {
-        case -1:
-            SkelAnime_InitFlex(globalCtx, &this->skelAnime, &gEnMbSkel_008F38, &gEnMbStandStillAnim, this->jointTable,
-                               this->morphTable, 28);
+        case -1: /* Spear Guard */
```
EllipticEllipsis, line 264: Make an enum for params, I think, then a few of these comments can go (keep the default).

EllipticEllipsis (commented):
Nice, this is much more comprehensible than before. Couple of minor things, although I confess I have not been through it with a fine-tooth comb.

I think some of the documentation you wrote in the PR description should be incorporated somewhere introductory—maybe at the top of the C file?

I also agree that using a midfix (?) based on type is a better idea: it makes it a lot more readable when the types have totally different function flow. I did something similar in... I think it was EnGe1.

src/overlays/actors/ovl_En_Mb/z_en_mb.c
```diff
@@ -7,10 +7,49 @@
...
+
 #define FLAGS 0x00000015
 
 #define THIS ((EnMb*)thisx)
 
+typedef enum {
+    /* -1 */ ENMB_TYPE_SPEAR_GUARD = -1,
+    /*  0 */ ENMB_TYPE_CLUB,
+    /*  1 */ ENMB_TYPE_SPEAR_PATROL
+} EnMbType;
```
EllipticEllipsis, line 25: Params enums usually go in the header so they can be accessed externally. Otherwise looks good!

Dragorn421: makes sense but I guess another enum would have to be created, as `ENMB_TYPE_SPEAR_PATROL == 1` is only used in code (it's set by code)
actors lists in room files use either `0x0000` for the Club Moblin or `0xPPFF` (`PP=pathId`) for the Spear Patrol Moblin
the lower byte can be anything as long as the whole variable isn't 0 or -1 so `ENMB_TYPE_SPEAR_PATROL == 1` would work, but apparently they used `0xFF`

so, do move it or just leave it and wait for another params pass

fig02: moving it is fine. even if some entries are not "public", in general you will be accessing type externally

fig02 (approved):

src/code/z_rcp.c
```diff
@@ -782,11 +782,11 @@ Gfx* Gfx_SetFog(Gfx* gfx, s32 r, s32 g, s32 b, s32 a, s32 n, s32 f) {
...
 
     if (n >= 1000) {
-        gSPFogFactor(gfx++, 0, 0); // gSPFogPosition(gfx++, ?, ?)
+        gSPFogFactor(gfx++, 0, 0); // gSPFogPosition(gfx++, 0, 10000000)
     } else if (n >= 997) {
-        gSPFogFactor(gfx++, 0x7FFF, 0x8100); // gSPFogPosition(gfx++, ?, ?)
+        gSPFogFactor(gfx++, 0x7FFF, 0x8100); // gSPFogPosition(gfx++, -3.92162, -0.0152593)
     } else if (n < 0) {
-        gSPFogFactor(gfx++, 0, 255); // gSPFogPosition(gfx++, ?, ?)
+        gSPFogFactor(gfx++, 0, 255); // gSPFogPosition(gfx++, -2422000, 10000);
```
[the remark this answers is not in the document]
Dragorn421: I'm pretty sure the og code just uses magic values with `gSPFogFactor` just like we have
I'm not sure why I didn't see this before... :

https://github.com/zeldaret/oot/blob/1f6a74e636f484fac92d5707352dc3e10b981bab/include/ultra64/gbi.h#L2601-L2614

so I could put
```c
        gSPFogFactor(gfx++, 0, 0); // always transparent fog
    } else if (n >= 997) {
        gSPFogFactor(gfx++, 0x7FFF, 0x8100); // mostly opaque fog
    } else if (n < 0) {
        gSPFogFactor(gfx++, 0, 255); // always opaque fog
```

I'm really not sure what `gSPFogFactor(gfx++, 0x7FFF, 0x8100);` wants to do, fog alpha would range from 0x101 (near z=-1) to 0x255 for `z>=-0.99...?`? with a high `n`/near value I'd have expected more no-fog than full-fog

src/code/z_eff_ss_dead.c
```diff
@@ -28,14 +28,14 @@ void func_80026230(GlobalContext* globalCtx, Color_RGBA8* color, s16 arg2, s16 a
     CLOSE_DISPS(globalCtx->state.gfxCtx, "../z_eff_ss_dead.c", 129);
 }
 
-void func_80026400(GlobalContext* globalCtx, Color_RGBA8* color, s16 arg2, s16 arg3) {
+void EffectSsDead_SetCloseFogColorOpa(GlobalContext* globalCtx, Color_RGBA8* color, s16 timeLeft, s16 timeTotal) {
```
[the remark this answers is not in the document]
Dragorn421: About the `EffectSsDead` prefix I can change it sure but to what?

what do they set more than just fog? I think the `SetCloseFogColorOpa`/Xlu part is descriptive enough

I used `timer` because the function is passed, well, a timer value, and it's used like a typical time in a cos() wave I thought (`Math_CosS((0x4000 / timeTotal) * timeLeft)`), but I could use `progressCur`/`progressMax` if you prefer that?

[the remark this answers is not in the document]
fig02: never occured to me that this file is prefixed with `effect_ss` when its not a softsprite effect lol

Anyway, we call this thing `ColorFilter` throughout the codebase, and abstract the fact that its actually achieved using fog to the user. Cause fog in that context doesnt make all too much sense.
Maybe we dont want to abstract it here, and only abstract it in z_actor. I dont have a solution to any of this naming, just putting it out there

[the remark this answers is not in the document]
Roman971: My point was that they set more than just the color itself. And they seem to have different implementations in the file, so we should be careful not to give names that aren't compatible with the other functions.

Anyway, at this point I'd say we should leave those functions unnamed for now since it's the only thing holding the PR afaict. Regarding the arguments, I'm not familiar with the math behind time defined like that in a cos wave, but we can probably go with these names if you think it's right.

Roman971 (approved):

PR #862, opened 2021-07-04
Change PreRenderContext to just be PreRender

Zelllll:
This pull request renames `PreRenderContext` to simply be `PreRender`, because PreRenders can be used anywhere, and there being a specific "context" for them is incorrect. `PreRender` should be a normal data type just like `Actor`, `SkelAnime`, etc.


fig02:
This is true for any of our "contexts" really. For example, file select has an instance of `EnvironmentContext` and `MessageContext` as well.
I think this is best left for when we discuss and rename all of these "contexts" in a more detailed documentation phase

Dragorn421:
I think I understand Zel's point, most Context structs seem to follow some kind of a singleton pattern (single global/message/actor context and having several doesn't make sense) which doesn't hold for a prerender.

I do agree it isn't a pressing matter though, it's not like it's misleading or entirely undocumented. The "context" word comes from romhacking history anyway (iirc) and may be get rid of entirely too eventually...

Roman971:
`PreRenderContext` is fine if we follow the loose definition of a context. But naming this specific struct `PreRender` could make sense since it's a bit like the `View` in the sense that they define 1 prerender or 1 view, and we don't use `ViewContext`.

Imo the other contexts are different and should keep the "context" part because they contain variables used to handle the state of a group of things (eg. the `ActorContext` being used to manage all actors), so they belong in their own category and using a common term for that makes sense.

fig02 (approved):
fair enough, yea I can see how it makes sense

Roman971 (approved):

PR #863, opened 2021-07-04
Global Context And General Cleanup

Zelllll:
- Rename all instances of `nb` to be `num` (see: https://discordapp.com/channels/688807550715560050/740615251011174401/861367462750846977 )
- Rename `Scene` to `SceneTableEntry`
- Add `DoorContext`
- Fix incorrect size for `SkyboxContext`, and add the missing variables.
- Fix incorrect size for `RoomContext` and add the missing variables.
- Made all texture pointers be `void*` instead of `u64*`, since soon these will probably be changed to not be `u64`.


include/z64.h
```diff
@@ -756,6 +758,7 @@ typedef struct {
     /* 0x38 */ DmaRequest dmaRequest;
     /* 0x58 */ OSMesgQueue loadQueue;
     /* 0x70 */ OSMesg loadMsg;
+    /* 0x74 */ s16 unk_74[2];
 } RoomContext; // size = 0x74
```
AngheloAlf, line 762: Are you sure about this? The size doesn't match

Zelllll: oh i forgot to update it lol

louist103 (approved):

src/code/z_scene.c
```diff
@@ -327,17 +327,17 @@ void func_80098C24(GlobalContext* globalCtx, SceneCmd* cmd) {
...
 void func_80098C68(GlobalContext* globalCtx, SceneCmd* cmd) {
-    globalCtx->nbTransitionActors = cmd->transiActorList.num;
-    globalCtx->transitionActorList = SEGMENTED_TO_VIRTUAL(cmd->transiActorList.segment);
+    globalCtx->doorCtx.numActors = cmd->transiActorList.num;
+    globalCtx->doorCtx.actorList = SEGMENTED_TO_VIRTUAL(cmd->transiActorList.segment);
 }
 
-void func_80098CBC(GlobalContext* globalCtx, u8* nbTransitionActors) {
-    *nbTransitionActors = 0;
+void Door_Init(GameState* state, DoorContext* doorCtx) {
```
Dragorn421, line 334: In #857 you used `Actor_InitContext`, maybe `Door_InitContext` would be appropriate here too, it would help not mistaking this function for `Actor_SpawnTransitionActors`

include/z64.h
```diff
@@ -940,6 +943,11 @@ typedef struct {
     /* 0x03 */ u8 byte3;
 } ElfMessage; // size = 0x4
 
+typedef struct {
+    /* 0x00 */ u8 numActors;
+    /* 0x04 */ TransitionActorEntry* actorList;
```
Dragorn421, line 948: I think this should stay named `transitionActorList`, even if it is now under `DoorContext` it may not be obvious to readers not familiar with the engine that this is in fact only transition actors

Zelllll: yeah, i had actually wanted to call the whole thing TransitionContext but that already exists for area fades

src/code/z_scene.c
Zelllll: yeah good idea, i think that every "context" should be named consistently, `InitContext` is fine i think

Zelllll:
Alright Dragorn's comments were addressed

Zelllll:
New change: All texture pointers that were u64* are now void*

Dragorn421 (commented):
So we should use `void*` for texture pointers?

In the past I used `UNK_PTR` or `UNK_TYPE*`, are such occurrences still present?

Zelllll:
`void*` is the new standard yeah

fig02 (approved):

src/overlays/actors/ovl_En_Bom_Bowl_Man/z_en_bom_bowl_man.c
```diff
@@ -523,9 +523,8 @@ s32 EnBomBowlMan_OverrideLimbDraw(GlobalContext* globalCtx, s32 limbIndex, Gfx**
     return false;
 }
 
-static u64* eyeTextures[3] = { gChuGirlEyeOpenTex, gChuGirlEyeHalfTex, gChuGirlEyeClosedTex };
-
 void EnBomBowlMan_Draw(Actor* thisx, GlobalContext* globalCtx) {
+    static void* eyeTextures[3] = { gChuGirlEyeOpenTex, gChuGirlEyeHalfTex, gChuGirlEyeClosedTex };
```
AngheloAlf, line 527: There's no need to put the size explicitly.
```suggestion
    static void* eyeTextures[] = { gChuGirlEyeOpenTex, gChuGirlEyeHalfTex, gChuGirlEyeClosedTex };
```

src/overlays/actors/ovl_En_Fr/z_en_fr.c
```diff
@@ -1058,7 +1058,7 @@ void EnFr_PostLimbDraw(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec
 }
 
 void EnFr_Draw(Actor* thisx, GlobalContext* globalCtx) {
-    static u64* eyeTextures[] = { 0x060059A0, 0x06005BA0 };
+    static void* eyeTextures[] = { 0x060059A0, 0x06005BA0 };
```
AngheloAlf, line 1061: Could you add symbols for those too?

Zelllll: Idk, if it's not decompiled then no.

AngheloAlf: If it isn't decompiled then you could do `extern void* D_060059A0;`

fig02: they are in hex all over the codebase. itll be up to whoever does this object to replace the symbols

Zelllll:
fixed the conflicts and updated a few things, should be ready to merge?

include/z64.h
```diff
@@ -327,7 +327,8 @@ typedef struct {
     /* 0x13C */ void* roomVtx;
     /* 0x140 */ s16  unk_140;
     /* 0x144 */ Vec3f rot;
-} SkyboxContext; // size = 0x150
+    /* 0x150 */ char unk_14A[0xE];
+} SkyboxContext; // size = 0x15E
```
Roman971, line 331: The field name and padding seems slightly incorrect here? (the resulting size being 0x160 either way)
```suggestion
    /* 0x150 */ char unk_150[0x10];
} SkyboxContext; // size = 0x160
```

include/z64.h
```diff
@@ -941,6 +943,11 @@ typedef struct {
     /* 0x03 */ u8 byte3;
 } ElfMessage; // size = 0x4
 
+typedef struct {
+    /* 0x00 */ u8 numTransitionActors;
+    /* 0x04 */ TransitionActorEntry* transitionActorList;
+} DoorContext;
```
Roman971, line 949: I think it would make more sense to name it TransitionActorContext or something like that since this is for transition actors in general, and we could probably shorten the name of its fields based on the context (eg. `numActors` and `list`)

Zelllll:
Ok I applied all of Roman's suggestions and fixed an incorrect type definition in z64scene.h

include/z64scene.h
```diff
@@ -258,17 +258,16 @@ typedef struct {
...
-    u8 diffuseClrA_R, diffuseClrA_G, diffuseClrA_B;
-    u8 diffuseDirA_X, diffuseDirA_Y, diffuseDirA_Z;
-    u8 diffuseClrB_R, diffuseClrB_G, diffuseClrB_B;
-    u8 diffuseDirB_X, diffuseDirB_Y, diffuseDirB_Z;
-    u8 fogClrR, fogClrG, fogClrB;
-    u16 unk;
-    u16 drawDistance;
-} LightSettings;
+    /* 0x00 */ Color_RGB8 ambientColor;
+    /* 0x03 */ s8 diffuseDir1[3];
```
fig02, line 263: these colors need to be u8 arrays to match in z_lights, and I believe in kankyo as well. So I dont think using `Color_RGB8 ` here is a good idea

fig02: not to mention the data structure for the lights in gbi is using u8 arrays as well

Zelllll: Where at specifically? 🤔

fig02: what do you mean, sorry

Zelllll: should we also move this struct into z64lights.h? would make more sense there, could also call it `EnvironmentLightSettings` but that would break ZAPD.

Zelllll: I meant where does it not match with the struct, but either way I just changed it

fig02: I think it will make more sense in kankyo, since thats where the settings are most relevant. I already have the struct there in my branch. So this struct is kind of temp tbh

Zelllll: alright, well it can stay here for the time being then.

fig02 (approved):

Roman971 (approved):

PR #865, opened 2021-07-05
Improve the state of handwritten assembly files

Thar0:
This PR aims to make the handwritten assembly files more understandable to readers by adding appropriate constants in place of magic numbers and offsets, along with comments describing what's going on. It also removes the 
`/* offset address instruction */` comments so as to make the files more friendly towards the eventual support of other versions, and spaces out the instruction operand(s) further from the mnemonic to make the code feel less dense.
This also facilitates the removal of OS hardware registers from `undefined_syms.txt`, of those now only the `0x800003XX` symbols remain.

Making this slightly early to get some opinions, I still have some files to finish.
 - What do we think of using more pseudo-instructions? They can make the asm more human-readable however the IDO assembler has sometimes reordered statements between instructions expanded from pseudo-instructions, making it not possible to match with pseudo-instructions in some places using GAS, so their use will be inconsistent at best.
 - ~~The `macro.inc` file has gotten somewhat expansive, does anyone have better suggestions for organising this?~~
 - ~~Is it fine for magic values to be duplicated? i.e. in both `macro.inc` and various `include/ultra64/` headers~~
 - For anyone knowledgeable on these functions, is there anything inaccurate in the comments?

~~To anyone who finds it concerning that I removed the symbol `D_80000002` in file select in favour of the raw address `0x80000002`, the function in question does not match if it were a symbol, it must be like `*((u8*)0x80000002)` to match. I know this as I matched that function (with fig's permission) a little while ago, however that is out of the scope of this PR, I only wanted it gone from `undefined_syms.txt` along with the rest of the symbols near the start of RAM.~~
(File select is now merged, so this consideration is now obsolete)


AngheloAlf:
What about splitting `macro.inc` a bit? For example making a file for cop0 stuff and so on

mzxrules:
> What about splitting `macro.inc` a bit? For example making a file for cop0 stuff and so on

seems unnecessary and just bloats dependencies. The file is far from being unwieldy and splitting it up makes it harder to discover registers. If a split were absolutely necessary, maybe I'd split out the n64os specific stuff

asm/__osDisableInt.s
```diff
@@ -8,34 +8,32 @@
...
+    mfc0    $t0, Status
+    li      $at, ~SR_IE
+    and     $t1, $t0, $at
+    mtc0    $t1, Status
+    andi    $v0, $t0, SR_IE
+    lw      $t0, ($t2)
+    andi    $t0, $t0, SR_IMASK
+    beq     $t0, $t3, .L800072E8
+     lui    $t2, %hi(__osRunningThread)
+     addiu  $t2, $t2, %lo(__osRunningThread)
```
Dragorn421, line 25: I'm not sure what indentation is used for, looks like to indicate delay slots? then the addiu shouldn't be indented? (I don't know)

Dragorn421 (commented):
> What do we think of using more pseudo-instructions? [...] their use will be inconsistent at best.

I don't think we should use macros or pseudo-instructions (that expand to 2 or more instructions) when their use can't be consistent, unless there really are only a few spots where they can't be used. For example I'm ok with `la -> lui + addiu` if it can be used in over ~95% of the `lui / addiu` spots, not counting branch delay slots spots (I assume pseudo instructions in delay slots would be confusing so to be avoided, but there's at least one, in bcmp.s line 29, so idk)

I think `macro.inc` is fine as is. As mzx said, it's nice to have all registers in one place.

> Is it fine for magic values to be duplicated?

I think yes, it's not like they can change.

> For anyone knowledgeable on these functions

I don't know enough about the N64. What can I read?

Did you use much of [libultra decomp](https://github.com/n64decomp/libreultra)?

asm/__osDisableInt.s
mzxrules: nah, just seems like indentation appears randomly throughout different asm files

Zelllll: It's because it's using both tabs and spaces, should probably be changed to all spaces

Thar0: The intention is to indent delay slots, but in manually formatting these files I have may have messed up in a few places.

Thar0:
@Dragorn421 a couple of sources for reading that I used:
I did make a lot of use of the libultra decomp for placement of constants
https://github.com/n64decomp/libreultra
I also referred to the vr4300 datasheet, for reading about the cpu cache, cop0 registers and exceptions & interrupts
http://datasheets.chipdb.org/NEC/Vr-Series/Vr43xx/U10504EJ7V0UMJ1.pdf
Also this page for rcp hardware registers
http://en64.shoutwiki.com/wiki/Memory_map_detailed
The n64 programming manual was also helpful in providing high level overviews for some of these functions
http://n64devkit.square7.ch/

include/ultra64/hardware.h
```diff
@@ -1,90 +1,121 @@
...
+#define PI_BSD_DOM1_RLS_REG     0x04600020 //PI dom1 release
+#define PI_BSD_DOM2_LAT_REG     0x04600024 //PI dom2 latency
+#define PI_DOMAIN2_REG          0x04600024
+#define PI_BSD_DOM2_PWD_REG     0x04600028 //PI dom2 pulse width
+#define PI_BSD_DOM2_PGS_REG     0x0460002C //PI dom2 page size
+#define PI_BSD_DOM2_RLS_REG     0x04600030 //PI dom2 release
+
+// PI_STATUS (read) bits
+#define PI_STATUS_BUSY          (1 << 0)
+#define PI_STATUS_IOBUSY        (1 << 1)
```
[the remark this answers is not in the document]
Thar0: Assemblers can compute constants like this just fine so there's no problem giving these to assembly files

JoshDuMan:
> * What do we think of using more pseudo-instructions? They can make the asm more human-readable however the IDO assembler has sometimes reordered statements between instructions expanded from pseudo-instructions, making it not possible to match with pseudo-instructions in some places using GAS, so their use will be inconsistent at best.

These reorderings can fixed by toggling noreorder as needed, in my experience. I would guess that is what the original devs did.

louist103 (commented):
This is more general but for the functions with the 4 NOPs of padding you can get rid of them and use `pad_text` in the spec to fill those in.

fig02 (approved):
thought i approved this ages ago when i looked at it, but i guess not

asm/__osGetFpcCsr.s
```diff
@@ -9,7 +10,8 @@
 
 .balign 16
 
-glabel __osGetFpcCsr
-/* 008680 80007A80 4442F800 */  cfc1  $v0, $31
-/* 008684 80007A84 03E00008 */  jr    $ra
-/* 008688 80007A88 00000000 */   nop   
+BEGIN __osGetFpcCsr
+    cfc1    $v0, FpCsr
```
Dragorn421, line 14: Is it `FpcCsr` as in `__osGetFpcCsr`, or `FpCsr`?

Fpc may stand for Floating Point Coprocessor
and Csr for Control/Status Register?

asm/guMtxF2L.s
```diff
@@ -9,34 +9,37 @@
...
-/* B7D6DC 8010653C 03E00008 */  jr    $ra
-/* B7D6E0 80106540 00000000 */   nop   
+# padding?
+    nop
+    nop
+    nop
+    nop
+
+BEGIN guMtxF2L
+    li      $at, 0x47800000 # 65536.0
```
Dragorn421, line 19:
```suggestion
    li      $at, 0x47800000 # 65536.0f
```
also in `guMtxL2F`, `guNormalize`, `guScale` and `guTranslate`

Makefile
```diff
@@ -1,5 +1,8 @@
 MAKEFLAGS += --no-builtin-rules
 
+SHELL = /bin/bash
+.SHELLFLAGS = -o pipefail -c
+
```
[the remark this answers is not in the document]
Thar0: This is so that `$(CPP) $(CPPFLAGS) -I include $< | $(AS) $(ASFLAGS) -o $@` will cause the build to fail if `cpp` errors (which happens for cases such as not being able to find an include)

asm/__osGetFpcCsr.s
Thar0: `FpcCsr`, will change

Dragorn421: I noticed in the fault.c PR, `Fault_PrintFPCSR` calls the register `FPCSR`
also a bunch of defines use `FPCSR`

are you changing all of this to fpc csr then?

asm/mio0.s
```diff
@@ -9,54 +9,59 @@
...
+    beqz    $t1, .read_comp
+     nop
+    lb      $t2, ($t9)      # read 1 byte from uncompressed data
+    addi    $t9, $t9, 1     # advance uncompressed start
+    addi    $a1, $a1, 1
+    b       .next_iter
+     sb     $t2, -1($a1)    # store uncompressed byte
+.read_comp:
+    lhu     $t2, ($a3)      # read 2 bytes from compressed data
+    addi    $a3, $a3, 2     # advance compresed start
```
Dragorn421, line 43:
```suggestion
    addi    $a3, $a3, 2     # advance compressed start
```

Makefile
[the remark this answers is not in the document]
Dragorn421: Could you add a comment near those lines to explain just that?

asm/__osGetFpcCsr.s
Thar0: I ended up not doing this since I've now switched to using `C1_FPCSR` from `r4300.h`

Makefile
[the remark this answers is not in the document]
Thar0: I've added a brief comment that's hopefully enough to clarify this

undefined_syms.txt
```diff
@@ -1,54 +1,20 @@
...
-D_A4800018 = 0xA4800018; // SI_STATUS_REG
+// z_player_lib
+D_06025218 = 0x06025218;
+D_060252D8 = 0x060252D8;
+D_06025438 = 0x06025438;
+D_06025598 = 0x06025598;
+D_06025658 = 0x06025658;
+D_060257B8 = 0x060257B8;
+D_0602A738 = 0x0602A738;
+D_0602CB48 = 0x0602CB48;
```
Dragorn421, line 17: regression/merge issue in here

asm/ipl3.s
```diff
@@ -7,4 +7,20 @@
...
+.endm
+
+IPL_SYMBOL osTvType,        0x80000300, 4
+IPL_SYMBOL osRomType,       0x80000304, 4
+IPL_SYMBOL osRomBase,       0x80000308, 4
+IPL_SYMBOL osResetType,     0x8000030C, 4
+IPL_SYMBOL osCicId,         0x80000310, 4
+IPL_SYMBOL osVersion,       0x80000314, 4
+IPL_SYMBOL osMemSize,       0x80000318, 4
+IPL_SYMBOL osAppNMIBuffer,  0x8000031C, 0x40
```
Dragorn421, line 24: For consistency with other acronyms like Rom, Dma, Tv, Jpeg, Cfb ... NMI should just be Nmi imo

include/ultra64/asm.h
```diff
@@ -0,0 +1,74 @@
+
+#define LEAF(x)                 \
+    .balign 4                  ;\
+    .globl x                   ;\
+    .type x, @function         ;\
+    x:                         ;\
```
Dragorn421, line 6: Could use a `#ifndef _LANGUAGE_C`, also include guards never hurt?

include/ultra64/gbi.h
```diff
@@ -954,7 +954,7 @@
 #define G_DL_NOPUSH     0x01
 
 /*
- * BEGIN C-specific section: (typedef's)
+ * LEAF(C-specific section: (typedef's))
```
Dragorn421, line 957: when regex does its job too well

include/ultra64/asm.h
```diff
@@ -0,0 +1,74 @@
+
+#define LEAF(x)                 \
+    .balign 4                  ;\
+    .globl x                   ;\
```
Dragorn421, line 4: `IPL_SYMBOL` uses `.global`, may want to decide on one or the other unless there's some compatibility reason?

Dragorn421 (commented):
No tharo my review progress NotLikeThis

Can you elaborate on asm.h ? Namely, what is the idea with the name of LEAF and do we need XLEAF, NESTED, XNESTED, IMPORT, EXPORT

include/ultra64/asm.h
Thar0: I'll also answer more generally about `asm.h` here.

`asm.h` is an sdk header (https://github.com/n64decomp/libreultra/blob/master/include/2.0I/sys/asm.h) which along with `regdef.h` (same sys directory) are designed to try to make assembly more portable between assemblers (in particular, libultra was assembled by both IDO's assembler and GAS and while they're very similar they're not quite the same) and MIPS ABIs (o32, n32, 64) by hiding a number of assembler-specific syntax behind preprocessor macros. I felt like BEGIN and such were so close to these that we might as well use the original, and I wanted to vacate `macro.inc` to only contain `glabel` again. 

The macros are lifted from `asm.h` with some minor tweaks like additional `.balign` and `.type` directives. I prefer to have more complete sdk headers even if some of their contents is not used, since others may find new use for them in modding.

`LEAF` as a name stems from "leaf function", one that doesn't call other functions, although this is something of a misnomer since you can use these just fine with non-leaf functions. I unfortunately can't refer you to any method to lodge a complaint to SGI about the choice of name.

I think `.globl` is supported by both IDO and GAS, while `.global` is only supported by the latter, but I haven't checked this since I'd need to set up `as0` in QEMU since there's no working recomp for it...
I didn't use .globl in ipl3 since I'm not sure if it's really portable anyway. Again this would be nice to check, but same reason above.

asm/ipl3.s
Thar0: NMI is how it's named in the sdk and generally we prioritize the sdk name over our usual style.

asm/osSetIntMask.s
```diff
@@ -1,123 +1,131 @@
...
+    xor     $t1, $t1, $at
+    andi    $t1, $t1, MI_INTR_MASK
+    or      $t2, $t2, $t1
+1:
+    sll     $t2, $t2, 0x10
+    or      $v0, $v0, $t2
+    lui     $at, MI_INTR_MASK
+    and     $t0, $a0, $at
+    and     $t0, $t0, $t3
+    srl     $t0, $t0, 0xf
```
Dragorn421, line 113: small inconsistencies among .s files, not a big deal:

the shift being hexadecimal or decimal in srl/sll instructions
hexadecimal being lower/upper case

Thar0: I've changed all the hex to use uppercase since as I understand it that's our convention for C files also. The discrepancy between hex and dec in shifts is due to <= 9 not being prefixed by 0x.

asm/__osGetFpcCsr.s
```diff
@@ -9,7 +10,8 @@
 
 .balign 16
 
-glabel __osGetFpcCsr
-/* 008680 80007A80 4442F800 */  cfc1  $v0, $31
-/* 008684 80007A84 03E00008 */  jr    $ra
-/* 008688 80007A88 00000000 */   nop   
+LEAF(__osGetFpcCsr)
+    cfc1    $v0, C1_FPCSR
```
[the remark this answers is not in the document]
Thar0: I liked them more too but I think it's for the best that we make full use of the existing sdk defines.

include/ultra64/asm.h
```diff
@@ -0,0 +1,86 @@
...
+#define NARGSAVE 4      // space for 4 args must be allocated
+#define ALSZ    (8-1)
+#define ALMASK ~(8-1)
+#elif (_MIPS_SIM == _ABIN32 || _MIPS_SIM == _ABI64)
+#define NARGSAVE 0      // no caller responsibilities
+#define ALSZ    (16-1)
+#define ALMASK ~(16-1)
+#endif
+
+#define FRAMESZ(size) (((size) + ALSZ) & ALMASK)
```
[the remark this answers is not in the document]
Thar0: These are in the original asm.h (https://github.com/n64decomp/libreultra/blob/master/include/2.0I/sys/asm.h#L184)

include/ultra64/exception.h
```diff
@@ -42,4 +47,15 @@ void __osResetGlobalIntMask(OSHWIntr mask);
 
 extern __osHwInt __osHwIntTable[];
 
+#else
+
+// __osHwInt struct member offsets
+#define HWINTR_CALLBACK 0x00
```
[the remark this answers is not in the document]
Thar0: I can change the define to `HWINT_CALLBACK` if that makes it any clearer? I'm not sure how much I can do to alleviate the issue though since it's all sdk stuff

asm/osWritebackDCache.s
```diff
@@ -9,38 +10,54 @@
...
-/* 005300 80004700 1420FFFD */  bnez  $at, .L800046F8
-/* 005304 80004704 25080010 */   addiu $t0, $t0, 0x10
-.L80004708:
-/* 005308 80004708 03E00008 */  jr    $ra
-/* 00530C 8000470C 00000000 */   nop   
+/**
+ *  void osWritebackDCache(void* vaddr, s32 nbytes);
+ *
+ *  Writes back the contents of the data cache for addresses vaddr
+ *  to vaddr + nbytes to main memory.
```
Dragorn421, line 17: The end address `vaddr + nbytes` is not included I think, maybe rephrase as "cache for nbytes of memory starting at vaddr" idk

asm/osInvalDCache.s
```diff
@@ -8,54 +9,88 @@
...
+ *  Invalidates the CPU Data Cache for `nbytes` at `vaddr`.
+ *  The cache is not automatically synced with physical memory, so cache
+ *  lines must be invalidated to ensure old data is not used in place of
+ *  newly available data supplied by an external agent in a DMA operation.
+ *
+ *  If `vaddr` is not aligned to a cache line boundary, or nbytes is not a
+ *  multiple of the data cache line size (16 bytes) a larger region is
+ *  invalidated.
+ *
+ *  If the amount to invalidate is greater than the data cache size, 8192,
```
Dragorn421, line 25: We have defines to avoid magic values 👺like `DCACHE_SIZE` here
Or the cache line size above

idk does this even matter it's not like there's any way it can change

also did you write these comments or where do they come from? maybe seems somewhat manual-inspired?

also
```suggestion
 *  If the amount to invalidate is at least the data cache size, 8192,
```

asm/osInvalDCache.s
```diff
@@ -8,54 +9,88 @@
...
+    # Hit-Writeback-Invalidate unaligned part
+    cache   (CACH_PD | C_HWBINV), DCACHE_LINESIZE($t1)
+    sltu    $at, $t1, $t0
+    # If that's all there is to do, return early
+    bnez    $at, .ret
+     nop
+    # Invalidate the rest
+1:
+    # Index-Invalidate
+    cache   (CACH_PD | C_HINV), ($t0)
```
Dragorn421, line 76:
```suggestion
    # Hit-Invalidate
    cache   (CACH_PD | C_HINV), ($t0)
```
the cache operations usage here looked like a complete mess at first but actually makes sense

asm/osInvalDCache.s
```diff
@@ -8,54 +9,88 @@
...
+ *
+ *  If `vaddr` is not aligned to a cache line boundary, or nbytes is not a
+ *  multiple of the data cache line size (16 bytes) a larger region is
+ *  invalidated.
+ *
+ *  If the amount to invalidate is greater than the data cache size, 8192,
+ *  the entire data cache is invalidated.
+ */
+LEAF(osInvalDCache)
+    # If the amount to invalidate is less or equal to 0, return immediately
```
Dragorn421, line 29:
```suggestion
    # If the amount to invalidate is less than or equal to 0, return immediately
```

Dragorn421 (changes requested):

asm/bzero.s
```diff
@@ -9,76 +9,65 @@
...
+LEAF(bzero)
+    slti    $at, $a1, 0xC
+    bnez    $at, .bytezero
+     negu   $v1, $a0
+    andi    $v1, $v1, 3
+    beqz    $v1, .blkzero
+     subu   $a1, $a1, $v1
+    swl     $zero, ($a0)
+    addu    $a0, $a0, $v1
 .blkzero:
```
Dragorn421, line 21: not your change but
```suggestion
.blockzero:
```
it took me a very long time to figure out what this blk meant

asm/osSetIntMask.s
```diff
@@ -1,123 +1,131 @@
...
-/* 005BC4 80004FC4 01816024 */  and   $t4, $t4, $at
-/* 005BC8 80004FC8 01886025 */  or    $t4, $t4, $t0
-/* 005BCC 80004FCC 408C6000 */  mtc0  $t4, $12
-/* 005BD0 80004FD0 00000000 */  nop   
-/* 005BD4 80004FD4 00000000 */  nop   
-/* 005BD8 80004FD8 03E00008 */  jr    $ra
-/* 005BDC 80004FDC 00000000 */   nop   
+/**
+ *  LUT to convert between MI_INTR and MI_INTR_MASK.
+ *  MI_INTR is status for each interrupt whereas MI_INTR_MASK has seperate bits for set/clr.
```
Dragorn421, line 16:
```suggestion
 *  MI_INTR is status for each interrupt whereas MI_INTR_MASK has separate bits for set/clr.
```

asm/exceptasm.s
```diff
@@ -69,646 +81,891 @@ glabel __osIntTable
...
+    lw      $s1, %lo(PHYS_TO_K1(MI_INTR_REG))($s1)
+    srl     $t0, $t0, 0x10
+    and     $s1, $s1, $t0
+
+/**
+ *  Signal Processor (SP) Interrupt
+ */
+sp:
+    # Test for sp interrupt
+    andi    $t1, $s1, MI_INTR_SP
```
Dragorn421, line 352: This is the only use of `MI_INTR_SP`/SI/AI/VI/PI/DP
What is the difference with `MI_INTR_MASK_SP`/... ?

If I go by `MI_INTR_SP` being for `MI_INTR_REG`, according to rcp.h comments, stuff with `$s1` should use the `MI_INTR_SP` and co. defines

asm/exceptasm.s
```diff
@@ -69,646 +81,891 @@ glabel __osIntTable
...
+
+/**
+ *  OSThread* __osPopThread(OSThread** threadQ);
+ *
+ *  Pops the highest priority thread from the top of the
+ *   thread queue `threadQ` and returns it
+ */
+LEAF(__osPopThread)
+    lw      $v0, ($a0)
+    lw      $t9, ($v0)
```
Dragorn421, line 837:
```suggestion
    lw      $t9, THREAD_NEXT($v0)
```
I think this is the one :thonk:

asm/osMapTLBRdb.s
```diff
@@ -9,26 +10,27 @@
...
-/* 00872C 80007B2C 40885000 */  mtc0  $t0, $10
-/* 008730 80007B30 03E00008 */  jr    $ra
-/* 008734 80007B34 00000000 */   nop   
+LEAF(osMapTLBRdb)
+    mfc0    $t0, C0_ENTRYHI
+    li      $t1, NTLBENTRIES
+    mtc0    $t1, C0_INX
+    mtc0    $zero, C0_PAGEMASK
+    li      $t2, (TLBLO_UNCACHED | TLBLO_D | TLBLO_V | TLBLO_G)
+    li      $t1, K2BASE
```
Dragorn421, line 19:
```suggestion
    li      $t1, (K2BASE & TLBHI_VPN2MASK)
```
not that it would make an unaligned address instead of `K2BASE` work but for completeness/clarity I guess? 🤷

asm/osMapTLBRdb.s
```diff
@@ -9,26 +10,27 @@
...
+    mtc0    $t1, C0_ENTRYHI
+    li      $t1, K0BASE
```
Dragorn421, line 21: this is the PFN so part of a physical address, does it make sense to use K0BASE for this?

asm/osUnmapTLBAll.s
```diff
@@ -9,22 +10,23 @@
...
-/* 006BEC 80005FEC 2129FFFF */  addi  $t1, $t1, -1
-/* 006BF0 80005FF0 0521FFF9 */  bgez  $t1, .L80005FD8
-/* 006BF4 80005FF4 00000000 */   nop   
-/* 006BF8 80005FF8 40885000 */  mtc0  $t0, $10
-/* 006BFC 80005FFC 03E00008 */  jr    $ra
-/* 006C00 80006000 00000000 */   nop   
+LEAF(osUnmapTLBAll)
+    mfc0    $t0, C0_ENTRYHI
+    li      $t1, (NTLBENTRIES - 1)
+    li      $t2, K0BASE
```
Dragorn421, line 16:
```suggestion
    li      $t2, (K0BASE & TLBHI_VPN2MASK)
```

asm/osUnmapTLBAll.s
```diff
@@ -9,22 +10,23 @@
...
+    mtc0    $zero, C0_ENTRYLO1
+1:
+    mtc0    $t1, C0_INX
+    nop
+    tlbwi
+    nop
+    nop
+    addi    $t1, $t1, -1
+    bgez    $t1, 1b
+     nop
```
Dragorn421, line 28: again not sure what these nops are for and why this isn't
```suggestion
    addi    $t1, $t1, -1
    bgez    $t1, 1b
     nop
```

Dragorn421 (commented):
I'm now a TLB expert 😎 (no)

Sources, which mostly include what you linked in the past (tfw no brand new documentation on a 25-years-old platform):
http://datasheets.chipdb.org/NEC/Vr-Series/Vr43xx/U10504EJ7V0UMJ1.pdf (chapter 5 (5.1, 5.2, 5.4); coprocessor 0 hazards (pages 620-621); tlbp, tlbr, tlwi (pages 532-534))
https://en.wikipedia.org/wiki/Memory_management_unit
https://n64.readthedocs.io/index.html#memory-map
http://n64devkit.square7.ch/pro-man/pro10/10-06.htm
http://n64devkit.square7.ch/n64man/os/osMapTLB.htm
https://ultra64.ca/files/tools/DETAILED_N64_MEMORY_MAP.txt
http://n64devkit.square7.ch/header/R4300.htm

asm/osMapTLBRdb.s
```diff
@@ -9,26 +10,27 @@
...
+    mtc0    $t3, C0_ENTRYLO0
+    li      $t1, TLBLO_G
+    mtc0    $t1, C0_ENTRYLO1
+    nop
+    tlbwi
+    nop
+    nop
+    nop
+    nop
+    mtc0    $t0, C0_ENTRYHI
```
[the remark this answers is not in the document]
Thar0: You were indeed looking at the right thing. For `tlbwi` and `tlbwr` the destination hazard is "TLB". That is, any instruction using the TLB will be subject to hazards following either of these two instructions (i.e. all load/store instructions use the TLB, see page 101 of the manual for a dissection of the pipeline operation of `lw`). The destination hazard is 8 PClock cycles, so using the calculation in the manual you would need 7 instructions in the worst case, and the worst case should be taken as we do not know what awaits after the function returns. The 4 extra nops are to make it so that the first instruction after the return is the 8th instruction, so there is no hazard. Similar should apply to unmap, although I believe there is one extra nop there that this doesn't explain.

asm/osMapTLBRdb.s
Thar0: I believe `K0BASE` is the most appropriate define here. It has to be something that results in `lui $t1, 0x8000`, it's not totally clear to me why it's written this way at all though, and what the consequences are if any.

Makefile
```diff
@@ -23,7 +27,7 @@ endif
 
 ifeq ($(COMPILER),gcc)
   CFLAGS += -DCOMPILER_GCC
-  CPPFLAGS += -DCOMPILER_GCC
+  CPPFLAGS += -DCOMPILER_GCC -DAVOID_UB
```
Dragorn421, line 30: I may be wrong but I don't think `CPPFLAGS` are passed to gcc, or that cpp output is passed to gcc, anywhere, so this looks like it would cause `AVOID_UB` to not be defined in gcc builds except for assembly (which uses cpp)

asm/__osProbeTLB.s
```diff
@@ -9,54 +10,79 @@
...
+    srl     $t3, $t3, 1
+    # & with vaddr
+    and     $t4, $t3, $a0
+    # Select C0_ENTRYLO0 or C0_ENTRYLO1
+    bnez    $t4, 1f
+     addi   $t3, $t3, -1 # make bitmask out of page size
+    mfc0    $v0, C0_ENTRYLO0
+    b       2f
+     nop
+1:
```
Dragorn421, line 62: why did you revert the label names? I thought get_lo1 / got_lo0 / ... were nice

asm/bzero.s
```diff
@@ -9,76 +9,59 @@
...
+LEAF(bzero)
+    slti    $at, $a1, 0xC
+    bnez    $at, bytezero
+     negu   $v1, $a0
+    andi    $v1, $v1, 3
+    beqz    $v1, blkzero
+     subu   $a1, $a1, $v1
+    swl     $zero, ($a0)
+    addu    $a0, $a0, $v1
+blkzero:
```
Dragorn421, line 21:
```suggestion
blockzero:
```

asm/guMtxL2F.s
```diff
@@ -7,33 +7,34 @@
...
-/* B7A194 80102FF4 E488FFF8 */  swc1  $f8, -8($a0)
-/* B7A198 80102FF8 14B8FFED */  bne   $a1, $t8, .L80102FB0
-/* B7A19C 80102FFC E492FFFC */   swc1  $f18, -4($a0)
-/* B7A1A0 80103000 03E00008 */  jr    $ra
-/* B7A1A4 80103004 00000000 */   nop   
+LEAF(guMtxL2F)
+    li      $at, 0x37800000 # 1.0f / 65536.0f
+    mtc1    $at, $f0
+    li      $t9, 0xFFFF0000
+    addiu   $t8, $a1, 0x40/2 # sizeof(mtx) / 2
```
Dragorn421, line 16: does it really matter that it's half of the matrix?
imho a `MTX_FRAC_PART 0x20` define would be better if you do want to highlight the "split" between int and frac parts

asm/parameters.s
```diff
@@ -0,0 +1,22 @@
...
+.section .text
+
+.macro IPL_SYMBOL name, address, size
+    .global \name
+    .set \name, \address
+    .type \name, @object
+    .size \name, \size
+.endm
+
+IPL_SYMBOL leoBootID,       0x800001a0, 4
```
Dragorn421, line 12: why move this in a new file? and why `parameters.s`

asm/interrupt.s
```diff
@@ -0,0 +1,53 @@
...
+    and     $t2, $t2, $t0
+    li      $at, ~SR_IMASK
+    and     $t1, $t1, $at
+    or      $t1, $t1, $t2
+    li      $at, ~SR_IE
+    and     $t1, $t1, $at
+    mtc0    $t1, C0_SR
+    nop
+    nop
+No_Change_Global_Int:
```
Dragorn421, line 40: not that I dislike this casing or style but most other labels are lowercase_snake_case
apart from exceptasm.s which may just need more passes labels should probably be consistent

Dragorn421 (commented):
my review progress again :crying_cat_face:

asm/__osProbeTLB.s
Thar0: We have the original label names in mdebug, so I changed all the labels to original names

asm/parameters.s
Thar0: This is where they are defined in libultra. I created this file in order to house the 0x60 bytes of 0 that used to be in bzero. With this, all the extra padding that was in the handwritten asm is accounted for properly.

asm/__osProbeTLB.s
Dragorn421: mdebug from which file again?

I understand keeping the sdk names for functions and structures and such,
but I don't see why we would stick to the original labels, as they don't matter outside the function.
To me they are closer to comments, to help understand the code

the original labels could be documented like debug-strings-leaked names in the source
like `get_lo1: # original name: "1"` or something

Thar0: mdebug from files inside `libultra_d.a`.
I'd rather have comments describing what's going on than relying on the labels, it makes it easier if we ever want to use ultralib for libultra.

Dragorn421: why does it make the switch easier?

Thar0: It makes it easier in that if we documented it now via the label names, those would need moving to comments anyway. Although I suppose if we switch we'd need to PR better documentation over there anyway since they're somewhat lacking compared to what I've done here, I still think that would be easier if we just used comments to document in the first place.

asm/__osProbeTLB.s
```diff
@@ -9,54 +10,78 @@
...
+     nop
+    # Read TLB, sets C0_ENTRYHI, C0_ENTRYLO0, C0_ENTRYLO1 and C0_PAGEMASK for the TLB
+    # entry indicated by the Index register
+    tlbr
+    nop
+    nop
+    nop
+    # Get C0_PAGEMASK
+    mfc0    $t3, C0_PAGEMASK
+    addi    $t3, $t3, DCACHE_SIZE
```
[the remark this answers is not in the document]
Thar0: Sorry missed these after making the change.
There's no define that really works here I don't think, so I just reverted it to 0x2000 and added a comment that it calculates the page size.
It's possible that the vr4300 in the N64 only supported those three page sizes, rather than the full range, but I'm not sure.
`OS_PM_16K` is part of `os_tlb.h`, an sdk header we do not have since we never needed it. The function `osMapTLB` is simply not present in the ROM.

Makefile
glankk: `CPPFLAGS` is traditionally meant to be passed to all translation programs that preprocess their input with cpp, e.g. translation programs for C, C++, Fortran, and preprocessed assembler. For example, the default C and C++ recipes for make are `$(CC) $(CPPFLAGS) $(CFLAGS) -c` and `$(CXX) $(CPPFLAGS) $(CXXFLAGS) -c` respectively.

Dragorn421: that makes sense
(but it's not what oot's makefile does afaict :smile: )

asm/exceptasm.s
```diff
@@ -69,646 +81,891 @@ glabel __osIntTable
...
+    lw      $t3, MQ_VALIDCOUNT($t1)
+    lw      $t4, MQ_MSGCOUNT($t1)
+    slt     $at, $t3, $t4
+    beqz    $at, send_done
+     nop
+    # Add validcount to first and modulo with msgcount
+    lw      $t5, MQ_FIRST($t1)
+    addu    $t5, $t5, $t3
+    # Modulo
+    div     $zero, $t5, $t4
```
Dragorn421, line 650: how does this `div` work, the various references I'm looking at show it used as `div a, b` and accomplishing `a = b * lo + hi`

asm/exceptasm.s
```diff
@@ -69,646 +81,891 @@ glabel __osIntTable
...
+    move    $a1, $k0
+    jal     __osEnqueueThread
+     addiu  $a0, $a0, %lo(__osRunQueue)
+    j       __osDispatchThread
+     nop
 
+/**
+ *  Dispatches the top of the run queue
+ */
 enqueueRunning:
```
Dragorn421, line 596: and if I understand it right it also shifts the "current thread"/saved interrupted thread context (k0) at the top of the run queue before dispatching, so really it's just resuming the thread isn't it?

(to be clear the comment should probably mention how this also changes the run queue and that the dispatched thread is the "current" one)

asm/osMapTLBRdb.s
Thar0: What I mean is it seems like the **most** appropriate define, not that it seems to make a lot of sense.
I'm not sure if it's a bug or not or exactly the behaviour, my guess is it will translate virtual addresses to 0x80000000 addresses and output them over the SysAD bus to the RCP, which will try to find a device at that physical address. Maybe this worked on n64s that had the additional components needed to use the libultra rdb, but I don't know.

I took another look through the headers, specifically `rdb.h` which we don't have in the repo currently, and it has some defines:
```
/* U64 side address */
#define	RDB_BASE_REG		0xc0000000
#define RDB_WRITE_INTR_REG	(RDB_BASE_REG + 0x8)
#define RDB_READ_INTR_REG	(RDB_BASE_REG + 0xc)
#define RDB_BASE_VIRTUAL_ADDR	0x80000000
```

I guess since this produces the tlb mapping for the rdb, the `K2BASE` and `K0BASE` should really actually be `RDB_BASE_REG` and `RDB_BASE_VIRTUAL_ADDR`.

asm/exceptasm.s
Thar0: Here it's used to calculate the remainder rather than the quotient which is more common, lower down it does `mfhi` rather than `mflo`. So your source is correct `a = b * lo + hi` is what `div` does, `lo` and `hi` are of course the quotient and remainder and are extracted via `mflo` and `mfhi` respectively.

asm/osSetIntMask.s
```diff
@@ -1,123 +1,132 @@
...
+    lui     $t2, %hi(__osRcpImTable)
+    addu    $t2, $t2, $t0
+    lhu     $t2, %lo(__osRcpImTable)($t2)
+    lui     $at, %hi(PHYS_TO_K1(MI_INTR_MASK_REG))
+    sw      $t2, %lo(PHYS_TO_K1(MI_INTR_MASK_REG))($at)
+    andi    $t0, $a0, OS_IM_CPU
+    andi    $t1, $t3, SR_IMASK
+    and     $t0, $t0, $t1
+    lui     $at, ((~SR_IMASK) >> 0x10) & 0xFFFF
+    ori     $at, (~SR_IMASK) & 0xFFFF
```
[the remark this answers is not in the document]
Thar0: I must have left this one since the others like this in exceptasm didn't match, but I've changed this one now.

fig02:
can we resolve conversations that have been addressed just to make this pr a little easier to navigate

asm/exceptasm.s
Dragorn421: but why is it `div     $zero, $t5, $t4` instead of idk maybe `div     $t5, $t4`

Thar0: Oh right - This is just how it was originally disassembled, for some reason the gnu assembler accepts 3 args to div, I don't think the first really does anything. I can see if removing it still works since it should.

Thar0: nvm, I found out both why it was disassembled this way and what it does. It's a gnu assembler thing for preventing generating div0 and overflow checks. As you can see around this code, there are checks for div0 and overflow, these were automatically generated by the sgi assembler. The gnu assembler actually does something similar (but it does not match), but if you add the leading `$zero` argument it prevents the assembler from adding the div0 and overflow checks.
So it has to stay this way as long as we use the gnu assembler.

asm/exceptasm.s
```diff
@@ -69,646 +81,893 @@ glabel __osIntTable
...
+    beqz    $t1, vi
+     nop
+    # Test for yielded or done signals in particular
+    lui     $t4, %hi(PHYS_TO_K1(SP_STATUS_REG))
+    lw      $t4, %lo(PHYS_TO_K1(SP_STATUS_REG))($t4)
+    li      $t1, (SP_CLR_INTR | SP_CLR_SIG3)
+    lui     $at, %hi(PHYS_TO_K1(SP_STATUS_REG))
+    andi    $t4, $t4, (SP_STATUS_YIELDED | SP_STATUS_TASKDONE)
+    # Mask out SP interrupt
+    andi    $s1, $s1, (MI_INTR_MASK_SI | MI_INTR_MASK_AI | MI_INTR_MASK_VI | MI_INTR_MASK_PI | MI_INTR_MASK_DP)
```
Dragorn421, line 362: since s1 comes from `MI_INTR_REG` and not `MI_INTR_MASK_REG` shouldn't it use the `MI_INTR_*` defines?
```suggestion
    andi    $s1, $s1, (MI_INTR_SI | MI_INTR_AI | MI_INTR_VI | MI_INTR_PI | MI_INTR_DP)
```

asm/exceptasm.s
```diff
@@ -69,646 +81,893 @@ glabel __osIntTable
...
+    beqz    $at, send_done
+     nop
+    # Add validcount to first and modulo with msgcount
+    lw      $t5, MQ_FIRST($t1)
+    addu    $t5, $t5, $t3
+    # Modulo
+    div     $zero, $t5, $t4
+    bnez    $t4, 1f
+     nop
+    break   7 # div0
```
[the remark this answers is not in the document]
Thar0: The "indy dump" is the commonly used name of a dump of the irix filesystem from an SGI Indy machine, which contains all the standard irix headers and libraries, as well as an installation of the n64 development kit. It's useful for cases like this where MIPS standard headers need referencing.

asm/exceptasm.s
```diff
@@ -69,646 +81,897 @@ glabel __osIntTable
...
-/* 004A78 80003E78 2401FEFF */  li    $at, -257
-/* 004A7C 80003E7C 01014024 */  and   $t0, $t0, $at
-/* 004A80 80003E80 40886800 */  mtc0  $t0, $13
-/* 004A84 80003E84 0C000FCB */  jal   send_mesg
-/* 004A88 80003E88 24040000 */   li    $a0, 0
-/* 004A8C 80003E8C 2401FEFF */  li    $at, -257
-/* 004A90 80003E90 1000FF53 */  b     next_interrupt
-/* 004A94 80003E94 02018024 */   and   $s0, $s0, $at
+    li      $at, ~CAUSE_SW1
+    # Clear interrupt
```
Dragorn421, line 553:
```suggestion
    # Clear interrupt
    li      $at, ~CAUSE_SW1
```
(same for sw2)


Also, is it guaranteed that t0 still hold Cause at this point? it's set to that in `handle_interrupt` but `rcp` also uses t0, and if both bits are set at the same time rcp runs before sw1 and sw2

So can two bits of the Cause interrupt pending mask be set at the same time?
The answer has to be yes since exceptasm accounts for that scenario (by branching back to `next_interrupt`)...

asm/osMapTLBRdb.s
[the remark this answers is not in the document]
Dragorn421: To get 7,
you get the 8 from the TLB Destination Hazard associated to tlbwi
and subtract (0 + 1)
and I'm not sure where does the 0 come from, the lowest source TLB-related is "Instruction fetch" and it's only at 2? what is the worse case than this

asm/exceptasm.s
```diff
@@ -69,646 +81,897 @@ glabel __osIntTable
...
+    lw      $t2, ($t1)
+    # Mask out pi interrupt
+    andi    $s1, $s1, (MI_INTR_SP | MI_INTR_SI | MI_INTR_AI | MI_INTR_VI | MI_INTR_DP)
+    # Skip callback if NULL
+    beqz    $t2, no_pi_callback
+     nop
+    # Set up a stack and run the callback
+    lw      $sp, HWINT_SP($t1)
+    jalr    $t2
+     move   $a0, $v0
```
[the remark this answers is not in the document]
Thar0: It isn't. It's likely some leftover instruction from more removed code but yeah idk

asm/exceptasm.s
Thar0: Yeah this is definitely a bug.. however the chances of both interrupts happening in the timeframe needed to trigger it is small I suppose, and also the software interrupts are (almost?) never asserted anyway in any code I've ever looked at

asm/interrupt.s
```diff
@@ -0,0 +1,53 @@
...
+    li      $at, ~SR_IE
+    and     $t1, $t0, $at
+    mtc0    $t1, C0_SR
+    andi    $v0, $t0, SR_IE
+    lw      $t0, ($t2)
+    andi    $t0, $t0, SR_IMASK
+    beq     $t0, $t3, No_Change_Global_Int
+     lui    $t2, %hi(__osRunningThread)
+    addiu   $t2, $t2, %lo(__osRunningThread)
+    lw      $t1, THREAD_SR($t2)
```
Dragorn421, line 29: :thinking: here,
`t2 = &__osRunningThread` is a `OSThread**`
so `lw $t1, THREAD_SR($t2)` does
`(OSThread*&__osRunningThread)->context.sr`
which doesn't make much sense to me
I would expect `__osRunningThread->context.sr`

`(OSThread*&__osRunningThread)->context.sr` is inside `osViModeMpalLan1`:
```
$ ./sym_info.py __osRunningThread
Symbol __osRunningThread (RAM: 0x8000AD50, ROM: 0x00B950, build/src/libultra/os/createthread.o)
```
`0x8000AD50 + THREAD_SR = 0x8000AE68`
```
$ ./sym_info.py 0x8000AE68
0x8000AE68 is at 0x18 bytes inside osViModeMpalLan1 (RAM 0x8000AE50, ROM 0xBA50, build/src/libultra/io/vimodempallan1.o)
```

is this stuff totally broken and if so how does it even work or what am I missing :cold_sweat:

Dragorn421 (commented):
The deeper I go the more doubts I have on libultra's os implementation

I would revert most comments related to `__OSGlobalIntMask` for now, at the very least it's mostly wrong to say it's used for masking the interrupt bits (afaict it only does so in the `rcp` label of exceptasm, and at the end of `__osDispatchThread`)

asm/interrupt.s
Thar0: Yes this is another bug, the `addiu` should definitely be `lw` instead.
That said, I'm reasonably sure the branch is always taken, so the broken code will never run. The branch is essentially `(__OSGlobalIntMask & SR_IMASK) == (__OSGlobalIntMask & SR_IMASK)`
The only way this could ever fail is if some interrupt occurred and changed `__OSGlobalIntMask`, however interrupts are disabled pretty quickly so I don't think such great timing would occur usually. I'm not really sure what their goal was with this code, it's actually missing in earlier libultra versions.

asm/exceptasm.s
```diff
@@ -69,646 +81,897 @@ glabel __osIntTable
...
+    sw      $k1, THREAD_FPCSR($a1)
+1:
+    lw      $k1, THREAD_SR($a1)
+    andi    $t1, $k1, SR_IMASK
+    beqz    $t1, 2f
+     nop
+    # If any CPU interrupts are enabled in SR, mask previous thread's SR
+    # with the CPU bits of the global interrupt mask
+    lui     $t0, %hi(__OSGlobalIntMask)
+    addiu   $t0, %lo(__OSGlobalIntMask)
```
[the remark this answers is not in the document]
Thar0: I've been avoiding using instruction macros that expand to multiple instructions since they can't be used consistently. If a working `as0` recomp ever happens I'd like to redo the asm files to have this sort of thing since we'd be able to do it properly there.

asm/exceptasm.s
```diff
@@ -69,646 +80,905 @@ glabel __osIntTable
...
+    sd      $sp, THREAD_SP($k0)
+    sd      $fp, THREAD_S8($k0)
+    sd      $ra, THREAD_RA($k0)
+    beqz    $t1, savercp
+     sd     $t0, THREAD_HI($k0)
+    # If any CPU interrupts are enabled in the previous thread's SR, bitwise-OR in the
+    # disabled CPU interrupts from the global interrupt mask. The COP0 SR that is
+    # saved here is really the thread's SR masked by the global interrupt mask, so this
+    # is an attempt to recover the thread's SR from the current SR. This is however
+    # broken, see comments for osSetIntMask.
```
Dragorn421, line 166: > The COP0 SR that is saved here is really the thread's SR masked by the global interrupt mask
the "that is saved here" feels wrong, since what is actually saved has the `|`d bits

also I don't think the distinction between the "COP0 SR" (Sr the thread wants & global mask, I assume) and the "thread's SR" (Sr the thread wants, I assume) would make sense for another reader

maybe both these remarks can be addressed by providing less details and leaving it up to `osSetIntMask` to explain:

```suggestion
    # disabled CPU interrupts from the global interrupt mask.
    # This is an attempt at reverting the effect of bitwise-AND-ing the SR
    # with the global interrupt mask.
    # This is however broken, see comments for osSetIntMask.
```

asm/exceptasm.s
```diff
@@ -69,646 +80,905 @@ glabel __osIntTable
...
-/* 004734 80003B34 3108003F */  andi  $t0, $t0, 0x3f
-/* 004738 80003B38 010C4024 */  and   $t0, $t0, $t4
-/* 00473C 80003B3C 01284825 */  or    $t1, $t1, $t0
+    # Save the currently masked RCP interrupts.
+    lui     $t1, %hi(PHYS_TO_K1(MI_INTR_MASK_REG))
+    lw      $t1, %lo(PHYS_TO_K1(MI_INTR_MASK_REG))($t1)
+    beqz    $t1, endrcp
+     nop
+    # Similar to the above comment, but for RCP interrupt enable bits rather than CPU.
+    # This suffers from the problem as above.
```
Dragorn421, line 190:
```suggestion
    # This suffers from the same problem as above.
```

asm/osMapTLBRdb.s
Dragorn421: then since we don't have `RDB_BASE_REG` and `RDB_BASE_VIRTUAL_ADDR`, I think at least `RDB_BASE_VIRTUAL_ADDR` should just be raw hex, using a virtual address as a physical address just doesn't make sense

asm/osSetIntMask.s
```diff
@@ -1,123 +1,170 @@
...
+    xor     $t0, $t3, $at
+    andi    $t0, $t0, SR_IMASK
+    or      $v0, $v0, $t0
+    # Fetch MI_INTR_MASK_REG
+    lui     $t2, %hi(PHYS_TO_K1(MI_INTR_MASK_REG))
+    lw      $t2, %lo(PHYS_TO_K1(MI_INTR_MASK_REG))($t2)
+    # If there are RCP interrupts masked
+    beqz    $t2, 1f
+     srl    $t1, $t3, RCP_IMASKSHIFT
+    # BItwise-OR in the disabled RCP bits of __OSGlobalIntMask
```
Dragorn421, line 136:
```suggestion
    # Bitwise-OR in the disabled RCP bits of __OSGlobalIntMask
```

Dragorn421 (commented):
minor things

Roman971 (commented):
Looks mostly good to me, I just have a few comments/suggestions

Makefile
```diff
@@ -112,10 +116,10 @@ else
   OPTFLAGS := -O2
 endif
 
-ASFLAGS := -march=vr4300 -32 -Iinclude
+ASFLAGS := -march=vr4300 -32 -no-pad-sections -I include
```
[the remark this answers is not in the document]
Thar0: The handwritten `gu` functions are aligned to 32 (icache line), and without this flag `as` will pad with too many nops after the function.

include/ultra64/asm.h
```diff
@@ -0,0 +1,85 @@
...
+    x:                         ;\
+        .ent x, 0              ;\
+        .frame $sp, fsize, ra
+
+#define XNESTED(x)              \
+    .globl x                   ;\
+    x:                         ;\
+        .aent x, 0
+
+#define END(x)                  \
```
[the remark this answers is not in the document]
Thar0: It is somewhat unfortunate but I think this is overall better. Without these the situation is really quite bad, you can't objdump the asm files properly which is now more worthwhile to do occasionally since they're now less literal.

include/ultra64/asm.h
```diff
@@ -0,0 +1,85 @@
...
+#ifdef __sgi
+#define _MIPS_ISA_MIPS1 1
+#define _MIPS_ISA_MIPS2 2
+#define _MIPS_ISA_MIPS3 3
+#define _MIPS_ISA_MIPS4 4
+#endif
+
+#ifndef _LANGUAGE_C
+
+#define LEAF(x)                 \
```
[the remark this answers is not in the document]
Thar0: Yeah we either use `LEAF` for everything as was presumably the case originally or add another macro that is just `LEAF` with a different name

src/libultra/io/pirawdma.c
```diff
@@ -1,10 +1,10 @@
 #include "global.h"
 
 s32 __osPiRawStartDma(s32 dir, u32 cartAddr, void* dramAddr, size_t size) {
-    register s32 status = HW_REG(PI_STATUS_REG, u32);
+    s32 status;
 
-    while (status & (PI_STATUS_BUSY | PI_STATUS_IOBUSY)) {
-        status = HW_REG(PI_STATUS_REG, u32);
+    while (status = HW_REG(PI_STATUS_REG, u32), status & (PI_STATUS_BUSY | PI_STATUS_IOBUSY)) {
+        ;
```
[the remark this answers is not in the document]
Roman971: I see, thanks. I would say having none of them do this is better (if it matches), but it's probably acceptable either way

include/boot.h
```diff
@@ -0,0 +1,6 @@
+#ifndef BOOT_H
+#define BOOT_H
+
+#define BOOT_STACK_SIZE 0x400
```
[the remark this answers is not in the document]
Roman971: Oh I didn't notice this was also needed in the entrypoint, that makes sense. In that case it's probably fine, tho I do wonder if the other stack sizes should be moved to headers eventually, maybe somewhere near the thread ids and priorities.

Makefile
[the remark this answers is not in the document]
Roman971: I see, I didn't know it would add more padding because of that.

An alternative would be to add this flag specifically for `gu` files (which I expect would be in a dedicated folder eventually), but I don't know if padding sections is really important anyway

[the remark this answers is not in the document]
Thar0: After this PR is over with I'd like to make another that moves them into src/ folders, the original libultra directory structure (visible in mdebug) has the handwritten gu functions in a folder called mgu, so we could restrict it at that point.
I don't believe this option should affect anything at least in the short term, maybe when we need GLOBAL_ASM again but by then we probably will need to sort out a bunch of other stuff too.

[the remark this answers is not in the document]
Roman971: I was more thinking along the lines of modding or portability for situations where padding could matter, but I don't know if it's worth worrying about since this is handwritten asm and not really subject to change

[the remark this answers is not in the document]
Thar0: Ah for modding I'm pretty sure it shouldn't matter at all. The linker should be able to resolve any problems that might arise with respect to alignment or padding

fig02 (approved):

Dragorn421 (approved):
👏 thanks for teaching me a lot about this low level stuff and providing all those links

Roman971 (approved):

PR #870, opened 2021-07-14
Only use the asm processor where it is needed

louist103:
These files dont have any ASM in them nor do they have any strings that needed to be converted to `EUC-JP`.  This saves about 10 seconds on average during a build.
I will make a tool that organizes the files into safe to gives directly to the compiler and now.  Opening this as a draft for now to get a conversion on it.


AngheloAlf (commented):
What is the `saveOverlays.mk` file?

louist103:
> 
> 
> What is the `saveOverlays.mk` file?

its going to hold the list of files that dont need to be run through the processor.

Dragorn421 (commented):
Saves 10-20 seconds on my machine (fully builds in ~2 minutes, where master builds in ~2 minutes and 15 seconds)


If I understand correclty what asm-processor does, another route for a "speed fix" could be to add early in `tools/asm_processor/build.py` (right after `in_file = compile_args[-1]`) something like
```py
with open(in_file) as f:
    has_GLOBAL_ASM = 'GLOBAL_ASM' in f.read()
if not has_GLOBAL_ASM:
    subprocess.check_call(compiler + compile_args)
    sys.exit(0)
```
(and delay imports for performance)
So I tried that to compare performance, and it didn't even produce an OK rom (that not-ok rom seems to run fine though). I'm not sure what went wrong. Just discard this idea if you wish I guess

Makefile
```diff
@@ -142,20 +142,20 @@ build/src/code/code_801067F0.o: OPTFLAGS := -g
...
-build/src/code/jpegutils.o: CC := python3 tools/asm_processor/build.py $(CC_OLD) -- $(AS) $(ASFLAGS) --
-build/src/code/jpegdecoder.o: CC := python3 tools/asm_processor/build.py $(CC_OLD) -- $(AS) $(ASFLAGS) --
+build/src/code/jpegutils.o: CC := $(CC_OLD)
+build/src/code/jpegdecoder.o: CC := $(CC_OLD)
 
 build/src/boot/%.o: CC := python3 tools/asm_processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
 build/src/code/%.o: CC := python3 tools/asm_processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
 build/src/overlays/actors/%.o: CC := python3 tools/asm_processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
 build/src/overlays/effects/%.o: CC := python3 tools/asm_processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
 build/src/overlays/gamestates/%.o: CC := python3 tools/asm_processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
```
Dragorn421, line 156: while you are at it, wouldn't it be nice to put `python3 tools/asm_processor/build.py $(CC) -- $(AS) $(ASFLAGS) --` into a variable like `ASM_PROCESSOR` (or `CC_ASM_PRE` idk),
and replace stuff with `CC := $(ASM_PROCESSOR)` to avoid duplicating the same command over several lines

safeCode.mk
```diff
@@ -0,0 +1,141 @@
+#Code filessafe to build without the ASM processor.
+#Dont change this unless you know what you are doing.
```
Dragorn421, line 2:
```suggestion
# Code filessafe to build without the ASM processor.
# Dont change this unless you know what you are doing.
```
it's me the space enthusiast

safeCode.mk
```diff
@@ -0,0 +1,141 @@
...
+
+SAFE_OVERLAY_FILES :=	 \
```
Dragorn421, line 4:
```suggestion
SAFE_CODE_FILES :=	 \
```
? (this is safeCode.mk)

Dragorn421 (commented):
force-pushing reverted removal of the two spaces that bothered only me (see my previous comment)

Why is safeCode in gitignore but not safeOverlays,
why is safeOverlays included in the Makefile but not safeCode?
(also why split those in two files anyway?)

I know this is a draft so I'd expect WIP things but the commit message says `works OK for now` so I do have some expectations 😅

louist103:
1.  Neither should be ignored so IDK what the deal with that is.
2. Because I haven't tested it yet.
3. Because its split already and I haven't felt like changing it yet.
and works OK right now means its only OK.  Not to be confused with :OK: as in all good.  I still have many improvements to make.

ethteck:
Just a thought: it might be better to list the files which *do* need asm-processor, rather than ones that don't. Unless I am forgetting something, only files with INCLUDE_ASMs need it, and those files will dwindle over time, right?

fig02:
Think this kind of change falls in line with our talks about cutscene floats, and how we want to handle preprocessing stuff in general. Can we close this and talk a bit more about how we want to approach this?

PR #874, opened 2021-07-19
ovl_Boss_Ganon2 (2 nonmatching)

Sirius902:
Finally done with Ganon! Morita really tried his best to give me a headache with this one.

As far as the nonmatchings go, I tested them both in game and they work fine (most likely equivalent). The first one, `func_808FD5F4`, is definitely equivalent but does have some dubious stuff I put in there to make it a closer match (without it some stuff isn't stored on the stack when it should be) so let me know if I should remove those (they should have comments).

I did try to get rid of as many obvious undefined syms as I could and replaced them with symbols from actual object files, leaving the ones that I'm not sure where they come from or if the symbol isn't in the asset xmls yet. Let me know if there's anything else I should do or if I've done anything wrong regarding that.


src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+    this->unk_398++;
+    this->unk_339 = 0x14;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Boss_Ganon2/func_80901020.s")
+    switch (this->unk_39C) {
+        case 0:
+            func_80064520(globalCtx, &globalCtx->csCtx);
+            this->unk_39E = Gameplay_CreateSubCamera(globalCtx);
+            Gameplay_ChangeCameraStatus(globalCtx, 0, 1);
+            Gameplay_ChangeCameraStatus(globalCtx, this->unk_39E, 7);
```
Dragorn421, line 1651:
```suggestion
            Gameplay_ChangeCameraStatus(globalCtx, MAIN_CAM, CAM_STAT_WAIT);
            Gameplay_ChangeCameraStatus(globalCtx, this->unk_39E, CAM_STAT_ACTIVE);
```
same in other places

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+            this->unk_3B0.x = sZelda->actor.world.pos.x;
+            this->unk_3B0.y = sZelda->actor.world.pos.y + 30.0f;
+            this->unk_3B0.z = sZelda->actor.world.pos.z - 10.0f;
+            Math_ApproachZeroF(&this->unk_324, 1.0f, 5.0f);
+            Math_ApproachF(&globalCtx->envCtx.unk_D8, 1.0f, 1.0f, 0.019607844f);
+            if (this->unk_1A2[1] == 0x50) {
+                func_8010B680(globalCtx, 0x70D7, NULL);
+            }
+            if ((this->unk_1A2[1] < 0x1E) && (func_8010BDBC(&globalCtx->msgCtx) == 0)) {
+                temp_v0 = Gameplay_GetCamera(globalCtx, 0);
```
Dragorn421, line 1693:
```suggestion
                temp_v0 = Gameplay_GetCamera(globalCtx, MAIN_CAM);
```
also in other places

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+                (this->unk_398 == 0x2D)) {
+                func_80078884(NA_SE_EN_MGANON_DAMAGE);
+                func_80078884(NA_SE_IT_SHIELD_BOUND);
+                globalCtx->envCtx.unk_D8 = 1.0f;
+            }
+            if ((this->unk_398 == 0x16) || (this->unk_398 == 0x23) || (this->unk_398 == 0x48) ||
+                (this->unk_398 == 0x2D)) {
+                func_8090109C(this, globalCtx);
+            }
+            if ((this->unk_398 >= 0x22U) && (this->unk_398 < 0x28U)) {
```
Dragorn421, line 2021:
```suggestion
            if ((this->unk_398 >= 0x22) && (this->unk_398 < 0x28)) {
```
I assume they aren't required?

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+                this->unk_3A4.x = 269.0f;
+                this->unk_3A4.y = 1112.0f;
+                this->unk_3A4.z = -28.0f;
+                this->unk_3B0.x = 234.0f;
+                this->unk_3B0.y = 1117.0f;
+                this->unk_3B0.z = -11.0f;
+            } else {
+                if (this->unk_398 < 0x1EU) {
+                    phi_a1 = 0;
+                } else if (this->unk_398 < 0x2BU) {
```
Dragorn421, line 2031:
```suggestion
                if (this->unk_398 < 0x1E) {
                    phi_a1 = 0;
                } else if (this->unk_398 < 0x2B) {
```

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+                    phi_a1 = 2;
+                }
+                this->unk_3A4.x = D_8090702C[phi_a1].x + (player->actor.world.pos.x - 50.0f);
+                this->unk_3A4.y = D_8090702C[phi_a1].y + (player->actor.world.pos.y + 50.0f);
+                this->unk_3A4.z = D_8090702C[phi_a1].z + (player->actor.world.pos.z + 40.0f);
+                this->unk_3B0.x = D_80907050[phi_a1].x + player->actor.world.pos.x;
+                this->unk_3B0.y = D_80907050[phi_a1].y + (player->actor.world.pos.y + 40.0f);
+                this->unk_3B0.z = D_80907050[phi_a1].z + player->actor.world.pos.z;
+            }
+            if (this->unk_398 >= 0x51U) {
```
Dragorn421, line 2046:
```suggestion
            if (this->unk_398 >= 0x51) {
```

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+                this->unk_398 = 0;
+                this->unk_194 = 1000.0f;
+            }
+            this->unk_3A4.x = 250;
+            this->unk_3A4.y = 1150.0f;
+            this->unk_3A4.z = 0.0f;
+            this->unk_3B0.x = this->unk_1B8.x;
+            this->unk_3B0.y = this->unk_1B8.y;
+            this->unk_3B0.z = this->unk_1B8.z;
+            if ((this->unk_398 < 0x3E8) && ((this->unk_398 & 0xF) == 0)) {
```
Dragorn421, line 2098: % 16 maybe?

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+void func_80902348(BossGanon2* this, GlobalContext* globalCtx) {
+    Player* player;
+    f32 temp_f2;
+    f32 temp_f12;
+    s16 i;
+    s16 j;
+    s16 phi_v0_2;
+
+    if (this->unk_316 == 0) {
+        for (i = 0; i < 2; i++) {
```
Dragorn421, line 2193:
```suggestion
        for (i = 0; i < ARRAY_COUNT(this->unk_864); i++) {
```
`unk_864` is the elements for `unk_444`

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+        }
+    }
+
+    if (this->unk_324 > 0.0f) {
+        player = PLAYER;
+        temp_f2 = -200.0f - player->actor.world.pos.x;
+        temp_f12 = -200.0f - player->actor.world.pos.z;
+
+        if (sqrtf(SQ(temp_f2) + SQ(temp_f12)) > 784.0f) {
+            for (j = 0; j < 18; j++) {
```
Dragorn421, line 2219:
```suggestion
            for (j = 0; j < ARRAY_COUNT(player->flameTimers); j++) {
```

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+
+void func_80902524(BossGanon2* this, GlobalContext* globalCtx) {
+    s8 temp_v0_4;
+    ColliderInfo* acHitInfo;
+    s16 i;
+    u8 phi_v1_2;
+
+    osSyncPrintf("this->no_hit_time %d\n", this->unk_316);
+    if (this->unk_316 != 0 || ((this->unk_334 == 0) && (this->actionFunc == func_80900890))) {
+        for (i = 0; i < 16; i++) {
```
Dragorn421, line 2238:
```suggestion
        for (i = 0; i < ARRAY_COUNT(this->unk_464); i++) {
```

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+            break;
+    }
+
+    if (this->unk_339 >= 0) {
+        this->unk_339 = 0;
+    }
+    if (D_80906D78 != 0) {
+        D_80906D78 = 0;
+
+        for (i2 = 0; i2 < 100; i2++) {
```
Dragorn421, line 2529:
```suggestion
        for (i2 = 0; i2 < ARRAY_COUNT(sParticles); i2++) {
```
Not sure about this one, kind of makes sense since this is one of the only two uses of sParticles (through `func_808FD27C` here) but also maybe you would want less particles than possible here, and having a number larger than 100 isn't an issue, just won't spawn the former effects

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+    s32 pad;
+
+    if (this->unk_324 > 0.0f) {
+        OPEN_DISPS(globalCtx->state.gfxCtx, "../z_boss_ganon2.c", 5131);
+
+        Matrix_Push();
+        gDPPipeSync(POLY_XLU_DISP++);
+        gSPSegment(POLY_XLU_DISP++, 0x08,
+                   Gfx_TwoTexScroll(globalCtx->state.gfxCtx, 0, (s32)globalCtx->gameplayFrames, 0, 32, 64, 1,
+                                    (0 - globalCtx->gameplayFrames) * 2, (0 - globalCtx->gameplayFrames) * 8, 32, 32));
```
Dragorn421, line 2756:
```suggestion
                                    -globalCtx->gameplayFrames * 2, -globalCtx->gameplayFrames * 8, 32, 32));
```
does that not match?

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+                        if (effect->velocity.y < -10.0f) {
+                            sp78 = effect->position;
+                            sp78.y = 1086.0f;
+                            func_80078884(NA_SE_IT_SHIELD_REFLECT_SW);
+                            CollisionCheck_SpawnShieldParticlesMetal(globalCtx, &sp78);
+                        }
+                        effect->velocity.y = 0.0f;
+                    }
+                    if ((SQ(player->actor.world.pos.x - effect->position.x) +
+                         SQ(player->actor.world.pos.z - effect->position.z)) < 625.0f) {
```
Dragorn421, line 3199:
```suggestion
                         SQ(player->actor.world.pos.z - effect->position.z)) < SQ(25.0f)) {
```

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+    f32 lerpx;
+    s16 j;
+    f32 lerpy;
+    f32 lerpz;
+    Vec3f sp70;
+    Vec3f sp64;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Boss_Ganon2/func_8090464C.s")
+    for (i = 0; i < 15; i++) {
+        //! @bug j is not initialized if arg2 == 0.0f, causing undefined behavior.
```
Dragorn421, line 3316: There's actually no bug here I think, if arg == 0 then the arg2 > 0 branch is never taken and all other branches either don't use j or initialize it

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+    gSPDisplayList(POLY_OPA_DISP++, ovl_Boss_Ganon2_DL_00B3D0);
+    gDPSetTextureImage(POLY_OPA_DISP++, G_IM_FMT_I, G_IM_SIZ_16b, 1, shadowTexture);
+    gDPSetTile(POLY_OPA_DISP++, G_IM_FMT_I, G_IM_SIZ_16b, 0, 0x0000, G_TX_LOADTILE, 0, G_TX_NOMIRROR | G_TX_CLAMP, 6,
+               G_TX_NOLOD, G_TX_NOMIRROR | G_TX_CLAMP, 6, G_TX_NOLOD);
+    gDPLoadSync(POLY_OPA_DISP++);
+    gDPLoadBlock(POLY_OPA_DISP++, G_TX_LOADTILE, 0, 0, 2047, 256);
+    gDPPipeSync(POLY_OPA_DISP++);
+    gDPSetTile(POLY_OPA_DISP++, G_IM_FMT_I, G_IM_SIZ_8b, 8, 0x0000, G_TX_RENDERTILE, 0, G_TX_NOMIRROR | G_TX_CLAMP, 6,
+               G_TX_NOLOD, G_TX_NOMIRROR | G_TX_CLAMP, 6, G_TX_NOLOD);
+    gDPSetTileSize(POLY_OPA_DISP++, G_TX_RENDERTILE, 0, 0, 0x00FC, 0x00FC);
```
Dragorn421, line 3424:
```suggestion
    gDPLoadTextureBlock(POLY_OPA_DISP++, shadowTexture, G_IM_FMT_I, G_IM_SIZ_8b, 64, 64, 0, G_TX_NOMIRROR | G_TX_CLAMP, G_TX_NOMIRROR | G_TX_CLAMP, 6, 6, G_TX_NOLOD, G_TX_NOLOD);
```
pretty sure, got this by compiling what you wrote and disassembling back with gfxdis

Dragorn421 (changes requested):
What a beast, great job
I don't have that much to say about code style/format, here we go

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -1,45 +1,65 @@
...
+#include "objects/object_ganon2/object_ganon2.h"
+#include "objects/object_ganon_anime3/object_ganon_anime3.h"
 
 #define FLAGS 0x00000035
 
 #define THIS ((BossGanon2*)thisx)
 
+typedef struct {
+    /* 0x00 */ u8 type;
+    /* 0x01 */ u8 unk_01;
```
louist103, line 14:
```suggestion
```
This is probably padding and can be removed.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+static s32 sSeed3;
+
+// unused
+static UNK_TYPE D_809105DC;
+
+static Vec3f D_809105D8[4];
+
+static Vec3f D_80910608[4];
+
+static s8 D_80910638;
```
louist103, line 413: It might be a good idea to make a _data.c file for all of this data (and bss).

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+                this->unk_398 = 0;
+                this->unk_410.x = 0.0f;
+            }
+            break;
+        case 3:
+            Math_ApproachF(&this->unk_3B0.y, player->actor.world.pos.y + 47.0f + 7.0f, 0.1f, 2.0f);
+            this->unk_339 = 4;
+            if (this->unk_398 == 0xA) {
+                func_80078914(&D_80906D6C, NA_SE_EV_STONE_BOUND);
+                Audio_SetBGM(0x100000FF);
```
louist103, line 664:
```suggestion
                Audio_SetBGM(NA_BGM_STOP);
```

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+    }
+}
+
+void func_80902524(BossGanon2* this, GlobalContext* globalCtx) {
+    s8 temp_v0_4;
+    ColliderInfo* acHitInfo;
+    s16 i;
+    u8 phi_v1_2;
+
+    osSyncPrintf("this->no_hit_time %d\n", this->unk_316);
```
louist103, line 2236: might as well name the variable since we have the name.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+    if ((limbIndex == 7) || (limbIndex == 13) || (limbIndex == 33) || (limbIndex == 34)) {
+        *dList = NULL;
+    }
+
+    CLOSE_DISPS(globalCtx->state.gfxCtx, "../z_boss_ganon2.c", 5431);
+    return 0;
+}
+
+void BossGanon2_PostLimbDraw(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3s* rot, void* thisx) {
+    s8 pad;
```
louist103, line 2876: Does `s32 pad` not work?

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+        gSPDisplayList(POLY_XLU_DISP++, ovl_Boss_Ganon2_DL_00B308);
+
+        for (i = 0; i < 11; i++) {
+            Matrix_Mult(&player->mf_9E0, MTXMODE_NEW);
+            Matrix_Translate((i * 250.0f) + 900.0f, 350.0f, 0.0f, MTXMODE_APPLY);
+
+            if (i < 7) {
+                phi_f20 = 1.0f;
+            } else {
+                phi_f20 = 1.0f - ((i - 7) * 0.2333333f);
```
louist103, line 3023:
```suggestion
                phi_f20 = 1.0f - ((i - 7) * (7/30.0f));
```
Should work.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+        gDPPipeSync(POLY_XLU_DISP++);
+        gSPSegment(POLY_XLU_DISP++, 0x08,
+                   Gfx_TwoTexScroll(globalCtx->state.gfxCtx, 0, this->unk_19C * -8, 0, 32, 64, 1, this->unk_19C * -4,
+                                    this->unk_19C * -8, 32, 32));
+        gDPSetPrimColor(POLY_XLU_DISP++, 0, 0, 255, 255, 170, (s16)this->unk_37C);
+        gDPSetEnvColor(POLY_XLU_DISP++, 255, 200, 0, 128);
+        Matrix_Translate(sZelda->actor.world.pos.x + 100.0f, sZelda->actor.world.pos.y + 35.0f + 7.0f,
+                         sZelda->actor.world.pos.z - 100.0f, MTXMODE_NEW);
+        Matrix_RotateY(-M_PI / 4.0f, MTXMODE_APPLY);
+        Matrix_Scale(0.040000003f, 0.040000003f, this->unk_380, MTXMODE_APPLY);
```
louist103, line 3074:
```suggestion
        Matrix_Scale(1/25.0f,1/25.0f, this->unk_380, MTXMODE_APPLY);
```

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+                  G_MTX_NOPUSH | G_MTX_LOAD | G_MTX_MODELVIEW);
+        gSPDisplayList(POLY_XLU_DISP++, SEGMENTED_TO_VIRTUAL(ovl_Boss_Ganon2_DL_00EC40));
+        Matrix_Pop();
+
+        CLOSE_DISPS(globalCtx->state.gfxCtx, "../z_boss_ganon2.c", 5817);
+    }
+}
+
+void BossGanon2_Draw(Actor* thisx, GlobalContext* globalCtx) {
+    u8* shadowTexture = Graph_Alloc(globalCtx->state.gfxCtx, 4096);
```
louist103, line 3086:
```suggestion
    void* shadowTexture = Graph_Alloc(globalCtx->state.gfxCtx, 4096);
```
We usually use void* for textures if this works.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+
+    CLOSE_DISPS(globalCtx->state.gfxCtx, "../z_boss_ganon2.c", 5983);
+
+    func_809060E8(globalCtx);
+}
+
+void func_80905DA8(BossGanon2* this, GlobalContext* globalCtx) {
+    s32 pad[5];
+    Player* player;
+    BossGanon2Effect* effect;
```
louist103, line 3165:
```suggestion
    Player player = PLAYER;
    BossGanon2Effect* effect = globalCtx->specialEffects;
```

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+                sp70.y = this->unk_234[i].y - this->actor.world.pos.y + 76.0f + 30.0f + 30.0f + 100.0f;
+                sp70.z = this->unk_234[i].z - this->actor.world.pos.z;
+            }
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Boss_Ganon2/func_809049A0.s")
+            Matrix_MultVec3f(&sp70, &sp64);
+            sp64.x *= 0.2f;
+            sp64.y *= 0.2f;
+            temp_a3 = sp64.x + 32.0f;
+            temp_t0 = (s16)sp64.y << 6;
```
louist103, line 3336: Can this be a multiplication?

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+                }
+            }
+        }
+    }
+}
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Boss_Ganon2/func_80904FC8.s")
+void BossGanon2_GenShadowTexture(u8* shadowTexture, BossGanon2* this, GlobalContext* globalCtx) {
+    s16 i;
+    u32* p = (u32*)shadowTexture;
```
louist103, line 3381: If this cant be `void*` it should be `u64*`.  If that works.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+static Vec3f D_80907138 = { 0.0f, 0.0f, 0.0f };
+
+static s8 D_80907144[] = {
+    0xFF, 0xFF, 0x01, 0xFF, 0x03, 0x04, 0x05, 0xFF, 0x06, 0x07, 0x08, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+    0xFF, 0xFF, 0xFF, 0x02, 0x0C, 0x0D, 0x0E, 0x09, 0x0A, 0x0B, 0xFF, 0xFF, 0xFF, 0xFF, 0x00, 0x00,
+};
+
+static Vec3f D_80907164 = { 800.0f, 420.0f, 100.0f };
+
+static void* sEyeTextures[] = {
```
louist103, line 366: We dont require segment addresses in arrays to be defined like this (though MM does) but since you already have them like this you can just leave them.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+static s16 D_809071A4[] = { 2, 3, 4, 4, 4, 4, 3, 2 };
+
+static s16 D_809071B4[] = { 2, 4, 5, 5, 6, 6, 6, 6, 5, 5, 4, 2 };
+
+static s16 D_809071CC[] = { 1, -1, 1, 1, 3, 4, 1, 6, 7, 2, 9, 10, 2, 12, 13 };
+
+static u8 D_809071EC[] = { 3, 2, 2, 1, 3, 3, 1, 3, 3, 1, 0, 3, 1, 0, 3 };
+
+// padding
+static u32 D_809071FC[2] = { 0 };
```
louist103, line 387: Is this actually needed?

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+
+void BossGanon2_InitRand(s32 seedInit0, s32 seedInit1, s32 seedInit2) {
+    sSeed1 = seedInit0;
+    sSeed2 = seedInit1;
+    sSeed3 = seedInit2;
+}
+
+f32 BossGanon2_RandZeroOne(void) {
+    // Wichmann-Hill algorithm
+    f32 rand_float;
```
louist103, line 423:
```suggestion
    f32 randFloat;
```

louist103 (changes requested):
Nice job on this.  Just a few things.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
Sirius902: You're right! I copied this comment from Sda https://github.com/zeldaret/oot/blob/master/src/overlays/actors/ovl_En_Sda/z_en_sda.c#L178 since the code was extremely similar but now that you mention it I don't think it has an issue either. Should I remove the comment from there as well?

Dragorn421: I guess, if it went through a PR once though I'd wait for a third opinion

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
Sirius902: I copied this from the Morpha PR which has it named `rand_float` still so we should probably suggest the same changes over there too.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
Sirius902: It's used: https://github.com/zeldaret/oot/blob/b93dafcada89c36befca3b11b4d05828762d25bd/src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c#L467.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+    } else {
+        target = 0;
+    }
+
+    Math_ApproachS(&this->unk_31A, target, 2, 0x4000);
+}
+
+void func_809000A0(BossGanon2* this, GlobalContext* globalCtx) {
+    Animation_MorphToLoop(&this->skelAnime, &object_ganon2_Anim_026510, -2.0f);
+    this->unk_194 = Animation_GetLastFrame(&object_ganon2_Anim_026510.common);
```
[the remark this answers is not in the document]
Sirius902: Should I make that change everywhere?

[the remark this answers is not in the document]
Dragorn421: I didn't notice other places? If there are, I guess

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
Dragorn421: I think petrie said they needed to go over morpha and barinade again to fix a bunch of stuff, so I wouldn't worry about those PRs, just change it here

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
Sirius902: Nope.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
Sirius902: Doesn't seem to match unfortunately.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
Sirius902: Doesn't match.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
Sirius902: I'm not sure, do you have any suggestions for me to try? Sda has something similar. https://github.com/zeldaret/oot/blob/master/src/overlays/actors/ovl_En_Sda/z_en_sda.c#L197

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -278,116 +304,3125 @@ static ColliderJntSphInit D_80907014 = {
...
+                this->unk_3A4.y = this->actor.world.pos.y + 60.0f;
+                this->unk_3A4.z = this->actor.world.pos.z - 15.0f;
+                this->unk_39C = 0xF;
+                this->unk_398 = 0;
+                this->unk_3B0.y = this->actor.world.pos.y + 77.0f + 100.0f;
+                this->unk_314 = 2;
+                this->unk_3B0.z = this->actor.world.pos.z + 5.0f;
+                this->unk_3B0.x = this->actor.world.pos.x;
+            }
+            if ((globalCtx->gameplayFrames & 0x1F) == 0) {
```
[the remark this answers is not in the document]
Sirius902: It didn't affect codegen anywhere so I went ahead and made the change everywhere. :)

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
Sirius902: Ah, didn't realize that whole thing was a macro. Thanks.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
Sirius902: Yes, doesn't even work as a single `u32` either. It has to be either `u32[2]` or `u64`.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
Sirius902: What part of it? The `void*[]`? I typed it like that because it was an array of textures. Should I have done something else?

louist103: usually what is done is we just leave the raw pointer.
`void* tex[] = {0x06001234, 0x06001233,etc...};`

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
louist103: Should be *64

Sirius902: Yep, that works. I guess earlier I was trying to multiply it as a float haha.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
Sirius902: Ah gotcha. I'll probably leave them `extern`ed since I know what type they are and left a comment next to them and you said it's fine, but if wanted I can change it.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
Sirius902: I'll hold off on this change since you're not sure, but I should definitely make this change here as well since it's directly iterating over the particles? https://github.com/zeldaret/oot/blob/cb7b9fdcfaa5666651fba4be15576d14ec6ad2e9/src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c#L3276-L3283

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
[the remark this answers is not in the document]
Sirius902: Ah oops, thought you meant the change in the opposite direction.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
Sirius902: I could do this but the Makefile is trying to build it as a standalone object file as well which means I have to move the definition for `BossGanon2Effect` over into the `_data.c`. Is this fine?

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
Dragorn421: Yeah that for loop you linked is definitely iterating over the whole array. I'm hesitating about the one I commented on because the code would also work with any value below 100 while in most other places it needs to be exactly 100 = ARRAY_COUNT

oh well just change both I guess

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
Dragorn421: don't worry about the Makefile building useless .o object files, it's fine as long as they aren't mentioned in the spec

Sirius902: Done, let me know if it looks good. I was more commenting on the fact that the effect struct definition has to be in the `_data.c` file or else the useless .o file fails to compile and that causes the whole build to fail. If it's fine that it's there then there's no problem but if not it'd either have to go in the header file (I'm not sure that's a good idea since other actors won't be using the particles in theory) or create another header file.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
Sirius902: Done, changed both.

Dragorn421 (approved):
Apart from the `@bug` where I don't think there's a bug, this is good to me

louist103 (approved):

fig02:
I agree that its not a bug and should be removed from both. I dont see a path where the variable is uninitialized

Sirius902:
oops I just @ someone didn't I, sorry to that person.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -52,342 +56,3008 @@ const ActorInit Boss_Ganon2_InitVars = {
...
+    sSeed3 = (sSeed3 * 170) % 30323;
+
+    randFloat = (sSeed1 / 30269.0f) + (sSeed2 / 30307.0f) + (sSeed3 / 30323.0f);
+    while (randFloat >= 1.0f) {
+        randFloat -= 1.0f;
+    }
+    return fabsf(randFloat);
+}
+
+void func_808FD080(s32 idx, ColliderJntSph* collider, Vec3f* arg2) {
```
lf94, line 82: BossGanon2_SetColliderOrigin ?

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -52,342 +56,3008 @@ const ActorInit Boss_Ganon2_InitVars = {
...
+        OPEN_DISPS(globalCtx->state.gfxCtx, "../z_boss_ganon2.c", 790);
+
+        gSPSegment(POLY_OPA_DISP++, 0x06, globalCtx->objectCtx.status[objectIdx].segment);
+        gSPSegment(POLY_XLU_DISP++, 0x06, globalCtx->objectCtx.status[objectIdx].segment);
+
+        CLOSE_DISPS(globalCtx->state.gfxCtx, "../z_boss_ganon2.c", 799);
+    }
+}
+
+void func_808FD210(GlobalContext* globalCtx, Vec3f* arg1) {
```
lf94, line 107: BossGanon2_SetEffectPosition ?

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -52,342 +56,3008 @@ const ActorInit Boss_Ganon2_InitVars = {
...
+    effect->unk_01 = 0;
+    effect->velocity.x = 25.0f;
+    effect->velocity.y = 15.0f;
+    effect->velocity.z = 0.0f;
+    effect->accel.x = 0.0f;
+    effect->accel.y = -1.0f;
+    effect->accel.z = 0.0f;
+}
+
+void func_808FD27C(GlobalContext* globalCtx, Vec3f* position, Vec3f* velocity, f32 scale) {
```
lf94, line 122: BossGanon2_SetParticles ?

fig02: this is for a specific type, not a generic "set"

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
fig02: this is doing more than setting a positioin

lf94: I guess I didn't read the whole function? Thought I did

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
lf94: Good point

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -52,342 +56,3008 @@ const ActorInit Boss_Ganon2_InitVars = {
...
+                this->actor.velocity.y = 0.0f;
+                Animation_MorphToPlayOnce(&this->skelAnime, &object_ganon_anime3_Anim_010380, 0.0f);
+                func_808FD4D4(this, globalCtx, 0, 3);
+                func_800A9F6C(0.0f, 0xC8, 0x14, 0x14);
+            }
+            break;
+        case 21:
+            this->unk_339 = 6;
+            SkelAnime_Update(&this->skelAnime);
+            this->unk_41C = Math_CosS(globalCtx->gameplayFrames << 0xF) * this->unk_420;
```
fig02, line 653: does multiplication work for this? think itd be * 0x8000

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -52,342 +56,3008 @@ const ActorInit Boss_Ganon2_InitVars = {
...
+            this->unk_3B0.y = (this->unk_1B8.y + 60.0f) - 40.0f;
+            if (this->unk_398 > 0xA6 && this->unk_398 < 0xAD) {
+                this->unk_312 = 2;
+            }
+            if (this->unk_398 > 0xBA && this->unk_398 < 0xC4) {
+                this->unk_312 = 1;
+            }
+            if (this->unk_398 > 0xCA && this->unk_398 < 0xD2) {
+                this->unk_312 = 2;
+            }
```
fig02, line 694: i think 398 is a timer value and all these can be decimal. but if youd rather wait for a documentation pass I think that probably fine too

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -52,342 +56,3008 @@ const ActorInit Boss_Ganon2_InitVars = {
...
+                    globalCtx->envCtx.unk_D8 = 0.0f;
+                    globalCtx->envCtx.unk_BE = 0;
+                    this->unk_339 = 0;
+                }
+            }
+            break;
+        case 24:
+            SkelAnime_Update(&this->skelAnime);
+            // might be fake?
+            if (1) {}
```
fig02, line 745: dont need to comment on if 1s imo, we can easily search them up if we want to try to get rid of them

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -52,342 +56,3008 @@ const ActorInit Boss_Ganon2_InitVars = {
...
+                while (D_8090EB30 != NULL) {
+                    if (D_8090EB30->id == ACTOR_EN_ELF) {
+                        this->unk_3A4.x = D_8090EB30->world.pos.x - 30.0f;
+                        this->unk_3A4.y = D_8090EB30->world.pos.y;
+                        this->unk_3A4.z = D_8090EB30->world.pos.z;
+                        this->unk_3B0.x = D_8090EB30->world.pos.x;
+                        this->unk_3B0.y = D_8090EB30->world.pos.y;
+                        this->unk_3B0.z = D_8090EB30->world.pos.z;
+                        break;
+                    }
```
fig02, line 820: nothing to add here, just upset that morita didnt use the navi pointer in the player struct lol

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -52,342 +56,3008 @@ const ActorInit Boss_Ganon2_InitVars = {
...
+    osSyncPrintf("this->no_hit_time %d\n", this->noHitTime);
+    if (this->noHitTime != 0 || ((this->unk_334 == 0) && (this->actionFunc == func_80900890))) {
+        for (i = 0; i < ARRAY_COUNT(this->unk_464); i++) {
+            this->unk_424.elements[i].info.bumperFlags &= ~2;
+        }
+    }
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_Boss_Ganon2/func_8090120C.s")
+    osSyncPrintf("this->look_on %d\n", this->lookOn);
+    if (this->lookOn) {
```
fig02, line 1890: I think i would personally leave these og names as comments in the struct and leave them unnamed. At least in the case of `noHitTime`, there will probably be a different name like `invulnerableTimer` or whatever, something less japanese-ey :kappa:

if you decide to commend them in the struct, you can leave them in quotes right beside the variable name. we do this in plenty of other places

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
fig02: yea this makes sense cause of alignment

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -52,342 +56,3008 @@ const ActorInit Boss_Ganon2_InitVars = {
...
+        Matrix_MultVec3f(&D_80907138, &this->unk_1C4);
+    }
+
+    temp_v0 = D_8090709C[limbIndex];
+    if (temp_v0 >= 0) {
+        Matrix_MultVec3f(&D_80906D60, &sp4C);
+        func_808FD080(temp_v0, &this->unk_424, &sp4C);
+    }
+
+    if ((limbIndex == 7) || (limbIndex == 0xD)) {
```
fig02, line 2556: second limb index in dec

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
fig02: sometimes changing what gets `f` or not can affect the results. Im not good at mangling these floats tho, just know that trick exists

fig02 (commented):
nice job on this

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
fig02: though itd probably make the cutscene functions easier to read in the mean time while things arent documented. so i do recommend it.

you can see a bunch of timers being decremented in the update function. Id try to make all uses of those variables decimal

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
Sirius902: Just tried it without the `f` but unfortunately in this case it ended up giving the same value.

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
fig02: just saw that this was done because of a review comment above. if i saw it earlier i would have commented lol
@louist103 in the future id avoid suggesting to use og names. It should just be named when a proper name is thought of through documentation

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
fig02: this is the last thing on my end and itd be good to go from me. Having them in decimal already will make docs alot easier down the road I think, so we should do it now.

I know of a good vscode extension that lets you highlight numbers and convert them between dec and hex, if youd like

Sirius902: Haha I uh just finished doing it all manually and it was pretty awful but I will keep that in mind for next time. Let me know if I missed anything.

fig02: nice lol
the plugin is NumberMonger. you highlight the number then press ctrl+p to bring up the command palette, and type "convert" and youll see a bunch of conversion options. Idk if its possible to bind it to a shortcut to skip the palette, prob is.

Its still a bit manual, but you at least dont need to tab out and go to a calculator everytime

fig02 (approved):

Dragorn421 (approved):

Roman971 (approved):

PR #879, opened 2021-07-30
convert remaining RGB5a1 to RGBA16

louist103:


AngheloAlf (approved):

EllipticEllipsis (approved):

fig02 (approved):
due to the limited use of that bitfield color type, i personally wouldnt be opposed to making it local to prerender. But we can do that later if we all agree

Roman971 (approved):

PR #881, opened 2021-07-31
Item00 BSS OK

louist103:


fig02:
Worried this fix will not be permanent, if some external factors that affect bss reordering occurs, like headers changing.
Would be nice to get it fully matching, but may want to leave a comment somewhere that the bss in this function is strange and (possibly?) subject to change

louist103:
> Worried this fix will not be permanent, if some external factors that affect bss reordering occurs, like headers changing.
> 
> Would be nice to get it fully matching, but may want to leave a comment somewhere that the bss in this function is strange and (possibly?) subject to change

I have a feeling that the BSS issues were due to the weird temp situation and nothing to do directly with BSS. Since it's accessed like an array it probably was one at one point and was reduced down to one element. I can still leave the comment though.

src/code/z_en_item00.c
```diff
@@ -732,36 +731,37 @@ void EnItem00_Update(Actor* thisx, GlobalContext* globalCtx) {
...
-                            ((dynaActor->world.pos.x != dynaActor->prevPos.x) ||
-                             (dynaActor->world.pos.y != dynaActor->prevPos.y) ||
+                        if (((dynaActor != NULL) && (dynaActor->update != NULL)) &&
+                            (((dynaActor->world.pos.x != dynaActor->prevPos.x) ||
+                              (dynaActor->world.pos.y != dynaActor->prevPos.y)) ||
                              (dynaActor->world.pos.z != dynaActor->prevPos.z))) {
-                            D_80157D94++;
+                            D_80157D94[0]++;
                             break;
                         }
```
Dragorn421, line 754: Not relevant to the PR but since you added parenthesis,
could this be broken into two ifs like:
```c
if ((dynaActor != NULL) && (dynaActor->update != NULL)) {
	if ((dynaActor->world.pos.x != dynaActor->prevPos.x) ||
		(dynaActor->world.pos.y != dynaActor->prevPos.y) ||
		(dynaActor->world.pos.z != dynaActor->prevPos.z)) {
		D_80157D94[0]++;
		break;
	}
}
```
looks way easier to read imo
(also I think it's easier to read `a && b && c` rather than `(a && b) && c` but that may be up to personal preference)

src/code/z_en_item00.c
```diff
@@ -776,10 +776,10 @@ void EnItem00_Update(Actor* thisx, GlobalContext* globalCtx) {
...
-        f32 newUnkBC = Math_CosS(this->actor.shape.rot.x) * 37.0f;
+        f32 cos = Math_CosS(this->actor.shape.rot.x) * 37.0f;
 
-        this->actor.shape.yOffset = newUnkBC;
-        if (newUnkBC >= 0.0f) {
+        this->actor.shape.yOffset = cos;
+        if (cos >= 0.0f) {
             this->actor.shape.yOffset = this->actor.shape.yOffset;
         } else {
             this->actor.shape.yOffset = -this->actor.shape.yOffset;
```
Dragorn421, line 785: I would name `cos` `yOffset` and the condition after looks a lot like `this->actor.shape.yOffset = ABS(this->actor.shape.yOffset)`

Dragorn421 (commented):
nice. I don't have anything to say about the actual bss/match stuff

src/code/z_en_item00.c
```diff
@@ -710,18 +716,18 @@ void func_8001E5C8(EnItem00* this, GlobalContext* globalCtx) {
         this->actor.world.pos.y += 20.0f;
     }
 }
-
-extern s32 D_80157D90;
-extern u32 D_80157D90_; // these must be defined separately for EnItem00_Update to match.
-extern s16 D_80157D94;
-
+/* The BSS in the function acted weird in the past. It is matching now but mighy cause issues in the future*/
```
Dragorn421, line 719:
```suggestion

/* The BSS in the function acted weird in the past. It is matching now but mighy cause issues in the future*/
```

src/code/z_en_item00.c
```diff
@@ -732,36 +738,37 @@ void EnItem00_Update(Actor* thisx, GlobalContext* globalCtx) {
...
+            if (*temp != globalCtx->gameplayFrames) {
+                D_80157D90 = globalCtx->gameplayFrames;
+                D_80157D94[0] = 0;
+                for (i = 0; i < 50; i++) {
                     if (globalCtx->colCtx.dyna.bgActorFlags[i] & 1) {
                         dynaActor = globalCtx->colCtx.dyna.bgActors[i].actor;
                         if ((dynaActor != NULL) && (dynaActor->update != NULL) &&
                             ((dynaActor->world.pos.x != dynaActor->prevPos.x) ||
                              (dynaActor->world.pos.y != dynaActor->prevPos.y) ||
                              (dynaActor->world.pos.z != dynaActor->prevPos.z))) {
```
Dragorn421, line 758: cf my previous comments, does this not match as two ifs or do you like the single if better?

src/code/z_en_item00.c
fig02: mighy -> might, and id use // personally

src/code/z_en_item00.c
fig02: fwiw i would prefer two ifs as well, the null checks on their own make sense. It should work id think

src/code/z_en_item00.c
```diff
@@ -402,64 +406,64 @@ void EnItem00_Init(Actor* thisx, GlobalContext* globalCtx) {
...
             break;
     }
 
     this->unk_156 = 0;
-    ActorShape_Init(&this->actor.shape, sp34, ActorShadow_DrawCircle, sp30);
+    ActorShape_Init(&this->actor.shape, yOffset, ActorShadow_DrawCircle, shadowScale);
     this->actor.shape.shadowAlpha = 180;
     this->actor.focus.pos = this->actor.world.pos;
-    this->unk_152 = 0;
+    this->getItemId = 0;
```
fig02, line 466: now that this is named, you can use the gid enum. relevant in a couple places

Dragorn421 (approved):

fig02 (approved):

Roman971 (approved):

PR #882, opened 2021-08-03
Mark func_800DC910 as NON_EQUIVALENT

louist103:
Also fixed a function call.

### DIFF
--- a/include/functions.h
+++ b/include/functions.h
@@ -2003,7 +2003,7 @@ s32 Audio_SyncLoadSample(s32 arg0, s32 arg1, u8* arg2);
 s32 Audio_SyncLoadSeq(s32 seqIdx, u8* ramAddr, u8* arg2);
 void Audio_AsyncLoadReqInit(void);
 void func_800E4D94(void);
-// ? func_800E4EEC(?);
+void func_800E4EEC(s32 arg0, s32 arg1, u8* arg2);
 void func_800E4F58(void);
 void func_800E4FB0(void);
 AudioTask* func_800E4FE0(void);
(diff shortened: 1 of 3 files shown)


Dragorn421 (approved):
imagine building with -D NON_MATCHING

<details>



</details>

this meme was made by the OK gang

AngheloAlf (approved):

fig02:
were we able to identify what exactly is non equivalent? 
i dont understand how the gcc testing being done was able to identify that this was non equiv. think itd be good to try to identify the exact problem and potentially try to fix it

louist103:
> 
> 
> were we able to identify what exactly is non equivalent?
> i dont understand how the gcc testing being done was able to identify that this was non equiv. But i havent really been keeping up with that stuff.
> Think itd be good to try to identify the exact problem and potentially try to fix it

The GCC testing is how I got the idea to test this.  IDO crashes with this as well if thats what you meant.  The only thing GCC has to do with this is that it enables NON_MATCHING and NON_EQUIVALENT. 
Ill take a look at the function but given its size IDK if I will be able to fix it.

fig02:
oh it crashes with ido, ok
yeah you dont need to look into it alot if you dont want to. but a quick look at the crash debugger might reveal an obvious problem

louist103:
> 
> 
> oh it crashes with ido, ok
> yeah you dont need to look into it alot if you dont want to. but a quick look at the crash debugger might reveal an obvious problem

yeah so the debugger doesn't show up even with the and removing the button thing.

fig02 (approved):

Roman971 (approved):

PR #885, opened 2021-08-04
Add a bug comment to bombchus

louist103:
Initial discussion starts at https://discord.com/channels/688807550715560050/688851317593997489/872302481110085733
Wanted to get it here to get more feedback. So far the only opposition is that you need to do glitches already to be in a state where this crash happens.

### DIFF
--- a/src/overlays/actors/ovl_En_Bom_Chu/z_en_bom_chu.c
+++ b/src/overlays/actors/ovl_En_Bom_Chu/z_en_bom_chu.c
@@ -219,6 +219,8 @@ void EnBomChu_WaitForRelease(EnBomChu* this, GlobalContext* globalCtx) {
         this->unk_16C.z = Math_CosS(this->actor.shape.rot.y + 0x4000);
 
         this->actor.speedXZ = 8.0f;
+        //! @bug there is no NULL check on the floor poly.  If the player is out of bounds the floor poly will be NULL
+        //! and will cause a crash inside this function.
         func_809C5BA8(this, this->actor.floorPoly, globalCtx);
         this->actor.flags |= 1; // make chu targetable
         func_8002F850(globalCtx, &this->actor);


fig02:
to add on to what i said in discord, for completeness sake I would want to include this. But I think this would be the first case of us marking a bug that you have to stack other bugs in order to accomplish. Seems like a rabbit hole to be jumping down lol

fig02:
To give an example, the remote control epona glitch. Sure it is possible to get bottle on B (bug 1), die and respawn in some water (bug 2? its already becoming shady lol), then press ocarina while on epona to get off her and keep controlling her. Are we gonna mark epona not handling ocarina as a bug?

Obviously a much more extreme example. But this is what i mean about a rabbit hole

fig02:
And dont get me started on bugs that require remote control epona as a prerequisite. We can keep adding more layers :Kappa:

mzxrules:
To that, my argument is that the actual conditions of the bug are fairly straightforward, and quite easy to achieve with modding. Just add a bombchu spawn over a void, and boom! crash. I don't think it's overly unreasonable to expect a bombchu to handle itself gracefully when spawned over a void.

Dragorn421:
Imo this is an oversight and a bug,

first I agree with mzx this is actually a common enough bug that may be encountered in mods,

and also if you take a quick look at every `floorPoly` usage in decompiled code there's always some kind of check that the floor poly should be set (eg `bgcheckflags & 2`, `floorY != bgcheck_min_y`, `floorPoly != NULL`).
So it looks like they just never even thought about bombchus not being on collision, while most of the code is safe about it

fig02:
fair enough yeah. like i was saying above i would like to see this included. just slightly worried about where we are going to draw the line as we get to the more complex bugs in the game lol

Dragorn421 (approved):

fig02:
Im still unsure if the placement of the bug comment here is best.  should it be the responsibility of that function to do a NULL check and react accordingly? or should the function thats using it do that instead? guess both work, but idk for sure

Dragorn421:
#905 may help, `EnBomChu_UpdateFloorPoly`/`func_809C5BA8` is otherwise called after bgcheck calls so the poly passed as argument isn't NULL, so I would agree with louis on the `@bug` location

not that good practice wouldn't put a NULL check in `func_809C5BA8` but they really forgot a NULL check specifically, in `EnBomChu_Move` when there's no collision nearby (all bgcheck calls fail to find collision) the bombchu defaults to exploding, the same would probably be better on bombchu release too

fig02:
alright, I agree too. thanks

fig02 (approved):

Roman971 (approved):

PR #887, opened 2021-08-04
Fix some issues in the vr_box non matching

Roman971:
Found some non equivalent things so I fixed them.
I also managed to fix the stack differences, but the rest is going to need more investigation.


Dragorn421 (approved):
"how can this have wrong line numbers"
and then: "oh it's a 550 lines function 😰"

fig02 (approved):

PR #889, opened 2021-08-05
Automatic objects 2

AngheloAlf:
Adds an XML to every remaining object and adds them to the spec.


fig02 (approved):

petrie911:
BTW, object_bv and object_mo are already part of their respective boss PRs.

AngheloAlf:
If they get merged first, I can remove them.
If this get merged first, they can either ignore this and use their XML, or use the XML from here to improve theirs in case they missed something. It isn't a big deal imo

EllipticEllipsis (approved):
IMO this shouldn't include the ones that are already in PRs, but it's not a big problem.

fig02:
Excluding the ones in prs is reasonable.
I only suggested getting them all in so we can be 100% sure we have them all. But I definitely expect all the ones open to get in sooner than later, so it would help avoid conflicts

AngheloAlf:
So, I would just need to exclude object_bv and object_mo?

AngheloAlf:
Do we want an XML for `gameplay_object_exchange_static`? I think it is the only one missing of the `gameplay_` stuff.
I know it only has zeroes, but we could just extract a small blob and add it to the spec.

Roman971 (approved):

fig02:
think we can probably wait on the exchange static stuff, we still need to talk about how we're going to make use of them and stuff. soon though

PR #890, opened 2021-08-06
Add linker file splits for objects

Zelllll:
This is a draft created for a few reasons:
- To display why assets should be properly split
- To show the the process for doing so
- And to show the barriers preventing this from fully working with ZAPD as it stands.
- More information is available in my write-up here: https://hackmd.io/A3sDKNLdT9SGLSkIv2k_hQ?view , as well as the split addresses for all files which I generated with a script I wrote, as well as a tool which creates a split XML: https://pastebin.com/KqPknrQj

Barriers:
- Currently ZAPD exports a new header for each split file--we want a single header. This is also why this PR will currently not build, because the `#include`'s are going to be broken.
- ZAPD does not properly export unaccounted data if a file is split. This is somewhat beneficial since the linker pads are now gone, however XMLs will need adjusting to have all of their data explicitly written.
- ZAPD does not have a way to detect textures from within other files that are linked into a single object. Some hacks can be done to bypass this, which are viewable here.

Benefits:
- The way that the files are split closely reflects the same type of files that modding tools like Fast64 can export (single files for collision, animations, skeletons and their models and textures, etc.)
- More modularity allows for easy optimizing of objects (to remove an animation, simply comment it out in `spec`).
- With files split like this, projects can decide which kinds of files they want to commit to their repos.
- If a texture is of the proper data type (i.e. u8, u16) it is less likely that there will be alignment issues since the textures are usually at the very top of a split file, and the linker will pad them automatically.
- By splitting files as intended with the linker, there is no need for messy "hub" c files that `#include` all of the assets, using the `spec` and/or a linker script will make managing assets cleaner and easier.

P.S. If an admin decides to close this PR, closing perfectly valid and constructive contributions does not benefit anyone.


NEstelami:
Zel, this PR may very well be constructive, but that does not change the fact that you've been banned from contributing, something you're well aware of. I'm going to have to close this PR. Please do not make any further pull requests.

PR #896, opened 2021-08-12
Match some audio functions

zelda2774:
Still working on this, but PR'ing early because the func_800E1618 non-matching
turns out to be non-equivalent in a way that seems like it could cause crashes
for NON_MATCHING builds. This PR matches that function plus a few more.


src/code/code_800E4FE0.c
```diff
@@ -536,7 +518,7 @@ s32 func_800E5EDC(void) {
 void func_800E5F34(void) {
     // macro?
     // clang-format off
-    s32 chk = -1; s32 sp28; do { } while (osRecvMesg(gAudioContext.audioResetQueueP, &sp28, 0) != chk);
+    s32 chk = -1;s32 sp28;do { } while (osRecvMesg(gAudioContext.audioResetQueueP, (OSMesg*) &sp28, OS_MESG_NOBLOCK) != chk);
```
Roman971, line 521: I think it should be formatted like this? I imagine it doesn't matter for codegen and it would match our normal formatting a bit more
```suggestion
    s32 chk = -1; s32 sp28; do { } while (osRecvMesg(gAudioContext.audioResetQueueP, (OSMesg*)&sp28, OS_MESG_NOBLOCK) != chk);
```

tools/permuter_settings.toml
```diff
@@ -1,6 +1,7 @@
 [preserve_macros]
 "g[DS]P.*" = "void"
 "gDma.*" = "void"
+"a[A-Z].*" = "void"
```
Dragorn421, line 4: not familiar with the permuter, is this fine to commit?

src/code/code_800EC960.c
```diff
@@ -1675,26 +2339,26 @@ void func_800F2A04(void) {
     }
 }
 
-void func_800F2D6C(s8* arg0, u16 arg1) {
+void func_800F2D6C(const s8* str, u16 num) {
```
Dragorn421, line 2342:
```suggestion
void func_800F2D6C(const char* str, u16 num) {
```

src/code/code_800EC960.c
```diff
@@ -169,10 +215,10 @@ s8 D_8016B7DC;
 f32 D_8016B7E0;
 u16 D_8016B7E4;
 typedef struct {
-    char unk_00[6];
-    s16 unk_06;
-} unk_D_8016B7E8;
-unk_D_8016B7E8 D_8016B7E8[0x19];
+    s8 str[5];
```
Dragorn421, line 218:
```suggestion
    char str[5];
```

tools/permuter_settings.toml
Roman971: It should be yeah, I assume it's to preserve abi commands (the equivalent of gbi commands for audio).
And I don't think it should conflict with any other macros since it wouldn't fit our naming scheme

fig02 (approved):

PR #897, opened 2021-08-12
Symbol cleanup

Dragorn421:
How much cleanup? It reduced `undefined_syms.txt` from 1704 lines to 302 so I guess a lot.

- Changed some hex offsets (like `0x06123456`) to symbols (like `object_XXX_DL_123456`)
- moved comments on some externs to xml
- changed undefined symbols (like `D_06123456`) to asset symbols (like `object_XXX_DL_123456`), where I could figure out what object it was from (some actors need more documentation to do it with a lesser chance of making a mistake)

Sorry for the large PR. However there is **no** documentation done here, only symbol usage changes. The only thing to watch out for really is for actors that use several objects (are the symbols from the right object)

It would be nice to merge this somewhat quickly (like, before october?) because this helps with the object decomp process. My idea with this PR is to make it easier to do object decomp without having to understand/touch C (just ctrl+h)

# interesting things I came across

I found some undefined syms that turned out to be symbols from documented objects (eg `D_060000F0` -> `gKm1Skel`)

symbols for `libultra OS symbols` / `OS hardware registers` aren't all used. I kept the unused ones still since I'm not too knowledgeable on that stuff

## z_bg_pushbox.c
```c
extern Gfx D_06000000[];
extern CollisionHeader D_06000350;
```
doesn't offset right into OBJECT_GAMEPLAY_DANGEON_KEEP

## z_bg_mjin.c
```c
extern UNK_TYPE D_06000000;
```
uses the same symbol for different objects

## z_en_a_keep.c
`D_06000730` doesn't offset right into gameplay_keep

## z_en_kusa.c
I feel like `Gfx_DrawDListOpa(globalCtx, D_060002E0);` is reachable without the current object being object_kusa (EDIT: after documenting en_kusa, I don't think so anymore)

`static Gfx* dLists[] = { gFieldBushDL, 0x06000140, 0x06000140 };` the first `0x06000140` should be `gHanaFlower3DL` (but the segment is 6, not 5)


Zelllll:
I will point out that #851 should probably be merged before this one, since it also fixes some of the stuff fixed here.

Dragorn421:
About if we should wait for #851 to consider merging this PR:

I tried merging this PR into zel's #851 branch locally and the only two conflicts are in `undefined_syms.txt` (not surprising since that's the focus of this PR)

And with the conflicts in `undefined_syms.txt` resolved, it builds OK without any other change. So basically at this point this PR and #851 do not conflict

fig02 (approved):

Roman971 (approved):

PR #898, opened 2021-08-13
Document en_viewer

Dragorn421:
Cutscene actors (impa, ganon, zelda, horses)

More documentation needed once relevant objects/scenes are decompiled

`params & 0xFF` goes unused but is set to various values in scene files.

### DIFF
--- a/src/overlays/actors/ovl_En_Viewer/z_en_viewer.c
+++ b/src/overlays/actors/ovl_En_Viewer/z_en_viewer.c
@@ -15,27 +15,30 @@
 
 #define THIS ((EnViewer*)thisx)
 
+#define TYPE (this->actor.params >> 8)
+
 void EnViewer_Init(Actor* thisx, GlobalContext* globalCtx);
 void EnViewer_Destroy(Actor* thisx, GlobalContext* globalCtx);
 void EnViewer_Update(Actor* thisx, GlobalContext* globalCtx);
 void EnViewer_Draw(Actor* thisx, GlobalContext* globalCtx);
 
-void EnViewer_GetCutsceneNextPos(EnViewer* this, GlobalContext* globalCtx);
-void func_80B2C8AC(EnViewer* this2, GlobalContext* globalCtx);
-void func_80B2CC1C(GlobalContext* globalCtx, EnViewer* this);
-void func_80B2A570(EnViewer* this, GlobalContext* globalCtx);
-void func_80B2A75C(EnViewer* this, GlobalContext* globalCtx);
+void EnViewer_UpdatePosition(EnViewer* this, GlobalContext* globalCtx);
+void EnViewer_DrawFireEffects(EnViewer* this2, GlobalContext* globalCtx);
+void EnViewer_UpdateGanonCape(GlobalContext* globalCtx, EnViewer* this);
+void EnViewer_InitImpl(EnViewer* this, GlobalContext* globalCtx);
+void EnViewer_UpdateImpl(EnViewer* this, GlobalContext* globalCtx);
 
-// sAnimFuncs
-void func_80B2A300(EnViewer* this, GlobalContext* globalCtx, void* skeletonHeaderSeg, AnimationHeader* animationSeg);
-void func_80B2A448(EnViewer* this, GlobalContext* globalCtx, void* skeletonHeaderSeg, AnimationHeader* animationSeg);
-void func_80B2A4D8(EnViewer* this, GlobalContext* globalCtx, void* skeletonHeaderSeg, AnimationHeader* animationSeg);
+// sInitDrawFuncs
+void EnViewer_InitDrawGanonOrZelda(EnViewer* this, GlobalContext* globalCtx, void* skeletonHeaderSeg,
+                                   AnimationHeader* anim);
+void EnViewer_InitDrawImpa(EnViewer* this, GlobalContext* globalCtx, void* skeletonHeaderSeg, AnimationHeader* anim);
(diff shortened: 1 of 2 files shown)


src/overlays/actors/ovl_En_Viewer/z_en_viewer.c
```diff
@@ -59,144 +62,169 @@ static InitChainEntry sInitChain[] = {
...
 };
+// clang-format on
 
-static EnViewerAnimFunc sAnimFuncs[] = {
-    func_80B2A300,
-    func_80B2A4D8,
-    func_80B2A300,
-    func_80B2A448,
+static EnViewerInitDrawFunc sInitDrawFuncs[] = {
+    EnViewer_InitDrawGanonOrZelda,
```
fig02, line 101: not too sure i agree with "InitDraw". i dont consider animation stuff really as a part of draw. it seems mostly concerned with skeletons and animations. So I would probably go with `InitSkelAnim` (without the "e", i mean initializing skeleton and animation, not the system itself lol)

src/overlays/actors/ovl_En_Viewer/z_en_viewer.c
```diff
@@ -19,29 +19,32 @@
...
+                                   AnimationHeader* anim);
+void EnViewer_InitDrawImpa(EnViewer* this, GlobalContext* globalCtx, void* skeletonHeaderSeg, AnimationHeader* anim);
+void EnViewer_InitDrawHorse(EnViewer* this, GlobalContext* globalCtx, void* skeletonHeaderSeg, AnimationHeader* anim);
 
 // sDrawFuncs
-void func_80B2B4A8(EnViewer* this, GlobalContext* globalCtx);
-void func_80B2B8FC(EnViewer* this, GlobalContext* globalCtx);
-void func_80B2BA38(EnViewer* this, GlobalContext* globalCtx);
-void func_80B2C130(EnViewer* this, GlobalContext* globalCtx);
+void EnViewer_DrawGanon(EnViewer* this, GlobalContext* globalCtx);
```
fig02, line 42: Ganon -> Ganondorf for all references in this file, imo

src/overlays/actors/ovl_En_Viewer/z_en_viewer.c
```diff
@@ -19,29 +19,32 @@
 
 #define THIS ((EnViewer*)thisx)
 
+#define TYPE (this->actor.params >> 8)
+
```
fig02, line 23: we still havent standardized how we are going to do this, so it would probably be best to wait until we do so. At the very least it would be a macro like GET_TYPE that takes params as an arg. but theres still some details to figure out with that

src/overlays/actors/ovl_En_Viewer/z_en_viewer.c
```diff
@@ -265,87 +295,88 @@ void func_80B2A75C(EnViewer* this, GlobalContext* globalCtx) {
...
-    if (D_80B2CFCC != 0) {
-        D_80B2CFCC--;
+    if (sTimer != 0) {
+        sTimer--;
     }
 
-    EnViewer_GetCutsceneNextPos(this, globalCtx);
-    Actor_MoveForward(&this->actor);
+    EnViewer_UpdatePosition(this, globalCtx);
+    Actor_MoveForward(&this->actor); // has no effect, speed is 0
```
fig02, line 303: believe this still applies gravity even if speed is 0 (also a point in favor of renaming this func from "move forward" to something else)

src/overlays/actors/ovl_En_Viewer/z_en_viewer.h
```diff
@@ -38,9 +65,9 @@ typedef struct EnViewer {
     /* 0x01DD */ u8 drawFuncIndex;
     /* 0x01E0 */ EnViewerActionFunc actionFunc;
     /* 0x01E4 */ u8 unused;
-    /* 0x01E5 */ u8 unk_1E5;
-    /* 0x01E6 */ u8 unk_1E6; // Conditional
-    /* 0x01E8 */ EnViewerUnkStruct unk_1E8[20];
+    /* 0x01E5 */ u8 state;
+    /* 0x01E6 */ u8 draw;
```
fig02, line 69: a name that makes it more clear that this is a bool like `isVisible` could be desirable

fig02 (commented):
apologies for the long wait on these doc prs

src/overlays/actors/ovl_En_Viewer/z_en_viewer.c
Dragorn421: Good point, however I think this actor has gravity 0? it doesn't set any and I don't think there is a default gravity besides the struct being bzero'd

fig02 (approved):

src/overlays/actors/ovl_En_Viewer/z_en_viewer.c
```diff
@@ -59,144 +60,170 @@ static InitChainEntry sInitChain[] = {
...
-    func_80B2A448,
+static EnViewerInitAnimFunc sInitAnimFuncs[] = {
+    EnViewer_InitAnimGanondorfOrZelda,
+    EnViewer_InitAnimHorse,
+    EnViewer_InitAnimGanondorfOrZelda,
+    EnViewer_InitAnimImpa,
 };
 
-static void* sActorShadowDrawFunc[] = {
+static ActorShadowFunc sShadowFuncs[] = {
```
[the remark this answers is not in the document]
fig02: it doesnt exist, but I think his point was just that its removing information that is helpful and would avoid possible confusion.
The typedef is a good point, I omitted it for length reasons and because the variable in the actor struct was named shadowDraw already. But if people feels its important to have we can talk about adding it

PR #901, opened 2021-08-13
en_kusa documentation

Dragorn421:
Bushes.

It looks like they really wanted to save some ram, some rotation values used for animation are in static variables and the "bush is cut" flag is an actor flag (or maybe it's just bad programming 😅 )

### DIFF
--- a/src/overlays/actors/ovl_En_Kusa/z_en_kusa.c
+++ b/src/overlays/actors/ovl_En_Kusa/z_en_kusa.c
@@ -194,8 +194,8 @@ void EnKusa_SpawnFragments(EnKusa* this, GlobalContext* globalCtx) {
 
         scaleIndex = (s32)(Rand_ZeroOne() * 111.1f) & 7;
 
-        EffectSsKakera_Spawn(globalCtx, &pos, &velocity, &pos, -100, 64, 40, 3, 0, sFragmentScales[scaleIndex], 0, 0, 80,
-                             KAKERA_COLOR_NONE, OBJECT_GAMEPLAY_KEEP, gCuttableShrubStalkDL);
+        EffectSsKakera_Spawn(globalCtx, &pos, &velocity, &pos, -100, 64, 40, 3, 0, sFragmentScales[scaleIndex], 0, 0,
+                             80, KAKERA_COLOR_NONE, OBJECT_GAMEPLAY_KEEP, gCuttableShrubStalkDL);
 
         pos.x = this->actor.world.pos.x + (dir->x * this->actor.scale.x * 40.0f);
         pos.y = this->actor.world.pos.y + (dir->y * this->actor.scale.y * 40.0f) + 10.0f;
@@ -207,8 +207,8 @@ void EnKusa_SpawnFragments(EnKusa* this, GlobalContext* globalCtx) {
 
         scaleIndex = (s32)(Rand_ZeroOne() * 111.1f) % 7;
 
-        EffectSsKakera_Spawn(globalCtx, &pos, &velocity, &pos, -100, 64, 40, 3, 0, sFragmentScales[scaleIndex], 0, 0, 80,
-                             KAKERA_COLOR_NONE, OBJECT_GAMEPLAY_KEEP, gCuttableShrubTipDL);
+        EffectSsKakera_Spawn(globalCtx, &pos, &velocity, &pos, -100, 64, 40, 3, 0, sFragmentScales[scaleIndex], 0, 0,
+                             80, KAKERA_COLOR_NONE, OBJECT_GAMEPLAY_KEEP, gCuttableShrubTipDL);
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/overlays/actors/ovl_En_Kusa/z_en_kusa.c
```diff
@@ -347,131 +351,132 @@ void EnKusa_SetupLiftedUp(EnKusa* this) {
...
+    EnKusa_SetupAction(this, EnKusa_Regrow);
+    EnKusa_SetScaleSmall(this);
     this->actor.shape.rot = this->actor.home.rot;
     this->actor.flags &= ~0x800;
 }
 
-void func_80A9C068(EnKusa* this, GlobalContext* globalCtx) {
-    s32 sp24;
+void EnKusa_Regrow(EnKusa* this, GlobalContext* globalCtx) {
+    s32 fullyGrown;
```
jpburnett, line 471:
```suggestion
    s32 isFullyGrown;
```

src/overlays/actors/ovl_En_Kusa/z_en_kusa.h
```diff
@@ -8,12 +8,18 @@ struct EnKusa;
...
+    /* 0 */ ENKUSA_TYPE_0,
+    /* 1 */ ENKUSA_TYPE_1,
+    /* 2 */ ENKUSA_TYPE_2
+} EnKusaType;
+
 typedef struct EnKusa {
     /* 0x0000 */ Actor actor;
     /* 0x014C */ EnKusaActionFunc actionFunc;
     /* 0x0150 */ ColliderCylinder collider;
     /* 0x019C */ s16 timer;
```
jpburnett, line 21:
```suggestion
    /* 0x019C */ s16 regrowTimer;
```
Since this timer is used in reference to when the bush regrows, would specifying what it times be useful?

Dragorn421: It gets reset in the generic SetupAction function and ticks up in the generic update function, so I wanted to call it `timerSinceLastActionChange` but that's too long 😅
I feel like the fact it appears in non-regrow code warrants not naming it after the regrow mechanic but idk

fig02: sounds like a general purpose timer to me, which means the name is just fine imo

fig02: can add a `// gets reset every action change` next to the var possibly

jpburnett: Looking closer, I can see what you mean, helped clear things up thanks :D

src/overlays/actors/ovl_En_Kusa/z_en_kusa.c
```diff
@@ -8,38 +8,43 @@
...
+#include "objects/object_kusa/object_kusa.h"
 #include "vt.h"
 
 #define FLAGS 0x00800010
 
 #define THIS ((EnKusa*)thisx)
 
+#define TYPE(params) ((params) & 3)
+#define SPAWN_BUGS ((this->actor.params >> 4) & 1)
+#define DROPS ((this->actor.params >> 8) & 0xF)
```
mzxrules, line 20:
```suggestion
#define ENKUSA_TYPE(params) ((params) & 3)
#define ENKUSA_SPAWN_BUGS(params) ((params >> 4) & 1)
#define ENKUSA_DROPS(params) ((params >> 8) & 0xF)
```
We haven't technically agreed to a style for `params` macros, but I feel like writing them like this is more appropriate for now.

fig02: agreed. would also prefer we just wait till we standardize a way to do this, as mentioned in the other pr

src/overlays/actors/ovl_En_Kusa/z_en_kusa.c
```diff
@@ -347,131 +350,135 @@ void EnKusa_SetupLiftedUp(EnKusa* this) {
...
 
-void func_80A9BC1C(EnKusa* this, GlobalContext* globalCtx) {
+void EnKusa_Fall(EnKusa* this, GlobalContext* globalCtx) {
     s32 pad;
     Vec3f contactPos;
 
-    if (this->actor.bgCheckFlags & 11) {
-        if (!(this->actor.bgCheckFlags & 32)) {
+    if (this->actor.bgCheckFlags & (1 | 2 | 8)) {
+        if (!(this->actor.bgCheckFlags & 0x20)) {
```
fig02, line 379: would prefer if it was just left as hex in all cases tbh, to make for easier search and replacing when we make defines

src/overlays/actors/ovl_En_Kusa/z_en_kusa.h
```diff
@@ -8,12 +8,18 @@ struct EnKusa;
...
-    /* 0x019E */ s8 kusaTexObjIndex;
+    /* 0x019E */ s8 objBankIndex;
```
fig02, line 22: think itd be a good idea to avoid having members named the same as one in the base actor struct

PR #902, opened 2021-08-14
Fix SKYBOX_HOUSE_LINK enum value.

ProjectRevoTPP:
It was recently discovered thanks to the efforts to enable GCC support that this value caused a non equivalent change in Skybox_Setup which caused the skybox image to not get loaded. This value is supposed to be 7 rather than 6. This PR corrects the value so the function recompiles correctly.


louist103 (approved):

include/z64.h
```diff
@@ -295,7 +295,7 @@ typedef enum {
     /* 0x03 */ SKYBOX_OVERCAST_SUNSET,
     /* 0x04 */ SKYBOX_MARKET_ADULT,
     /* 0x05 */ SKYBOX_CUTSCENE_MAP,
-    /* 0x06 */ SKYBOX_HOUSE_LINK,
+    /* 0x06 */ SKYBOX_HOUSE_LINK = 7,
```
fig02, line 298: can you change the number in the comment as well

fig02 (approved):

Roman971 (approved):

PR #903, opened 2021-08-15
Update asm-processor and diff.py

Thar0:
This version of asm-processor doesn't drop .mdebug and instead relocates the previously problematic file-relative offsets, so 
`--source` diffing can be used on files that still have `GLOBAL_ASM` blocks.


AngheloAlf:
Could you update asm-differ too? Or that should be a different PR?

Thar0:
Done

Dragorn421 (approved):
if like me your distro doesn't have the cutting-edge binutils version there's also `--source-old-binutils` instead of `--source`

AngheloAlf (approved):

Roman971 (approved):

fig02 (approved):

PR #906, opened 2021-08-16
Implement define tables for objects, actors and effect soft sprites

Roman971:
This follows a discussion we had in discord some time ago and introduces common tables in `include/tables/`, which are used to define enums and anything related in one place, with the same pattern used in sm64 decomp for levels and courses. The basic idea is to include the table where needed with defines configured to generate what we need for each entry in the table (see the new stuff in `z64actor.h` for example).

For now I only implemented it for objects, actors and effects ss, but eventually I think it could be used for other things like scenes and entrances. Items are also a possible candidate which could allow us to unify the item enum with lists like `gItemIcons`, `gItemSlots` and `sItemActionParams`. Get Items could also be a good candidate if we want to unify the GI enum and its table in `z_player.c`, same thing for the GID enum and its table in `z_draw.c`.

I also moved the linker symbol and init vars externs to where they are needed instead of having them in `segment_symbols.h` and `initvars.h`, but this is just to simplify things and improve performance and it's not strictly needed.

Note: The new tables are `.h` files and I put them in `include/tables` because this made the most sense to me, but they could technically be any type and in src/ or a dedicated folder if we preferred.


ethteck (approved):

mzxrules:
Only concern is just Intellisense failing to find the definition for the enum, but it seems to be figuring it out.

fig02:
Took me a second to understand what was going on here, but this is a really cool change. Think if anyone has tried adding an actor by hand yet, you know how annoying it is to figure out the million other things you need to update accordingly. This lets you update it in one place and it automatically updates it in many other places, like the overlay table, enum, init vars etc

Because of that, +1 to using this for items in the future. After working with adding new items already, there are *so* many places you need to add new things, *and* ensure that they line up in item id/action parameter order. Its actually insane. And that one will be less easier to tool (at least imo) than something like adding a new actor

fig02 (approved):

Zelllll (approved):
I like this change, could something similar be done for dmadata?

include/tables/object_table.h
```diff
@@ -0,0 +1,411 @@
+/**
+ * Object Table
+ *
+ * DEFINE_OBJECT should be used for normal objects
+ *    - Argument 1: Name of the object segment in spec
+ *    - Argument 2: Enum value for this object
+ *
+ * DEFINE_OBJECT_UNSET and DEFINE_OBJECT_NULL are needed to define empty entries from the original game
+ */
+/* 0x0000 */ DEFINE_OBJECT_UNSET(OBJECT_UNSET_0)
```
Dragorn421, line 10: I have a vague memory that object 0 cannot be used but I don't remember how I would know that

I looked at the code and this may back it up:

https://github.com/zeldaret/oot/blob/7551dc2b71f7c6587f96c498cbd62cfd05aa292c/src/code/z_scene.c#L86

another hint may be https://github.com/zeldaret/oot/blob/7551dc2b71f7c6587f96c498cbd62cfd05aa292c/src/code/z_scene.c#L279

I can't really get a definitive opinion on if object id 0 is usable or not from code (but I kinda remember not being able to use it for some reason), if it indeed isn't it would be nice to document it here

Dragorn421 (commented):
Great work, it looks incredibly better and easier to use modding-wise

I'm not sure about putting this in `include` since it (indirectly) represents data that makes it into the rom. I do feel like `src` would be a better place for the `tables` folder to live in

A dedicated folder would look out of place I think (eg if it was called "data") since it wouldn't have all the various tables and data scattered throughout `src/`

I would have suggested using those tables in the spec/dma table too but the order isn't the same :(

fig02:
I dont feel too strongly about it being in include. Its not the actual data to me, just helps construct it. And it does get included into the places that it needs to go in the actual data.
But ye I understand why it might feel a bit weird

some more opinions would be good

mzxrules:
Considering you can't compile the file into an obj, I don't know why you'd want it in `src/`

fig02:
yea thats kinda enough reason for me to keep it in include personally

Dragorn421:
In my opinion `include/` should only have typical header files about structs and function prototypes (which describe stuff), while this table data is not about defining anything you'd expect in a header file, it represents data (doesn't describe anything like a typical header).

An assumption often made about header files that these table.h breaks, is being able to include a header file (almost) anywhere without directly changing the compiled object. When editing these table.h files the changes would directly impact the data in compiled objects and rom. As I already said, these tables do end up in the ROM (contrarily to all other files in `include/`) which is why I'd argue `src/` makes more sense

*But editing structs in includes also changes the compiled objects*: well yeah the includes can't have no bearing on code obviously, however changing struct offsets will have smaller (and transparent) impacts on code sections, while editing table.h files would have direct and meaningful impacts on data sections

fig02:
Think we will need some more opinions other than us 3 before moving forward.

The topic of discussion for anyone not following so far is about the placement of the `tables` folder, which contains the information used to construct various tables and enums in the codebase.
Currently roman has that in `/include`, but some people feel this is not appropriate. Read context above for more information

Dragorn421:
To summarize the discussion on Discord (starting at https://discord.com/channels/688807550715560050/752770632357511239/877998519980523560 ):

No consensus was reached about `include/` vs `src/` (among the few people who were there)

`.inc.c` would be a preferred file extension (but no real consensus either)

The few people who were there seemed to agree on the idea of putting these table files and future ones in a `tables/` folder at the root level.
Fig:
> This is warranted for me cause down the line there will be many tables in the folder that modders will be wanting to modify (either by hand or with a tool), and having easy access to them will be convenient. And it also acts as a compromise on the include vs src debate.

The discussion isn't closed of course

include/tables/object_table.h
Roman971: I think you're right that object id 0 is no good because the system uses the negative value of the id to mark an object as not being loaded yet, so I 0 could easily cause issues even if it somehow worked

fig02:
Pretty stuck on how to make everyone happy with this tbh
Unfortunately felt like in our discussion everyone was kinda talking past each other about what they think is right, without working toward a solution. But I may just be misjudging things.

A possible idea is to keep it the way it is now and see how it goes in practice? @Dragorn421 you said before that its likely not to change after its decided, but I think I disagree a bit. If we find the current configuration doesnt work for everyone I think we could make changes pretty easily. But I may be missing something

include/tables/object_table.h
Roman971: I can add a comment about this in the table, but maybe we should change the enum value for 0 to `OBJECT_NONE` or something and potentially use that in the code where there is `id = 0`?

Dragorn421 (approved):
Roman (what a hero) single-handedly turned my opinion around:
> including a file from src/ in a header located in include/ like z64actor.h feels really wrong to me

Approving for dramatic effect but in reality there's still the object 0 thing to address (just a comment and an appropriate enum name)

Roman971:
Made the object 0 changes based on the discussion above

Dragorn421 (approved):

fig02:
To summarize discord convo: I think the best course of action is to leave it in `include/` like it is now, and revisit it at a later time if people still feel strongly about this after using it a bit in practice.
Will be merging soon, but will give time for anyone to speak up who hasnt seen our recent conversations.

Roman971:
I renamed it to `OBJECT_INVALID` which should be enough to make it clear I think

Dragorn421 (approved):

PR #909, opened 2021-08-18
Remove fake match in EnInsect

louist103:
This fake match did not change IDO code gen but was causing a GCC build to crash if the player dropped bugs far away from a soft soil spot.


fig02:
going to merge without waiting for approval because of how small it is

AngheloAlf:
> going to merge without waiting for approval because of how small it is

_Jenkins fails_

fig02:
approval != jenkins FeelsOkMan

PR #910, opened 2021-08-18
Remove and change some fake matches

louist103:
Removes some fake matches and change ones that cant into `if(1)` when possible.

### DIFF
--- a/src/overlays/actors/ovl_En_Insect/z_en_insect.c
+++ b/src/overlays/actors/ovl_En_Insect/z_en_insect.c
@@ -590,8 +590,6 @@ void func_80A7D460(EnInsect* this, GlobalContext* globalCtx) {
         sp40 = 40.0f;
     }
 
-    if (!this->soilActor->actor.params) {}
-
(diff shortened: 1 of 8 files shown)


src/overlays/actors/ovl_En_Insect/z_en_insect.c
```diff
@@ -164,11 +164,12 @@ void func_80A7C058(EnInsect* this) {
     }
 }
 
-void EnInsect_Init(Actor* thisx, GlobalContext* globalCtx) {
+void EnInsect_Init(Actor* thisx, GlobalContext* globalCtx2) {
     EnInsect* this = THIS;
     f32 rand;
     s16 temp_s2;
     s32 count;
+    GlobalContext* globalCtx = globalCtx2;
```
fig02, line 172: can this be near the top of the block, either above or below `this`

Dragorn421 (approved):
fig is the fastest merger in the west, they call him Fig "Green Clicker" 02

fig02 (approved):

Roman971 (approved):

PR #915, opened 2021-08-21
Move to the spec the cutscenes of Bg_Treemough and En_Zl1

AngheloAlf:


fig02 (approved):

EllipticEllipsis (approved):

Dragorn421 (approved):

src/overlays/actors/ovl_Bg_Treemouth/z_bg_treemouth.c
```diff
@@ -26,7 +26,10 @@ void func_808BC8B8(BgTreemouth* this, GlobalContext* globalCtx);
 void func_808BC9EC(BgTreemouth* this, GlobalContext* globalCtx);
 void func_808BCAF0(BgTreemouth* this, GlobalContext* globalCtx);
 
-#include "z_bg_treemouth_cutscene_data.c" EARLY
+extern CutsceneData D_808BCE20[];
+extern CutsceneData D_808BD2A0[];
+extern CutsceneData D_808BD520[];
+extern CutsceneData D_808BD790[];
```
Dragorn421, line 32: I think it's out of the scope of this PR but would we want to have these symbols in a `z_bg_treemouth_cutscene_data.h` header file

Thinking ahead about extraction (?) and tooling it seems to make sense

(also github ate this comment in my approval review)

Roman971: I agree this is a bit out of scope so we should probably discuss it in discord

Roman971 (approved):

fig02:
So zl1 has a bunch of Camera float data in it that isnt really relevant to the cutscene script. Imo we should probably move this back to the overlay and have the separately linked file be just the cs script. But curious what others think

Roman971:
I don't really have an opinion about this, so I'm fine with either. But if our plan is to have all cutscenes in their own file then I agree with moving it back to the overlay or at least another file.

fig02:
My idea mainly is that a tool generates the cutscene script and links it as its own thing. Obviously we know this doesnt hold true in the whole codebase, cause some actors have it in the middle of their data section without a file boundary. But I think it makes sense to hold that organizational structure when possible

would be fine with a seperate file for the float data as well if people felt it was too intrusive. But for an already pretty irrelevant actor I dont really care too much either way

Roman971:
Yeah moving it back to the overlay seems fine to me in this case

PR #916, opened 2021-08-21
Match func_800EEA50 + some surrounding doc

zelda2774:


include/z64audio.h
```diff
@@ -1068,13 +1068,13 @@ typedef struct {
...
-    BANK_SYSTEM,
-    BANK_OCARINA,
-    BANK_VOICE
+    BANK_PLAYER = 0,
+    BANK_ITEM = 1,
+    BANK_ENV = 2,
+    BANK_ENEMY = 3,
+    BANK_SYSTEM = 4,
+    BANK_OCARINA = 5,
+    BANK_VOICE = 6
```
louist103, line 1077: We usually just comment the index for these.

src/code/code_800EC960.c
```diff
@@ -3,91 +3,105 @@
...
+    PAGE_INTERFACE_INFO = 7,
+    PAGE_SE_FLAG_SWAP = 8,
+    PAGE_BLOCK_CHANGE_BGM = 9,
+    PAGE_NATURAL_SOUND_CONTROL = 10, // unused
+    PAGE_OCARINA_TEST = 11,
+    PAGE_SE_PARAMETER_CHANGE = 12,
+    PAGE_SCROLL_PRINT = 13,
+    PAGE_FREE_AREA = 14,
+    PAGE_LIM = 15
+};
```
louist103, line 44: Same here with the comments instead of setting them like this

louist103 (changes requested):
Don't really know much about audio so this is all I saw.

include/z64audio.h
EllipticEllipsis: Specifically, on the left. You can see plenty of examples in the codebase.

src/code/code_800F7260.c
```diff
@@ -60,34 +60,34 @@ u8 D_801333CC = 0;
...
 
     for (bankId = 0; bankId < ARRAY_COUNT(gSoundBanks); bankId++) {
         if (arg0 & 1) {
-            D_8016E264[bankId] = 1;
+            gSoundBankMuted[bankId] = true;
         } else {
-            D_8016E264[bankId] = 0;
+            gSoundBankMuted[bankId] = false;
         }
         arg0 = (arg0 >> 1) & 0xFFFF;
```
Dragorn421, line 92:
```suggestion
        arg0 = arg0 >> 1;
```
for whatever reason `arg0 >>= 1;` doesn't match

src/code/code_800EC960.c
```diff
@@ -1151,336 +1159,319 @@ void func_800EE930(void) {
...
     Input inputs[4];
     u32 btn;
 
-    PadMgr_RequestPadData(&gPadMgr, &inputs, 0);
+    PadMgr_RequestPadData(&gPadMgr, inputs, 0);
     btn = inputs[3].cur.button;
-    D_8016BAB0 = btn & 0xFFFF;
-    sDebugPadPress = (btn ^ sDebugPadPressRaw) & btn;
-    sDebugPadPressRaw = btn;
+    sDebugPadHold = btn & 0xFFFF;
```
Dragorn421, line 1168:
```suggestion
    sDebugPadHold = (u16)btn;
```

src/code/code_800EC960.c
```diff
@@ -1151,336 +1159,319 @@ void func_800EE930(void) {
...
-                            GfxPrint_Printf(printer, "%2X %5d %5d %5d %02X %04X %04X", idx, (s32)*entry->posX,
-                                            (s32)*entry->posY, (s32)*entry->posZ, entry->unk_24, entry->unk_26,
-                                            entry->unk_28);
+                GfxPrint_SetPos(printer, 2 + sAudioIntInfoX, 4 + ind + sAudioIntInfoY);
+                GfxPrint_Printf(printer, "%s <%d>", sSoundBankNames[k], sAudioIntInfoBankPage[k]);
+
+                for (k2 = 0; k2 < D_80130578[D_801333CC][k]; k2++) {
+#define idx (D_8016E1B8[k][k2].unk_4)
+#define bank (&gSoundBanks[k][idx])
+#define chan (gAudioContext.seqPlayers[2].channels[bank->unk_2E])
```
Dragorn421, line 1340:
```suggestion
#define bankEntryIndex (D_8016E1B8[k][k2].unk_4)
#define bankEntry (&gSoundBanks[k][idx])
#define chan (gAudioContext.seqPlayers[2].channels[bank->unk_2E])
```

src/code/code_800EC960.c
```diff
@@ -1751,710 +1771,717 @@ void func_800EEA50(GfxPrint* printer) {
...
+        if (sAudioSndContWorkLims[sAudioSndContSel] >= 16) {
+            step = 8;
         }
     } else {
-        if (D_8016BAB0 & 8) {
-            D_80131E08[D_80131CA4] = 0;
+        if (CHECK_BTN_ANY(sDebugPadHold, BTN_CUP)) {
+            sAudioSndContWork[sAudioSndContSel] = 0;
         }
     }
```
Dragorn421, line 1885:
```suggestion
    } else if (CHECK_BTN_ANY(sDebugPadHold, BTN_CUP)) {
        sAudioSndContWork[sAudioSndContSel] = 0;
    }
```

src/code/code_800EC960.c
```diff
@@ -2718,10 +2736,10 @@ u8 func_800F37B8(f32 arg0, SoundBankEntry* arg1, s8 arg2) {
             break;
     }
 
-    if (arg1->unk_1C > 1923.077f) {
+    if (arg1->dist > 1923.077f) {
         phi_f12 = 1923.077f;
     } else {
-        phi_f12 = arg1->unk_1C;
+        phi_f12 = arg1->dist;
     }
```
Dragorn421, line 2743:
```suggestion
    phi_f12 = CLAMP_MAX(arg1->dist, 1923.077f);
```

src/code/code_800EC960.c
```diff
@@ -2820,31 +2837,35 @@ void func_800F3A08(u8 bankIdx, u8 entryIdx, u8 channelIdx) {
...
+    if (reverb != D_8016B8B8[channelIdx].oldReverb) {
         Audio_QueueCmdS8(0x5020000 | (channelIdx << 8), reverb);
-        D_8016B8B8[channelIdx].unk_08 = reverb;
+        D_8016B8B8[channelIdx].oldReverb = reverb;
     }
-    if (freqScale != D_8016B8B8[channelIdx].unk_04) {
+    if (freqScale != D_8016B8B8[channelIdx].oldFreqScale) {
         Audio_QueueCmdF32(0x4020000 | (channelIdx << 8), freqScale);
-        D_8016B8B8[channelIdx].unk_04 = freqScale;
+        D_8016B8B8[channelIdx].oldFreqScale = freqScale;
```
Dragorn421, line 2848: I think `prev` (or `cur` depending on how you look at it) is better than `old`

src/code/code_800EC960.c
```diff
@@ -2909,27 +2930,27 @@ void func_800F4010(Vec3f* arg0, u16 arg1, f32 arg2) {
...
 
 void func_800F4190(Vec3f* pos, u16 sfxId) {
     Audio_PlaySoundGeneral(sfxId, pos, 4, &D_801305B0, &D_801333E0, &D_801305B4);
 }
-void func_800F41E0(Vec3f* pos, u16 baseSfx, u8 arg2) {
-    u8 offset = Audio_NextRandom() % arg2;
-    Audio_PlaySoundGeneral(baseSfx + offset, pos, 4, &D_801333E0, &D_801333E0, &D_801333E8);
+void Audio_PlaySoundRandom(Vec3f* pos, u16 baseSfxId, u8 randLim) {
+    u8 offset = Audio_NextRandom() % randLim;
+    Audio_PlaySoundGeneral(baseSfxId + offset, pos, 4, &D_801333E0, &D_801333E0, &D_801333E8);
```
Dragorn421, line 2953:
```suggestion
    u8 offset = Audio_NextRandom() % randLim;

    Audio_PlaySoundGeneral(baseSfxId + offset, pos, 4, &D_801333E0, &D_801333E0, &D_801333E8);
```

src/code/code_800EC960.c
```diff
@@ -3210,116 +3226,116 @@ void func_800F4E30(Vec3f* arg0, f32 arg1) {
...
-        func_800F5E18(3, arg1, 0, 7, 2);
+    dist = sqrtf(SQ(pos->z) + SQ(pos->x));
+    if (sSariaBgmPtr == NULL) {
+        sSariaBgmPtr = pos;
+        func_800F5E18(3, seqId, 0, 7, 2);
     } else {
-        temp_f0 = sqrtf(SQ(D_8013064C->z) + SQ(D_8013064C->x));
-        if (phi_f14 < temp_f0) {
-            D_8013064C = arg0;
+        oldDist = sqrtf(SQ(sSariaBgmPtr->z) + SQ(sSariaBgmPtr->x));
```
Dragorn421, line 3299:
```suggestion
        prevDist = sqrtf(SQ(sSariaBgmPtr->z) + SQ(sSariaBgmPtr->x));
```

src/code/code_800F7260.c
```diff
@@ -361,101 +358,101 @@ void func_800F7CEC(u8 arg0) {
...
+            if (entry->unk_26 & 0x10) {
+                entry->unk_20 = SQ(0xFF - spA9) * 0x1690;
             } else {
-                if (temp_s2->unk_1C > 0x7FFFFFD0) {
-                    temp_s2->unk_1C = 0x70000008;
-                    osSyncPrintf(D_80133344, temp_s2->unk_28, temp_s2->posX, temp_s2->posZ, *temp_s2->posX,
-                                 *temp_s2->posY, *temp_s2->posZ);
+                if (entry->dist > 0x7FFFFFD0) {
+                    entry->dist = 0x70000008;
+                    // VT_COL(RED, WHITE) "<INAGAKI CHECK> dist over! flag:%04X ptr:%08X pos:%f-%f-%f" VT_RST "\n"
```
Dragorn421, line 375:
```suggestion
```
I assume this will be improved once the .data situation improves, in the mean time people can just look at `D_80133344`

src/overlays/actors/ovl_En_Fr/z_en_fr.c
```diff
@@ -581,15 +581,15 @@ void EnFr_UpdateActive(Actor* thisx, GlobalContext* globalCtx) {
...
 
     if (frog != NULL && frog->isJumpingUp == false) {
-        audioPitch = frog->growingScaleIndex == 3 ? sPitchLargeFrog[frogIndex] : sPitchSmallFrog[frogIndex];
+        semitone = frog->growingScaleIndex == 3 ? sLargeFrogNotes[frogIndex] : sSmallFrogNotes[frogIndex];
         if (this->songIndex == FROG_CHOIR_SONG) {
             frog->isJumpingToFrogSong = true;
         }
         frog->isJumpingUp = true;
-        func_800F4BF4(&frog->actor.projectedPos, NA_SE_EV_FROG_JUMP, (s8)audioPitch);
+        Audio_PlaySoundTransposed(&frog->actor.projectedPos, NA_SE_EV_FROG_JUMP, (s8)semitone);
```
Dragorn421, line 592:
```suggestion
        Audio_PlaySoundTransposed(&frog->actor.projectedPos, NA_SE_EV_FROG_JUMP, semitone);
```

src/overlays/actors/ovl_En_In/z_en_in.c
```diff
@@ -597,7 +597,7 @@ void func_80A79FB0(EnIn* this, GlobalContext* globalCtx) {
 void func_80A7A304(EnIn* this, GlobalContext* globalCtx) {
     if (this->skelAnime.animation == &D_06015814 || this->skelAnime.animation == &D_0601646C) {
         if (this->skelAnime.curFrame == 8.0f) {
-            func_800F41E0(&this->actor.projectedPos, NA_SE_VO_IN_LASH_0, 2);
+            Audio_PlaySoundRandom(&this->actor.projectedPos, NA_SE_VO_IN_LASH_0, 2);
```
Dragorn421, line 600:
```suggestion
            Audio_PlaySoundRandom(&this->actor.projectedPos, NA_SE_VO_IN_LASH_0,
                                  NA_SE_VO_IN_LASH_1 - NA_SE_VO_IN_LASH_0 + 1);
```
to help documenting that the two sfx ids should be in a row?

src/overlays/actors/ovl_En_Zl3/z_en_zl3.c
```diff
@@ -1018,7 +1018,7 @@ void func_80B55808(EnZl3* this) {
 }
 
 void func_80B5582C(EnZl3* this) {
-    func_800F41E0(&D_80B5A488, NA_SE_VO_Z1_CRY_0, 2);
+    Audio_PlaySoundRandom(&D_80B5A488, NA_SE_VO_Z1_CRY_0, 2);
```
Dragorn421, line 1021:
```suggestion
    Audio_PlaySoundRandom(&D_80B5A488, NA_SE_VO_Z1_CRY_0, NA_SE_VO_Z1_CRY_1 - NA_SE_VO_Z1_CRY_0 + 1);
```
same
I would suggest a macro if there was more usage but there are only two uses so shrug

Dragorn421 (changes requested):
This is an impressive amount of documentation (fueled by matching `AudioDebug_Draw` I assume?), fantastic work!

Unfortunately some (a lot?) of it flies over my head but I do pick up on some parts

src/code/code_800EC960.c
zelda2774: dropped the prefix completely

src/code/code_800EC960.c
zelda2774: Don't really agree with this, sDebugPadHold is a u32. We're intentionally throwing away bits but not for the purpose of fitting into a smaller data type

src/code/code_800EC960.c
```diff
@@ -3,91 +3,105 @@
...
+extern u32 D_80130638;               // = 0
+extern u8 D_8013063C;                // = 0
+extern u8 D_80130640;                // = 0
+extern u8 D_80130644;                // = 0
+extern u8 D_80130648;                // = 0
+extern Vec3f* sSariaBgmPtr;          // = 0
+extern f32 D_80130650;               // = 2000.0f
+extern u8 D_80130654;                // = 0
+extern u8 D_80130658[];              // 0x6D <= size <= 0x70
+extern s8 sSpecReverbs[20];          // = {0, 0, 0, 0, 0, 0, 0, 40, 0, 15, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
```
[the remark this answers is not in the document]
zelda2774: there's nothing "asset"y about this data, so it should be fine to write it out in C (and will need to be so once data is migrated, this is just in preparation of that). I'm sure the incbins were just because the work to identify data types wasn't done yet.

src/code/code_800EC960.c
zelda2774: went with entryIndex/entry, bankEntry gets a bit too verbose

src/code/code_800EC960.c
```diff
@@ -3,91 +3,105 @@
...
+    PAGE_CHANNEL_INFO = 6, // unused
+    PAGE_INTERFACE_INFO = 7,
+    PAGE_SE_FLAG_SWAP = 8,
+    PAGE_BLOCK_CHANGE_BGM = 9,
+    PAGE_NATURAL_SOUND_CONTROL = 10, // unused
+    PAGE_OCARINA_TEST = 11,
+    PAGE_SE_PARAMETER_CHANGE = 12,
+    PAGE_SCROLL_PRINT = 13,
+    PAGE_FREE_AREA = 14,
+    PAGE_LIM = 15
```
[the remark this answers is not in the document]
zelda2774: Done, though I'm not completely sold on the name "MAX"... it sounds like it equals the max of all real enum variants while in reality it's that + 1.

src/code/code_800EC960.c
```diff
@@ -3,91 +3,105 @@
...
-    f32 unk_00;
-    f32 unk_04;
-    f32 unk_08;
-    s32 unk_0C;
-} unk_s2;
+    f32 value;
+    f32 target;
+    f32 step;
+    s32 remainingFrames;
+} Lerp;
```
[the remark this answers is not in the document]
zelda2774: It's local to the file, so the risk seems small to me, but changed to FreqLerp

src/code/code_800EC960.c
```diff
@@ -3774,41 +3785,41 @@ void func_800F6828(u8 arg0) {
     }
 }
 
-void func_800F68BC(s8 arg0) {
-    D_8013060C = arg0;
+void Audio_SetDemoActive(s8 active) {
+    sAudioDemoActive = active;
 }
```
fig02, line 3790: even though they originally called cutscenes "demos" we generally try to avoid it when possible.
is this only used for scripted cutscenes? (ie those controlled by z_demo.c), or is it "cutscene" in a broader sense? there are a few different kind of cutscenes in the game

src/code/code_800EC960.c
```diff
@@ -3774,41 +3785,41 @@ void func_800F6828(u8 arg0) {
...
 
-void func_800F68D4(u16 arg0, Vec3f* arg1, u8 arg2, u32* arg3, u32* arg4, u32* arg5) {
-    if (D_8013060C == 0) {
-        Audio_PlaySoundGeneral(arg0, arg1, arg2, arg3, arg4, arg5);
+void Audio_PlaySoundGeneralIfNotDemo(u16 sfxId, Vec3f* pos, u8 arg2, f32* freqScale, f32* arg4, s8* reverbAdd) {
+    if (!sAudioDemoActive) {
```
fig02, line 3793: think the name is a bit verbose, but we can look into improving it at a later time.
Am curious on your thoughts of `PlaySoundGeneral` in general though. this function was named super early on with almost no knowledge of the audio system afaik. So Im curious if we have a better name for it

src/code/code_800EC960.c
```diff
@@ -1751,710 +1771,715 @@ void func_800EEA50(GfxPrint* printer) {
...
-            GfxPrint_Printf(printer, "DEMO FLAG %d", D_8013060C);
+            GfxPrint_Printf(printer, "DEMO FLAG %d", sAudioDemoActive);
 
             GfxPrint_SetPos(printer, 3, 12);
-            if (D_8016BAA8 == 1) {
-                GfxPrint_Printf(printer, "MARON BGM DIST %f", D_8016BAAC);
-                D_8016BAA8 = 0;
+            if (sAudioHasMaronBgm == true) {
+                GfxPrint_Printf(printer, "MARON BGM DIST %f", sAudioMaronBgmDist);
+                sAudioHasMaronBgm = false;
```
fig02, line 1843: ah I see, these names are coming from stings lol
In this case Maron would be Malon. But I think we should probably try to avoid just blindly using the strings, and name them on our own terms. But yeah if we want to do this in another pass id be fine with it

src/code/code_800EC960.c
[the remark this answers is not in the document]
fig02: afaik the _MAX convention started because its what they originally did in some cacses
`ASSERT(gMaxActorId == ACTOR_ID_MAX, "MaxProfile == ACTOR_DLF_MAX", "../z_actor.c", UNK_LINE);`

im fine with it personally, but some people have been confused from it in the past. so it may be worth talking about again

src/code/code_800EC960.c
zelda2774: Oh, haha, shows how well I know the game for not noticing. I think in this case it's probably a decent name with the misspelling fixed. But yeah, none of the symbol of function names here are final, this code will need several more doc passes.

src/code/code_800EC960.c
fig02: mainly wondering things like: does `Audio_PlaySfx` make sense? is the word general needed? if playsfx is too broad, is there a way to make it a bit more specific other than saying the word "general"?
stuff like that

zelda2774: I think `PlaySoundGeneral` is a decent name (I read the "General" as "can play any sfxId, with any sound settings"), but what we should do is add some macros for `Audio_PlaySoundGeneral(sfxId, pos, 4, &D_801333E0, &D_801333E0, &D_801333E8)` (playing sound with default pitch and reverb) and `Audio_PlaySoundGeneral(sfxId, &D_801333D4, 4, &D_801333E0, &D_801333E0, &D_801333E8)` (default source position). (The `4` in those calls doesn't mean anything btw, the parameter is completely ignored afaict.)

The naming of sfx vs sound is a good question. I think "sfx" *might* mean something else internally within audio, like a separate kind of instrument from "program"/"drum"? But I don't know too much about that and whether it could be confusing. Certainly given the name "sfxId" it would be good to unify terminology here...

src/code/code_800EC960.c
zelda2774: Only used from z_demo, yes. Would Audio_SetCutsceneActive be the right name for that, or do we want something more specific?

src/code/code_800EC960.c
fig02: Alright cool. yea as we get further along with audio docs I think we'll be able to unify things better. just something to keep in the back of our minds as we keep documenting stuff

src/code/code_800EC960.c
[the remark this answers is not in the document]
EllipticEllipsis: `COUNT` or something related would be an obvious alternative (as in `ARRAY_COUNT`).

src/code/code_800EC960.c
Dragorn421: "Cutscene" sounds better but I'm not so sure about the "Active" part, I would just use generic "Flag" like `SetCutsceneFlag`, like what the debug string has (`DEMO FLAG`) and wait for further documentation to better understand the contexts where that function is used

fig02: that sounds fine for now. my main concern is distinguishing a scripted cutscene from stuff like one point demos or player cutscenes, but I think anything labeled as a generic cutscene currently is assumed to mean the script system (ie CutsceneContext and related things)

src/code/code_800EC960.c
```diff
@@ -1751,710 +1771,715 @@ void func_800EEA50(GfxPrint* printer) {
...
             GfxPrint_Printf(printer, "SWITCH BGM MODE %d %d %d (FLAG %d)", D_8013061C, D_80130634, D_80130638,
                             D_80130654);
 
             GfxPrint_SetPos(printer, 3, 10);
-            GfxPrint_Printf(printer, "ENEMY DIST %f VOL %3d", D_80130620, D_80130624);
+            GfxPrint_Printf(printer, "ENEMY DIST %f VOL %3d", sAudioEnemyDist, sAudioEnemyVol);
 
             GfxPrint_SetPos(printer, 3, 11);
-            GfxPrint_Printf(printer, "GANON DIST VOL %3d", D_8016B8B4);
+            GfxPrint_Printf(printer, "GANON DIST VOL %3d", sAudioGanonDistVol);
```
fig02, line 1835: another reason to be wary of just using string names verbatim: what does this really mean? they sometimes refer to ganondorf as ganon (since thats the actor name). and looking at uses of this, I can only see it being touched by the river_sound actor, which afaik has nothing to do with ganon. unless the ganon fight uses river sound for something that I have never learned about before

fig02 (approved):

include/functions.h
```diff
@@ -2171,21 +2150,19 @@ void func_800F5B58(void);
...
+void Audio_SetEnvReverb(s8 reverb);
+void Audio_SetCodeReverb(s8 reverb);
 void func_800F6700(s8 outputMode);
 void func_800F67A0(u8);
 void func_800F6828(u8);
-void func_800F68BC(s8);
-// ? func_800F68D4(?);
-// ? func_800F691C(?);
+void Audio_SetCutsceneFlag(s8 flag);
+void Audio_PlaySoundIfNotDemo(u16 sfxId);
```
Dragorn421, line 2165:
```suggestion
void Audio_PlayCutsceneSound(u16 sfxId);
```
at the very least Demo -> Cutscene but I think this name looks better too
`Audio_PlayCutsceneOnlySound` may be more explicit if the cutscene flag really means "currently in a cutscene", idk

include/functions.h
```diff
@@ -2199,12 +2176,12 @@ void func_800F70F8();
...
 // ? func_800F71BC(?);
-void func_800F7260(u16);
+void Audio_SetSoundBanksMute(u16 muteMask);
 void Audio_QueueSeqCmdMute(u8);
 void Audio_ClearBGMMute(u8);
-void Audio_PlaySoundGeneral(u16 sfxId, Vec3f* a1, u8 a2, f32* a3, f32* a4, s8* a5);
-// ? func_800F74E0(?);
-void func_800F7680(void);
+void Audio_PlaySoundGeneral(u16 sfxId, Vec3f* pos, u8 a2, f32* freqScale, f32* a4, s8* reverbAdd);
+// ? Audio_RemoveMatchingSoundRequests(?);
```
Dragorn421, line 2183:
```suggestion
```

src/code/code_800EC960.c
```diff
@@ -252,7 +260,7 @@ s32 D_8016BA1C;
 u8 sCurOcarinaSong[8];
 u8 sOcarinaSongAppendPos;
 u8 sOcarinaHasStartedSong;
-u8 gOcarinaSongNotestartIdx;
+u8 sOcarinaSongNotestartIdx;
```
Dragorn421, line 263:
```suggestion
u8 sOcarinaSongNoteStartIdx;
```

src/code/code_800EC960.c
```diff
@@ -3774,41 +3786,41 @@ void func_800F6828(u8 arg0) {
...
-void func_800F68BC(s8 arg0) {
-    D_8013060C = arg0;
+void Audio_SetCutsceneFlag(s8 flag) {
+    sAudioCutsceneFlag = flag;
 }
 
-void func_800F68D4(u16 arg0, Vec3f* arg1, u8 arg2, u32* arg3, u32* arg4, u32* arg5) {
-    if (D_8013060C == 0) {
-        Audio_PlaySoundGeneral(arg0, arg1, arg2, arg3, arg4, arg5);
+void Audio_PlaySoundGeneralIfNotDemo(u16 sfxId, Vec3f* pos, u8 arg2, f32* freqScale, f32* arg4, s8* reverbAdd) {
```
Dragorn421, line 3793: same as `Audio_PlaySoundIfNotDemo`

src/code/code_800F7260.c
```diff
@@ -32,9 +33,8 @@ char D_80133390[] = "SEQ H";
 char D_80133398[] = "    L";
 
 // data
-u8 D_801333A0 = 0;
-
-u8 D_801333A4 = 0;
+u8 sSoundRequestHead = 0;
+u8 sSoundRequestTail = 0;
```
Dragorn421, line 37: are you sure these aren't swapped?

`Audio_ProcessSoundRequests` does `sSoundRequestTail++`
and `Audio_PlaySoundGeneral` does `sSoundRequestHead++`

that seems weird but maybe I'm mistaken

include/functions.h
zelda2774: Oops, forgot to change that. Audio_PlayCutsceneSound would indicate the opposite thing, but I'll change it to PlaySoundIfNotInCutscene. It's only called from one place so the name doesn't have to be perfect.

src/code/code_800F7260.c
zelda2774: Question of definitions. My intuition is that the tail lags behind the head, and the head is where it extends forwards, like in Snake. :snake:

Dragorn421: took me a few minutes but I get how you see it now, then to prevent confusion what about
`sSoundRequestTail` -> `sSoundRequestFirst` (or Start)
`sSoundRequestHead` -> `sSoundRequestNext`

I think head/tail is confusing because traversal starts at the tail which at least for me is counter-intuitive

zelda2774: Changed to ReadIndex/WriteIndex. (But I really do like the concise head/tail terminology personally.)

src/code/code_800EC960.c
```diff
@@ -3774,41 +3786,41 @@ void func_800F6828(u8 arg0) {
...
 void func_800F6964(u16 arg0) {
-    s32 phi_v1;
+    s32 skip;
     u8 i;
 
     Audio_SeqCmd1(0, (arg0 * 3) / 2);
     Audio_SeqCmd1(1, (arg0 * 3) / 2);
     for (i = 0; i < 0x10; i++) {
-        phi_v1 = 0;
+        skip = 0;
```
Dragorn421, line 3810:
```suggestion
        skip = false;
```

src/code/code_800F7260.c
```diff
@@ -361,101 +360,100 @@ void func_800F7CEC(u8 arg0) {
...
+                entry->unk_20 = (u32)entry->dist + (SQ(0xFF - spA9) * 0x1690) + temp3 - temp3;
+                if (*entry->posZ < 0.0f) {
+                    entry->unk_20 += (s32)(-*entry->posZ * 6.0f);
                 }
             }
-            if (temp_s2->unk_1C > 1.0e10f) {
-                if (temp_s2->unk_2A == 4) {
-                    Audio_QueueCmdS8((temp_s2->unk_2E << 8) | 0x6020000, 0);
-                    if (temp_s2->unk_28 & 0xC00) {
+            if (entry->dist > 1.0e10f) {
```
Dragorn421, line 386:
```suggestion
            if (entry->dist > SQ(1.0e5f)) {
```
it's kinda weird how `entry->dist` is sometimes the square value and sometimes the actual value, in some other functions `10000 = SQ(100)`/`2500 = SQ(50)` appear near it but I can't tell if dist is squared or not at those points. Just writing for my curiosity, it can definitely wait for more documentation

src/code/code_800EC960.c
```diff
@@ -2740,58 +2752,59 @@ s8 func_800F3990(f32 arg0, u16 arg1) {
...
-    sp3A = 0;
+    panSigned = 0x40;
+    reverbFlg = 0;
     sp39 = 0;
     sp38 = 0;
     sp33 = 0;
     freqScale = 1.0f;
     sp44 = 1.0f;
-    temp_a3 = &gSoundBanks[bankIdx][entryIdx];
+    entry = &gSoundBanks[bankIdx][entryIdx];
```
Dragorn421, line 2776:
```c
    f32 sp44 = 1.0f;
    s8 phi_a1;
    s8 reverb = 0;
    f32 freqScale = 1.0f;
    s8 panSigned = 0x40;
    u8 reverbFlg = 0;
    u8 sp39 = 0;
    s8 sp38 = 0;
    f32 sp34;
    u8 sp33 = 0;
    SoundBankEntry* entry = &gSoundBanks[bankIdx][entryIdx];

```

src/code/code_800EC960.c
```diff
@@ -2740,58 +2752,59 @@ s8 func_800F3990(f32 arg0, u16 arg1) {
     return ret | 1;
 }
 
-void func_800F3A08(u8 bankIdx, u8 entryIdx, u8 channelIdx) {
+void Audio_SetSoundProps(u8 bankIdx, u8 entryIdx, u8 channelIdx) {
```
Dragorn421, line 2755: does `Props` stand for "properties"? if yes, since it's also an actual word (as in, accessories) I think "Settings" would be better even if slightly longer

(or "Info", "Data"? not exactly sure what does this function do, apply sound info of `entryIdx` in `bankIdx` to `channelIdx`?)

src/code/code_800EC960.c
```diff
@@ -2855,21 +2870,21 @@ void func_800F3ED4(void) {
     for (i = 0; i < 16; i++) {
```
Dragorn421, line 2870:
```suggestion
    for (i = 0; i < ARRAY_COUNT(D_8016B8B8); i++) {
```

src/code/code_800F7260.c
```diff
@@ -60,36 +62,36 @@ u8 D_801333CC = 0;
...
+f32 D_801333E0 = 1.0f; // default freqScale
 
-UNK_TYPE D_801333E4 = 0;
+UNK_TYPE D_801333E4 = 0; // unused
 
-s8 D_801333E8 = 0;
+s8 D_801333E8 = 0; // default reverbAdd
 
-UNK_TYPE D_801333EC = 0;
+UNK_TYPE D_801333EC = 0; // unused
```
Dragorn421, line 73: unused `UNK_TYPE` stuff can just be `s32`, `UNK_TYPE` is more meant as a placeholder when waiting further documentation

Dragorn421 (commented):
I'm nitpicking at this point, just discard whatever comment you want if you feel like a future documentation pass would be better to handle them
(but there's also so much stuff to look at I may as well point out what I see when I see it)

src/code/code_800EC960.c
fig02: if properties is a better fitting word, i dont think its an issue to just spell out the whole thing personally

zelda2774: Yep, it means properties: reverb, volume, pitch, pan, etc. Settings seems like a worse name. I changed props -> properties.

> not exactly sure what does this function do, apply sound info of entryIdx in bankIdx to channelIdx?

it computes sounds properties loosely based on the `entryIdx` sound request, and then applies those properties to a sequence channel. (Sequence players/channels/layers is what produces sound, by playing notes with various properties; here we communicate with that subsystem.)

src/code/code_800F7260.c
zelda2774: It's silly, yeah. The instances of 2500 and 10000 in code_800EC960 appear to come after `entry->dist = sqrtf(entry->dist);`.

src/code/code_800EC960.c
zelda2774: Not sure about this actually: what we really should have is a define for the number of channels in a sequence player. Using ARRAY_COUNT makes it look like this code works regardless of the array size, but then the `D_8016B8B8[0xD].unk_0C = 0;` looks out of place. (I have no idea what that line is doing there fwiw...)

Dragorn421: alright a define sounds better than the array_count macro so scratch that

for `D_8016B8B8[13]` I guess it would be a job for a future `assert(13 < NUM_CHANNELS)`, like many similar things everywhere in the code

src/overlays/actors/ovl_En_G_Switch/z_en_g_switch.c
```diff
@@ -210,14 +210,14 @@ void EnGSwitch_WaitForObject(EnGSwitch* this, GlobalContext* globalCtx) {
 }
 
 void EnGSwitch_SilverRupeeTracker(EnGSwitch* this, GlobalContext* globalCtx) {
-    static s8 rupeePitches[] = { 0, 2, 4, 5, 7 };
+    static s8 majorScale[] = { 0, 2, 4, 5, 7 };
 
-    if (this->pitchIndex < sCollectedCount) {
+    if (this->noteIndex < sCollectedCount) {
         if (sCollectedCount < 5) {
             // sound?
```
Dragorn421, line 217:
```suggestion
```

zelda2774: No, this is just a translation of the japanese string. Should arguably have quotes around I think, but that's a larger change, and would need to be done consistently across the project.

Dragorn421 (approved):

src/overlays/actors/ovl_En_G_Switch/z_en_g_switch.c
fig02: fwiw most debug strings do have quotes on them, and we usually tell people to add them in review. there are certainly a bunch that got missed though

fig02: but yeah, you can add the quotes now if you like

fig02 (approved):

Roman971 (approved):

louist103 (approved):

PR #917, opened 2021-08-22
Decomp the remaining functions in vr_box and import its data

petrie911:
The loop unrolls at the start of each of the new functions are badly non-matching, but I can't figure out how to make it work right. Also fixed a small non-equivalence in DbCamera_Update.


include/z64.h
```diff
@@ -322,7 +322,8 @@ typedef enum {
 typedef struct {
     /* 0x000 */ char unk_00[0x128];
     /* 0x128 */ void* staticSegments[3];
-    /* 0x134 */ Gfx* dListBuf;
+    /* 0x134 */ union{ Gfx* dListBuf;
+                    Gfx (*dListBuf2)[150];};
```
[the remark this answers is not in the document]
Dragorn421: A union is not required if the code currently using `dListBuf` adapts to the `dListBuf2` type (use `&skyboxCtx->dListBuf2[n]` instead of `skyboxCtx->dListBuf + 150 * n`)

src/code/z_vr_box.c
```diff
@@ -32,12 +97,258 @@ typedef struct {
 
 extern Struct_8011FD3C D_8011FD3C[];
 
+#ifdef NON_MATCHING
```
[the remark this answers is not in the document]
petrie911: It certainly seems to be equivalent, as when compiled with these two functions as non-matching everything in the game looks fine.

include/z64.h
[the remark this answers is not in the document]
petrie911: Oh, yeah, that was just a hack to make the indices work out. I'll go back and fix that.

src/code/z_vr_box.c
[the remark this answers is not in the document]
Roman971: I see, then I'd say the comments should be updated to say something like "should be equivalent" and the other function should be made NON_MATCHING as well.

fig02:
this looks good to me other than romans comments needing to be addressed

fig02:
feel like the setupdl thing probably needs some more discussion and shouldnt be included in this set of changes

src/code/z_vr_box.c
```diff
@@ -652,24 +960,24 @@ void Skybox_Init(GlobalContext* globalCtx, SkyboxContext* skyboxCtx, s16 skyboxI
         osSyncPrintf(VT_FGCOL(GREEN));
 
         if (skyboxCtx->unk_140 != 0) {
-            skyboxCtx->dListBuf = GameState_Alloc(&globalCtx->state, 1200 * sizeof(Gfx), "../z_vr_box.c", 1636);
+            skyboxCtx->dListBuf = GameState_Alloc(&globalCtx->state, 8 * 150 * sizeof(Gfx), "../z_vr_box.c", 1636);
             ASSERT(skyboxCtx->dListBuf != NULL, "vr_box->dpList != NULL", "../z_vr_box.c", 1637);
 
-            skyboxCtx->roomVtx = GameState_Alloc(&globalCtx->state, 256 * sizeof(Vtx), "../z_vr_box.c", 1639);
+            skyboxCtx->roomVtx = (Vtx*)GameState_Alloc(&globalCtx->state, 256 * sizeof(Vtx), "../z_vr_box.c", 1639);
```
Roman971, line 966: We generally don't cast the result of alloc functions, so I think we should do the same here
```suggestion
            skyboxCtx->roomVtx = GameState_Alloc(&globalCtx->state, 256 * sizeof(Vtx), "../z_vr_box.c", 1639);
```

src/code/z_vr_box.c
```diff
@@ -652,24 +960,24 @@ void Skybox_Init(GlobalContext* globalCtx, SkyboxContext* skyboxCtx, s16 skyboxI
...
 
             func_800AEFC8(skyboxCtx, skyboxId);
         } else {
-            skyboxCtx->dListBuf = GameState_Alloc(&globalCtx->state, 1800 * sizeof(Gfx), "../z_vr_box.c", 1643);
+            skyboxCtx->dListBuf = GameState_Alloc(&globalCtx->state, 12 * 150 * sizeof(Gfx), "../z_vr_box.c", 1643);
             ASSERT(skyboxCtx->dListBuf != NULL, "vr_box->dpList != NULL", "../z_vr_box.c", 1644);
 
             if (skyboxId == SKYBOX_CUTSCENE_MAP) {
-                skyboxCtx->roomVtx = GameState_Alloc(&globalCtx->state, 192 * sizeof(Vtx), "../z_vr_box.c", 1648);
+                skyboxCtx->roomVtx = (Vtx*)GameState_Alloc(&globalCtx->state, 192 * sizeof(Vtx), "../z_vr_box.c", 1648);
```
Roman971, line 975:
```suggestion
                skyboxCtx->roomVtx = GameState_Alloc(&globalCtx->state, 192 * sizeof(Vtx), "../z_vr_box.c", 1648);
```

src/code/z_vr_box.c
```diff
@@ -652,24 +960,24 @@ void Skybox_Init(GlobalContext* globalCtx, SkyboxContext* skyboxCtx, s16 skyboxI
...
                 ASSERT(skyboxCtx->roomVtx != NULL, "vr_box->roomVtx != NULL", "../z_vr_box.c", 1649);
 
                 func_800AF178(skyboxCtx, 6);
             } else {
-                skyboxCtx->roomVtx = GameState_Alloc(&globalCtx->state, 160 * sizeof(Vtx), "../z_vr_box.c", 1653);
+                skyboxCtx->roomVtx = (Vtx*)GameState_Alloc(&globalCtx->state, 160 * sizeof(Vtx), "../z_vr_box.c", 1653);
```
Roman971, line 980:
```suggestion
                skyboxCtx->roomVtx = GameState_Alloc(&globalCtx->state, 160 * sizeof(Vtx), "../z_vr_box.c", 1653);
```

src/code/z_vr_box_draw.c
```diff
@@ -1,5 +1,7 @@
 #include "global.h"
 
+#define SKYBOX_PALETTES ((u16(*)[256])skyboxCtx->staticSegments[2])
```
Roman971, line 3: I'm not sure this macro really improves things tbh. I feel like it hides important details, and it relies on a specific local variable which we're tring to avoid nowadays.

The previous pointer arithmetic seemed fine to me but we probably need more opinions on this.

Dragorn421: I like the macro much better, I think it is a lot more readable this way

Would you settle on `SKYBOX_PALETTES[n]` -> `SKYBOX_PALETTES(skyboxCtx)[3]` ?

But, does `staticSegments` have to be an array? I know that's what debug strings say it was, but if it can match as individual struct members it would be better

Roman971: I personally think hiding important details for readability is a bad idea, but maybe that's just me.

If we really don't want pointer arithmetic for this specific thing, then a decent compromise would be to change `staticSegments[2]` to an individual member (while keeping the other 2 pointers as a generic staticSegments array).

fig02: Im kind of in the middle on this. I understand wanting to make it more readable, but also feel the same about hiding the details.
I think in this case it doesnt really help with much, there isnt any ambiguity of what it is. They are being directly used in tlut commands, so itd be very clear at first glance that they are palletes.

Dragorn421:
```c
    /* 0x128 */ void* staticSegments[3];
```
also matches as
```c
    /* 0x128 */ void* staticSegments0;
    /* 0x12C */ void* staticSegments1;
    /* 0x130 */ u16 (*skyboxPalettes)[256];
```
would that be fine or do we really need to go with what the debug string says?

fig02: id be fine with this personally

Roman971: I would be okay with the variant of this I suggested above, where the first 2 are kept as an array (which could easily have more entries in a modding scenario)

fig02 (approved):

Dragorn421 (approved):

src/code/z_vr_box.c
```diff
@@ -126,43 +126,104 @@ s32 func_800ADBB0(SkyboxContext* skyboxCtx, Vtx* roomVtx, s32 arg2, s32 arg3, s3
...
             }
+            // for (i = 0; i < 5; i++) {
+            //     for (j = 0; j < 5; j++) {
+            //         sp358[i][j] = arg3 + j * arg6;
+            //         sp2A4[i][j] = arg4 + i * arg7;
+            //         sp1F0[i][j] = arg5;
+            //         sp13C[i][j] = D_8012AD20[j];
+            //         sp88[i][j] = D_8012AD2C[i];
+            //     }
+            // }
```
Dragorn421, line 166: this commented out code and the two other commented out loops look like they should be removed since they rely on more than what's commented out (eg 2d arrays, and idk why i only goes up to 5 instead of 9)

petrie911: I'm leaving it in because it's more clear about what the loop is actually doing. The improved matching code I have in there is pretty hard to read and it's entirely possible the correct code looks nothing like it. The relevant arrays are listed as being `[5 * 5]` and `[9 * 5]`, which should give a hint as to what the 2D array's dimensions would be.

Dragorn421: what about `for (i = 0; i < 5; i++)` using 5 instead of 9

Dragorn421: also it wouldn't hurt leaving a comment to explain the 2d arrays and about this code is equivalent but matches less well

petrie911: oh, right, that was a copy/paste error. Fixing.

src/code/z_vr_box.c
```diff
@@ -32,12 +97,383 @@ typedef struct {
...
+    s32 pad42C;
+
+    switch (arg8) {
+        case 0:
+        case 1:
+            temp358 = sp2B4;
+            temp2A4 = sp250;
+            temp1F0 = sp1EC;
+            temp13C = sp188;
+            temp88 = sp124;
```
Dragorn421, line 315: this is omega confusing because the `temp*` and `sp*` variables are labeled differently
eg `temp358 = sp2B4;` should probably be `temp2B4 = sp2B4;`

Dragorn421 (approved):

src/code/z_vr_box.c
```diff
@@ -121,11 +553,12 @@ void Skybox_Setup(GlobalContext* globalCtx, SkyboxContext* skyboxCtx, s16 skybox
...
-            // reorderings in the rest of this case
+            if (((sp41 & 4) >> 2) != (sp41 & 1)) {
+                D_8011FD3C[sp41].unk_C = D_8011FD3C[sp41].unk_C; // fake but greatly improves match.
+
+                size = (D_8011FD3C[sp41].unk_C - D_8011FD3C[sp41].unk_8) * 2;
 
-            if (((u8)(sp41 & 4) >> 2) != (sp41 & 1)) {
-                size = D_8011FD3C[sp41].unk_C - D_8011FD3C[sp41].unk_8;
-                skyboxCtx->staticSegments[2] = GameState_Alloc(&globalCtx->state, size * 2, "../z_vr_box.c", 1072);
+                skyboxCtx->staticSegments[2] = GameState_Alloc(&globalCtx->state, size, "../z_vr_box.c", 1072);
```
Roman971, line 561: All `skyboxCtx->staticSegments[2]` accesses should be updated to the palettes member (2 is out of bounds now)

Roman971 (approved):

PR #918, opened 2021-08-22
Document object_pu_box and related actors

Dragorn421:
("reopens" #899 , don't rename branches in PRs kids, I was wondering what it would do and I learned)

Small documentation on these unused actors and object

I assume `z_bg_pushbox` used `object_pu_box` at some point in development, even though its init vars say it "now" uses the dungeon keep object

<details>

<summary>click to show dlists and collision</summary>

(I spaced the four dlists/collision models apart so they can be seen better)

dlists (z64utils is too accurate to display them since they have errors):
dlists imported into blender

collision:
collision models importer into blender

the collision has no special properties apart from setting the no_grab flag (can't grab ledges)

</details>


jpburnett (approved):

fig02 (approved):
not a fan of using Impl in this context, but dont have any suggestions for now

PR #919, opened 2021-08-22
Only use the ASM processor where it is needed

louist103:
When I closed the other PR I forgot that I had done this first.  This builds an OK ROM as is and should be fine to merge as is unless we want to wait for the changes to the build script to make all the changes at once.


Dragorn421 (commented):
The only reason I see to not skip asm-processor is euc-jp encoding, if you were to add non-ascii strings in the files not using asm-processor and send them to "is64" (eg osSyncPrintf), it would break either the is64 emulator or the is64 code in the game itself (?)

But since this is only an issue about non-ascii strings in not-modding-friendly files I think this potential issue can just be completely ignored

Is there any argument to make against skipping asm-processor?

louist103:
The processor also deals with the CS files and files with ASM. The files skipped by this PR don't have any non English prints (if any prints at all) so are OK to skip. This is really more of a temporary thing to speed up build times until we get a more permanent solution.

louist103:
Also skipping the processor on files with with Japanese characters doesn't match so it's required for now

AngheloAlf (approved):

Roman971 (commented):
Asset files technically need to run through the asm processor, but cs data isn't extracted properly by zapd atm so it builds regardless. The zapd issue is supposedly easy to fix tho, so I think we should leave the asm processor on these files instead of changing it again in the near future.

Not running libultra files through the asm processor seems fine to me tho.

Roman971 (approved):

fig02 (approved):

PR #924, opened 2021-08-28
Cleanup translation comments

Dragorn421:
For example,
```c
        // Arena is not initalized
        osSyncPrintf("アリーナは初期化されていません\n");

        // Translation: Arena is not initalized
        osSyncPrintf("アリーナは初期化されていません\n");

        // Translates to: "Arena is not initalized"
        osSyncPrintf("アリーナは初期化されていません\n");

        osSyncPrintf("アリーナは初期化されていません\n"); // Arena is not initalized

        osSyncPrintf("アリーナは初期化されていません\n"); // "Arena is not initalized"
```
->
```c
        // "Arena is not initalized"
        osSyncPrintf("アリーナは初期化されていません\n");
```

I think I got everything in `src/code/`. There is still more in `src/overlays/`


Roman971:
Personally I think having the comment on the same line is better if it fits, especially when there are multiple small prints in a row.
Otherwise having a consistent format is nice so I agree with the change to always have `// "translation"`

While on the topic of translation comments, we were using all caps at first and we've moved away from that over time, but there are still many comments that use the old style (eg. in `z_actor.c`). It could be a good idea to fix that at the same time, if you want to do it ofc, otherwise I'll do it myself later since most of that is my own mess anyway 😅

Dragorn421:
A regex like `\n\s*// ".{1,10}"\n.{1,50}\n` can easily find "short" translations with "short" printfs but how do we define "short"?

some examples:

```c
    // "Copy end"
    osSyncPrintf("ＣＯＰＹ終了\n");

    osSyncPrintf("ＣＯＰＹ終了\n"); // "Copy end"
```

```c
    // "game constructor end"
    osSyncPrintf("game コンストラクタ終了\n");

    osSyncPrintf("game コンストラクタ終了\n"); // "game constructor end"
```

```c
        // "Stack overflow or dangerous"
        osSyncPrintf("スタックがオーバーフローしたか危険な状態です\n");

        osSyncPrintf("スタックがオーバーフローしたか危険な状態です\n"); // "Stack overflow or dangerous"
```

```c
    // "System heap initalization"
    osSyncPrintf("システムヒープ初期化 %08x-%08x %08x\n", sysHeap, fb, gSystemHeapSize);

    osSyncPrintf("システムヒープ初期化 %08x-%08x %08x\n", sysHeap, fb, gSystemHeapSize); // "System heap initalization"
```

fig02:
I think an easy enough rule for us to follow is if it can fit beside it without going over the 120 char line limit, then it can stay there

Dragorn421:
TIL clang-format's `ColumnLimit` is not a character count but a byte count so kanjis count for 2 (or so it seems)

I used 100 as the line-byte-count limit to keep some leeway in case there are function or struct member renames

Using `// "..."` makes it easy to automate formatting

<details>

<summary>
Python script used to move translation comments:
</summary>

```py
import os
for root, dirs, files in os.walk("src"):
    for name in files:
        file = os.path.join(root, name)
        with open(file) as f:
            lines = []
            translationCommentLine = None
            for line in f:
                # only process lines with indentation (inside functions),
                # avoids comments in data
                hasIndentation = line.startswith("    ")
                if not hasIndentation:
                    if translationCommentLine is not None:
                        print("! Found a line without indentation right after unused translationCommentLine =", translationCommentLine)
                        lines.append(translationCommentLine)
                        translationCommentLine = None
                    lines.append(line)
                    continue
                
                lineStripped = line.strip()
                # if line is a translation comment, like '// "..."'
                if lineStripped.startswith('// "') and lineStripped.endswith('"'):
                    if translationCommentLine is not None:
                        print("! Found another translation right after unused translationCommentLine =", translationCommentLine)
                        lines.append(translationCommentLine)
                        translationCommentLine = None
                    translationCommentLine = line
                # prevent appending a translation comment to a comment
                elif lineStripped.startswith("//"):
                    if translationCommentLine is not None:
                        print("! Found a comment right after unused translationCommentLine =", translationCommentLine)
                        lines.append(translationCommentLine)
                        translationCommentLine = None
                    lines.append(line)
                # previous line was a translation comment
                elif translationCommentLine is not None:
                    translationCommentLineStripped = translationCommentLine.strip()
                    lineStart = line[:-1] if line.endswith("\n") else line # line without \n
                    lineEnd = line[len(lineStart):] # \n
                    mergedLine = f"{lineStart} {translationCommentLineStripped}{lineEnd}"
                    
                    clangMaxColumns = 120
                    # if the merged line fits under 100 bytes
                    if len(mergedLine.encode("euc-jp")) - len(lineEnd) <= clangMaxColumns - 20:
                        #print("mergedLine =", mergedLine)
                        lines.append(mergedLine)
                    # the merged line doesn't fit
                    else:
                        lines.append(translationCommentLine)
                        lines.append(line)
                    translationCommentLine = None
                # no translation comment in the previous line
                else:
                    lines.append(line)
        with open(file, "w") as f:
            f.writelines(lines)
```

</details>

src/code/z_play.c
```diff
@@ -362,8 +362,8 @@ void Gameplay_Init(GameState* thisx) {
     zAlloc = GameState_Alloc(&globalCtx->state, zAllocSize, "../z_play.c", 2918);
     zAllocAligned = (zAlloc + 8) & ~0xF;
     ZeldaArena_Init(zAllocAligned, zAllocSize - zAllocAligned + zAlloc);
-    osSyncPrintf("ゼルダヒープ %08x-%08x\n", zAllocAligned,
-                 (s32)(zAllocAligned + zAllocSize) - (s32)(zAllocAligned - zAlloc)); // "Zelda Heap"
+    osSyncPrintf("ゼルダヒープ %08x-%08x\n", zAllocAligned, // "Zelda Heap"
```
Roman971, line 365: Last small thing I noticed. I assume it was unintended
```suggestion
    // "Zelda Heap"
    osSyncPrintf("ゼルダヒープ %08x-%08x\n", zAllocAligned,
```

fig02 (approved):

Roman971 (approved):

PR #925, opened 2021-08-28
Handle multiprocessing exceptions in extract_assets.py

AngheloAlf:
There are some weird edge cases on some machines (specially MacOS machines) were extracting assets with python's multiprocessing module fails.
This PR adds a check for this kind of error and fallbacks to plain and simple single-processor extraction, so people with that kind of problem can build the repo anyways.


Roman971 (approved):

Dragorn421 (commented):
What kind of `ProcessError` does this catch exactly?

AngheloAlf:
`ProcessError` is the base class of the Python multiprocessing module. So this is trying to catch every exception generated by that module, which seem to be the cause of some people on specific macs not being able to extract assets properly.
https://docs.python.org/3/library/multiprocessing.html#multiprocessing.ProcessError

fig02 (approved):

PR #927, opened 2021-08-28
Texture alignment macro and a bit of cleaning

AngheloAlf:
Add the macro `ALIGNED8`. It allows to specify the alignment of a variable in compilers which support it, and ignoring it in every other compiler.
Add `CMD_F`. It allows to use a float designated initializer for cutscene macros in compilers which support that.
A few minor fixes: fix a typo in `ovl_Boss_Sst`'s XML, and properly extract all the gfx data of `ovl_Magic_Dark`.


assets/xml/overlays/ovl_Magic_Dark.xml
```diff
@@ -1,8 +1,12 @@
...
-        <DList Name="sDiamondVertsDL" Offset="0x1700"/>-->
+    <File Name="ovl_Magic_Dark" BaseAddress="0x80B873D0" RangeStart="0xD10" RangeEnd="0x1740" Segment="128">
+        <Texture Name="sMagicDarkDiamondTex" OutName="diamond_texture" Format="i8" Width="32" Height="64" Offset="0xD10"/>
+
+        <Array Name="gMagicDarkDiamondVtx" Count="20" Offset="0x1510">
+            <Vtx/>
+        </Array>
+
+        <DList Name="sMagicDarkDiamondTexDL" Offset="0x1650"/>
+        <DList Name="sMagicDarkDiamondVertsDL" Offset="0x1700"/>
```
Dragorn421, line 10:
```suggestion
        <DList Name="sMagicDarkDiamondGeometryDL" Offset="0x1700"/>
```
Vertices + Faces

assets/xml/overlays/ovl_Magic_Dark.xml
```diff
@@ -1,8 +1,12 @@
...
-        <!--<DList Name="sDiamondTexDL" Offset="0x1650"/>
-        <DList Name="sDiamondVertsDL" Offset="0x1700"/>-->
+    <File Name="ovl_Magic_Dark" BaseAddress="0x80B873D0" RangeStart="0xD10" RangeEnd="0x1740" Segment="128">
+        <Texture Name="sMagicDarkDiamondTex" OutName="diamond_texture" Format="i8" Width="32" Height="64" Offset="0xD10"/>
+
+        <Array Name="gMagicDarkDiamondVtx" Count="20" Offset="0x1510">
+            <Vtx/>
+        </Array>
+
+        <DList Name="sMagicDarkDiamondTexDL" Offset="0x1650"/>
```
Dragorn421, line 9:
```suggestion
        <DList Name="sMagicDarkDiamondSetupDL" Offset="0x1650"/>
```
Doesn't just load the images, "Setup" sounds more generic

include/command_macros_base.h
```diff
@@ -6,16 +6,22 @@
...
+#define CMD_BBH(a, b, c) {(_SHIFTL(a, 24, 8) | _SHIFTL(b, 16, 8) | _SHIFTL(c, 0, 16))}
 
-#define CMD_HBB(a, b, c) (_SHIFTL(a, 16, 16) | _SHIFTL(b, 8, 8) | _SHIFTL(c, 0, 8))
+#define CMD_HBB(a, b, c) {(_SHIFTL(a, 16, 16) | _SHIFTL(b, 8, 8) | _SHIFTL(c, 0, 8))}
 
-#define CMD_HH(a, b) (_SHIFTL(a, 16, 16) | _SHIFTL(b, 0, 16))
+#define CMD_HH(a, b) {(_SHIFTL(a, 16, 16) | _SHIFTL(b, 0, 16))}
 
-#define CMD_W(a) (a)
+#define CMD_W(a) {(a)}
```
Dragorn421, line 17: changing these to use a struct initialization syntax could mean they would now belong more in a cutscene-specific header than a general-purpose header? especially with `CMD_F` using a named field (`CutsceneData#f`)

include/macros.h
```diff
@@ -161,4 +161,10 @@ extern GraphicsContext* __gfxCtx;
                        ((height)-1) << G_TEXTURE_IMAGE_FRAC);                                                          \
     } while (0)
 
+#ifdef __GNUC__
+#define ALIGNED8 __attribute__ ((aligned (8)))
+#else
+#define ALIGNED8
```
Dragorn421, line 167: so what's the point of this? planning ahead for gcc support?

AngheloAlf: Yeah, this is because of a discussion we had on discord. The idea is to have better compatibility with gcc. ZAPD in the future will start using this macro for the extracted assets, so we will need it in the repo.

include/command_macros_base.h
AngheloAlf: Well, yeah. `CMD_F` should be used only with `CutsceneData`. This will allow to _not_ use the asm processor in any non-IDO build in the future, like modding contexts, building with gcc, etc.

fig02 (approved):

Dragorn421:
Is this fine with `include/z64scene.h`'s usage of the `CMD_` macros? looks kind of weird having eg
```c
SCENE_CMD_SPECIAL_FILES(0x02, 0x0003),
```
expand to
```c
{ SCENE_CMD_ID_SPECIAL_FILES, 0x02, CMD_W(0x0003) }
```
```c
{ SCENE_CMD_ID_SPECIAL_FILES, 0x02, {(0x0003)} }
```
to fill a
```c
typedef struct {
    /* 0x00 */ u8  code;
    /* 0x01 */ u8  data1;
    /* 0x04 */ u32 data2;
} SCmdBase;
```
though it's fine apparently

fig02:
yeah, braces wrapping a single element would just be ignored i think
do we need to have it though?

Dragorn421 (commented):
`CMD_F` should be in `z64cutscene_commands.h` instead of `command_macros_base.h` (which is used by more than cutscenes, eg scene commands),
because `CMD_F` relies on the `f` field of `CutsceneData` for the designated initializer variant

include/command_macros_base.h
Roman971: I agree with Dragorn, this definitely doesn't belong in this header because it's intended as general purpose macros to build commands upon, not just for cutscenes (in fact they are already used in scene commands as well).

For this reason, I think we should leave these macros as is and have wrappers with the braces and all in the cutscene command headers.

Roman971 (approved):
I find it odd to have `CMD_F` relying on the `f` field like Dragorn mentioned, but it's only used for cs data rn so I guess it's fine. And otherwise this seems good to me

fig02:
should be good to merge after conflict fixes

PR #929, opened 2021-08-28
translations

GlitterBerri:
Edited translations.

### DIFF
--- a/src/code/z_play.c
+++ b/src/code/z_play.c
@@ -295,9 +295,11 @@ void Gameplay_Init(GameState* thisx) {
     // The emulator constantly checks whether PC is 0x81000000, so this works even though it's not a valid address.
     if ((gEntranceTable[((void)0, gSaveContext.entranceIndex)].scene == SCENE_SPOT09) &&
         gSaveContext.sceneSetupIndex == 6) {
-        osSyncPrintf("エンディングはじまるよー\n"); // "The ending starts"
+        // "Here goes the ending!"
+        osSyncPrintf("エンディングはじまるよー\n");
         ((void (*)())0x81000000)();
-        osSyncPrintf("出戻り？\n"); // "Return?"
+        osSyncPrintf("出戻り？\n");
+        // "Return?"
     }
 
     Cutscene_HandleEntranceTriggers(globalCtx);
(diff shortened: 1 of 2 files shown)


src/code/z_play.c
```diff
@@ -295,9 +295,11 @@ void Gameplay_Init(GameState* thisx) {
     // The emulator constantly checks whether PC is 0x81000000, so this works even though it's not a valid address.
     if ((gEntranceTable[((void)0, gSaveContext.entranceIndex)].scene == SCENE_SPOT09) &&
         gSaveContext.sceneSetupIndex == 6) {
-        osSyncPrintf("エンディングはじまるよー\n"); // "The ending starts"
+        // "Here goes the ending!"
+        osSyncPrintf("エンディングはじまるよー\n");
         ((void (*)())0x81000000)();
-        osSyncPrintf("出戻り？\n"); // "Return?"
+        osSyncPrintf("出戻り？\n");
+        // "Return?"
```
AngheloAlf, line 302: I personally prefer to have the comment at the top of the printf, instead of the bottom
```suggestion
        // "Return?"
        osSyncPrintf("出戻り？\n");
```

src/code/z_play.c
```diff
@@ -905,10 +912,10 @@ void Gameplay_Update(GlobalContext* globalCtx) {
             if (globalCtx->unk_1242B != 0) {
                 if (CHECK_BTN_ALL(input[0].press.button, BTN_CUP)) {
                     if ((globalCtx->pauseCtx.state != 0) || (globalCtx->pauseCtx.debugState != 0)) {
-                        // Translates to: "Changing viewpoint is prohibited due to the kaleidoscope"
+                        // "changing viewpoint is prohibited due to being in the kaleidoscope"
                         osSyncPrintf(VT_FGCOL(CYAN) "カレイドスコープ中につき視点変更を禁止しております\n" VT_RST);
                     } else if (Player_InCsMode(globalCtx)) {
-                        // Translates to: "Changing viewpoint is prohibited during the cutscene"
+                        // "changing viewpoint is prohibited due to being in a demo"
```
Dragorn421, line 918: We use "cutscene" instead of "demo" wherever that is what is implied, for example here the line right above has `Player_InCsMode` where `Cs` stands for "Cutscene"

I'm not sure if we also want to apply this to translations, more opinions needed

Dragorn421 (commented):
Looks fine, don't worry too much about the formatting (where to put the translations), as long as it's clear what is being translated we can move the comments ourselves once we settled on something

PR #930, opened 2021-08-29
Match the remaining function in Boss_Tw

Roman971:
Tricking the compiler because I can't find a proper way to match it. This is peak decompilation right here

Note: The `if (1) {}` is only there to fix the regalloc change caused by the `if (zero)` block, it's not technically part of the trick.


Dragorn421 (approved):
Roman just got back from Hogwarts

fig02 (approved):

PR #936, opened 2021-08-31
Document some of EnHy (Hylian NPCs), mostly its data 

Dragorn421:
FeelsUnkMan


src/overlays/actors/ovl_En_Hy/z_en_hy.c
```diff
@@ -61,242 +69,380 @@ static ColliderCylinderInit D_80A71EC0 = {
     { 20, 46, 0, { 0, 0, 0 } },
 };
 
-static CollisionCheckInfoInit2 D_80A71EEC = { 0x00, 0x0000, 0x0000, 0x0000, 0xFF };
+static CollisionCheckInfoInit2 sColChkInfoInit = { 0x00, 0x0000, 0x0000, 0x0000, MASS_IMMOVABLE };
```
fig02, line 72: rest of those numbers can be dec too

src/overlays/actors/ovl_En_Hy/z_en_hy.c
```diff
@@ -6,29 +6,37 @@
...
+#include "objects/object_cne/object_cne.h"
+#include "objects/object_cob/object_cob.h"
 #include "objects/object_os_anime/object_os_anime.h"
 
 #define FLAGS 0x00000019
 
 #define THIS ((EnHy*)thisx)
 
+#define TYPE (this->actor.params & 0x7F)
+#define PATH_ID ((this->actor.params & 0x780) >> 7)
```
fig02, line 23: same comment about this again, would prefer to not do anything yet, but should take an arg at the very least if we must have it now

src/overlays/actors/ovl_En_Hy/z_en_hy.c
```diff
@@ -61,242 +69,380 @@ static ColliderCylinderInit D_80A71EC0 = {
...
-    { 20.0f, { 0.0f, 0.0f, -200.0f }, 0.01f, 0x06, 30.0f },
+    /* 0x00 */ f32 shadowScale;
+    /* 0x04 */ Vec3f modelOffset;
+    /* 0x10 */ f32 scale;
+    /* 0x14 */ s8 targetMode;
+    /* 0x18 */ f32 unkRange;
+} EnHyInit2Info; // size = 0x1C
+
+// indexed by EnHyType
+static EnHyInit2Info sInit2Infos[] = {
```
fig02, line 322: might be an unpopular opinion but i dont think we need to note what its indexed by when context makes that clear

src/overlays/actors/ovl_En_Hy/z_en_hy.c
```diff
@@ -61,242 +69,380 @@ static ColliderCylinderInit D_80A71EC0 = {
...
+    /* 0x00 */ u8 headInfoIndex;       // EnHyHeadIndex
+    /* 0x01 */ u8 unusedSkelInfoIndex; // EnHySkeletonIndex, see EnHy#objBankIndexSkelUnused
+    /* 0x02 */ Color_RGBA8 envColorSeg8;
+    /* 0x06 */ u8 skelInfoIndex; // EnHySkeletonIndex
+    /* 0x07 */ Color_RGBA8 envColorSeg9;
+    /* 0x0B */ u8 animInfoIndex; // EnHyAnimationIndex
+} EnHyModelInfo;                 // size = 0xC
+
+// indexed by EnHyType
+static EnHyModelInfo sModelInfos[] = {
```
fig02, line 226: info as plural is a bit strange i think. goes for the others as well

Dragorn421: would it be fine to just use `sModelInfo` even though there are several "info structs" inside the array? idk what to call the arrays

src/overlays/actors/ovl_En_Hy/z_en_hy.c
```diff
@@ -61,202 +66,339 @@ static ColliderCylinderInit D_80A71EC0 = {
...
+    /* 11 */ ENHY_HEAD_CNE_11,
+    /* 12 */ ENHY_HEAD_CNE_12,
+    /* 13 */ ENHY_HEAD_BJI_13,
+    /* 14 */ ENHY_HEAD_BJI_14,
+    /* 15 */ ENHY_HEAD_COB
+} EnHyHeadIndex;
+
+// indexed by EnHyHeadIndex
+static EnHyHeadInfo sHeadInfo[] = {
+    /*  0 */ { OBJECT_AOB, gDogLadyHeadDL, sEyeTexturesAOB },
```
Roman971, line 106: I think the comments in all these arrays should use the enum values since the numbers themselves don't mean anything. And if we use the enums then the "indexed by" comments are probably not necessary either.

src/overlays/actors/ovl_En_Hy/z_en_hy.c
```diff
@@ -635,57 +778,58 @@ void func_80A70978(EnHy* this, GlobalContext* globalCtx) {
...
-                ((this->actor.params & 0x7F) != 12) && ((this->actor.params & 0x7F) != 2) &&
-                ((this->actor.params & 0x7F) != 7)) {
+            if (!((this->actor.params & 0x7F) == ENHY_TYPE_BOJ_9 || (this->actor.params & 0x7F) == ENHY_TYPE_BOJ_10 ||
+                  (this->actor.params & 0x7F) == ENHY_TYPE_BOJ_12 || (this->actor.params & 0x7F) == ENHY_TYPE_AHG_2 ||
+                  (this->actor.params & 0x7F) == ENHY_TYPE_BJI_7)) {
                 return true;
             } else if (!LINK_IS_ADULT) {
                 return true;
-            } else if (((this->actor.params & 0x7F) != 12) && IS_NIGHT) {
+            } else if ((this->actor.params & 0x7F) != ENHY_TYPE_BOJ_12 && IS_NIGHT) {
```
Roman971, line 802: Pretty minor but I personally think parens around conditions actually help readability, so removing them is a slight downgrade imo

src/overlays/actors/ovl_En_Hy/z_en_hy.h
```diff
@@ -12,25 +37,25 @@ typedef struct EnHy {
...
-    /* 0x0194 */ char unk_194;
-    /* 0x0195 */ u8 unk_195;
-    /* 0x0196 */ s8 unk_196;
-    /* 0x0197 */ s8 unk_197;
-    /* 0x0198 */ s8 unk_198;
-    /* 0x0199 */ s8 unk_199;
+    /* 0x0194 */ char unk_194; // unused
+    /* 0x0195 */ u8 pathReverse;
+    /* 0x0196 */ s8 objBankIndexHead;
+    /* 0x0197 */ s8 objBankIndexSkelUnused; // always the same as objBankIndexSkel, and doesn't achieve anything
```
Roman971, line 43: Mostly curious but did you try to investigate what happens if this isn't the same as `objBankIndexSkel`? It seems to be used for a specific purpose in `EnHy_PostLimbDraw`, even tho it's not relevant in the final game (or at least not in this version). Also not sure about calling it "unused" for this reason.

src/overlays/actors/ovl_En_Hy/z_en_hy.c
Dragorn421: We should enforce some rule about this, I'm on the "no-parentheses" side (it is more readable without them to me, especially in large conditionals), but I would be fully okay with going the opposite for the sake of consistency

src/overlays/actors/ovl_En_Hy/z_en_hy.h
Dragorn421: I'm not sure what my reasoning was at the time. Maybe I thought that `SkelAnime_DrawFlexLimbOpa` was setting the segments back to their original values somehow, but it isn't

So the "doesn't achieve anything" part is only true due to the object being the same, but actually if it was to be different then half the skeleton (limbs 8-15 where 1 is root since the segment change is after limb 7) would use `objBankIndexSkelUnused` instead of `objBankIndexHead`

I'm confused what this was intended for. Maybe they planned to use half-skeletons of each hylian npc to boost the variety. This could make sense since it seems all hylian npcs used by EnHy use the same skeleton (eg limb 15 is the head limb for all of them)

<details>

<summary>
It sounded too fun to not try, right (click for details)
</summary>

```c
// at the top of EnHy_FindSkelAndHeadObjects
unusedSkelInfoIndex = Rand_ZeroOne() * 7.99;
```

Results (it's very prone to crashing, this was a lucky run):





Vanilla for reference:



</details>

So yeah anyway I'm probably renaming (and in `EnHyModelInfo` too)
`objBankIndexSkel` -> `objBankIndexSkelFirstHalf`
`objBankIndexSkelUnused` -> `objBankIndexSkelSecondHalf`
?

fig02: is it clear which part of the body each is for? upper and lower, for upper and lower body seems fine

fig02 (approved):

src/overlays/actors/ovl_En_Hy/z_en_hy.c
Roman971: I'm fine with enforcing it, but we would probably need to discuss it with everyone first.
Also fwiw the vast majority of the codebase should have parens atm, but I guess it's not fully consistent either

Roman971 (approved):

PR #938, opened 2021-09-01
Cleanup symbols from undefined_syms.txt (3rd round)

Dragorn421:
Looks like that's as much cleanup as can be done for now

Most remaining symbols are from undecompiled stuff (boss_ganon, en_wf, en_zf, kankyo, adult link's object)


fig02 (approved):

assets/xml/scenes/overworld/spot05.xml
```diff
@@ -1,6 +1,7 @@
 <Root>
     <File Name="spot05_scene" Segment="2">
         <Cutscene Name="gMinuetCs" Offset="0x3F80"/>
+        <Cutscene Name="spot05_sceneCutsceneData0x005730" Offset="0x5730"/>
```
Roman971, line 4:
```suggestion
        <Cutscene Name="spot05_scene_Cs_005730" Offset="0x5730"/>
```

assets/xml/scenes/overworld/spot02.xml
```diff
@@ -1,6 +1,8 @@
 <Root>
     <File Name="spot02_scene" Segment="2">
         <Scene Name="spot02_scene" Offset="0x0"/>
+        <Cutscene Name="spot02_sceneCutsceneData0x003C80" Offset="0x3C80"/>
```
Roman971, line 4:
```suggestion
        <Cutscene Name="spot02_scene_Cs_003C80" Offset="0x3C80"/>
```

assets/xml/scenes/overworld/spot05.xml
Dragorn421: Unfortunately, this symbol name needs to be exactly that to circumvent a ZAPD bug

https://github.com/zeldaret/ZAPD/issues/177

Roman971: I see. In that case I feel like we should wait for the bug to be fixed in ZAPD before using these symbols, if possible. They are autogenerated by ZAPD so they could technically change name at any time, and I think we should strongly avoid that in general.

Roman971: Tho I might be the only one feeling this way, so I won't stand against it if we really want the symbols anyway

Dragorn421: I don't think they can break unless ZAPD updates and at that point it's kind of ZAPD's responsibility to be compatible

But if others would also like these symbols changes to be reverted I'd be fine with it too. We already have a waiting room for ZAPD fixes in undefined_syms.txt anyway

Dragorn421: I added the names we would want as comments, and saying it's a zapd issue

Anghelo thinks it's better as is https://discord.com/channels/688807550715560050/838852326231769089/884461872827465808

Roman971 (approved):

PR #942, opened 2021-09-02
Improve actorfixer.py (2nd round) and rename to namefixer.py

Dragorn421:
Or an attempt at it at least.

Deletes animfixer and mathfixer, and moves their content into actorfixer.

Fixes `PLAYER` being replaced by `GET_PLAYER(globalCtx)` in the `"PLAYER"` string in audio data. Not entirely convinced about the way I fixed it tbh, tried to make it somewhat generic in case we have another similar issue in the future.

Other options to fix it would include:

- remove the PLAYER replace from actorfixer
- change "PLAYER" in audio data to "PLAYE""R"
- making the script see `"` as a word character
- hardcoding special behavior for PLAYER

The script seems to perform the same with or without the fix, `time ./tools/actorfixer.py .` reporting ~700ms for me.


Roman971:
Should we rename the script to `namefixer.py` or something similar since it's not just about actors anymore?

fig02:
not a fan of changing that audio string tbh
do we want to just exclude this file from checking for the player change? itd be an unfortunate thing to hardcode, but audio code probably isnt going to be referencing player much, if at all

fig02 (approved):

Dragorn421:
Since no one else voiced interest about the name of the script I renamed it to `namefixer.py`, despite not liking that name myself

Hopefully when squashed git still figures out the rename, but not a big deal if it doesn't

Roman971 (approved):

PR #943, opened 2021-09-03
Fix UB in func_80B537E8.

ProjectRevoTPP:
Hey folks with another GCC UB discovery. Found this while doing Werror=return-type stuff, and it turns out it actually *is* fixable while retaining matching. A bit weird, maybe it implies a weird macro with the Math_SmoothStepToS stuff.


Dragorn421 (approved):
I see rotations in decimal 👺
Anyway nice find

Roman971 (approved):

ethteck (approved):

fig02 (approved):

PR #947, opened 2021-09-05
Fix compiler warnings in z_camera/db_camera

zelda2774:


src/code/z_camera.c
```diff
@@ -7020,10 +7020,19 @@ void Camera_PrintSettings(Camera* camera) {
...
             sp50[i++] = '-';
         }
+
+        // @bug: this code was clearly meaning to print camera->camDataIdx as a two-digit
+        // number, not i.
+        // "sp50[i++] = ..." matches here, but is undefined behavior due to conflicting
+        // reads/writes between sequence points, triggering warnings. Work around by
+        // putting i++ afterwards while on the same line.
+        // clang-format off
         if (camera->camDataIdx / 0xA != 0) {
```
Dragorn421, line 7030:
```suggestion
        if (camera->camDataIdx / 10 != 0) {
```

src/code/z_camera.c
```diff
@@ -7020,10 +7020,19 @@ void Camera_PrintSettings(Camera* camera) {
         if (camera->camDataIdx < 0) {
             sp50[i++] = '-';
         }
+
+        // @bug: this code was clearly meaning to print camera->camDataIdx as a two-digit
+        // number, not i.
```
Dragorn421, line 7025:
```suggestion
        //! @bug: this code was clearly meaning to print `camera->camDataIdx` as a two-digit
        //! number, instead of `i`.
```
I think this is the right way to format bug comments?

also this is probably not the only thing wrong, there's a check for `camera->camDataIdx < 0` but I don't think the following code would work correctly with a negative number?

Dragorn421 (approved):

fig02 (approved):

Roman971 (approved):
Side note: we need to find a better solution than using the ((void)0, foo) trick for cases like this one where it's clearly fake

PR #951, opened 2021-09-07
Make sSetupDLs a Gfx[][6]

petrie911:
This array makes more sense as a 2D array, although one function has to do some fake stuff to match. It's possible the game didn't have it as a 2D array originally, though I can't imagine why that would be.

### DIFF
--- a/src/code/z_rcp.c
+++ b/src/code/z_rcp.c
@@ -13,7 +13,6 @@ Gfx sSetupDL[][6] = {
         gsSPLoadGeometryMode(G_ZBUFFER | G_SHADE | G_CULL_BACK | G_FOG | G_SHADING_SMOOTH),
         gsSPEndDisplayList(),
     },
-
     {
         /* 0x01 */
         gsDPPipeSync(),
@@ -25,7 +24,6 @@ Gfx sSetupDL[][6] = {
         gsSPLoadGeometryMode(G_SHADE | G_CULL_BACK | G_FOG | G_LIGHTING | G_SHADING_SMOOTH),
         gsSPEndDisplayList(),
     },
-
     {
         /* 0x02 */
         gsDPPipeSync(),
@@ -37,7 +35,6 @@ Gfx sSetupDL[][6] = {
         gsSPLoadGeometryMode(G_ZBUFFER | G_SHADE | G_CULL_BACK | G_LIGHTING | G_SHADING_SMOOTH),
         gsSPEndDisplayList(),
     },
-
     {
         /* 0x03 */
         gsDPPipeSync(),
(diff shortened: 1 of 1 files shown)


src/code/z_rcp.c
```diff
@@ -1,737 +1,881 @@
...
+        gsDPSetCombineLERP(PRIMITIVE, ENVIRONMENT, TEXEL0, ENVIRONMENT, PRIMITIVE, 0, TEXEL0, 0, 0, 0, 0, COMBINED, 0,
+                           0, 0, COMBINED),
+        gsDPSetOtherMode(G_AD_NOISE | G_CD_NOISE | G_CK_NONE | G_TC_FILT | G_TF_BILERP | G_TT_NONE | G_TL_TILE |
+                             G_TD_CLAMP | G_TP_PERSP | G_CYC_2CYCLE | G_PM_NPRIMITIVE,
+                         G_AC_NONE | G_ZS_PIXEL | G_RM_FOG_SHADE_A | G_RM_ZB_CLD_SURF2),
+        gsSPLoadGeometryMode(G_ZBUFFER | G_SHADE | G_CULL_BACK | G_FOG | G_SHADING_SMOOTH),
+        gsSPEndDisplayList(),
+    },
+
+    {
```
Dragorn421, line 17: I would get rid of the empty lines between the entries

AngheloAlf (approved):

src/code/z_rcp.c
```diff
@@ -821,7 +895,10 @@ Gfx* Gfx_SetFog2(Gfx* gfx, s32 r, s32 g, s32 b, s32 a, s32 n, s32 f) {
 }
 
 Gfx* Gfx_CallSetupDLImpl(Gfx* gfx, u32 i) {
-    gSPDisplayList(gfx++, &sSetupDL[6 * i]);
+    s32 dListIndex = 6 * i;
+
+    gSPDisplayList(gfx++, &((Gfx*)sSetupDL)[dListIndex]);
```
fig02, line 900: i still feel like doing stuff like this is forcing it to be something that its not. but i guess we've always taken our liberties with changing the structure of something to make it better if it doesnt effect matching. 
just still dont feel 100% about it in this case

petrie911: I might be forcing it to be something it wasn't, but I'm not forcing it to be something it's not. It *is* a n x 6 array of Gfx commands. That Nintendo chose a strange way to handle it shouldn't make a difference.

fig02: fair enough, we're looking to do the same thing with vrbox anyway. Still dont feel 100% about it, but will go with the majority here.
Which btw if you want to make the changes to the skybox palletes, is prob the only thing holding that up

fig02 (approved):

Dragorn421 (approved):

PR #954, opened 2021-09-08
Mark a bug in En_Tana

louist103:


src/overlays/actors/ovl_En_Tana/z_en_tana.c
```diff
@@ -55,6 +55,8 @@ static void* sStoneTextures[] = {
 void EnTana_Init(Actor* thisx, GlobalContext* globalCtx) {
     EnTana* this = THIS;
 
+    //! @bug In the zora shop params is set to 2. This causes an out of bounds read on the array of shelf types. With
+    //! IDO it just happens to not crash, however with different compilers re-ordering may cause a crash.
```
fig02, line 59: imo the bug tag makes more sense above the array itself, since the problem is really that its missing a string. the access is fine

fig02: because of that I also think the details about zora shop being params 2 isnt necessarily important for the bug description. just the fact that its missing a third entry

src/overlays/actors/ovl_En_Tana/z_en_tana.c
```diff
@@ -29,7 +29,7 @@ const ActorInit En_Tana_InitVars = {
     NULL,
 };
 
-static char* sShelfTypes[] = {
+static char* const sShelfTypes[] = {
     "木の棚", // "Wooden Shelves"
```
fig02, line 33: youd generally see `const char*` not `char* const`, right

louist103: Yeah this was a leftover from testing.  I'll take it out.

fig02 (approved):

Roman971 (approved):

PR #955, opened 2021-09-08
Document undocumented matrix functions

Dragorn421:


fig02:
is there a preference on whether this goes first or #952?

Dragorn421:
let's say #952 first since discussion on `TranslateRotate***` vs `Rotate***Translate` has started there already

Dragorn421:
Thanks for the feedback on #952
But wait, there's more!

src/code/sys_matrix.c
```diff
@@ -470,59 +470,64 @@ void Matrix_TranslateRotateZYX(Vec3f* translation, Vec3s* rotation) {
     }
 }
 
-void func_800D1694(f32 x, f32 y, f32 z, Vec3s* vec) {
+/**
+ * Translate and rotate using YXZ Tait-Bryan angles.
+ * This means a (column) vector is first rotated around Z, then around X, then around Y, then translated, then gets
+ * transformed according to whatever the matrix was previously.
+ */
+void Matrix_TranslateRotateYXZ(f32 translateX, f32 translateY, f32 translateZ, Vec3s* rot) {
```
Roman971, line 478: This one feels wrong to me because the function isn't quite "equivalent" to `Matrix_TranslateRotateZYX` implementation-wise. More specifically, the previous matrix seems to only be partially preserved, and only rotation Z carries over? It seems like that to me when comparing the implementaton of each section, but I could be wrong.
If I'm right, then I think at least the comment needs to be adjusted, and we might want a more specific name, because with the current name this function sounds like another version of `Matrix_TranslateRotateZYX` with a different rotation order, which could be misleading.

Roman971 (approved):
Seems good to me code-wise, but I can't say much on the matrix math side of things beyond the names which make sense to me overall

fig02 (approved):
agreed. we should perhaps get some of the matrix math people to approve to make sure they are OK with the contents of the pr

src/code/sys_matrix.c
```diff
@@ -698,47 +702,70 @@ void Matrix_Transpose(MtxF* mf) {
     mf->yz = temp;
 }
 
-void func_800D1FD4(MtxF* mf) {
+/**
+ * Replaces the rotation in the matrix at the top of the stack, keeping scale and translation untouched.
```
zbanks, line 706: I think this function also combines the scales from `mf` with `sCurrentMatrix`?
(I don't think `mf`'s 3x3 section is normalized?)
This would match the final line of this comment, "*then the 3x3 part of `sCurrentMatrix` becomes `\f$ \texttt{mf} \cdot S \f$`*".

I might rewrite the first line to something like this? 
```suggestion
 * Modifies matrix at the top of the stack, by setting the rotation to zero, then applying the scale and rotation from `mf`. This keeps the translation untouched.
```
(I also think it helps to mention `mf` explicitly, to make it more clear that `sCurrentMatrix` is the thing being modified?)

(This is a conceptually kind of a weird operation, isn't it?)

zbanks (commented):
Yeah this documentation looks great to me! The names seem consistent with the project, and from what I can tell, the algebra matches up with the function names.

The one exception is `Matrix_ReplaceRotation` -- I left a comment inline.

(I only reviewed `functions.h`, `sys_matrix.c`, and `z_skin_matrix.c` -- I think the rest of the spots were just the name changes? I'm assuming someone else will double-check that.)

EllipticEllipsis:
I checked all the other files, they are just name changes. Still making my way through the 2 actual matrix files.

src/code/sys_matrix.c
Dragorn421: Yes `mf` can actually be anything, not necessarily a rotation or even a rotation*scale

I may have named this based on usage tbh, since it is exclusively used as `Matrix_ReplaceRotation(&globalCtx->billboardMtxF);`
billboarding is for making geometry face the camera, `billboardMtxF` is a rotation matrix which is the inverted view matrix

So a more accurate name would be `Matrix_Set3x3KeepScale` but it's also not a very good name is it

src/code/sys_matrix.c
```diff
@@ -698,47 +702,70 @@ void Matrix_Transpose(MtxF* mf) {
...
+ *
+ * Considering only the 3x3 part (ignoring the last row and column which are left untouched), assuming
+ * \f[ \texttt{sCurrentMatrix} = R \cdot S \f]
+ *
+ * with \f$
+ *  S = \begin{bmatrix}
+ *      \alpha & 0 & 0 \\
+ *      0 & \beta & 0 \\
+ *      0 & 0 & \gamma
+ *      \end{bmatrix}
```
Dragorn421, line 716: Btw this is how this doc looks in doxygen

<details>

<summary>
Click me (careful dark mode users the screenshot is white)
</summary>



</details>

I'm not a fan of how bad it looks in the source code though

src/code/sys_matrix.c
```diff
@@ -973,165 +998,166 @@ void func_800D2A34(MtxF* mf, f32 arg1, f32 arg2, f32 arg3, f32 arg4) {
...
+    mf->yy = scale;
+    mf->zz = scale;
+    mf->xw = translateX;
+    mf->yw = translateY;
+    mf->zw = translateZ;
     mf->ww = 1.0f;
 }
 
-void func_800D2A98(Mtx* mtx, f32 arg1, f32 arg2, f32 arg3, f32 arg4) {
+void Matrix_SetTranslateUniformScaleMtx(Mtx* mtx, f32 scale, f32 translateX, f32 translateY, f32 translateZ) {
```
engineer124, line 1010: You use the convention:
`Matrix_SetTranslateUniformScaleMtx`
`Matrix_SetTranslateUniformScaleMtx2`
for this set of equivalent functions, but then use:
`Matrix_SetTranslateScaleMtx1`
`Matrix_SetTranslateScaleMtx2`
For the set of equivalent functions below. I wonder if they should be consistent. Either `Mtx/Mtx2` or `Mtx1/Mtx2`. I don't have a preference either way

Dragorn421: I think it's because `Matrix_SetTranslateUniformScaleMtx` is the Mtx version of the MtxF function `Matrix_SetTranslateUniformScaleMtxF` (just a wrapper)

src/code/sys_matrix.c
Roman971: It might just be me and I didn't mention it for this PR, but I really don't think we should be using LaTeX or write things specific to doxygen in general.
I get the point of doxygen for something like a library, but I honestly don't see anyone using it when working with a codebase like this. So it's a cool bonus if it just uses comments and documentation we would be writing anyway, but I find it unfortunate to be writing things specifically for doxygen instead of, say, having it in a more basic format which is meant to be read from the code itself (eg. with `|`s for a matrix layout).

Also worth noting that any doxygen syntax like this with `\` and such isn't generally parsable by IDEs so it just gets cut-off. Here is how this one looks like in vscode for instance:
image
I did look for an extension or setting that supports it, but couldn't find anything

fig02: fwiw I agree, I would rather have something readable in the codebase directly over needing to generate and use doxygen.
It may seem counter productive since we added doxygen to the project, but I see it as an optional thing to view the documentation if you want, not the only way

fig02:
not sure where we stand on matrix-competent peoples reviews on this.

only input i still have is i would prefer to not use doxygen syntax for the docs here. skin matrix already has some, so there is some that would need to be removed if we agree to do so. but yea

I still feel like the documentation should be intelligible in the codebase without extra tools if the user chooses to

src/code/sys_matrix.c
zbanks: I don't mind the function name -- this function is weird enough that any name will be weird.

I do think the comment should be updated though because it does not keep the scale untouched.

zbanks (approved):

fig02 (approved):

Roman971 (approved):

PR #957, opened 2021-09-09
Fix all headers to comply with C standard

louist103:


Dragorn421 (approved):

fig02 (approved):

include/stdbool.h
```diff
@@ -1,5 +1,5 @@
-#ifndef _STDBOOL
-#define _STDBOOL
+#ifndef STDBOOL_H
+#define STDBOOL_H
```
Roman971, line 2: This file also has a comment on the last endif which should be updated

Roman971 (commented):
This seems to be missing a change for z64animation.h

Roman971 (approved):

PR #958, opened 2021-09-11
Fix return UB

louist103:
Fix some more undefined behavior due to missing returns.

### DIFF
--- a/src/overlays/actors/ovl_player_actor/z_player.c
+++ b/src/overlays/actors/ovl_player_actor/z_player.c
@@ -1138,13 +1138,13 @@ static LinkAnimationHeader* D_808543D4[] = {
     &gPlayerAnim_002CC0,
 };
 
-// return type can't be void due to regalloc in func_8084FCAC
+// return type must be s32 due to regalloc in func_8084FCAC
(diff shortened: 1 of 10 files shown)


src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -1138,13 +1138,13 @@ static LinkAnimationHeader* D_808543D4[] = {
     &gPlayerAnim_002CC0,
 };
 
-// return type can't be void due to regalloc in func_8084FCAC
+// return type must be s32 due to regalloc in func_8084FCAC
```
fig02, line 1141: we dont know it has to be s32 specifically, so i think the original wording was fine

louist103: I had changed it to f32 to try a different return and it still didn't match (even when returning nothing)

fig02: any integer type should be exchangeable here. float is a special case cause it wouldnt necessarily use v0

fig02 (approved):

Roman971 (approved):

PR #966, opened 2021-09-14
Document `ovl_select`

AngheloAlf:
I named almost every member of the `SelectContext` struct and added translations to the scene names.
I made this because I decompiled and documented `ovl_select` in MM (zeldaret/mm/pull/310) and since the OoT version so similar I decided to back-port the same documentation.

### DIFF
--- a/src/overlays/gamestates/ovl_select/z_select.c
+++ b/src/overlays/gamestates/ovl_select/z_select.c
@@ -312,9 +312,9 @@ void Select_UpdateMenu(SelectContext* this) {
     }
 
     if (CHECK_BTN_ALL(controller1->press.button, BTN_L)) {
-        this->unk_1DC++;
-        this->unk_1DC = (this->unk_1DC + ARRAY_COUNT(this->unk_1E0)) % ARRAY_COUNT(this->unk_1E0);
-        this->currentScene = this->topScreen = this->unk_1E0[this->unk_1DC];
+        this->pageDownIndex++;
+        this->pageDownIndex = (this->pageDownIndex + ARRAY_COUNT(this->pageDownStops)) % ARRAY_COUNT(this->pageDownStops);
+        this->currentScene = this->topScreen = this->pageDownStops[this->pageDownIndex];
     }
 
     this->unk_21C += this->unk_220;
@@ -355,7 +355,7 @@ void Select_UpdateMenu(SelectContext* this) {
 
     dREG(80) = this->currentScene;
     dREG(81) = this->topScreen;
-    dREG(82) = this->unk_1DC;
+    dREG(82) = this->pageDownIndex;
 
     if (this->timerUp != 0) {
         this->timerUp--;
@@ -590,14 +590,14 @@ void Select_Init(GameState* thisx) {
     this->scenes = sScenes;
     this->topScreen = 0;
     this->currentScene = 0;
-    this->unk_1E0[0] = 0;  // " 1:SPOT00"
-    this->unk_1E0[1] = 19; // "20:\x8Dﾄｷﾉﾏ"
-    this->unk_1E0[2] = 37; // "38:\x8Dﾀｶﾗﾊﾞｺﾔ"
-    this->unk_1E0[3] = 51; // "52:\x8Dﾊｶﾓﾘﾉｲｴ"
-    this->unk_1E0[4] = 59; // "60:\x8Cｿﾞｰﾗ\x8Dﾉﾐｾ"
-    this->unk_1E0[5] = 73; // "74:\x8Dｲﾄﾞｼﾀ \x8Cﾀﾞﾝｼﾞｮﾝ"
-    this->unk_1E0[6] = 91; // "92:\x8Cｶﾞﾉﾝ\x8Dﾉﾄｳ ｿﾉｺﾞ 3"
-    this->unk_1DC = 0;
+    this->pageDownStops[0] = 0;  // " 1:SPOT00"
+    this->pageDownStops[1] = 19; // "20:\x8Dﾄｷﾉﾏ"
+    this->pageDownStops[2] = 37; // "38:\x8Dﾀｶﾗﾊﾞｺﾔ"
+    this->pageDownStops[3] = 51; // "52:\x8Dﾊｶﾓﾘﾉｲｴ"
+    this->pageDownStops[4] = 59; // "60:\x8Cｿﾞｰﾗ\x8Dﾉﾐｾ"
+    this->pageDownStops[5] = 73; // "74:\x8Dｲﾄﾞｼﾀ \x8Cﾀﾞﾝｼﾞｮﾝ"
+    this->pageDownStops[6] = 91; // "92:\x8Cｶﾞﾉﾝ\x8Dﾉﾄｳ ｿﾉｺﾞ 3"
+    this->pageDownIndex = 0;
     this->opt = 0;
     this->count = 126;
     View_Init(&this->view, this->state.gfxCtx);
@@ -615,7 +615,7 @@ void Select_Init(GameState* thisx) {
     if ((dREG(80) >= 0) && (dREG(80) < this->count)) {
         this->currentScene = dREG(80);
         this->topScreen = dREG(81);
-        this->unk_1DC = dREG(82);
+        this->pageDownIndex = dREG(82);
     }
     R_UPDATE_RATE = 1;
 
--- a/src/overlays/gamestates/ovl_select/z_select.c
+++ b/src/overlays/gamestates/ovl_select/z_select.c
@@ -178,7 +178,7 @@ void Select_UpdateMenu(SelectContext* this) {
     s32 pad;
     SceneSelectEntry* selectedScene;
 
-    if (this->unk_21C == 0) {
+    if (this->verticalInputAccumulator == 0) {
         if (CHECK_BTN_ALL(controller1->press.button, BTN_A) || CHECK_BTN_ALL(controller1->press.button, BTN_START)) {
             selectedScene = &this->scenes[this->currentScene];
             if (selectedScene->loadFunc != NULL) {
@@ -273,13 +273,13 @@ void Select_UpdateMenu(SelectContext* this) {
                 this->timerUp = 20;
                 this->lockUp = true;
                 Audio_PlaySoundGeneral(NA_SE_IT_SWORD_IMPACT, &D_801333D4, 4, &D_801333E0, &D_801333E0, &D_801333E8);
-                this->unk_220 = R_UPDATE_RATE;
+                this->verticalInput = R_UPDATE_RATE;
             }
         }
 
         if (CHECK_BTN_ALL(controller1->cur.button, BTN_DUP) && this->timerUp == 0) {
             Audio_PlaySoundGeneral(NA_SE_IT_SWORD_IMPACT, &D_801333D4, 4, &D_801333E0, &D_801333E0, &D_801333E8);
-            this->unk_220 = R_UPDATE_RATE * 3;
+            this->verticalInput = R_UPDATE_RATE * 3;
         }
 
         if (CHECK_BTN_ALL(controller1->press.button, BTN_DDOWN)) {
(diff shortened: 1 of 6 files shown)


src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -409,37 +535,52 @@ void Select_PrintMenu(SelectContext* this, GfxPrint* printer) {
...
-    "\x8Cｺｰﾋｰ ﾌﾞﾚｲｸ",    // "Coffee Break"
-    "\x8C"
-    "Bﾒﾝｦｾｯﾄｼﾃｸﾀﾞｻｲ",                      // "Please set B side"
-    "\x8Dｼﾞｯﾄ\x8Cｶﾞﾏﾝ\x8Dﾉ\x8Cｺ\x8Dﾃﾞｱｯﾀ", // "Be patient, now"
-    "\x8Dｲﾏｼﾊﾞﾗｸｵﾏﾁｸﾀﾞｻｲ",                 // "Please wait just a minute"
-    "\x8Dｱﾜﾃﾅｲｱﾜﾃﾅｲ｡ﾋﾄﾔｽﾐﾋﾄﾔｽﾐ｡",          // "Don't worry, don't worry. Take a break, take a break"
+// clang-format off
+static const char* sLoadingMessages[] = {
+    // "Please wait a minute",
+    "\x8Dｼﾊﾞﾗｸｵﾏﾁｸﾀﾞｻｲ",
```
Dragorn421, line 541: I liked these on the side better tbh, it looked less noisy

src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -588,34 +730,34 @@ void Select_Init(GameState* thisx) {
...
-    this->unk_1E0[5] = 0x49;
-    this->unk_1E0[6] = 0x5B;
-    this->unk_1DC = 0;
+    this->pageDownStops[0] = 0;  // " 1:SPOT00"
+    this->pageDownStops[1] = 19; // "20:\x8Dﾄｷﾉﾏ"
+    this->pageDownStops[2] = 37; // "38:\x8Dﾀｶﾗﾊﾞｺﾔ"
+    this->pageDownStops[3] = 51; // "52:\x8Dﾊｶﾓﾘﾉｲｴ"
+    this->pageDownStops[4] = 59; // "60:\x8Cｿﾞｰﾗ\x8Dﾉﾐｾ"
+    this->pageDownStops[5] = 73; // "74:\x8Dｲﾄﾞｼﾀ \x8Cﾀﾞﾝｼﾞｮﾝ"
+    this->pageDownStops[6] = 91; // "92:\x8Cｶﾞﾉﾝ\x8Dﾉﾄｳ ｿﾉｺﾞ 3"
```
Dragorn421, line 741: I don't think these comments are of much use

src/overlays/gamestates/ovl_select/z_select.c
fig02: agreed. would say the same for the translation comments next to the entries in the table

src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -43,144 +43,143 @@ void Select_LoadGame(SelectContext* this, s32 entranceIndex) {
...
-    { "121:\x8Cﾎﾞｽｽﾀﾛﾌｫｽ\x8Dﾍﾞﾔ", Select_LoadGame, 0x0020 },
-    { "122:Sutaru", Select_LoadGame, 0x0047 },
-    { "123:jikkenjyou", Select_LoadGame, 0x02EA },
-    { "124:depth\x8Cﾃｽﾄ", Select_LoadGame, 0x00B6 },
-    { "125:\x8Cﾊｲﾗﾙ\x8Dﾆﾜ\x8Cｹﾞｰﾑ2", Select_LoadGame, 0x0076 },
-    { "title", Select_LoadTitle, 0x0000 },
+    { " 1:SPOT00", Select_LoadGame, 0x00CD }, // "1: SPOT00" ("1: Hyrule Field")
+    { " 2:SPOT01", Select_LoadGame, 0x00DB }, // "2: SPOT01" ("2: Kakariko Village")
+    { " 3:SPOT02", Select_LoadGame, 0x00E4 }, // "3: SPOT02" ("3: Graveyard")
+    { " 4:SPOT03", Select_LoadGame, 0x00EA }, // "4: SPOT03" ("4: Zora's River")
```
fig02, line 51: imo, no need for quotes and index numbers on the "real" name
`// "1: SPOT00" (Hyrule Field)`
would be my personal preference

src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -409,37 +409,40 @@ void Select_PrintMenu(SelectContext* this, GfxPrint* printer) {
...
-    "\x8D"
-    "5(ﾜｶｽｷﾞ)", // "5(very young)"
+// clang-format off
+static const char* sAgeLabels[] = {
+    // "17(young)"
+    "\x8D" "17(ﾜｶﾓﾉ)",
+    // "5(very young)"
+    "\x8D" "5(ﾜｶｽｷﾞ)",
 };
+// clang-format on
```
fig02, line 445: i would move these to the side as well (and remove clang comments)

src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -575,7 +578,8 @@ void Select_Main(GameState* thisx) {
 }
 
 void Select_Destroy(GameState* thisx) {
-    osSyncPrintf("%c", 7);
+    // ASCII's 0x07 is the BEL character
+    osSyncPrintf("%c", '\x07');
```
fig02, line 582: side also

src/overlays/gamestates/ovl_select/z_select.c
fig02: i think if the comments said the "useful" place name without the index it could be of use, but with the japanese and control codes it doesnt add any more helpful information

as a side note i would possibly suggest `pageLimit` as a name for pageDownStops, and call the index into it the `pageIndex`

fig02: so yea my suggestion is 
```
this->pageDownStops[3] = 51; // Gravekeeper's Hut
this->pageDownStops[4] = 59; // Goron Shop
```

src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -43,144 +43,143 @@ void Select_LoadGame(SelectContext* this, s32 entranceIndex) {
...
+    { "44:\x8Dﾘﾝｸﾉｲｴ", Select_LoadGame, 0x00BB }, // "44: Link's House" ("44: Link's House")
+    { "45:\x8Cｶｶﾘｺ\x8Dﾑﾗﾉﾅｶﾞﾔ", Select_LoadGame, 0x02FD }, // "45: Kakariko Village Row House" ("45: Carpenter Boss's House")
+    { "46:\x8Dｳﾗﾛｼﾞﾉ ｲｴ", Select_LoadGame, 0x043B }, // "46: Back Alley House" ("46: Back Alley House (Man in Green)")
+    { "47:\x8Dｺｷﾘﾉﾑﾗ ﾓﾉｼﾘｷｮｳﾀﾞｲﾉｲｴ", Select_LoadGame, 0x00C9 }, // "47: Kokiri Village - Know-It-All Brothers' House" ("47: Know-It-All Brothers' House")
+    { "48:\x8Dｺｷﾘﾉﾑﾗ ﾌﾀｺﾞﾉｲｴ", Select_LoadGame, 0x009C }, // "48: Kokiri Village - Twins' House" ("48: Twins' House")
+    { "49:\x8Dｺｷﾘﾉﾑﾗ \x8Cﾐﾄﾞ\x8Dﾉｲｴ", Select_LoadGame, 0x0433 }, // "49: Kokiri Village - Mido's House" ("49: Mido's House")
+    { "50:\x8Dｺｷﾘﾉﾑﾗ \x8Cｻﾘｱ\x8Dﾉｲｴ", Select_LoadGame, 0x0437 }, // "50: Kokiri Village - Saria's House" ("50: Saria's House")
+    { "51:\x8Dｳﾏｺﾞﾔ", Select_LoadGame, 0x02F9 }, // "51: Stable" ("51: Stable")
+    { "52:\x8Dﾊｶﾓﾘﾉｲｴ", Select_LoadGame, 0x030D }, // "52: Grave Keeper's House" ("52: Gravekeeper's Hut")
+    { "53:\x8Dｳﾗﾛｼﾞ ｲﾇｵﾊﾞｻﾝﾉｲｴ", Select_LoadGame, 0x0398 }, // "53: Back Alley - Dog Lady's House" ("53: Back Alley House (Mamamu Yan)")
```
fig02, line 100: Mamamu Yan is apparently the name this character got in mm. Would not include it in this game since afaik she is unnamed
Today i learned she had a name in mm though lol

src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -43,144 +43,143 @@ void Select_LoadGame(SelectContext* this, s32 entranceIndex) {
...
+    { "109:\x8Dｶｸｼﾄﾋﾞｺﾐｱﾅ 8", Select_LoadGame, 0x05B4 }, // "109: Hidden Dive Hole 8" ("109: Grotto 8 - 2 Wolfos")
+    { "110:\x8Dｶｸｼﾄﾋﾞｺﾐｱﾅ 9", Select_LoadGame, 0x05B8 }, // "110: Hidden Dive Hole 9" ("110: Grotto 9 - Bombable Walls")
+    { "111:\x8Dｶｸｼﾄﾋﾞｺﾐｱﾅ 10", Select_LoadGame, 0x05BC }, // "111: Hidden Dive Hole 10" ("111: Grotto 10 - 2 Business Scrubs")
+    { "112:\x8Dｶｸｼﾄﾋﾞｺﾐｱﾅ 11", Select_LoadGame, 0x05C0 }, // "112: Hidden Dive Hole 11" ("112: Grotto 11 - Tektite & Heart Piece")
+    { "113:\x8Dｶｸｼﾄﾋﾞｺﾐｱﾅ 12", Select_LoadGame, 0x05C4 }, // "113: Hidden Dive Hole 12" ("113: Grotto 12 - Deku Stage")
+    { "114:\x8Dｶｸｼﾄﾋﾞｺﾐｱﾅ 13", Select_LoadGame, 0x05FC }, // "114: Hidden Dive Hole 13" ("114: Grotto 13 - Rupees & Cow")
+    { "115:\x8Cﾊｲﾗﾙ ﾃﾞﾓ", Select_LoadGame, 0x00A0 }, // "115: Hyrule Cutscenes" ("115: Cutscene Map")
+
+    { "116:\x8Dﾍﾞｯｼﾂ (ﾀｶﾗﾊﾞｺ\x8Cﾜｰﾌﾟ)", Select_LoadGame, 0x0520 }, // "116: Special Room (Treasure Chest Warp)" ("116: Ganondorf Test Room")
+    { "117:\x8Dｻｻ\x8Cﾃｽﾄ", Select_LoadGame, 0x0018 }, // "117: Sasaki Test" ("117: Sasa Test")
```
fig02, line 165: oh and i would remove extra break

src/overlays/gamestates/ovl_select/z_select.c
AngheloAlf: `pageLimit` would be wrong, because it isn't limiting anything, it is jumping to one of those positions each time you press L

fig02: those are the limits of each page to me, marking the page boundaries

EllipticEllipsis: The differences aren't even, though, it's something like 19,18,14,8,14,18

fig02: i dont think that fact is really relevant to the name "limit". to me its basically a synonym to "page down stops", the limit of which the page is finished
maybe there is a better word to describe my meaning, but i think in general we can improve pagedownstops

src/overlays/gamestates/ovl_select/z_select.c
AngheloAlf: Removing the clang-format comments does this. Is it fine?
```c

static const char* sAgeLabels[] = {
    "\x8D"
    "17(ﾜｶﾓﾉ)", // "17(young)"
    "\x8D"
    "5(ﾜｶｽｷﾞ)", // "5(very young)"
};

```

fig02: i would say so. in general we try to avoid disabling clang unless we really need to.
we may experiment with other formatters or format rules in the future which will fix tihs problem (and the many other issues some people have with it)
but for now this isnt too much of a big deal imo

src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -43,144 +43,163 @@ void Select_LoadGame(SelectContext* this, s32 entranceIndex) {
...
-    { "118:\x8Cﾃｽﾄﾏｯﾌﾟ", Select_LoadGame, 0x0094 },
-    { "119:\x8Cﾃｽﾄﾙｰﾑ", Select_LoadGame, 0x0024 },
-    { "120:\x8Dﾁｭｳ\x8Cｽﾀﾛﾌｫｽ\x8Dﾍﾞﾔ", Select_LoadGame, 0x001C },
-    { "121:\x8Cﾎﾞｽｽﾀﾛﾌｫｽ\x8Dﾍﾞﾔ", Select_LoadGame, 0x0020 },
-    { "122:Sutaru", Select_LoadGame, 0x0047 },
-    { "123:jikkenjyou", Select_LoadGame, 0x02EA },
-    { "124:depth\x8Cﾃｽﾄ", Select_LoadGame, 0x00B6 },
-    { "125:\x8Cﾊｲﾗﾙ\x8Dﾆﾜ\x8Cｹﾞｰﾑ2", Select_LoadGame, 0x0076 },
-    { "title", Select_LoadTitle, 0x0000 },
+    { " 1:SPOT00", Select_LoadGame, 0x00CD },                 // "1: SPOT00" (Hyrule Field)
```
Zelllll, line 48: I don't see the comments as being incredibly necessary. The better thing to do would be simply documenting the entrances, (like I already helpfully did here: https://github.com/Zelllll/oot/blob/ent/src/overlays/gamestates/ovl_select/z_select.c#L47 ), though overall it doesn't matter too much. Also saying that SPOT00 is Hyrule Field isn't really needed either considering that eventually the enum will just be `SCENE_HYRULE_FIELD` and `ENT_HYRULE_FIELD_0` probably.

src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -409,32 +429,32 @@ void Select_PrintMenu(SelectContext* this, GfxPrint* printer) {
...
-    "\x8Dｼﾊﾞﾗｸｵﾏﾁｸﾀﾞｻｲ", // "Please wait a minute"
-    "\x8Dﾁｮｯﾄ ﾏｯﾃﾈ",     // "Hold on a sec"
-    "\x8Cｳｪｲﾄ ｱ ﾓｰﾒﾝﾄ",  // "Wait a moment"
-    "\x8Cﾛｰﾄﾞ\x8Dﾁｭｳ",   // "Loading"
-    "\x8Dﾅｳ ﾜｰｷﾝｸﾞ",     // "Now working"
-    "\x8Dｲﾏ ﾂｸｯﾃﾏｽ",     // "Now creating"
-    "\x8Dｺｼｮｳｼﾞｬﾅｲﾖ",    // "It's not broken"
-    "\x8Cｺｰﾋｰ ﾌﾞﾚｲｸ",    // "Coffee Break"
+static const char* sLoadingMessages[] = {
+    "\x8Dｼﾊﾞﾗｸｵﾏﾁｸﾀﾞｻｲ", // "Please wait a minute",
```
Zelllll, line 433: Are the commas here really needed? I am assuming that the idea would be to uncomment them and the commas are already there for translating it, but eventually that could probably be an ifdef anyways. Generally I don't think we put commas in comments though like this.

src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -409,32 +429,32 @@ void Select_PrintMenu(SelectContext* this, GfxPrint* printer) {
...
 
-    GfxPrint_SetPos(printer, 0xA, 0xF);
+    GfxPrint_SetPos(printer, 10, 15);
     GfxPrint_SetColor(printer, 255, 255, 255, 255);
     randomMsg = Rand_ZeroOne() * ARRAY_COUNT(sLoadingMessages);
     GfxPrint_Printf(printer, "%s", sLoadingMessages[randomMsg]);
 }
 
-static char* sAgeLabels[] = {
+static const char* sAgeLabels[] = {
```
Zelllll, line 457: This should be inside of `Select_PrintAgeSetting` rather than above it, and renamed to `ageLabels` since we do not use the `s` prefix on variables that are inside of functions.

src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -409,32 +429,32 @@ void Select_PrintMenu(SelectContext* this, GfxPrint* printer) {
...
-static char* sLoadingMessages[] = {
-    "\x8Dｼﾊﾞﾗｸｵﾏﾁｸﾀﾞｻｲ", // "Please wait a minute"
-    "\x8Dﾁｮｯﾄ ﾏｯﾃﾈ",     // "Hold on a sec"
-    "\x8Cｳｪｲﾄ ｱ ﾓｰﾒﾝﾄ",  // "Wait a moment"
-    "\x8Cﾛｰﾄﾞ\x8Dﾁｭｳ",   // "Loading"
-    "\x8Dﾅｳ ﾜｰｷﾝｸﾞ",     // "Now working"
-    "\x8Dｲﾏ ﾂｸｯﾃﾏｽ",     // "Now creating"
-    "\x8Dｺｼｮｳｼﾞｬﾅｲﾖ",    // "It's not broken"
-    "\x8Cｺｰﾋｰ ﾌﾞﾚｲｸ",    // "Coffee Break"
+static const char* sLoadingMessages[] = {
```
Zelllll, line 432: This should be inside of `Select_PrintLoadingMessage` and renamed to `loadingMessages` since we do not use the `s` prefix for variables inside of functions.

fig02 (approved):
still think we need to work on the translation comments a bit more to be less redundant and a but more clean. but can take care of that some other time. everything else is fine

PR #972, opened 2021-09-16
Clean up some float literals

petrie911:
Cleaned up some float literals that were rational numbers or rational multiples of `M_PI`, and added comments for some that were trig literals.


src/code/audio_playback.c
```diff
@@ -79,9 +79,9 @@ void Audio_InitNoteSub(Note* note, NoteSubEu* sub, NoteSubAttributes* attrs) {
     } else if (gAudioContext.soundMode == 3) {
         sub->bitField0.s.stereoHeadsetEffects = false;
         sub->bitField0.s.usesHeadsetPanEffects = false;
-        volLeft = 0.707f;
+        volLeft = 0.707f; // approx 1/sqrt(2)
         volRight = 0.707f;
-    } else {
```
Dragorn421, line 84: this is probably what broke matching

src/code/audio_playback.c
```diff
@@ -79,9 +79,9 @@ void Audio_InitNoteSub(Note* note, NoteSubEu* sub, NoteSubAttributes* attrs) {
     } else if (gAudioContext.soundMode == 3) {
         sub->bitField0.s.stereoHeadsetEffects = false;
         sub->bitField0.s.usesHeadsetPanEffects = false;
-        volLeft = 0.707f;
+        volLeft = 0.707f; // approx 1/sqrt(2)
```
Dragorn421, line 82: in another place you used "about" instead of "approx", should probably decide on something for consistency

or sometimes there's no "approx" or equivalent at all: `0.2333333f); // 7 / 30`, `0.7071) { // cos(45 degrees)`

so just no "approx"/"about" ?

Dragorn421: what about
```suggestion
        volLeft = 0.707f; // ~ 1/sqrt(2)
```
or `~=`
or `≈` but I don't think IDO would like that character

fig02 (approved):
agreed on cleaning up "approx" vs "about", but is something we can take care of later

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -1400,7 +1400,7 @@ void func_80900890(BossGanon2* this, GlobalContext* globalCtx) {
             break;
         case 2:
             Math_ApproachF(&this->unk_324, 255.0f, 1.0f, 10.0f);
-            Math_ApproachZeroF(&globalCtx->envCtx.unk_D8, 1.0f, 0.039215688f);
+            Math_ApproachZeroF(&globalCtx->envCtx.unk_D8, 1.0f, 2.0f / 51.0f);
```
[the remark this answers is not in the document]
fig02: maybe we leave this for another float pass, cause there will surely be more. will merge in a few hours unless you feel like getting to this @petrie911

PR #978, opened 2021-09-19
Match remaining z_room functions

Synray:


src/code/z_room.c
```diff
@@ -167,45 +164,45 @@ void func_80095D04(GlobalContext* globalCtx, Room* room, u32 flags) {
         }
     }
 
-    iREG(87) = polygon2->num;
+    iREG(87) = polygon2->num & 0xffff & 0xffff & 0xffff; // if this is real then I might not be
```
Dragorn421, line 167:
```suggestion
    iREG(87) = polygon2->num & 0xFFFF & 0xFFFF & 0xFFFF; // if this is real then I might not be
```
at least you can be unreal in all caps

Dragorn421 (changes requested):

Dragorn421 (approved):

Roman971 (approved):

fig02 (approved):

PR #983, opened 2021-09-21
Fix/cleanup/rephrase miscellaneous stuff

Dragorn421:
Can I make this more vague? Cf commit messages for actual changes


src/boot/is_debug.c
```diff
@@ -2,14 +2,13 @@
...
-#define gISVDbgPrnAdrs ((ISVDbg*)0xb3ff0000)
-#define ASCII_TO_U32(a, b, c, d) ((u32)((a << 24) | (b << 16) | (c << 8) | (d << 0)))
+#define gISVDbgPrnAdrs ((ISVDbg*)0xB3FF0000)
 
 void isPrintfInit(void) {
     sISVHandle = osCartRomInit();
     osEPiWriteIo(sISVHandle, (u32)&gISVDbgPrnAdrs->put, 0);
     osEPiWriteIo(sISVHandle, (u32)&gISVDbgPrnAdrs->get, 0);
-    osEPiWriteIo(sISVHandle, (u32)&gISVDbgPrnAdrs->magic, ASCII_TO_U32('I', 'S', '6', '4'));
+    osEPiWriteIo(sISVHandle, (u32)&gISVDbgPrnAdrs->magic, 'IS64');
```
Roman971, line 11: I don't know how I feel about actually using multichars. It raises a gcc warning and it feels a bit like using an unsupported part of the C language (iirc it's implementation defined at best).

Dragorn421: For some reason I got the impression this was well defined but yeah it's implementation-defined, reverting

Roman971:
`PLAYER_` enums aren't actually supposed to be used in the context of inventory stuff, they are meant for stuff like `Player::currentTunic`. I know some incorrect checks got in the repo over time in that regard, so the fix should probably be to remove those.

An enum for upgrade values and such could exist, but we don't have any right now, and it would have to be a separate thing imo.

Roman971 (approved):

fig02 (approved):

PR #987, opened 2021-09-25
Improve `fixbaserom.py`

Dragorn421:
* Set bytes past ROM's end to `0xFF`
* Recognize the infamous ZeldaEdit-edited rom
* also look at `baserom_original.ext` with `ext` in uppercase

### DIFF
--- a/fixbaserom.py
+++ b/fixbaserom.py
@@ -17,14 +17,19 @@ if path.exists("baserom.z64"):
             sys.exit(0)
 
 # Determine if we have a ROM file
-romFileName = ""
-if path.exists("baserom_original.z64"):
-    romFileName = "baserom_original.z64"
-elif path.exists("baserom_original.n64"):
-    romFileName = "baserom_original.n64"
-elif path.exists("baserom_original.v64"):
-    romFileName = "baserom_original.v64"
-else:
+romFileExtensions = ["z64", "n64", "v64"]
+
+def find_baserom_original():
+    for romFileExtLower in romFileExtensions:
+        for romFileExt in (romFileExtLower, romFileExtLower.upper()):
+            romFileNameCandidate = "baserom_original." + romFileExt
+            if path.exists(romFileNameCandidate):
+                return romFileNameCandidate
+    return None
+
+romFileName = find_baserom_original()
+
+if romFileName is None:
     print("Error: Could not find baserom_original.z64/baserom_original.n64/baserom_original.v64.")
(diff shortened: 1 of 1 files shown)


fixbaserom.py
```diff
@@ -68,11 +73,19 @@ def get_str_hash(byte_array):
 print("Patching header...")
 fileContent[0x3E] = 0x50
 
+for i in range(0x35CF000, len(fileContent)):
+    fileContent[i] = 0xFF
```
AngheloAlf, line 77: What's the advantage of doing this?

Dragorn421: the whole goal with this PR is make it easier for people to use whatever equivalent rom they have

there are roms floating around that have 0x00 padding instead of 0xFF at the end of the rom

(the rom "elf" ends at 0x35CF000, cf map file `0x00000000035cf000                _RomEnd = _RomSize`)

fixbaserom.py
```diff
@@ -68,11 +73,19 @@ def get_str_hash(byte_array):
...
+
 # Check to see if the ROM is a "vanilla" Debug ROM
 str_hash = get_str_hash(bytearray(fileContent))
 if str_hash != "f0b7f35375f9cc8ca1b2d59d78e35405":
     print("Error: Expected a hash of f0b7f35375f9cc8ca1b2d59d78e35405 but got " + str_hash + ". " +
           "The baserom has probably been tampered, find a new one")
+
+    if str_hash == "32fe2770c0f9b1a9cd2a4d449348c1cb":
+        print("The provided baserom is a rom which has been edited with ZeldaEdit and is not suitable for use with decomp. " +
+              "Find a new one.")
```
Roman971, line 87: Isn't this just redundant with the message we already provide?

Dragorn421: It is, but imo acknowledging an unsupported ROM is better than just saying the ROM isn't supported (I may even want to add a check for the oot 1.0 n64 rom and other versions...)
And it also helps make a user aware there are several variants of the rom

It's common for the average user to ignore an error message, regardless of how nice and friendly it is. Hopefully duplicating the "find a new one" statement gets the message across better
It's not like this is cluttering the `make setup -j` output by producing 3 times as much prints

fig02 (approved):

Roman971 (approved):
I don't think we should add checks for specific ROMs like I mentioned in my previous comments, but I'm going to approve anyway since it's not too important rn

PR #988, opened 2021-09-26
EnZf OK and partially documented

EllipticEllipsis:
It took 8 months, but it's OK and PR'd. Thanks to @petrie911 for encouraging me to do it in the first place and matching those awful platform functions that we all hate, @AngheloAlf for getting me started with some ideas about the struct variables, and @Dragorn421 for getting me to read EnWf carefully, which helped a lot with this thing.

I got a lot farther with documenting than I was expecting, but a few of the actions remain a bit elusive, and I'm a bit stuck on names for other things, both functions and struct variables. Didn't put much effort into naming all the temps since I assume we're going to try and make the fighter actors consistent anyway. I started the ball rolling on that by attempting to unify all the weird curframe things that play sound, although I didn't try to name those consistently yet either.

I'll probably carry on doing some documentation, particularly the textures in the object, but thought it was worth PRing now to get some more eyes on it.

What a mess of an actor. I've never seen setup functions that sometimes set a different function instead 1UL.


EllipticEllipsis:
Oh, almost forgot:
image
Yuck!

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+
+    this->alpha = 255;
+    thisx->colChkInfo.cylRadius = 40;
+    thisx->colChkInfo.cylHeight = 100;
+    Collider_InitCylinder(globalCtx, &this->bodyCollider);
+    Collider_SetCylinder(globalCtx, &this->bodyCollider, thisx, &sBodyCylinderInit);
+    Collider_InitQuad(globalCtx, &this->swordCollider);
+    Collider_SetQuad(globalCtx, &this->swordCollider, thisx, &sSwordQuadInit);
+
+    if (thisx->params == ENZF_TYPE_DINOLFOS) { // Dinalfos
```
Dragorn421, line 319:
```suggestion
    if (thisx->params == ENZF_TYPE_DINOLFOS) { // Dinolfos
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+        thisx->colChkInfo.health = 12;
+        thisx->naviEnemyId = 0x10;
+        SkelAnime_Init(globalCtx, &this->skelAnime, &gZfDinolfosSkel, &gZFCryingAnim, this->jointTable,
+                       this->morphTable, ENZF_LIMB_MAX);
+    } else { // Lizalfos
+        thisx->colChkInfo.health = 6;
+        thisx->naviEnemyId = 0xF;
```
Dragorn421, line 326:
```suggestion
        thisx->naviEnemyId = 0x0F;
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+    if ((startIndex != -1) && ((sPlatformPositions[startIndex].y - 150.0f) <= pos->y) &&
+        (pos->y <= (sPlatformPositions[startIndex].y + 150.0f)) &&
+        ((sPlatformPositions[startIndex].x - rangeXZ) <= pos->x) &&
+        (pos->x <= (sPlatformPositions[startIndex].x + rangeXZ)) &&
+        ((sPlatformPositions[startIndex].z - rangeXZ) <= pos->z) &&
+        (pos->z <= (sPlatformPositions[startIndex].z + rangeXZ))) {
+        return startIndex;
+    }
+
+    for (i = ARRAY_COUNT(sPlatformPositions) - 1; i > -1; i--) {
```
Dragorn421, line 390:
```suggestion
    for (i = ARRAY_COUNT(sPlatformPositions) - 1; i >= 0; i--) {
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+                if (Actor_IsFacingPlayer(&this->actor, 30 * 0x10000 / 360)) {
+                    if ((this->actor.xzDistToPlayer < 200.0f) && (this->actor.xzDistToPlayer > 100.0f) &&
+                        (Rand_ZeroOne() < 0.3f)) {
+                        if (this->actor.params == ENZF_TYPE_DINOLFOS) {
+                            this->actor.world.rot.y = this->actor.shape.rot.y = this->actor.yawTowardsPlayer;
+                            EnZf_SetupJumpForward(this);
+                        } else {
+                            func_80B483E4(this, globalCtx);
+                        }
+                    } else if ((Rand_ZeroOne() > 0.3f)) {
```
Dragorn421, line 737:
```suggestion
                    } else if (Rand_ZeroOne() > 0.3f) {
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+        }
+
+        sp54 = this->skelAnime.curFrame;
+        SkelAnime_Update(&this->skelAnime);
+        sp50 = this->skelAnime.curFrame - ABS(this->skelAnime.playSpeed);
+        sp30 = (f32)ABS(this->skelAnime.playSpeed);
+
+        if (sp48 == this->curPlatform) {
+            if (!Actor_IsFacingPlayer(&this->actor, 0x11C7)) {
+                if ((Rand_ZeroOne() > 0.5f)) {
```
Dragorn421, line 858:
```suggestion
                if (Rand_ZeroOne() > 0.5f) {
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+                    func_80B462E4(this, globalCtx);
+                } else {
+                    func_80B45384(this);
+                }
+            } else if ((this->actor.xzDistToPlayer < 100.0f)) {
```
Dragorn421, line 863:
```suggestion
            } else if (this->actor.xzDistToPlayer < 100.0f) {
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+                if ((Rand_ZeroOne() > 0.05f) && EnZf_MayAttack(globalCtx, this)) {
+                    EnZf_SetupSlash(this);
+                } else if ((Rand_ZeroOne() > 0.5f)) {
```
Dragorn421, line 866:
```suggestion
                } else if (Rand_ZeroOne() > 0.5f) {
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+
+// Jump between platforms
+void EnZf_SetupJumpForward(EnZf* this) {
+    Animation_Change(&this->skelAnime, &gZfJumpingAnim, 1.0f, 0.0f, 3.0f, 2, -3.0f);
+    this->unk_3F0 = 0;
+    this->hopAnimIndex = 1;
+    this->actor.velocity.y = 15.0f;
+
+    // miniboss
+    if (this->actor.params >= ENZF_TYPE_LIZALFOS_MINIBOSS_A) {
```
Dragorn421, line 922:
```suggestion
    if (this->actor.params >= ENZF_TYPE_LIZALFOS_MINIBOSS_A) { // miniboss
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+            // miniboss
+            if (this->actor.params >= ENZF_TYPE_LIZALFOS_MINIBOSS_A) {
+                this->curPlatform = EnZf_FindPlatform(&this->actor.world.pos, this->curPlatform);
+                if (this->curPlatform != EnZf_FindPlatform(&player->actor.world.pos, -1)) {
+                    EnZf_SetupApproachPlayer(this, globalCtx);
+                    return;
+                }
+            }
+
+            if (Actor_IsFacingPlayer(&this->actor, 30 * 0x10000 / 360)) {
```
Dragorn421, line 1015:
```suggestion
            if (Actor_IsFacingPlayer(&this->actor, DEGF_TO_BINANG(30))) {
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+                Audio_PlayActorSound2(&this->actor, NA_SE_EN_RIZA_CRY);
+            }
+        }
+    }
+}
+
+// Conditional setup function
+void func_80B462E4(EnZf* this, GlobalContext* globalCtx) {
+    // miniboss
+    if ((this->actor.params < ENZF_TYPE_LIZALFOS_MINIBOSS_A) ||
```
Dragorn421, line 1032:
```suggestion
    if ((this->actor.params < ENZF_TYPE_LIZALFOS_MINIBOSS_A) || /* not miniboss */
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+            }
+        } else if (Math_SinS(angleBehindPlayer - this->actor.shape.rot.y) < 0.0f) { // Superfluous check
+            this->actor.speedXZ += 0.25f;
+            if (this->actor.speedXZ > 8.0f) {
+                this->actor.speedXZ = 8.0f;
+            }
+        }
+
+        // miniboss
+        if (this->actor.params >= ENZF_TYPE_LIZALFOS_MINIBOSS_A) {
```
Dragorn421, line 1078:
```suggestion
        if (this->actor.params >= ENZF_TYPE_LIZALFOS_MINIBOSS_A) { // miniboss
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+            }
+        }
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Zf/func_80B446A8.s")
+        if ((globalCtx->gameplayFrames & 0x5F) == 0) {
+            Audio_PlayActorSound2(&this->actor, NA_SE_EN_RIZA_CRY);
+        }
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Zf/func_80B44870.s")
+        if ((Math_CosS((angleBehindPlayer - this->actor.shape.rot.y)) < -0.85f) || (this->unk_3F0 == 0)) {
```
Dragorn421, line 1149:
```suggestion
        if ((Math_CosS(angleBehindPlayer - this->actor.shape.rot.y) < -0.85f) || (this->unk_3F0 == 0)) {
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Zf/func_80B45E30.s")
+                if (rotDiff <= 10000) {
+                    yawDiff = this->actor.yawTowardsPlayer - this->actor.shape.rot.y;
+                    yawDiff = ABS(yawDiff);
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Zf/func_80B45EF0.s")
+                    if (yawDiff > 16000) {
+                        this->actor.world.rot.y = this->actor.yawTowardsPlayer;
+                        func_80B483E4(this, globalCtx);
+                    } else if ((player->stateFlags1 & 0x6010)) {
```
Dragorn421, line 1215:
```suggestion
                    } else if ((player->stateFlags1 & 0x6010) != 0) {
```
I like != 0 when several flags, if not at least remove parens

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+        if (this->actor.speedXZ < 0.0f) {
+            this->actor.speedXZ += 0.05f;
+        }
+        this->hopAnimIndex = 0;
+    }
+
+    if ((this->actor.colorFilterTimer == 0) && (this->actor.bgCheckFlags & 1)) {
+        if (this->actor.colChkInfo.health == 0) {
+            EnZf_SetupDie(this);
+        } else if ((this->actor.params != ENZF_TYPE_DINOLFOS) || (!EnZf_ChooseAction(globalCtx, this))) {
```
Dragorn421, line 1324:
```suggestion
        } else if ((this->actor.params != ENZF_TYPE_DINOLFOS) || !EnZf_ChooseAction(globalCtx, this)) {
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+    this->unk_408 = 0.0f;
+    EnZf_SetupAction(this, EnZf_HopAndTaunt);
+}
+
+void EnZf_HopAndTaunt(EnZf* this, GlobalContext* globalCtx) {
+    f32 lastFrame;
+    f32 maxDist = 400.0f;
+
+    Math_SmoothStepToF(&this->actor.speedXZ, 0.0f, 1.0f, 0.5f, 0.0f);
+    Math_SmoothStepToS(&this->actor.shape.rot.y, (this->actor.yawTowardsPlayer + 0x8000), 1, 4000, 0);
```
Dragorn421, line 1396:
```suggestion
    Math_SmoothStepToS(&this->actor.shape.rot.y, this->actor.yawTowardsPlayer + 0x8000, 1, 4000, 0);
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+                    this->unk_40C = 0.0f;
+                    break;
+
+                case 2:
+                    this->actor.world.pos.y = this->actor.floorHeight;
+                    lastFrame = 3.0f;
+                    break;
+
+                default:
+                    break;
```
Dragorn421, line 1445: is this needed for matching?

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+            this->actor.shape.rot.y = sp5A + 0x8000;
+            D_80B4AB30 = 0;
+            this->homePlatform = this->curPlatform;
+            temp_v1_2 = !EnZf_PrimaryFloorCheck(this, globalCtx, 107.0f);
+            temp_v1_2 |= !EnZf_PrimaryFloorCheck(this, globalCtx, 220.0f) << 1;
+            this->hopAnimIndex++;
+
+            switch (temp_v1_2) {
+                case 1:
+                case 1 | 2:
```
Dragorn421, line 1506:
```suggestion
                case 0b01:
                case 0b11:
```
may make sense to use here

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+        case 1:
+            if ((this->actor.bgCheckFlags & 2) || (this->actor.bgCheckFlags & 1)) {
+                Audio_PlayActorSound2(&this->actor, NA_SE_EN_RIZA_ONGND);
+                this->actor.velocity.y = 0.0f;
+                this->actor.world.pos.y = this->actor.floorHeight;
+                this->actor.speedXZ = 0.0f;
+                Actor_SpawnFloorDustRing(globalCtx, &this->actor, &this->leftFootPos, 3.0f, 2, 2.0f, 0, 0, 0);
+                Actor_SpawnFloorDustRing(globalCtx, &this->actor, &this->rightFootPos, 3.0f, 2, 2.0f, 0, 0, 0);
+
+                if ((phi_f20 <= this->actor.xzDistToPlayer)) {
```
Dragorn421, line 1550:
```suggestion
                if (phi_f20 <= this->actor.xzDistToPlayer) {
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+                    EnZf_SetupHopAndTaunt(this);
+                } else if ((sp74 < 80.0f)) {
```
Dragorn421, line 1552:
```suggestion
                } else if (sp74 < 80.0f) {
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+    this->action = ENZF_ACTION_JUMP_UP;
+    this->actor.velocity.y = 22.0f;
+    this->actor.speedXZ = 7.5f;
+    Audio_PlayActorSound2(&this->actor, NA_SE_EN_RIZA_JUMP);
+    this->actor.world.rot.y = this->actor.shape.rot.y;
+    EnZf_SetupAction(this, EnZf_JumpUp);
+}
+
+void EnZf_JumpUp(EnZf* this, GlobalContext* globalCtx) {
+    s16 temp_yaw; // Need this temp for stack size
```
Dragorn421, line 1712:
```suggestion
    s16 tempYaw; // Need this temp for stack size
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+        Actor_TestFloorInDirection(&this->actor, globalCtx, 40.0f, (s16)(this->actor.shape.rot.y + 0x3FFF)) ||
+        Actor_TestFloorInDirection(&this->actor, globalCtx, -40.0f, (s16)(this->actor.shape.rot.y + 0x3FFF))) {
+        Animation_PlayLoop(&this->skelAnime, &gZfSidesteppingAnim);
+        player = GET_PLAYER(globalCtx);
+        Math_SmoothStepToS(&this->actor.shape.rot.y, this->actor.yawTowardsPlayer, 1, 4000, 1);
+        playerRotY = player->actor.shape.rot.y;
+
+        if (Math_SinS(playerRotY - this->actor.shape.rot.y) >= 0.0f) {
+            this->actor.speedXZ = -6.0f;
+        } else if (Math_SinS(playerRotY - this->actor.shape.rot.y) < 0.0f) { // Yes, this check is superfluous
```
Dragorn421, line 1754:
```suggestion
        } else if (Math_SinS(playerRotY - this->actor.shape.rot.y) < 0.0f) { // Superfluous check
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+        Math_SmoothStepToF(&this->unk_408, -4.0f, 1.0f, 1.5f, 0.0f);
+    } else if ((90.0f + baseRange) < this->actor.xzDistToPlayer) {
+        Math_SmoothStepToF(&this->unk_408, 4.0f, 1.0f, 1.5f, 0.0f);
+    } else {
+        Math_SmoothStepToF(&this->unk_408, 0.0f, 1.0f, 5.65f, 0.0f);
+    }
+
+    if ((this->unk_408 != 0.0f) && !EnZf_SecondaryFloorCheck(this, globalCtx, this->unk_408)) {
+        this->actor.world.pos.x += (Math_SinS(this->actor.shape.rot.y) * this->unk_408);
+        this->actor.world.pos.z += (Math_CosS(this->actor.shape.rot.y) * this->unk_408);
```
Dragorn421, line 1833:
```suggestion
        this->actor.world.pos.x += Math_SinS(this->actor.shape.rot.y) * this->unk_408;
        this->actor.world.pos.z += Math_CosS(this->actor.shape.rot.y) * this->unk_408;
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+        this->actor.speedXZ = 0.0f;
+        // miniboss
+        if ((this->actor.params >= ENZF_TYPE_LIZALFOS_MINIBOSS_A) && (D_80B4A1B4 == this->actor.params)) {
+            EnZf_SetupHopAndTaunt(this);
+            return;
+        }
+        EnZf_SetupApproachPlayer(this, globalCtx);
+        return;
+    }
+    if ((this->actor.params != ENZF_TYPE_DINOLFOS) || !EnZf_ChooseAction(globalCtx, this)) {
```
Dragorn421, line 1861:
```suggestion
            EnZf_SetupHopAndTaunt(this);
        } else {
            EnZf_SetupApproachPlayer(this, globalCtx);
        }
    } else if ((this->actor.params != ENZF_TYPE_DINOLFOS) || !EnZf_ChooseAction(globalCtx, this)) {
```
fits better the whole function not using early returns otherwise

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+            }
+            Actor_ChangeCategory(globalCtx, &globalCtx->actorCtx, &this->actor, ACTORCAT_PROP);
+        }
+
+        if (this->alpha != 0) {
+            this->actor.shape.shadowAlpha = this->alpha -= 5;
+
+        } else {
+            Actor_Kill(&this->actor);
+            return;
```
Dragorn421, line 1966:
```suggestion
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+                *dList = gZfEmptyHandDL;
+            }
+            break;
+        case ENZF_LIMB_SCABBARD:
+            if (this->swordSheathed) {
+                *dList = gZfSheathedSwordDL;
+            }
+            break;
+        default:
+            break;
```
Dragorn421, line 2151: needed?

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+                bodyPart = 6;
+                break;
+            case ENZF_LIMB_RIGHT_SHIN:
+                bodyPart = 7;
+                break;
+            case ENZF_LIMB_LEFT_SHIN_ROOT:
+                bodyPart = 8;
+                break;
+            default:
+                break;
```
Dragorn421, line 2233: needed?

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +151,2291 @@ static ColliderQuadInit D_80B4A204 = {
...
+            EnZf_SetupJumpUp(this);
+            return true;
+        }
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Zf/func_80B48CEC.s")
+        this->actor.world.rot.y = this->actor.shape.rot.y + 0x3FFF;
 
-#pragma GLOBAL_ASM("asm/non_matchings/overlays/actors/ovl_En_Zf/func_80B48E50.s")
+        if (phi_t0 == 0) {
+            phi_v1 = (globalCtx->gameplayFrames % 2);
```
Dragorn421, line 2342:
```suggestion
            phi_v1 = globalCtx->gameplayFrames % 2;
```

Dragorn421 (changes requested):
Good job

(nice EnWf copypaste (or the other way around, or whatever enemy) by the dev)

src/overlays/actors/ovl_En_Zf/z_en_zf.c
EllipticEllipsis: I think it's better to always have a default anyway, just in case.

src/overlays/actors/ovl_En_Zf/z_en_zf.c
EllipticEllipsis: Hm, hadn't thought of that. I don't know if this sort of thing happens elsewhere in the codebase.

src/overlays/actors/ovl_En_Zf/z_en_zf.c
EllipticEllipsis: This pattern is used so often that I'm thinking it would be better to have a define for `IS_MINIBOSS`, but I suspect people will think that's less clear? I'd argue the opposite, actually, which is why I've added all these comments.

Dragorn421: yeah I thought about it, I would be ok with it. but I also don't think we can reach consensus about using a define, let's wait for more opinions

src/overlays/actors/ovl_En_Zf/z_en_zf.c
EllipticEllipsis: I prefer early return after running the setup funcs: it makes it clear that there is no more execution happening in this function, which is rather less obvious with the elses IMO.

src/overlays/actors/ovl_En_Zf/z_en_zf.c
Dragorn421: Not that I know of. there also isn't a single use of the `0b` binary literal anywhere

src/overlays/actors/ovl_En_Zf/z_en_zf.c
EllipticEllipsis: This is normal in the codebase after an `Actor_Kill`, isn't it? There's one where I didn't put one, but it's right at the bottom of the function.

src/overlays/actors/ovl_En_Zf/z_en_zf.c
Dragorn421: I get that, but the pattern should be consistent at least in a single function, in this function most setup calls aren't followed with a return

src/overlays/actors/ovl_En_Zf/z_en_zf.c
Dragorn421: it's normal when the alternative is using an else, but here it's equivalent to have or not have the return;, I think in these cases it's better not to have it

src/overlays/actors/ovl_En_Zf/z_en_zf.c
EllipticEllipsis: Hm, I'll have another look. IIRC a lot of these early returns are needed because the audio bit at the bottom is separate.

src/overlays/actors/ovl_En_Zf/z_en_zf.c
EllipticEllipsis: Given that `-1` is used as a sort of "default" in this function, I think it's better to keep it (Slightly tempting to use a define, even...)

src/overlays/actors/ovl_En_Zf/z_en_zf.c
EllipticEllipsis: I don't think this will work since it's not a float calculation.

src/overlays/actors/ovl_En_Zf/z_en_zf.c
EllipticEllipsis: IDO may not even support this: https://stackoverflow.com/questions/18244726/why-doesnt-c-have-binary-literals

EllipticEllipsis: And it doesn't:
```
cfe: Error: src/overlays/actors/ovl_En_Zf/z_en_zf.c, line 2405: Syntax Error
                     case 0b11  :
 --------------------------^
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
EllipticEllipsis: It does match. I'll do that until Fig tells me to put it back 1UL.

src/overlays/actors/ovl_En_Zf/z_en_zf.c
fig02: yea to elaborate an early return makes sense in cases like
```
if (condition) {
    Actor_Kill
    return;
}
// Rest of the code
```
vs
```
if (condition) {
    Actor_Kill
} else {
    // Rest of the code
}
```
but here the return isnt doing anything, the kill is already in an else

src/overlays/actors/ovl_En_Zf/z_en_zf.c
fig02: I think I agree with an early return here. health being 0 and going to die is an exit condition to me, similar to how an actor kill is an abrubt end to the current execution and warrants it. I dont think it matters what the other actions are doing as far as consistency goes, to me its about what is happening in terms of the execution/chain of actions

if this was a longer if else chain checking the against params or something similar, i would agree. but the if here is a health check, and the else if is a type check. they arent really heavily related imo

fig02: wait, i was looking at the wrong thing. line numbers changed lol

fig02: deleted my previous post,  it makes sense to me here

Roman971 (commented):
Nice work on this. Just a few things

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +152,2281 @@ static ColliderQuadInit D_80B4A204 = {
...
+        if (this->actor.speedXZ < 0.0f) {
+            this->actor.speedXZ += 0.05f;
+        }
+        this->hopAnimIndex = 0;
+    }
+
+    if ((this->actor.colorFilterTimer == 0) && (this->actor.bgCheckFlags & 1)) {
+        if (this->actor.colChkInfo.health == 0) {
+            EnZf_SetupDie(this);
+        } else if ((this->actor.params != ENZF_TYPE_DINOLFOS) || !EnZf_ChooseAction(globalCtx, this)) {
```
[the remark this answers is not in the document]
EllipticEllipsis: Looking through again, I wonder if a lot of the `if (!EnZf_DodgeRanged(...))`s should actually be `if (EnZf_DodgeRanged(...)) { return; }`? It would certainly improve the indentation, and it happens in a lot of these actionFuncs

src/overlays/actors/ovl_En_Zf/z_en_zf.h
```diff
@@ -6,9 +6,119 @@
...
-    /* 0x014C */ char unk_14C[0x41C];
+    /* 0x014C */ SkelAnime skelAnime;
+    /* 0x0190 */ Vec3s jointTable[ENZF_LIMB_MAX];
+    /* 0x02B6 */ Vec3s morphTable[ENZF_LIMB_MAX];
+    /* 0x03DC */ s32 action;
+    /* 0x03E0 */ s32 unk_3E0; // Set but unused
+    /* 0x03E4 */ s32 hopAnimIndex;
+    /* 0x03E8 */ EnZfActionFunc actionFunc;
+    /* 0x03EC */ s16 headRot;
+    /* 0x03EE */ s16 headRotTemp; // Only used in one function
```
fig02, line 102: also not worth commenting on imo, unused is a special case worth mentioning but not anything else

EllipticEllipsis:
I did an experiment with a few things we could do to Zf in terms of the type and whether or not to use returns: https://github.com/EllipticEllipsis/oot/blob/438cb9ab64466ed1a885b89625bb9d7b36d10b0b/src/overlays/actors/ovl_En_Zf/z_en_zf.c Be interested to know what people think (diff with what's currently here is https://github.com/EllipticEllipsis/oot/compare/Zf...EllipticEllipsis:Zf_experiment?expand=1, but rather difficult to read) (made a new branch to avoid messing up this one).

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -82,115 +152,2272 @@ static ColliderQuadInit D_80B4A204 = {
...
+
+        this->actor.world.pos.y = this->actor.floorHeight + 300.0f;
+    } else if (this->alpha < 255) {
+        this->alpha += 51;
+    }
+
+    if ((this->actor.bgCheckFlags & 3) && (this->hopAnimIndex != 0)) {
+        Audio_PlayActorSound2(&this->actor, NA_SE_EN_RIZA_ONGND);
+        Animation_Change(&this->skelAnime, &gZfLandingAnim, 1.0f, 0.0f, 17.0f, ANIMMODE_ONCE, 0.0f);
+        this->hopAnimIndex = 0;
```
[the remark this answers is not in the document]
Dragorn421: Out of the 45 uses of `hopAnimIndex`, only 18, maybe 20, seem relevant to hopping, so I'm not sure we can really brush off those other uses

fig02 (approved):
feel like lots of the open ended conversations here are things we will figure out as we standardize actor documentation in general. so its fine on my end

Roman971 (approved):

fig02:
Some conflicts to fix but after that it's good to merge

PR #990, opened 2021-09-27
Use clang when possible for tools

louist103:
Makefile changes taken from ZAPD zeldaret/ZAPD#195


EllipticEllipsis:
You need to define `LLD ?= 0` or something as well, otherwise `make` will count it as not equal to 0 and will try to use `ld.lld`; that's why this failed.

Dragorn421 (approved):

Roman971:
So with this change, clang would become the default when available, and not just an option. Is there a clear advantage to using clang over gcc in our context? And what are the cons?

louist103:
Clang gives:
Faster compile and link times,
Smaller binaries,
And slightly faster performance on the tools.
Clang also uses less Ram than GCC does on average which is more of an issue for WSL since it has a lot of memory isses.

Roman971 (approved):
I'll be honest I find it hard to believe that clang is just better than gcc in all regards and there is no catch, but I guess we can revert it if it ever starts causing issues

PR #992, opened 2021-10-01
Speed up `extract_assets.py`

Dragorn421:
## Why `extract_assets.py` is slow

Currently in `extract_assets.py` the last-extracted-time of assets is stored in a file `.extracted-assets.json` to avoid extracting assets again when xmls didn't change
**Reading takes around 1.5 seconds, writing around 1 second.**
An empty run (when no assets are extracted, all of them are up-to-date) takes 3.5 seconds (and extracting a few assets doesn't take much longer)

So basically when few assets are extracted `.extracted-assets.json`-related IO takes 70% of the script execution time

My first instinct was to replace json by something more efficient like pickle (I did it [here](https://github.com/Dragorn421/oot/blob/fix_extract_assets_slow/extract_assets.py)), but it turns out **the true bottleneck is using `multiprocessing.Manager.dict`**. (I'm not familiar with `multiprocessing` I assume it's needed for synchronization or something like that)

## A fix

So I got rid of 99.8% of the `multiprocessing.Manager.dict` calls by changing the json structure from
```json
{
    "assets/xml/objects/object_gi_soldout.xml": {
        "timestamp": 1632918073
    },
```
to
```json
{
    "assets/xml/objects/object_gi_soldout.xml": 1632918073,
```

and **now reading takes around 10 milliseconds, writing around 100 milliseconds.** So, consequent speedup

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,7 +47,7 @@ graphs/
 *.mtl
 *.fbx
 !*_custom*
-.extracted-assets.json
+.extracted-assets-v2.json
 
 # Per-user configuration
 .python-version
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ graphs/
 *.mtl
 *.fbx
 !*_custom*
+.extracted-assets.json
 .extracted-assets-v2.json
 
 # Per-user configuration
--- a/extract_assets.py
+++ b/extract_assets.py
@@ -76,15 +76,10 @@ def main():
     manager = Manager()
     signal.signal(signal.SIGINT, SignalHandler)
 
-    startTimePerf = time.time()
-
     extractedAssetsTracker = manager.dict()
     if os.path.exists(EXTRACTED_ASSETS_NAMEFILE) and not args.force:
         with open(EXTRACTED_ASSETS_NAMEFILE, encoding='utf-8') as f:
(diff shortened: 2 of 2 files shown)


AngheloAlf:
Yeah, `multiprocessing.Manager` is required to have shared memory among threads. It is slow, and even slowerer because it uses a dict of dicts.
The rationale to use it was to allow the format to be future proof and easily scalable in the case we wanted to add more metadata to each extracted XML (besides timestamp).
Right now I can't really think of any other metadata we may want to track right now, so I think this change may be OK.

fig02 (approved):
i have no personal opinion on the thoughts above, but think its prob something we should consider. this looks good to me either way

Dragorn421:
What I thought would be the simplest way to edit the script for speed turned out to not be consensual at all.

I agree with Roman's comment on having versioning in the json, but writing more code conflicts with the initial intent of being a quick fix.

So I'm going to close this and maybe revisit it at a later time.

What can be done I think is to keep the json format as is, pass the (non-synchronized) sub-dicts into each worker, make those worker return a dict and put that dict into the main dict. Synchronized dicts for each worker doesn't make sense since only one worker uses the dict. (may be a poor explanation but good enough notes for me)

PR #995, opened 2021-10-03
Use void* instead of u32 for segmented addresses in debug display

louist103:


src/code/z_debug_display.c
```diff
@@ -88,7 +88,7 @@ void DebugDisplay_DrawSpriteI8(DebugDispObject* dispObj, u32 texture, GlobalCont
     CLOSE_DISPS(globalCtx->state.gfxCtx, "../z_debug_display.c", 192);
 }
 
-void DebugDisplay_DrawPolygon(DebugDispObject* dispObj, u32 dlist, GlobalContext* globalCtx) {
+void DebugDisplay_DrawPolygon(DebugDispObject* dispObj, void* dlist, GlobalContext* globalCtx) {
```
Dragorn421, line 91: So `Gfx*` would be better but I assume we prefer `void*` to match the `DebugDispObject_DrawFunc` function pointer type, to avoid casting the function in `sDebugObjectDrawFuncTable`

src/code/z_debug_display.c
```diff
@@ -2,14 +2,14 @@
 #include "objects/gameplay_keep/gameplay_keep.h"
 
 typedef struct {
-    /* 0x00 */ s16 drawType; // indicates which draw function to use when displaying the object
-    /* 0x04 */ u32 drawArg;  // segment address (display list or texture) passed to the draw funciton when called
-} DebugDispObjectInfo;       // size = 0x8
+    /* 0x00 */ s16 drawType;  // indicates which draw function to use when displaying the object
+    /* 0x04 */ void* drawArg; // segment address (display list or texture) passed to the draw funciton when called
```
Dragorn421, line 6:
```suggestion
    /* 0x04 */ void* drawArg; // segment address (display list or texture) passed to the draw function when called
```
While you're at it

src/code/z_debug_display.c
louist103: Yeah. Unless there is a way to have both

Dragorn421 (approved):

fig02 (approved):

PR #998, opened 2021-10-10
Fix a few warnings

Dragorn421:
This fixes a few warnings.

I have a branch which builds with `-Werror` and many `-Wno-error` exceptions [Dragorn421/oot/build_werror](https://github.com/Dragorn421/oot/tree/build_werror) (see the [Makefile changes](https://github.com/Dragorn421/oot/commit/7a449b0190b1c60e69db73f375929509a163152c))

Warnings that can't be fixed include:
- `#include "..." EARLY`: `EARLY` throws off gcc
- fake matches
- conditionals that are always true or always false due to type range limits (eg checking if a s16 angle is `< 0x8000` is always true) (it may be worth documenting these in code, maybe even `@bug` in some places, not sure)


include/variables.h
```diff
@@ -2817,10 +2817,10 @@ extern s16 D_80153960[];
...
 //extern ? D_80154AFC;
 //extern ? D_80154B00;
-extern u8 gAudioBankTable[];
-extern u8 D_80155340[];
-extern u8 gSequenceTable[];
-extern u8 gAudioTable[];
+extern AudioBankTable gAudioBankTable;
+extern u16 D_80155340[];
+extern SequenceTable gSequenceTable;
+extern AudioTable gAudioTable;
```
zelda2774, line 2823: u8[] is more correct here: it's a blob of bytes that later gets interpreted as various audio tables. A cast at the usage site is a better fix. AudioBankTable etc. are dubious because they can't be initialized due to the trailing `[1]` mock flexible array member

src/code/audio_load.c
```diff
@@ -1099,15 +1099,15 @@ void Audio_ContextInit(void* heap, u32 heapSize) {
...
 
     gAudioContext.audioResetSpecIdToLoad = 0;
     gAudioContext.resetStatus = 1;
 
     Audio_ResetStep();
-    Audio_InitAudioTable(gAudioContext.sequenceTable, _AudioseqSegmentRomStart, 0);
-    Audio_InitAudioTable(gAudioContext.audioBankTable, _AudiobankSegmentRomStart, 0);
+    Audio_InitAudioTable((AudioTable*)gAudioContext.sequenceTable, _AudioseqSegmentRomStart, 0);
+    Audio_InitAudioTable((AudioTable*)gAudioContext.audioBankTable, _AudiobankSegmentRomStart, 0);
     Audio_InitAudioTable(gAudioContext.audioTable, _AudiotableSegmentRomStart, 0);
```
zelda2774, line 1111: the proper fix for this (that doesn't violate strict aliasing) is to make all of these the same type; I do that in my more-audio branch which hasn't made it to a PR yet. Totally fine to land this as temporary warning fix though

PR #999, opened 2021-10-10
Create macros for and decompile elf_message scripts

Thar0:


louist103 (approved):

src/code/z_elf_message.c
```diff
@@ -1,46 +1,60 @@
...
+/* 000C   3 */ ELF_MESSAGE_FLAG(CHECK, 0x0162, false, 0x25), /* eventChkInf[2] & 0x20 */
+/* 0010   4 */ ELF_MESSAGE_FLAG(CHECK, 0x0163, false, 0x37), /* eventChkInf[3] & 0x80 */
+/* 0014   5 */ ELF_MESSAGE_FLAG(CHECK, 0x0165, false, 0x43), /* eventChkInf[4] & 0x8 */
+/* 0018   6 */ ELF_MESSAGE_MEDALLION(CHECK, 0x0166, false, ITEM_MEDALLION_FOREST),
+/* 001C   7 */ ELF_MESSAGE_MEDALLION(CHECK, 0x0166, false, ITEM_MEDALLION_FIRE),
+/* 0020   8 */ ELF_MESSAGE_MEDALLION(CHECK, 0x0166, false, ITEM_MEDALLION_WATER),
+/* 0024   9 */ ELF_MESSAGE_SONG(CHECK, 0x0167, false, ITEM_SONG_STORMS),
+/* 0028  10 */ ELF_MESSAGE_MEDALLION(CHECK, 0x0168, false, ITEM_MEDALLION_SPIRIT),
+/* 002C  11 */ ELF_MESSAGE_MEDALLION(CHECK, 0x0168, false, ITEM_MEDALLION_SHADOW),
+/* 0030  12 */ ELF_MESSAGE_END(0x0169),
```
Roman971, line 18: I feel like the comments for each entry are unnecessary and could be omitted to keep things clean and easy to work with in modding purposes

src/elf_message/elf_message_field.c
```diff
@@ -0,0 +1,33 @@
...
+#include "z64elf_message.h"
+
+ElfMessage gOverworldNaviMsgs[] = {
+/* 0000   0 */ ELF_MESSAGE_FLAG(CHECK, 0x0140, false, 0x05), /* eventChkInf[0] & 0x20 */
+/* 0004   1 */ ELF_MESSAGE_FLAG(CHECK, 0x0141, false, 0x09), /* eventChkInf[0] & 0x200 */
+/* 0008   2 */ ELF_MESSAGE_FLAG(CHECK, 0x0142, false, 0x12), /* eventChkInf[1] & 0x4 */
+/* 000C   3 */ ELF_MESSAGE_FLAG(CHECK, 0x0143, false, 0x14), /* eventChkInf[1] & 0x10 */
+/* 0010   4 */ ELF_MESSAGE_FLAG(CHECK, 0x0144, false, 0x40), /* eventChkInf[4] & 0x1 */
+/* 0014   5 */ ELF_MESSAGE_SONG(CHECK, 0x0145, false, ITEM_SONG_SARIA),
+/* 0018   6 */ ELF_MESSAGE_STRENGTH_UPG(CHECK, 0x0146, true, 0), // TODO goron bracelet
```
Roman971, line 11: Since we don't know yet if we're even going to have an enum or defines for upgrade values, I personally think we should omit the TODOs here for now

src/elf_message/elf_message_field.c
```diff
@@ -0,0 +1,33 @@
+#include "global.h"
+#include "z64elf_message.h"
+
+ElfMessage gOverworldNaviMsgs[] = {
+/* 0000   0 */ ELF_MESSAGE_FLAG(CHECK, 0x0140, false, 0x05), /* eventChkInf[0] & 0x20 */
```
Roman971, line 5: Just a suggestion to keep things more concise, but we could shorten all these macros to `ELF_MSG_`

src/code/z_elf_message.c
```diff
@@ -1,46 +1,60 @@
...
-// TODO: use macros to define elf messages once the format is fully documented
 ElfMessage sChildSariaMsgs[] = {
-    { 0x66, 0x00, 0x03, 0x00 }, { 0x00, 0x37, 0x61, 0x00 }, { 0xE0, 0x00, 0x64, 0x00 }, { 0x00, 0x25, 0x62, 0x00 },
-    { 0x00, 0x37, 0x63, 0x00 }, { 0x00, 0x43, 0x65, 0x00 }, { 0x06, 0x30, 0x66, 0x66 }, { 0x06, 0x30, 0x66, 0x67 },
-    { 0x06, 0x30, 0x66, 0x68 }, { 0x06, 0x20, 0x67, 0x65 }, { 0x06, 0x30, 0x68, 0x69 }, { 0x06, 0x30, 0x68, 0x6A },
-    { 0xE0, 0x00, 0x69, 0x00 },
+/* 0000   0 */ ELF_MESSAGE_STRENGTH_UPG(BRANCHING, 3, false, 0), // TODO goron bracelet
+/* 0004   1 */ ELF_MESSAGE_FLAG(CHECK, 0x0161, false, 0x37), /* eventChkInf[3] & 0x80 */
+/* 0008   2 */ ELF_MESSAGE_END(0x0164),
+
```
Roman971, line 8: This empty line is probably not necessary either

spec
```diff
@@ -6807,13 +6807,15 @@ endseg
 beginseg
     name "elf_message_field"
     romalign 0x1000
-    include "build/baserom/elf_message_field.o"
+    include "build/src/elf_message/elf_message_field.o"
+    number 0
```
Roman971, line 6811: Do we need to define these segments as having number 0? I thought I remembered these were never accessed in memory, in which case they probably don't need a number at all?

src/code/z_elf_message.c
```diff
@@ -1,46 +1,60 @@
...
+#include "z64elf_message.h"
 
-// TODO: use macros to define elf messages once the format is fully documented
 ElfMessage sChildSariaMsgs[] = {
-    { 0x66, 0x00, 0x03, 0x00 }, { 0x00, 0x37, 0x61, 0x00 }, { 0xE0, 0x00, 0x64, 0x00 }, { 0x00, 0x25, 0x62, 0x00 },
-    { 0x00, 0x37, 0x63, 0x00 }, { 0x00, 0x43, 0x65, 0x00 }, { 0x06, 0x30, 0x66, 0x66 }, { 0x06, 0x30, 0x66, 0x67 },
-    { 0x06, 0x30, 0x66, 0x68 }, { 0x06, 0x20, 0x67, 0x65 }, { 0x06, 0x30, 0x68, 0x69 }, { 0x06, 0x30, 0x68, 0x6A },
-    { 0xE0, 0x00, 0x69, 0x00 },
+/* 0000   0 */ ELF_MESSAGE_STRENGTH_UPG(BRANCHING, 3, false, 0), // TODO goron bracelet
+/* 0004   1 */ ELF_MESSAGE_FLAG(CHECK, 0x0161, false, 0x37), /* eventChkInf[3] & 0x80 */
```
Roman971, line 6: Having the full text ID in these commands seems very misleading since only the lowest byte will actually be used, especially in a modding scenario where people will probably tend to add entries in the 0x09XX which wouldn't work here.

So I think we should only have the lowest byte for every text ID (eg. 0x61 here)

Roman971 (commented):
Sorry for the delayed review. I don't have any major comments on this atm, just a few suggestions on the command layout and spec segments

include/z64elf_message.h
```diff
@@ -0,0 +1,129 @@
...
+#include "ultra64.h"
+
+// Checks the condition and exits the script if the check passes
+#define ELF_MESSAGE_TYPE_CHECK      0
+// ? (unused)
+#define ELF_MESSAGE_TYPE_UNK_1      1
+// ? (unused)
+#define ELF_MESSAGE_TYPE_UNK_2      2
+// Checks the condition and branches to another command if the check passes
+#define ELF_MESSAGE_TYPE_BRANCHING  3
```
Roman971, line 13: Also this is more of a general point, but I don't know if "branching" is the right term since afaict it only skips a certain number of commands as opposed to going to a specific one (see the relevant case in `ElfMessage_GetTextFromMsgs`)

src/code/z_elf_message.c
mzxrules: well, are we leaving Text IDs as raw numbers?

spec
Thar0: We don't need to, however I believe it is correct to do so, otherwise they'll be put at some weird address based on whatever was set prior in the spec, which imo is strictly worse than just declaring a number, or even an address, even if the address isn't currently relevant to anything in the codebase.

Roman971: I can see that yeah. In that case it's probably fine, tho I'm worried about the noise it would add if we applied it everywhere, and how it would look like with defines for all numbers, but we can probably figure it out later

PR #1004, opened 2021-10-17
`object_human` OK

AngheloAlf:
Named almost every asset of `object_human`.
- There are a few textures that I couldn't properly name because of the lack of context (like the multiples mouth textures), or having no idea what those are supposed to be (like some textures or most limbs).

I also decided to split it into multiples files guided by the files boundaries found in the object during extraction (the typical small array of zeroes).
- This is the reason why the spec looks so weird

At some point in time (probably when I did the automatic objects PR) I messed up a PR and ended up including the half done  `object_human` I had at that time, that's why it mostly done in master.

Random stuff:
- For some reason, this object has _many_ eyes, for looking in different directions, and each direction has a blinking texture too.
- Each eye texture is repeated twice.
- This object tried very hard to have multiples facial expressions (see the multiples, eyes, eyebrows and mouths textures), which is weird considering the low resolution the N64 has.
- Every dlist has a wrapper, and that wrapper has a second wrapper. Every limb that uses a dlist references the second wrapper.
- This object has a lot of limbs (41 to be precise), and most of them are just small displacement/rotation points (joints?).
- This object has 12 animations. Some of them are some kind of "starting X animation", "looping X animation" and "ending X animation"


assets/xml/objects/object_human.xml
```diff
@@ -19,10 +22,13 @@
         <Texture Name="gHumanEyeLookForwardOpen2Tex" OutName="eye_look_forward_open_2" Format="ci8" Width="32" Height="32" TlutOffset="0x7A00" Offset="0x2A00"/>
         <Texture Name="gHumanEyeLookForwardHalf2Tex" OutName="eye_look_forward_half_2" Format="ci8" Width="32" Height="32" TlutOffset="0x7A00" Offset="0x2E00"/>
         <Texture Name="gHumanEyeLookForwardClosed2Tex" OutName="eye_look_forward_closed_2" Format="ci8" Width="32" Height="32" TlutOffset="0x7A00" Offset="0x3200"/>
+
         <Texture Name="gHumanEyeLookRightOpen2Tex" OutName="eye_look_right_open_2" Format="ci8" Width="32" Height="32" TlutOffset="0x7A00" Offset="0x3600"/>
         <Texture Name="gHumanEyeLookRightHalf2Tex" OutName="eye_look_right_half_2" Format="ci8" Width="32" Height="32" TlutOffset="0x7A00" Offset="0x3A00"/>
+
         <Texture Name="gHumanEyeLookLeftOpen2Tex" OutName="eye_look_left_open_2" Format="ci8" Width="32" Height="32" TlutOffset="0x7A00" Offset="0x3E00"/>
         <Texture Name="gHumanEyeLookLeftHalf2Tex" OutName="eye_look_left_half_2" Format="ci8" Width="32" Height="32" TlutOffset="0x7A00" Offset="0x4200"/>
```
Dragorn421, line 30: these seem more like look_right ?

assets/xml/objects/object_human.xml
```diff
@@ -19,10 +22,13 @@
         <Texture Name="gHumanEyeLookForwardOpen2Tex" OutName="eye_look_forward_open_2" Format="ci8" Width="32" Height="32" TlutOffset="0x7A00" Offset="0x2A00"/>
         <Texture Name="gHumanEyeLookForwardHalf2Tex" OutName="eye_look_forward_half_2" Format="ci8" Width="32" Height="32" TlutOffset="0x7A00" Offset="0x2E00"/>
         <Texture Name="gHumanEyeLookForwardClosed2Tex" OutName="eye_look_forward_closed_2" Format="ci8" Width="32" Height="32" TlutOffset="0x7A00" Offset="0x3200"/>
+
         <Texture Name="gHumanEyeLookRightOpen2Tex" OutName="eye_look_right_open_2" Format="ci8" Width="32" Height="32" TlutOffset="0x7A00" Offset="0x3600"/>
         <Texture Name="gHumanEyeLookRightHalf2Tex" OutName="eye_look_right_half_2" Format="ci8" Width="32" Height="32" TlutOffset="0x7A00" Offset="0x3A00"/>
```
Dragorn421, line 27: these look more like look_left ?

Dragorn421 (commented):
How are you looking at this one?

AngheloAlf:
Thanks, I messed up those :sweat_smile:

assets/xml/objects/object_human.xml
```diff
@@ -1,81 +1,104 @@
...
+
         <Texture Name="gHumanEyeLookUpOpenTex" OutName="eye_look_up_open" Format="ci8" Width="32" Height="32" Offset="0x1C00" TlutOffset="0x7A00"/>
         <Texture Name="gHumanEyeLookDownOpenTex" OutName="eye_look_down_open" Format="ci8" Width="32" Height="32" Offset="0x2000" TlutOffset="0x7A00"/>
 
         <Texture Name="gHumanEyebrowNormalTex" OutName="eyebrow_normal" Format="ci8" Width="32" Height="16" Offset="0x2400" TlutOffset="0x7A00"/>
         <Texture Name="gHumanEyebrowSurprisedTex" OutName="eyebrow_surprised" Format="ci8" Width="32" Height="16" Offset="0x2600" TlutOffset="0x7A00"/>
         <Texture Name="gHumanEyebrowSadTex" OutName="eyebrow_sad" Format="ci8" Width="32" Height="16" Offset="0x2800" TlutOffset="0x7A00"/>
 
-        <!-- Those are exaclty the same are above eye textures -->
+        <!-- Those are exaclty the same as above eye textures -->
```
Dragorn421, line 21: minor typo I just noticed passing by
```suggestion
        <!-- Those are exactly the same as above eye textures -->
```
also on at least one other line I saw

louist103 (approved):

assets/xml/objects/object_human.xml
```diff
@@ -1,81 +1,104 @@
...
+        <Texture Name="gHumanMouthOpen4Tex" OutName="mouth_open_4" Format="ci8" Width="32" Height="32" Offset="0x6A00" TlutOffset="0x7A00"/>
+        <Texture Name="gHumanMouthClosed3Tex" OutName="mouth_closed_3" Format="ci8" Width="32" Height="32" Offset="0x6E00" TlutOffset="0x7A00"/>
+        <Texture Name="gHumanMouthOpen5Tex" OutName="mouth_open_5" Format="ci8" Width="32" Height="32" Offset="0x7200" TlutOffset="0x7A00"/>
+        <Texture Name="gHumanMouthOpen6Tex" OutName="mouth_open_6" Format="ci8" Width="32" Height="32" Offset="0x7600" TlutOffset="0x7A00"/>
+
+        <!-- TLUT -->
         <Texture Name="gHumanTLUT" OutName="face_palette" Format="rgba16" Width="16" Height="16" Offset="0x7A00"/>
+    </File>
 
+    <File Name="object_human" OutName="anim_looking_back" RangeStart="0x7A40" RangeEnd="0x8360" Segment="6">
```
Roman971, line 58: Considering the amount of file splits, I don't think it's very viable to use the "File" node with the full information every time (ie. having to specific the object name and the segment, as well as a full range). To me it also kinda 'looks like' an abuse of the way ZAPD currently works. I think it might make more sense to have nodes for sub files or something along those lines.
That being said, it could also wait for a separate ZAPD update + PR to improve things, since it's more of a concern about the way we handle splits like this in the future.

Roman971 (commented):
Sorry for the late review. I have been meaning to comment on this, but didn't get around to it for a long time. The general changes look good to me, but I have some concerns with the way file splits are handled.

assets/xml/objects/object_human.xml
AngheloAlf: In that case, I think this PR will have to wait until somebody implements that in ZAPD.

Roman971: To be clear, I meant we could also merge this PR as is if we don't want to wait for a ZAPD update, but I think we should make some improvements or at least discuss things before doing splits like this in other objects

fig02: thats fair. can we detail out more specifically what we think should change about how multi file objects are handled, possibly in a zapd issue? 
I believe mm noticed an issue about it with the way it handles headers/includes that we'll need to look at as well. mm people correct me if im wrong tho

fig02: and in that case, we should prob merge this as is, with the notion that we will not continue with multi-file objects until we settle things zap-side

AngheloAlf: In that case I think I could un-split this object and split it again in the future when stuff changes in the ZAPD side

fig02: thats an option too. I think there is some value in having the example to see what we want to fix about it. but i guess it doesnt have to be in the repo for it to serve that purpose

tom-overton: RE: the header stuff, I also split an object in MM here: https://github.com/zeldaret/mm/pull/569
Each file that gets split in the object creates a different header file, which results in a lot more includes in the C file, like so:
```c
#include "objects/object_obj_tokeidai/termina_field_walls.h"
#include "objects/object_obj_tokeidai/counterweight.h"
#include "objects/object_obj_tokeidai/exterior_gear.h"
#include "objects/object_obj_tokeidai/clock_center_and_hand.h"
#include "objects/object_obj_tokeidai/sun_and_moon_panel.h"
#include "objects/object_obj_tokeidai/minute_ring.h"
#include "objects/object_obj_tokeidai/unused_tower_wall.h"
#include "objects/object_obj_tokeidai/staircase_to_rooftop.h"
#include "objects/object_obj_tokeidai/clock_tower_clock_face.h"
#include "objects/object_obj_tokeidai/wall_clock_clock_face.h"
```
The concern is that for objects that contain a lot of animations (like Majora), there will be dozens upon dozens of includes. This is probably something ZAPD could fix by just having a single header that contains all assets (like it does for unsplit objects), but obviously that's something that isn't ready at the moment. So I'd agree with Anghelo that it probably makes more sense to un-split these objects for now and split them later when ZAPD supports everything we want it to.

fig02: sure. we should compile all of our concerns about this into a zapd issue so we know exactly what we want to change

Roman971: That's a good point about the separate headers. Ideally we should still use a single header (at least in most cases), so we either need a different solution on the ZAPD side, or we need to wait until we push headers and such in the repo at which point we'll have control over the headers and their structure.

Btw I'm still not sure splitting things in spec files first is the right approach. I feel like we should start by splitting individual assets first so we can push main C file(s) to the repo, and only group things in sub files and/or directories if it becomes necessary in complex objects, since that last part only helps with the organization and isn't needed in most objects.

AngheloAlf: The idea of splitting object files in the spec comes from the obvious file splits each object have in them, which I though was an already accepted idea and how we would end up working with this (That's one of the reasons of this PR, to show how it currently works and to discuss it).

I'm not really sure what you mean by "splitting individual assets" tbh

Roman971: Following original sub file splits can make sense, but only if it's worth it and we don't lose anything while doing it, otherwise we could just choose not to do it (possibly on a case-by-case basis). I find it pretty unnecessary for the most part tbh, so that's why I personally don't feel it should be prioritized, even tho it's probably fine in most cases.

On the other hand, what we really need imo is to split individual assets to their own files, and by that I mean having a dedicated C file for every texture, display list, vertex array, cutscene, etc... Basically it would mean having ZAPD generate a C file for each "Texture"/"DList"/etc node, which would contain only that asset. The idea is to make it so the "main" C file(s) would only have includes to all these assets, which means they no longer have to be generated by ZAPD during the setup and they can be pushed to the repo along with the headers. Among other things, this could allow assets to be added, moved or removed from existing asset files when modding, in a way that's compatible with git, and that's the biggest issue for modding afaik. It also wouldn't be exclusive with splitting files in the spec, to be clear, since that could just mean having multiple "main" files.
Now, to be fair, this would probably need a lot of changes both in ZAPD and in the repo, but it's very important imo, and I'd be willing to help working on it if needed (in the repo itself but I could also try to help with ZAPD stuff)

(Sorry for the long answer but I tried to be as complete as possible so we're on the same page 😅)

AngheloAlf: Personally I think we should follow the original subfile splits when possible for assets, but I can understand not prioritizing it or not wanting to do it, specially with the current ZAPD implementation ot it and the points people has mentioned (we should write those in a ZAPD issue so those don't get lost in the discussion).

On the other hand, yeah I think ZAPD extracting each asset to a file and the main C file including those should be prioritized. To be honest I though this was one of the rejected ideas for how to commit those C files to the repo (I'm not really sure why I though that, I may be misremembering). I personally think this should be made entirely by ZAPD, the same way it is already extracting textures and vtxs.

fig02: its been the plan for a while, so it might have been a misunderstanding.
what we struggled to agree on is what should be extracted to its own file. some felt dlists dont need to be in their own file and should be included in the repo

at this point im fine with either solution, as long as its modular and easy for modding

AngheloAlf: Just to be sure we are on the same page, I meant to extract _every_ asset to its own file, and the main C file is just a skeleton that includes everything else (and which will be pushed to the repo). 
This way, ZAPD can just extract every asset during extraction. ZAPD could be used to generate the C skeleton file once and then we can manually take care and discuss for each asset type if it should be included in the main C skeleton file or not.

fig02: yes exactly, that is what we have wanted to do for a while. The disagreements have always been about what an asset is, and what needs to be in its own file. The argument being that not everything in an object file is an asset that needs to be separate.

AngheloAlf: Ohh nice. If that's the case I could take a look into implementing it in the ZAPD side

AngheloAlf:
un-splitted object as requested

Roman971 (approved):

PR #1007, opened 2021-10-19
Misc Cleanup 2

Dragorn421:
I'm adding stuff to this as I see it, so i'll keep it as draft until it grows too large


assets/xml/textures/place_title_cards.xml
```diff
@@ -1,5 +1,5 @@
 <Root>
-    <!-- These dont actually go into Segment 0 but its needed to prevent an error from ZAPD -->
+    <!-- These dont actually go into Segment 0 but it's needed to prevent an error from ZAPD -->
```
AngheloAlf, line 2: I think this comment can be removed now

Roman971:
Are there more things planned for this PR? I briefly looked over and the changes so far seem good to me at least

fig02:
agreed with roman here, do you think this is ready to come off of being a draft? @Dragorn421

Dragorn421:
It can be ready-for-review as is sure, but I plan to do more, for example I want to proofread every header, and I have a list of minor things to improve

fig02:
we can do those in separate PRs then. Very small prs that accomplish only a couple tasks have been moving alot quicker through the queue these days and it would be good to keep that up

Roman971 (approved):

fig02 (approved):

PR #1008, opened 2021-10-20
Zapd hotfix

louist103:
zeldaret/ZAPD#204
Also fixed an issue where there was a warning that was only picked up by clang-14


louist103:
I made a mess trying to fix this.

PR #1018, opened 2021-11-04
Add symbol and file for softsprite_matrix_static

louist103:


include/variables.h
```diff
@@ -4,7 +4,7 @@
 #include "z64.h"
 #include "segment_symbols.h"
 
-extern Mtx D_01000000;
+extern Mtx gSoftSpriteMartixStatic;
```
Dragorn421, line 7:
```suggestion
extern Mtx gSoftSpriteMatrixStatic;
```
you had one job Sadge

src/code/softsprite_matrix_static.c
```diff
@@ -0,0 +1,10 @@
+#include "ultra64/types.h"
+
+//clang-format off
+Mtx gSoftSpriteMartixStatic = { 
+    65536 , 0     , 1     , 0     , 
+    0     , 65536 , 0     , 1     , 
+    0     , 0     , 0     , 0     , 
+    0     , 0     , 0     , 0     , 
+};
+//clang-format on
```
Dragorn421, line 10: what does it look like without clang-format off/on?

(it doesn't make more sense to format it as a 4x4 grid, the first 8 values are the integer parts, the last 8 the fractional parts)

spec
```diff
@@ -10466,5 +10466,6 @@ endseg
 beginseg
     name "softsprite_matrix_static"
     romalign 0x1000
-    include "build/baserom/softsprite_matrix_static.o"
+    include "build/src/code/softsprite_matrix_static.o"
```
Dragorn421, line 10469: probably shouldn't be code but no clue where tbh
there are more of these so maybe a folder `src/segments/` ? no clue

Dragorn421 (changes requested):

tools/ZAPD/ExporterTest/Jenkinsfile
```diff
@@ -0,0 +1,97 @@
+pipeline {
+    agent {
+        label 'ZAPD'
+    }
+
+    stages {
+        // Non-parallel ZAPD stage
+        stage('Build ZAPD') {
```
AngheloAlf, line 8: Looks like you messed up something here

spec
AngheloAlf: Yeah, this definitely shouldn't be in `code/`. 
I like the  `segments/` idea

PR #1019, opened 2021-11-07
Doc dodongo's cavern stuff

Dragorn421:
Themed PR around Dodongo's Cavern actors and related code.

I can't wait to bomb some dodongos


Dragorn421:
I'll actually bomb some dodongos in another PR to keep with the short PRs trend

src/overlays/actors/ovl_Bg_Dodoago/z_bg_dodoago.c
```diff
@@ -17,10 +17,10 @@ void BgDodoago_Destroy(Actor* thisx, GlobalContext* globalCtx);
 void BgDodoago_Update(Actor* thisx, GlobalContext* globalCtx);
 void BgDodoago_Draw(Actor* thisx, GlobalContext* globalCtx);
 
-void func_80871CF4(BgDodoago* this, GlobalContext* globalCtx);
-void func_80871FB8(BgDodoago* this, GlobalContext* globalCtx);
+void BgDodoago_WaitExplosives_(BgDodoago* this, GlobalContext* globalCtx);
```
fig02, line 20: accidental trailing underscore

Dragorn421: that's how i indicate a not-final name when I work on docs, but yeah accidentally left it in :+1:

include/z64.h
```diff
@@ -931,7 +931,7 @@ typedef struct {
     /* 0x38 */ DmaRequest dmaRequest;
     /* 0x58 */ OSMesgQueue loadQueue;
     /* 0x70 */ OSMesg loadMsg;
-    /* 0x74 */ s16 unk_74[2];
+    /* 0x74 */ s16 unk_74[2]; // context-specific data used by the current scene draw config
```
fig02, line 934: this is pretty interesting, had no idea this was a thing. would a general name like `data` `roomData` `params` or something work?
if you dont want to name this now thats fine

fig02 (approved):

include/z64.h
Dragorn421: The purpose is clear (imo) but the use cases not so much, so I think we should wait for more documentation on stuff using it (especially scene draw configs)

src/code/z_actor.c
```diff
@@ -3155,20 +3155,20 @@ void Actor_SpawnFloorDustRing(GlobalContext* globalCtx, Actor* actor, Vec3f* pos
     }
 }
 
-void func_80033480(GlobalContext* globalCtx, Vec3f* arg1, f32 arg2, s32 arg3, s16 arg4, s16 scaleStep, u8 arg6) {
+void func_80033480(GlobalContext* globalCtx, Vec3f* posBase, f32 randRangeDiameter, s32 amountMinusOne, s16 scaleBase, s16 scaleStep, u8 arg6) {
```
Roman971, line 3158: Looks like this file needs a format run. Same for functions.h probably

src/code/z_scene_table.c
```diff
@@ -1025,11 +1027,11 @@ void* sDCLavaFloorTextures[] = {
     gDCLavaFloor5Tex, gDCLavaFloor6Tex, gDCLavaFloor7Tex, gDCLavaFloor8Tex,
 };
 
-// Scene Draw Config 20
+// Scene Draw Config 20 - Dodongo's Cavern
 void func_80099878(GlobalContext* globalCtx) {
     u32 gameplayFrames;
     s32 pad;
-    Gfx* displayListHead = Graph_Alloc(globalCtx->state.gfxCtx, 6 * sizeof(Gfx));
+    Gfx* dlistBuffer = Graph_Alloc(globalCtx->state.gfxCtx, 2 * sizeof(Gfx[3]));
```
Roman971, line 1034: I think `displayListHead` (or `dListHead`) is better here since it's used as a pointer to the head and incremented over the function, whereas "buffer" makes it sound like it's a pointer to a buffer which isn't entirely true

Dragorn421: unlike in other places this holds two display lists, not just one, so I thought "buffer" made more sense

Roman971: It can be seen like that, but I'd argue the variable still holds a pointer to the head of either display lists so a name like `dListHead` makes more sense either way. The alternative would probably be `gfx` if you want something more generic, but it seems unnecessary to me.

Roman971 (approved):

PR #1024, opened 2021-11-23
Add pad_text to mkldscript and update progress.py

Roman971:
Implements `pad_text` from MM and uses it instead of asm nops for the padding in ucode_disas.
Also updates `progress.py` to account for the change (RIP 1000000 code size), and to have a green badge for 100%.


progress.py
```diff
@@ -95,7 +95,7 @@ def GetNonMatchingSize(path):
 ovl -= nonMatchingASMOvl
 
 bootSize = 31408 # decompilable code only
-codeSize = 1000000 # decompilable code only (1.00mb)
+codeSize = 999984 # decompilable code only (1.00mb)
```
AngheloAlf, line 98: I think the 1.00mb comment should go
```suggestion
codeSize = 999984 # decompilable code only
```

louist103 (approved):

AngheloAlf (approved):

PR #1027, opened 2021-11-25
Some small Makefile changes

louist103:
This flag is not needed since we use our own reloc system already.
Removing the optimization flags save between 1-3 seconds when compiling assets.


Thar0:
I'm not sure I like the removal of `--emit-relocs`, the relocs in the final elf would be useful for debuggers to have access to so if their inclusion barely impacts build time I don't think it's worth removing them if there's no tangible advantage.
Of course, it's always possible to be re-added for use with a debugger, I just think if there's no real negative to keeping it that we might as well.

fig02:
agreed, dont think theres an issue with leaving it

Makefile
```diff
@@ -226,8 +226,7 @@ build/assets/text/nes_message_data_static.o: build/assets/text/message_data.enc.
 build/assets/text/staff_message_data_static.o: build/assets/text/message_data_staff.enc.h
 
 build/assets/%.o: assets/%.c
-	$(CC) -c $(CFLAGS) $(MIPS_VERSION) $(OPTFLAGS) -o $@ $<
-	$(OBJCOPY) -O binary $@ $@.bin
+	$(CC) -c $(CFLAGS) $(MIPS_VERSION) -O1 -o $@ $<
```
Dragorn421, line 229: about optimization flags sure why not,

but having the objects as binary can in some occasions be useful so I'm not a fan of its plain removal,
idk if making it optional like by setting something like `BUILD_ASSETS_BIN ?= 0` would be a popular option

does removing the objcopy even gain much time?

EllipticEllipsis: Are they any different from what we get out of the baserom folder in the first place?

Dragorn421: well yes, `baserom/` files are from the base rom and built assets objcopied to binary are the built assets
if you were to edit an asset you may want to look at it with a hex editor, z64utils or whatever, which needs a binary file, not an object file

EllipticEllipsis: Oh, my memory was the data was just in the `.o` in the same form after the header, although it's been a while since I last looked. It's like the objdumped assembly file in my mind: not nearly so useful as it could be because it can't track the VRAM addresses etc.

And z64utils can view any binary? I remember needing that and it not having it.

Dragorn421 (approved):

louist103:
Closing until after #1056 is merged so there isn't any duplicate work with the asset building.

PR #1032, opened 2021-11-27
Preliminary fixes for boot/libultra

Thar0:
More boot/libultra/handwritten asm cleanup is to follow in future PRs, this PR only includes things that are either important to get in sooner, overlap with both handwritten asm and libultra, or are sufficiently small.

 - Fixes `entry.s` stack pointer assignment, the boot stack size will be moved to a header when we start pre-processing asm files
 - Standardizes `u32 prevInt`, before there was a mix of `s32` and `u32`
 - Removed some fake symbols from cache functions, and removes related undefined syms
 - Fixes the definitions of `__osGetHWIntrRoutine` and `__osSetHWIntrRoutine`
 - Fixes `PI_STATUS_ERROR`
 - Removes `viMesgStruct`, it was `OSIoMesg` all along
 - Sets the size of `osAppNmiBuffer` in the extern, so it can be used in the size arg for `bzero`


Roman971 (approved):

fig02 (approved):

EllipticEllipsis:
I'd like to look at this before merge if possible, but don't have time rn.

asm/osInvalDCache.s
```diff
@@ -49,13 +49,13 @@ glabel osInvalDCache
...
-/* 006E88 80006288 3C088000 */  lui   $t0, %hi(D_80000010) # $t0, 0x8000
+/* 006E88 80006288 3C088000 */  lui   $t0, 0x8000
 /* 006E8C 8000628C 010B4821 */  addu  $t1, $t0, $t3
 /* 006E90 80006290 2529FFF0 */  addiu $t1, $t1, -0x10
 .L80006294:
 /* 006E94 80006294 BD010000 */  cache 1, ($t0)
 /* 006E98 80006298 0109082B */  sltu  $at, $t0, $t1
 /* 006E9C 8000629C 1420FFFD */  bnez  $at, .L80006294
-/* 006EA0 800062A0 25080010 */   addiu $t0, %lo(D_80000010) # addiu $t0, $t0, 0x10
+/* 006EA0 800062A0 25080010 */   addiu $t0, 0x10
```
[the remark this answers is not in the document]
Thar0: This piece of handwritten code just so happens to look like a %hi/%lo pair so was mistaken for one when these were originally disassembled.
The first one is specifically the start of n64 ram, but this is just a dummy address and doesn't really matter I don't think, it's just so the `cache` op has a valid address. The other is a loop increment, this isn't actually loading a pointer like a %hi/%lo pair might have you think of.

undefined_syms.txt
```diff
@@ -1,11 +1,4 @@
...
-D_80000000 = 0x80000000;
-D_80000002 = 0x80000002;
-D_80000004 = 0x80000004;
-D_80000008 = 0x80000008;
-D_8000000C = 0x8000000C;
-D_80000010 = 0x80000010;
-D_80000020 = 0x80000020;
 osTvType = 0x80000300;
 osRomBase = 0x80000308;
 osResetType = 0x8000030C;
```
EllipticEllipsis, line 4: Are these going to stay as undefined_syms forever?

Thar0: We have two options. We can use a separate linker file, identical to `undefined_syms`, but containing only these symbols with a more appropriate name. Or, we can create a spec segment like the buffers segment starting at `0x80000300` and include a file with these as bss symbols. Creating a file for them may be slightly misleading however, since they're uneditable.

AngheloAlf: I personally like the idea of a separate linker file for those

EllipticEllipsis: Since, as far as I understand, these and the A4...s are used to talk to the hardware directly and are uneditable, they probably should be but somewhere that reflects that.

Thar0: The A4...s at least are removable, I was able to successfully get rid of those in the handwritten asm PR

EllipticEllipsis: Ah, so it would just be these few. Hmm, well same thing applies, although the file'll be shorter. I think an undefined_syms equivalent it the better option, since it looks less editable.

fig02:
@AngheloAlf @EllipticEllipsis are the discussions above concluded?

AngheloAlf (approved):

EllipticEllipsis (approved):
Green button, go!

PR #1034, opened 2021-11-27
Fix the rupee count in progress.py

Roman971:
Float issues are causing the rupee count to not always be 0 when the decompilable size lines up with a heart piece (eg. at 80 heart pieces). It should be fixed with this change.

### DIFF
--- a/progress.py
+++ b/progress.py
@@ -104,7 +104,7 @@ codePct = 100 * code / codeSize
 bootPct = 100 * boot / bootSize
 ovlPct = 100 * ovl / ovlSize
 
-bytesPerHeartPiece = total / 80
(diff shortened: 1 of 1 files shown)


progress.py
```diff
@@ -104,7 +104,7 @@ def GetNonMatchingSize(path):
 bootPct = 100 * boot / bootSize
 ovlPct = 100 * ovl / ovlSize
 
-bytesPerHeartPiece = total / 80
+bytesPerHeartPiece = int(total / 80)
```
AngheloAlf, line 107: Python has `//` for integer divisions, which I think it is a bit cleaner
```suggestion
bytesPerHeartPiece = total // 80
```

AngheloAlf (approved):

PR #1035, opened 2021-11-28
Pin docker image using image SHA instead of tags. Configure apt-get to never prompt for user input.

vancrozier:
Signed-off-by: vancrozier <5666264+vancrozier@users.noreply.github.com>


ethteck:
Hi, thanks for your PR! I'm not sure the tag is better from a user perspective, as now it's impossible to tell what exactly this is pulling from. I'm not a docker expert though - care to explain the reasoning for this?

vancrozier:
While attempting to follow the build instructions for docker, I ran into a few issues. Clang failed to install on the old docker image that was specified (18.04), using latest fixed that particular issue. I pinned the image using the SHA so that reproducibility is enhanced.

After upgrading, apt-get was prompting to select the system timezone, so I disabled interactivity in apt-get by using the `DEBIAN_FRONTEND=noninteractive` environment variable.

ethteck:
Unfortunately, the docker setup has been completely neglected, as Docker is not really necessary to build the project anymore, compared to earlier when I set it up. 

So this isn't using an 18.04 image? How can I tell what this is actually pulling from?

vancrozier:
> So this isn't using an 18.04 image? How can I tell what this is actually pulling from?

Hi, thanks for reviewing. Obviously not the most critical of PRs, but this will be useful for those that like to use Docker instead of installing tools on a VM / dev machine.

My PR specifies the latest official `ubuntu:latest` tag at the time of writing. I use the SHA, rather than the `ubuntu` repo's `latest` tag, so that when the `latest` tag changes later the exact same base image will be pulled regardless. This article probably does a better job of explaining why this is good practice: https://rockbag.medium.com/why-you-should-pin-your-docker-images-with-sha-instead-of-tags-fd132443b8a6

You can verify the sha by using this command, it will show the SHA of `ubuntu:latest` (and this is how I found the SHA that I use in the PR)

`docker inspect --format='{{index .RepoDigests 0}}' ubuntu:latest`

If you still feel this doesn't add any value, please feel free to close this PR.

raulferras:
Tagging the docker image you want to use is, as @vancrozier already said, actually a very good practice. 
It mainly avoids suddenly the project not building because a new docker image has been released and having missing or different components.

Tagging it, you are sure you always get the very same ubuntu docker image version.
Using `:latest` on the other hand, will get the last version available, and this does not guarantee you will get the same ubuntu version.

ethteck:
I see the benefit to pinning the base image to a specific hash, but I find it a bit unfortunate that it makes it impossible to eyeball what you're grabbing. 18.04 is a thing I can understand just by looking at it, but some long hash isn't. Is there no way to say ubuntu:20.04@hash (or something like that)? Not a huge deal, but yeah was just pointing that out.

As for this PR in, it looks great to me in general. However, as you can tell, we've kind of neglected to maintain the Dockerfile / surrounding stuff (as you can tell). It isn't as necessary lately since the game's compiler can now be run on many more operating systems than before. For those reasons, we're thinking about removing the Dockerfile entirely. I'll write an update soon. If enough people find it useful or necessary to use the repo, perhaps we can keep it and make a better effort to maintain it. But I suspect people won't need it

peterwiggers:
Just use ubuntu:18.04@[the-hash]

It’s readable and immutable.

ethteck:
Oh, this works? then yeah, we should do that

vancrozier:
> Just use ubuntu:18.04@[the-hash]
> It’s readable and immutable.

I just tried that, it works, however as far as I can tell the `:18.04` part is not honored; docker will pull via the hash, but the `18.04` is not parsed and can be literally anything. If you're not going to delete the Dockerfile altogether, you could include in a comment which common tag the hash is at the current date - e.g.

`
FROM ubuntu@sha256:626ffe58f6e7566e00254b638eb7e0f3b11d4da9675088f4781a50ae288f3322 as build
# The above specifies the SHA of ubuntu:20.04 as of 2021-11-28. 
`

Also, just for some context as to why I made this change - I saw this project on Reddit, and I wanted to try building the software on my own, purely out of curiosity. As a matter of security / cleanliness / isolation, I prefer to use docker containers - when they're provided by the software developer, you usually get a well-designed docker image that will "just work" as far as building the software; you won't run into issues with missing dependencies, conflicting packages, etc.

raulferras:
Just to add to what @vancrozier already said:
even if the toolchain works on major operating systems, consider the benefit of just maintaining one of them plus the benefit of having a frozen environment that will not change unexpectedly due to an OS update or third party libraries.
I use docker extensively and can tell it saves lots of headaches (when talking about executing something cross-systems).

ethteck:
Cool. Thank you both for your input. I'm totally in favor of merging this, then, especially if people are using it and find it useful.  But it would be nice to have the comment there, as you suggested @vancrozier .

vancrozier:
Sounds good, I added a comment specifying as much (after rebasing my branch)

Dragorn421 (approved):
I'm no docker expert but the points and the solutions raised sound sensible.

-----

Not a huge deal but the python pip installs may be outdated?

```dockerfile
RUN python3 -m pip install --user colorama ansiwrap attrs watchdog python-Levenshtein
```
this is probably the diff.py requirements? (if so, only `colorama watchdog python-Levenshtein` are required https://github.com/simonlindholm/asm-differ/blob/b41f967ff2493914e4aa8fab7b66ee9e38806f80/diff.py#L343 )

```dockerfile
RUN python3 -m pip install --upgrade attrs pycparser
```
I'm not sure what this is for. pycparser hints that it would be the permuter ( https://github.com/simonlindholm/decomp-permuter , currently requires `pycparser pynacl toml`)

However neither of these (diff.py and permuter) are required to build the rom, they are only tools to help decompiling, so idk if it's really relevant to set up for them here. The permuter isn't even included in the repo (but diff.py is)

Anyway this isn't really relevant to the PR. we just never get a chance to talk about docker so I took the opportunity

fig02:
closing this PR because a new one has been opened that addresses this issue+more 
https://github.com/zeldaret/oot/pull/1062

appreciate the contribution regardless, and feel free to weigh in on the other pr if you like

PR #1036, opened 2021-11-28
Rework handling of dmadata

Thar0:
- Moved dmadata to c
 - Created a table for the matching layout for PAL MQ Debug (and possibly any versions with identical layout)
 - Split spec parsing code from `mkldscript.c` into `spec.c`
 - Added a new tool using `spec.c`, `mkdmadata.c`, to automatically generate a dmadata layout from the spec for `NON_MATCHING` builds
 - Used the dmadata table to generate the contents of `gDmaDataTable` and `sDmaMgrFileNames`


fig02 (approved):

Makefile
```diff
@@ -229,6 +232,12 @@ build/assets/%.o: assets/%.c
 	$(CC) -c $(CFLAGS) $(MIPS_VERSION) $(OPTFLAGS) -o $@ $<
 	$(OBJCOPY) -O binary $@ $@.bin
 
+build/include/tables/dmadata_table.h: build/$(SPEC)
```
Roman971, line 235: I'm not sure it makes sense to make that file in `build/include/`, it seems like it could be misleading. I'd say we could just have it in root as `build/dmadata_table.h` like other build files we create specifically from the makefile (or maybe `build/dmadata_table_spec.h` or something along those lines so it's not confused with the real `dmadata_table.h`)

Makefile
```diff
@@ -229,6 +232,12 @@ build/assets/%.o: assets/%.c
...
+	$(MKDMADATA) $< $@
+
+build/src/boot/z_std_dma.o: build/include/tables/dmadata_table.h
+build/src/dmadata/dmadata.o: build/include/tables/dmadata_table.h
```
Roman971, line 239: Very minor but this should probably be wrapped in a `ifeq ($(NON_MATCHING),1)` so it doesn't trigger rebuilds for matching builds

src/dmadata/dmadata.c
```diff
@@ -0,0 +1,30 @@
...
+
+// dmadata Table definition
+#define DEFINE_DMA_ENTRY(name) \
+    { (u32)_##name##SegmentRomStart, (u32)_##name##SegmentRomEnd, (u32)_##name##SegmentRomStart, 0 },
+
+#define DEFINE_DMA_ENTRY_END() \
+    { 0 },
+
+DmaEntry gDmaDataTable[] = {
+    #include "tables/dmadata_table.h"
```
Roman971, line 23: Iirc the formatter removes the indentation for these (which is fine, it just means the files likely need a format run)

Roman971 (commented):
Looks good to me overall, just a few comments/suggestions

Makefile
Thar0: Not sure I can do this, since `CC_CHECK` fails as it always builds with `NON_MATCHING`, so the generated file needs to present for it

Roman971: Yeah that's true. In that case maybe we should wrap the spec dmadata recipe as well in the if, and have a recipe for it in the else which just outputs a default file so CC_CHECK doesn't fail. Or perhaps we should have a C define specific to using spec dmadata instead of NON_MATCHING, which would be enabled if NON_MATCHING==1 in the makefile. 
I'm not sure what's the best option, but I feel like it's important to negate the impact for matching builds where the spec dmadata goes completely unused.

Roman971: Thinking about it more, this stuff might not matter to the build performance overall, especially once we start making relocs based on spec (which would trigger its own rebuilds based on changes to the spec).
So assuming mkdmadata is almost instant, we could probably leave it always running along with these rebuilds, and it would have the one advantage of actually running during normal builds so errors don't go silent.

Thar0: mkdmadata is essentially instant, I get on average 0.007s/0.008s when timing:
`time ./tools/mkdmadata build/spec build/dmadata_table_spec.h`
so it should be consistently low for all kinds of systems
```
real    0m0.007s
user    0m0.000s
sys     0m0.000s
```

tools/mkdmadata.c
```diff
@@ -0,0 +1,55 @@
...
+        usage(argv[0]);
+        return 1;
+    }
+
+    spec = util_read_whole_file(argv[1], &size);
+    parse_rom_spec(spec, &g_segments, &g_segmentsCount);
+
+    dmaout = fopen(argv[2], "w");
+    if (dmaout == NULL)
+        util_fatal_error("failed to open file '%s' for writing", argv[3]);
```
Dragorn421, line 49:
```suggestion
        util_fatal_error("failed to open file '%s' for writing", argv[2]);
```

Dragorn421 (changes requested):

include/tables/dmadata_table.h
```diff
@@ -0,0 +1,9 @@
+/**
+ * Select dmadata table for version
+ */
+#ifdef NON_MATCHING
+// For non-matching builds, dmadata is generated from the specfile segments
+#include "dmadata_table_spec.h"
+#else
```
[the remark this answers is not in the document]
Thar0: The spec is designed to match the order files appear in the rom. The ordering of files in dmadata is not the same (though is almost the same) as the ordering in rom, so by extension is not the same as the ordering in the spec. The only way around this that I can think of is to add more spec directives to define the dmadata ordering, which is what I originally thought to do, however after discussing in discord we thought it may be better to just hardcode the matching dmadata ordering rather than adding all this extra machinery. Something that also complicates matters somewhat is the buffers segment, since it doesn't exist in dmadata at all as it is only bss, this would need special handling as well if we wanted to generate a matching dmadata.

src/boot/z_std_dma.c
```diff
@@ -13,1541 +13,15 @@ u32 D_80009460 = 0;
...
-    "anime_texture_3_static",
-    "anime_texture_4_static",
-    "anime_texture_5_static",
-    "anime_texture_6_static",
-    "softsprite_matrix_static",
+// dmadata filenames
+#define DEFINE_DMA_ENTRY(name) #name,
+
+const char* sDmaMgrFileNames[] = {
+#include "tables/dmadata_table.h"
```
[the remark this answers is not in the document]
Thar0: This issue is pervasive in the other tables we currently have as well. I agree this should be resolved but I'd like to get some more input on a resolution that works for the other tables also.

fig02 (approved):

Dragorn421 (approved):

Roman971 (approved):

PR #1038, opened 2021-11-29
libultra files and directories restructure

Thar0:
Renames and sorts libultra files into their respective folders. Also began working on headers, however I felt that this was getting large enough that it was worth splitting up, this will be continued in a future PR.
The information on original names and directory structure is from the [libultra mdebug](https://github.com/Rozelette/libultra-mdebug) repo and other libultra decompilations such as sm64's, majoras mask's and libreultra.


src/libultra/io/motor.c
```diff
@@ -45,7 +45,7 @@ s32 osSetRumble(OSPfs* pfs, u32 vibrate) {
     return ret;
 }
 
-void osSetUpMempakWrite(s32 channel, OSPifRam* buf) {
+void _MakeMotorData(s32 channel, OSPifRam* buf) {
```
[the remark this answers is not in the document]
Thar0: I looked into this before for Majora's Mask libultra, this is the original name. It has such an unusual name because it was originally marked static, so wouldn't have been visible outside of this file. I confirmed this by looking in the mdebug section of a libultra.a that contained motor.o

fig02 (approved):

src/libultra/io/aisetnextbuf.c
```diff
@@ -14,12 +17,15 @@ s32 osAiSetNextBuffer(void* buf, u32 size) {
...
+    // Originally a call to __osAiDeviceBusy
     status = HW_REG(AI_STATUS_REG, s32);
     if (status & AI_STATUS_AI_FULL) {
         return -1;
     }
 
-    HW_REG(AI_DRAM_ADDR_REG, u32) = PHYSICAL_TO_VIRTUAL(bufAdjusted);
+    // VIRTUAL_TO_PHYSICAL replaces osVirtualToPhysical, this replacement
+    // assumes that only KSEG0 addresses are given
+    HW_REG(AI_DRAM_ADDR_REG, u32) = VIRTUAL_TO_PHYSICAL(bufAdjusted);
```
Roman971, line 28: Since this is libultra, should it use `OS_K0_TO_PHYSICAL` to avoid depending on our own `VIRTUAL_TO_PHYSICAL` macro? I imagine it wouldn't make sense for a libultra file to include a project specific header either

include/ultra64.h
```diff
@@ -1,13 +1,16 @@
...
+#include "ultra64/types.h"
 #include "unk.h"
 
-#include "ultra64/types.h"
-#include "ultra64/interrupt.h"
+#include "libc/stdarg.h"
+#include "libc/stdbool.h"
+#include "libc/stddef.h"
+#include "libc/stdlib.h"
+#include "libc/math.h"
```
Roman971, line 11: Just a thought and we don't need to do anything about this right now, but does it make sense to have these includes in `ultra64.h`? Same question for `unk.h` tho that one is probably easier to answer.

Roman971 (commented):
Looks good to me overall, just a couple things

include/ultra64.h
Thar0: The libc headers are needed in libultra, and of course in other parts of the codebase. I think it's appropriate currently to have them in `ultra64.h` but this may change depending on how libultra headers go. It also depends how we want to include libultra headers in non-libultra files, whether we just include `ultra64.h` or if we want to be more specific with our libultra includes. If the former, I think the libc headers are good to stay here.

Roman971 (approved):

include/ultra64/exception.h
```diff
@@ -0,0 +1,45 @@
...
+#define OS_IM_RDBREAD   0x00004401
+#define OS_IM_COUNTER   0x00008401
+#define OS_IM_CPU       0x0000ff01
+#define OS_IM_SP        0x00010401
+#define OS_IM_SI        0x00020401
+#define OS_IM_AI        0x00040401
+#define OS_IM_VI        0x00080401
+#define OS_IM_PI        0x00100401
+#define OS_IM_DP        0x00200401
+#define OS_IM_ALL       0x003fff01
```
Dragorn421, line 22: we've been using uppercase for hex
also with `OS_IM_CPU`, `RCP_IMASK`
and in `r4300.h`

src/libultra/os/createthread.c
```diff
@@ -18,12 +18,13 @@ void osCreateThread(OSThread* thread, OSId id, void (*entry)(void*), void* arg,
...
     thread->context.sp = (u64)(s32)sp - 16;
     thread->context.ra = __osCleanupThread;
-    t8 = 0x3FFF01;
-    thread->context.sr = (t8 & 0xFF01) | 2;
-    thread->context.rcp = (t8 & 0x3F0000) >> 16;
-    thread->context.fpcsr = 0x01000800;
+
+    mask = 0x3FFF01;
+    thread->context.sr = (mask & 0xFF01) | 2;
+    thread->context.rcp = (mask & 0x3F0000) >> 16;
```
Dragorn421, line 24:
```suggestion
    mask = OS_IM_ALL;
    thread->context.sr = (mask & OS_IM_CPU) | 2;
    thread->context.rcp = (mask & RCP_IMASK) >> 16;
```

src/libultra/os/initialize.c
```diff
@@ -10,7 +10,7 @@ typedef struct {
 u64 osClockRate = OS_CLOCK_RATE;
 s32 osViClock = VI_NTSC_CLOCK;
 u32 __osShutdown = 0;
-u32 __OSGlobalIntMask = 0x003FFF01;
+OSHWIntr __OSGlobalIntMask = 0x003FFF01;
```
Dragorn421, line 13:
```suggestion
OSHWIntr __OSGlobalIntMask = OS_IM_ALL;
```

Dragorn421 (commented):
Do you plan on organizing libultra stuff located under `asm/`?

Thar0:
I plan to tackle organizing the contents of `asm/` more once #865 is done, which I intend to return to after this

fig02 (approved):

Dragorn421 (approved):

PR #1045, opened 2021-12-01
Document `Actor_DrawDoorLock`

Dragorn421:


fig02 (approved):

Roman971 (approved):

EllipticEllipsis:
In this the one that displays garbage if you look at the back in boss rooms because the wrong object is loaded?

PR #1047, opened 2021-12-01
Remove THIS macro

fig02:
This is something we've mentioned in passing here and there, but now that all actors are done I think its a good time.
Not everyone may be aware of this planned change, so this PR also serves as a discussion starter if need be.

My opinion on the matter:
This macro made sense to have while we were decompiling things, its a construct we had to type often. But long term, I think it hides information that is useful to see on the surface without needing a macro. Seeing that upcast from Actor* to the specific actor is important, and not really that verbose.
Additionally, the same thing will happen for gamestate -> globalCtx, and I dont think we plan to introduce a macro for that.

Thanks to tharo for helping me make a script to do this, im a pepega when it comes to making tools lol


Dragorn421 (approved):
I thought there were a lot of actors using inlined `THIS` due to regalloc or stack not allowing a `this` temp,
but that actually only happens in `BgMizuMovebg_Init` (which now looks extra bad)

fig02:
yea, and that one we believe to be a fake match and can hopefully be removed

Roman971 (approved):

mzxrules:
I'm gonna miss THIS

louist103 (approved):
The end of an era.

petrie911 (approved):

engineer124 (approved):

PR #1050, opened 2021-12-01
Symbol cleanup 4

Dragorn421:
Replace some hardcoded constants (eg `0x06000118`) with proper asset symbols (`&object_bwall_Col_000118`)

I think this may be only around 10% (?) of all the symbols that need to be used so prepare for more PRs

It is often non-trivial to find the symbol, either because the actor uses several objects, or the symbol has to be added to the zapd xml and that gets more complicated when the symbol is a texture passed "by segment" to a dlist, anyway.


assets/xml/objects/object_tw.xml
```diff
@@ -50,6 +50,7 @@
         <Texture Name="object_tw_Tex_009E38" OutName="tex_009E38" Format="rgba16" Width="16" Height="16" Offset="0x9E38"/>
         <Texture Name="object_tw_Tex_00A038" OutName="tex_00A038" Format="rgba16" Width="16" Height="16" Offset="0xA038"/>
         <Texture Name="object_tw_Tex_00A238" OutName="tex_00A238" Format="rgba16" Width="16" Height="16" Offset="0xA238"/>
+        <Texture Name="object_tw_Tex_00A438" OutName="tex_00A438" Format="ci8" Width="32" Height="32" Offset="0xA438" TlutOffset="0x9490"/>
         <!--Blob Name="object_tw_Blob_00A438" Size="0x400" Offset="0xA438" /-->
```
EllipticEllipsis, line 54: Looks like you covered at least some of this.

Roman971 (approved):

EllipticEllipsis (commented):
Think you kept a blob comment accidentally, everything else looks okay to me.

fig02 (approved):

PR #1053, opened 2021-12-02
Symbol cleanup 5

Dragorn421:
#1050 but continued


src/overlays/actors/ovl_En_Ossan/z_en_ossan.c
```diff
@@ -2353,7 +2353,11 @@ void EnOssan_DrawBazaarShopkeeper(Actor* thisx, GlobalContext* globalCtx) {
 
 s32 EnOssan_OverrideLimbDrawKokiriShopkeeper(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3f* pos,
                                              Vec3s* rot, void* thisx) {
-    static void* sKokiriShopkeeperEyeTextures[] = { 0x06001570, 0x060001F0, 0x06000B30 };
+    static void* sKokiriShopkeeperEyeTextures[] = {
+        gKokiriShopkeeperEyeDefaultTex,
```
fig02, line 2357: what does default mean in the context of an eye texture?

Dragorn421: I don't know (I wasn't the one who named this symbol) so I'm going to guess

I assume it may be because it's referenced in a dlist (`gKokiriShopkeeperHeadDL`) while the other two are not.

But it's kind of weird, that `gKokiriShopkeeperEyeDefaultTex` texture is used for two triangles and the "dynamic" eye texture for eight others

I assume they somehow didn't set those two triangles to be "segment 10 textured"


-------

so it should probably be "open" instead of default

(interesting asset "bug" though)

fig02: oh i didnt realize it wasnt your name, in that case we can reevaluate later. thanks for taking a look at it though

fig02 (approved):

Roman971 (approved):

PR #1056, opened 2021-12-03
Add GCC compiler option.

ProjectRevoTPP:
Folks, this has been a very long time coming. Let me just state that there have been many, many hardships debugging this (especially that GCC reloc **nonsense**) but the wait is finally over. Here is the PR which officially adds a COMPILER variable and the ability to use the GCC compiler right out the gate as long as gcc-mips-linux-gnu is installed.

While this is nice and all, I can already spot some concerns which may need the appropriate amount of discussion and review:

* Currently, I add an iconv option that you can toggle on or off. This is very slow to use with GCC and I only had to do this because the debug prints will break otherwise since GCC seems to completely refuse to recognize the EUC-JP encoding natively. Perhaps there's a better option than this?
* When defining GCC, we assume NON_MATCHING, to prevent GLOBAL_ASM gaps from being struck by GCC. This should be currently fine to avoid regressions, however when it comes time to add version support there may be repeated GCC regressions as a result of a WIP of function implementation. Is expecting this option to break normally expected?
* There is also an extra use of iconv for the data files that does not use the CONV_SOURCE_TEXT check. I tried to make it clear this flag is tied strictly to piping the converted output into GCC, but perhaps this is not ideal.
* As a result of missing implicits and libraries to download, many functions are missing, which have to be defined in asm/llmuldiv_gcc and src/boot/missing_gcc_functions respectively. While you can technically use the game's MemCopy and such, they HAVE to be in boot, or the game will crash due to GCC compiling memsets and memcpys in the boot section. Again, not quite ideal? Ideally, I would like to avoid adding new code explicitly for GCC, but I digress, and this is the best I could do.
* I also had to add an AVOID_UB flag due to the game crashing in a few places. Currently, I was only able to identify 2 crashes and as such only 2 explicit UB fixes when using GCC. The game will crash on boot if the audio_heap.c one isn't fixed, and the Zora Shop will crash due to a missing 3rd string table entry in z_en_tana.

Your commentary on this PR 1y+ in the making is very much appreciated.


Makefile
```diff
@@ -77,10 +104,17 @@ ZAPD       := tools/ZAPD/ZAPD.out
 
 OPTFLAGS := -O2
 ASFLAGS := -march=vr4300 -32 -Iinclude
-MIPS_VERSION := -mips2
 
-# we support Microsoft extensions such as anonymous structs, which the compiler does support but warns for their usage. Surpress the warnings with -woff.
-CFLAGS += -G 0 -non_shared -Xfullwarn -Xcpluscomm $(INC) -Wab,-r4300_mul -woff 649,838,712
+ifeq ($(COMPILER),gcc)
+  CFLAGS += -c -G 0 -nostdinc -Iinclude -Isrc -Iassets -Ibuild -I. -DNON_MATCHING=1 -DMODDING=1 -DNORMAL_GAMEPLAY=1 -DAVOID_UB=1 -mno-shared -march=vr4300 -mfix4300 -mabi=32 -mhard-float -mdivide-breaks -fno-stack-protector -fno-common -fno-zero-initialized-in-bss -mno-abicalls -fno-strict-aliasing -fno-inline-functions -fno-inline-small-functions -fno-toplevel-reorder -ffreestanding -fwrapv $(CHECK_WARNINGS) -g -mno-explicit-relocs -mno-split-addresses -funsigned-char
```
AngheloAlf, line 109:
```suggestion
  CFLAGS += -c -G 0 -nostdinc $(INC) -DNON_MATCHING=1 -DMODDING=1 -DNORMAL_GAMEPLAY=1 -DAVOID_UB=1 -mno-shared -march=vr4300 -mfix4300 -mabi=32 -mhard-float -mdivide-breaks -fno-stack-protector -fno-common -fno-zero-initialized-in-bss -mno-abicalls -fno-strict-aliasing -fno-inline-functions -fno-inline-small-functions -fno-toplevel-reorder -ffreestanding -fwrapv $(CHECK_WARNINGS) -g -mno-explicit-relocs -mno-split-addresses -funsigned-char
```

AngheloAlf: Also, could you explain the `MODDING` and `NORMAL_GAMEPLAY` defines?

Makefile
```diff
@@ -241,25 +277,25 @@ build/src/boot/z_std_dma.o: build/dmadata_table_spec.h
 build/src/dmadata/dmadata.o: build/dmadata_table_spec.h
 
 build/src/overlays/%.o: src/overlays/%.c
-	$(CC) -c $(CFLAGS) $(MIPS_VERSION) $(OPTFLAGS) -o $@ $<
+	$(CC) -c $(CFLAGS) -I $(<D) $(MIPS_VERSION) $(OPTFLAGS) -o $@ $(PIPEIN)
 	$(CC_CHECK) $<
-	$(ZAPD) bovl -eh -i $@ -cfg $< --outputpath $(@D)/$(notdir $(@D))_reloc.s
+	$(ZAPD) bovl -eh -i $@ -cfg $< --outputpath $(@D)/$(notdir $(@D))_reloc.s --gcc-compat
```
AngheloAlf, line 282: ` --gcc-compat` should not be enabled on IDO builds, because it is significantly slower compared to not use it (that's the reason why it is a flag instead of enabled by default)

Makefile
```diff
@@ -231,7 +267,7 @@ build/assets/text/nes_message_data_static.o: build/assets/text/message_data.enc.
 build/assets/text/staff_message_data_static.o: build/assets/text/message_data_staff.enc.h
 
 build/assets/%.o: assets/%.c
-	$(CC) -c $(CFLAGS) $(MIPS_VERSION) $(OPTFLAGS) -o $@ $<
+	$(CC) -c $(CFLAGS) -I $(<D) $(MIPS_VERSION) $(OPTFLAGS) -o $@ $(PIPEIN)
```
AngheloAlf, line 270: What's the motivation of adding `-I $(<D)`?

Makefile
ProjectRevoTPP: Ah, whoops! I didn't mean to leave those in.

Makefile
ProjectRevoTPP: GCC will error being unable to find file_choose.h if you don't.

fig02: why that one specifically, do you know?

ProjectRevoTPP: Cant reproduce it actually. Swear I was seeing this earlier, even during GCC research. Removing these now.

Makefile
```diff
@@ -8,10 +8,20 @@ COMPARE ?= 1
 NON_MATCHING ?= 0
 # If ORIG_COMPILER is 1, compile with QEMU_IRIX and the original compiler
 ORIG_COMPILER ?= 0
+# If COMPILER is GCC, compile with GCC instead of IDO.
+COMPILER ?= ido
```
Dragorn421, line 12: we may want to add a check that `ORIG_COMPILER` can only be non-0 with IDO

Makefile
```diff
@@ -8,10 +8,20 @@ COMPARE ?= 1
...
+
+# If gcc is used, define the NON_MATCHING flag respectively so the files that
+# are safe to be used can avoid using GLOBAL_ASM which doesn't work with gcc.
+ifeq ($(COMPILER),gcc)
+  NON_MATCHING := 1
+  CPPFLAGS := -DCOMPILER_GCC
+  ZAPDFLAGS := --gcc-compat
```
Dragorn421, line 19: I think ideally there should be `ZAPDFLAGS ?=` at the top level (not in an if) to "declare" the variable (for makefile documentation purpose mostly), and this would be +=

README.md
```diff
@@ -90,6 +90,10 @@ sudo apt-get update
 sudo apt-get install git build-essential binutils-mips-linux-gnu python3 libpng-dev
 ```
 
+If you are using GCC as the compiler for Ocarina of Time, you will also need:
+
+* gcc-mips-linux-gnu
```
Dragorn421, line 95: the makefile says `mips-linux-gnu-gcc`

Dragorn421 (changes requested):

Dragorn421 (commented):
Looks good to me but i'm kind of scared of the `missing_gcc_functions.c` and `llmuldiv_gcc.s` files.
They feel like modding, which doesn't fit in the repo imo, and especially don't fit in src/ or asm/ folders which have only contained stuff from the vanilla game for the longest time

ProjectRevoTPP:
Including "new" files was a concern of mine, but I cant seem to find any reasonable way to link these in instead of adding new files. Any suggestions to avoid having to do this is welcome. Maybe I can put them somewhere else?

louist103:
This is more general so I'll just say it here. Since CC_CHECK is just GCC you can probably skip that step when compiling with mips-gcc since it will likely be the same.

ProjectRevoTPP:
Waiting on #1016 before updating this.

fig02 (approved):
Ive tested this setup by playing the whole game *twice* using two different toolchains (`mips-linux-gnu` and `mips64-gcc`), and was able to play a majority of the game without issues.

Of course we havent found all the UB that could be problematic, but this is a great start and I think this is usable for modding in its current state.

Highly suggest those involved with gcc over the last couple of weeks take a look over this as well to make sure we've consolidated all of our ideas properly @Thar0 @EllipticEllipsis @AngheloAlf and whoever else is interested.

fig02 (approved):
ido builds work now. missing_gcc_functions was separated to its own folder and is only built if the compiler used is gcc

src/gcc-fix/missing_gcc_functions.c
```diff
@@ -0,0 +1,211 @@
...
+    sw    $a0, ($sp)                        \n\
+    sw    $a1, 4($sp)                       \n\
+    sw    $a2, 8($sp)                       \n\
+    sw    $a3, 0xc($sp)                     \n\
+    ld    $t7, 8($sp)                       \n\
+    ld    $t6, ($sp)                        \n\
+    ddivu $zero, $t6, $t7                   \n\
+    bnez  $t7, 1f                           \n\
+     nop                                    \n\
+    break 7                                 \n\
```
AngheloAlf, line 122: Why use inline asm instead of just a .s file?

fig02: the idea is to reduce how intrusive these changes are. the only downside is less readability, but there arent meant to be read and modified anyway. it keeps all of the gcc related hacks in one place

tools/mkldscript.c
```diff
@@ -73,7 +75,8 @@ static void write_ld_script(FILE *fout)
         for (j = 0; j < seg->includesCount; j++)
         {
             if (!seg->includes[j].dataWithRodata)
-                fprintf(fout, "            %s (.data)\n", seg->includes[j].fpath);
+                fprintf(fout, "            %s (.data)\n"
+                              "        . = ALIGN(0x10);\n", seg->includes[j].fpath);
```
EllipticEllipsis, line 79: We might consider adding a comment on the current necessity of section alignment somewhere, since this file is now festooned in `ALIGN(0x10)`s.

tools/mkldscript.c
```diff
@@ -129,23 +135,22 @@ static void write_ld_script(FILE *fout)
...
             fprintf(fout, "            %s (.ovl)\n", seg->includes[j].fpath);
 
-        fprintf(fout, "        . = ALIGN(0x10);\n");
-
         fprintf(fout, "        _%sSegmentOvlEnd = .;\n", seg->name);
 
         if (seg->fields & (1 << STMT_increment))
             fprintf(fout, "    . += 0x%08X;\n", seg->increment);
 
+        // fprintf(fout, "        . = ALIGN(0x%X);\n", seg->align);
```
EllipticEllipsis, line 143: Why is this here but commented out?

tools/mkldscript.c
```diff
@@ -193,13 +208,55 @@ static void write_ld_script(FILE *fout)
         //fprintf(fout, "\n    }\n");
     }
 
-
-    fputs("    _RomEnd = _RomSize;\n}\n", fout);
+    fputs("    _RomEnd = _RomSize;\n\n", fout);
+
+    // Debugging sections
```
EllipticEllipsis, line 213: I'm rather surprised that all this has to be set up manually. Is that because we're using an explicit linker script?

tools/mkldscript.c
AngheloAlf: Yeah, noting somewhere why the `ALIGN(0x10)`s  where added sound like a good idea. Specially considering when the other rodata sections have an explanation

Makefile
```diff
@@ -8,14 +8,32 @@ COMPARE ?= 1
...
+
+# ORIG_COMPILER cannot be combined with a non-IDO compiler. Check for this case and error out if found.
+ifneq ($(COMPILER),ido)
+  ifeq ($(ORIG_COMPILER),1)
+    $(error ORIG_COMPILER can only be used with the IDO compiler. Please check your Makefile variables and try again)
+  endif
+endif
+
+# If gcc is used, define the NON_MATCHING flag respectively so the files that
+# are safe to be used can avoid using GLOBAL_ASM which doesn't work with gcc.
```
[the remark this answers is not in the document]
fig02: tbh this comment can probably be removed. NON_MATCHING also allows the build to do other things, like use the auto generated dma table and such. We may want to document all of what NON_MATCHING allows, but dont need to single out this specific thing

tools/mkldscript.c
fig02: do you guys have suggestions for what and where to put the comment (there are alignment changes all over the file, picking a spot is hard)
you guys are more well versed in the reloc issue than I am

fig02:
Sure yeah thats fair. Agreed on cross that bridge when we get there

tools/mkldscript.c
EllipticEllipsis: Given that it's all over the place, it seems like it would be best at the top of the file rather than duplicating the comment everywhere. Really this whole thing could do with documenting, if not refactoring: at the moment it's pretty hard to follow, especially should another project want to use it.

fig02: agreed there, we also have commented out stuff everywhere. Figure if we were moving forward with the z64 shared build tools, we can run the formatter and clean things up then

EllipticEllipsis: I discussed this with @AngheloAlf and we came up with the following stopgap for now:
> ```Note: *SECTION ALIGNMENT* Object files built with a compiler such as GCC can by default use narrower alignment for sections than IDO's 0x10. To properly generate relocations relative to section starts, sections currently need to be aligned to 0x10, because otherwise the linker inserts padding in the wrong place.```

It is perhaps overkill, but hopefully it's clear enough.

Makefile
```diff
@@ -8,14 +8,32 @@ COMPARE ?= 1
 NON_MATCHING ?= 0
 # If ORIG_COMPILER is 1, compile with QEMU_IRIX and the original compiler
 ORIG_COMPILER ?= 0
+# If COMPILER is GCC, compile with GCC instead of IDO.
+COMPILER ?= ido
+# Declare CPPFLAGS used for the preprocessor.
+CPPFLAGS ?=
```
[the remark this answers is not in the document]
fig02: so just put `C_FLAGS ?=` right above this? does anything else need changing?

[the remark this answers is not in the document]
EllipticEllipsis: I think that should suffice, although I defer to @Roman971 for in-depth makefile stuff.

tools/mkldscript.c
Thar0: Can be removed

tools/mkldscript.c
Thar0: I changed the /DISCARD/ at the bottom to remove all sections that aren't explicitly mentioned. From what I understand this is better practice and is a straight improvement from just discarding a couple of known problematic sections when there could be more. Due to this though, I had to specify the debugging sections explicitly so they aren't also discarded.

tools/mkldscript.c
```diff
@@ -88,14 +95,17 @@ static void write_ld_script(FILE *fout)
         fprintf(fout, "        _%sSegmentDataEnd = .;\n", seg->name);
 
         fprintf(fout, "    _%sSegmentDataSize = ABSOLUTE( _%sSegmentDataEnd - _%sSegmentDataStart );\n", seg->name, seg->name, seg->name);
-
+
```
Thar0, line 98:
```suggestion

```

Thar0 (approved):

EllipticEllipsis (approved):

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -452,9 +452,15 @@ s16 EnZf_FindNextPlatformAwayFromPlayer(Vec3f* pos, s16 curPlatform, s16 arg2, G
         }
     }
 
-    // These functions have no side effects, so these two calls do nothing
+    //! @bug `altNextPlatform` can be -1 in certain conditions and cause an out of bounds access.
+    //! Under normal conditions, this doesn't cause problems because the data before `sPlatformPositions`
+    //! is section padding between .text and .data, so 0 gets read as a float.
+    //! Other compilers can shift things and result in a crash. Defining `AVOID_UB` will remove these function calls
+    //! entirely as they are dead code and do not have any effect.
```
Dragorn421, line 459:
```suggestion
    //! entirely as they have no side effect.
```
it's not dead code (it gets run), and the proper computer science term is indeed "side effect" like in the current comment

tools/mkldscript.c
```diff
@@ -9,6 +9,10 @@
 #include "spec.h"
 #include "util.h"
 
+// Note: *SECTION ALIGNMENT* Object files built with a compiler such as GCC can, by default, use narrower alignment 
+// for sections compared to IDO's 0x10. To properly generate relocations relative to section starts, sections currently 
+// need to be aligned to 0x10, because otherwise the linker inserts padding in the wrong place.
```
Dragorn421, line 14: when this says "currently", is it a Fado limitation or a link-time-related limitation?

src/gcc-fix/missing_gcc_functions.c
```diff
@@ -0,0 +1,211 @@
...
+        if (b == 0) {
+            break;
+        }
+
+        a *= a;
+    }
+    return recip ? 1 / r : r;
+}
+
+__asm__("                                   \n\
```
Dragorn421, line 103: so if I got it right the only reason this uses __asm__ is to avoid clutter?
I think since gcc-fix is a subfolder already it can hold one more file, and otherwise it's just one more line in the spec, looks worth the readability of using a .s file

but it's not a big deal

Dragorn421 (commented):
So I hate to kill the mood but can I be convinced that it makes sense for the repo to include code entirely unrelated to the reverse-engineering effort?
Not just `src/gcc-fix/`, also `AVOID_UB`. To me it looks more fitting in a modding base repo than in the game's decompilation. If we start fixing issues, if we write new code inside the reversed-engineered code, where do we stop?

I think it's really cool to have gcc support and the steps have been interesting to follow, but I'm worried about where this is taking the repo. Taking it away from the path of being "just" a reverse-engineering project is my feeling.

tools/mkldscript.c
EllipticEllipsis: It is, as this says, because the linker puts the padding in the wrong place, namely after the section ends. Fado would be able to handle it if the section sizes were computed including padding, but they aren't (and you can't expect Fado to know what the linker script feels like doing). Indeed, it has an experimental option to do use the alignment the elf files specify.

fig02:
Its fair to feel like this isnt appropriate for the repo. In my opinion this isn't something that can be feasibly maintained outside of the repo as a separate entity. Its too closely involved with the build system and would be a nightmare to maintain over time if its kept seperate. Its going to (likely) be something used by everyone who mods this codebase. I think its an exception worth making.

As far as AVOID_UB goes, I think theres a pretty clearly defined line here. Its goal is simply to maintain the same functionality of the original code while getting rid of undefined behavior. Anyone using this codebase for something beyond just reading it would want to make use of this. 
Its not like say, an enhancement for modding that changes how things work. I dont think it can really escalate beyond what it currently is.

Dragorn421:
I'm not worried about readability of the repo as getting rid of the avoid_ub parts can be automated easily, it's more about the direction taken by the repo. It shifts the goal from just reverse-engineering into (admittedly very basic and straightforward) modding.
I know an implicit goal we have is to be modding friendly, but that doesn't mean including mods, which is what custom code is

We also really can't tell what may end up needing modded code to fix UB, be it required or eventually just "for safety" to address potential issues. This is where I'm worried about it slipping to more and more custom code

For example `z_bg_mjin.c` probably needs, for the same reasons as the link object symbols UB,
```c
#ifdef AVOID_UB
static void* sMedallionPlatformTexPointers = {
    gLightMedallionPlatformTex, gShadowMedallionPlatformTex, gFireMedallionPlatformTex,
    gWaterMedallionPlatformTex, gSpiritMedallionPlatformTex, gForestMedallionPlatformTex,
};
#endif

#ifdef AVOID_UB
        gSPSegment(POLY_OPA_DISP++, 0x08, SEGMENTED_TO_VIRTUAL(sMedallionPlatformTexPointers[thisx->params - 1]));
#else
        gSPSegment(POLY_OPA_DISP++, 0x08, SEGMENTED_TO_VIRTUAL(&D_06000000));
#endif
```
who can say if this should be included, and how much is left to fix?

tools/mkldscript.c
AngheloAlf: ## The problem

The alignment problem is related to the `_SEGMENTNAMESegmentCAT` symbols generated by the linker.
For example, when compiling with `mips64-gcc` and not applying the `ALIGN(0x10)` patch in the linker script, you get this in the map:
```
 build/src/overlays/actors/ovl_Object_Kankyo/z_object_kankyo.o(.data)
 .data          0x0000000080b72e6c       0x76 build/src/overlays/actors/ovl_Object_Kankyo/z_object_kankyo.o
 build/src/overlays/actors/ovl_Object_Kankyo/ovl_Object_Kankyo_reloc.o(.data)
 .data          0x0000000080b72ee2        0x0 build/src/overlays/actors/ovl_Object_Kankyo/ovl_Object_Kankyo_reloc.o
                0x0000000080b72ee2                _ovl_Object_KankyoSegmentDataEnd = .
                0x0000000000000076                _ovl_Object_KankyoSegmentDataSize = ABSOLUTE ((_ovl_Object_KankyoSegmentDataEnd - _ovl_Object_KankyoSegmentDataStart))
                0x0000000080b72ee2                _ovl_Object_KankyoSegmentRoDataStart = .
 build/src/overlays/actors/ovl_Object_Kankyo/z_object_kankyo.o(.rodata)
 *fill*         0x0000000080b72ee2        0x6 
 .rodata        0x0000000080b72ee8      0x1b8 build/src/overlays/actors/ovl_Object_Kankyo/z_object_kankyo.o
                0x0000000080b72ee8                Object_Kankyo_InitVars
 build/src/overlays/actors/ovl_Object_Kankyo/z_object_kankyo.o(.rodata.str1.4)
 build/src/overlays/actors/ovl_Object_Kankyo/z_object_kankyo.o(.rodata.cst4)

```
This causes `_ovl_Object_KankyoSegmentRoDataStart` to not be properly aligned (0x0000000080b72ee2). This is a problem because `relocation.c` expects word alignment on sections (it does pointer arithmetic and just dereferences the pointers without checking for them to be correctly aligned).
Also, that number is lying, because, as you can see, the first rodata file is at VRAM `0x0000000080b72ee8`, this causes the relocs to be "wrong" compared to the linked program, and `relocation.c` will use the wrong offsets for its calculations.

## Why this happens?

First, look at the size of the .data section of z_object_kankyo
```
 build/src/overlays/actors/ovl_Object_Kankyo/z_object_kankyo.o(.data)
 .data          0x0000000080b72e6c       0x76 build/src/overlays/actors/ovl_Object_Kankyo/z_object_kankyo.o
 ```
It has size 0x76, which is not word aligned. Weird, right?, let's take a look at `readelf` (`readelf -a  build/src/overlays/actors/ovl_Object_Kankyo/z_object_kankyo.o`)

```
Section Headers:
  [Nr] Name              Type            Addr     Off    Size   ES Flg Lk Inf Al
  // other sections...
  [ 3] .data             PROGBITS        00000000 002710 000076 00  WA  0   0  4
  // other sections...
  [11] .rodata           PROGBITS        00000000 002a18 0001b8 00   A  0   0  8
```

The object itself says the .data section has size 0x76, with a 4 byte alignment (last column). The starting address for the data is already 4-aligned (`0x0000000080b72e6c`), so the linker doesn't have to adjust anything. The address is just incremented to account for this size (ending up not word aligned).

Then, according to our linker script, the linker has to produce the `SegmentDataEnd` and `SegmentRoDataStart` symbols, just taking the "current" processed address.

After that, the rodata is processed. As you can see in the readelf output, it requires 8-byte aligment, and since the current address (`0x0000000080b72ee2`) is not 8-byte aligned, then it adds a `*fill*` of 6 bytes. This `*fill*` introduce an empty space between the `SegmentRoDataStart` and the real rodata start.

## Can it be fixed in the side of Fado?

Not really. Fado currently forces each section to be 0x10 aligned, but as Elliptic said, there's an experimental option to account for the alignment declared in the .o file which is being processed.

But, Fado relies on the `SegmentSECTIONSize` symbols to be correct, which are produced by the linker, so there isn't much Fado can do it this regard.

## Can it be fixed in the linker script (mkldscript)?

Maybe? I'm no linker script expert, but I think it would be hard to patch in the mkldscript.c itself. Mainly because the alignment is accounted by the linker _after_ the `SegmentSECTIONStart` symbol has been emitted, and I don't see any way to fix the alignment of the section, then emitting the symbol and then writing the sections themselves

Comment (Dragorn421) [reply tools/mkldscript.c:14]: :sunglasses: thanks for the explanation!

So in the comment in mkldscript, "use narrower alignment ... for sections" refers to alignment of the section sizes?

So for example the data section of `z_object_kankyo.o` in IDO is always padded to a 0x10-alignment? which means the linker has no further alignment to do since an align size added to an aligned address is an aligned address
And with gcc the data section is of unaligned/unpadded size so the linker has to introduce alignment on the resulting address

If that's right the mkldscript comment on the aligns may be fleshed out a bit about aligning size vs aligning addresses or something, it's too fresh for me to word something

I will also say, it looks more worthwhile to indeed just align everything, rather than try to get stuff to work with the linker-made padding

Comment (AngheloAlf) [reply tools/mkldscript.c:14]: > So in the comment in mkldscript, "use narrower alignment ... for sections" refers to alignment of the section sizes?

Yeah, pretty much.

> So for example the data section of z_object_kankyo.o in IDO is always padded to a 0x10-alignment?

Yes, IDO always aligns each section to 0x10. Actually, that's the basic way to find file splits in multi-file segments, like `code`, `boot` and some overlays.

> which means the linker has no further alignment to do since an align size added to an aligned address is an aligned address
> And with gcc the data section is of unaligned/unpadded size so the linker has to introduce alignment on the resulting address

Exactly

> I will also say, it looks more worthwhile to indeed just align everything, rather than try to get stuff to work with the linker-made padding

Yeah, that what every `ALIGN(0x10)` in every section is doing

File: src/gcc-fix/missing_gcc_functions.c (hunk 16 above)
Comment (fig02) [reply src/gcc-fix/missing_gcc_functions.c:103]: This change was made before gcc stuff was in its own file. So it was more important to isolate things then. But yeah I see no reason to go back personally, this being readable isnt too important

File: Makefile [hunk 17]
```diff
@@ -8,14 +8,31 @@ COMPARE ?= 1
 NON_MATCHING ?= 0
 # If ORIG_COMPILER is 1, compile with QEMU_IRIX and the original compiler
 ORIG_COMPILER ?= 0
+# If COMPILER is GCC, compile with GCC instead of IDO.
+COMPILER ?= ido
```
Dragorn421, line 12: very minor remark but make's comparison is case-sensitive so COMPILER=GCC wouldn't work, COMPILER=gcc would
```suggestion
# If COMPILER is gcc, compile with GCC instead of IDO.
COMPILER ?= ido
```
may want to use quotes too idk

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -452,9 +452,15 @@ s16 EnZf_FindNextPlatformAwayFromPlayer(Vec3f* pos, s16 curPlatform, s16 arg2, G
         }
     }
 
-    // These functions have no side effects, so these two calls do nothing
+    //! @bug `altNextPlatform` can be -1 in certain conditions and cause an out of bounds access.
+    //! Under normal conditions, this doesn't cause problems because the data before `sPlatformPositions`
+    //! is section padding between .text and .data, so 0 gets read as a float.
+    //! Other compilers can shift things and result in a crash. Defining `AVOID_UB` will remove these function calls
+    //! entirely as they don't have any effect (return values are not used).
```
Dragorn421, line 459: I'm nitpicking but it's not just that return values are not used, `Math_Vec3f_DistXYZ` also doesn't have side effects
At the very least
```suggestion
    //! entirely as they don't have any effect (and return values are not used).
```
(it's not because the return values are not used that they don't have any effect)

Dragorn421 (commented):
I want to be convinced this is a good thing for the repo but I'm just not, so if anyone has anything to add to what fig said feel free to share

Makefile
```diff
@@ -80,12 +107,22 @@ ELF2ROM    := tools/elf2rom
...
+
 ASFLAGS := -march=vr4300 -32 -Iinclude
-MIPS_VERSION := -mips2
 
-# we support Microsoft extensions such as anonymous structs, which the compiler does support but warns for their usage. Surpress the warnings with -woff.
-CFLAGS += -G 0 -non_shared -Xfullwarn -Xcpluscomm $(INC) -Wab,-r4300_mul -woff 649,838,712
+ifeq ($(COMPILER),gcc)
+  CFLAGS += -G 0 -nostdinc $(INC) -DAVOID_UB -march=vr4300 -mfix4300 -mabi=32 -mno-abicalls -mdivide-breaks -fno-zero-initialized-in-bss -fno-toplevel-reorder -ffreestanding -fno-common -fno-merge-constants -mno-explicit-relocs -mno-split-addresses $(CHECK_WARNINGS) -funsigned-char
+  MIPS_VERSION := -mips3
+else
```
AngheloAlf, line 121: Since the linker is explicitly adding debugging sections, shouldn't we add debugging flags when compiling with gcc?

fig02: feels like something to me the user should opt into (by adding it themsleves) if they want to make use of it.
The changes in the linker script simply say to include them if they are there, doesnt mean they have to be (imo)

AngheloAlf: According to the gcc manual, the default debugging level is 2. https://gcc.gnu.org/onlinedocs/gcc/Debugging-Options.html#Debugging-Options

But I don't really trust gcc documentation, so I did a few manual tests.

(Using `mips-linux-gnu-gcc` + `-S` (produce assembly instead of a .o file) + the flags described here )

no `-gN` behaves the same as `-g0`. `-g1` adds a bit of ~~bloat~~ debugging information. `-g2` goes nuts and `-g3` is a bit absurd...

Conclusion: the documentation is lying again.

AngheloAlf: > feels like something to me the user should opt into (by adding it themsleves) if they want to make use of it. The changes in the linker script simply say to include them if they are there, doesnt mean they have to be (imo)

Yeah, no worries. I just though the flag was intended to be used, but somebody forgot to add it.

fig02:
@Dragorn421 Do you take issue with the build system having gcc as an option, or with the AVOID_UB fixes specifically?

`AVOID_UB` is not happening *because* of gcc specifically, its something that has been planned for a while to allow for portability in general. Its just being implemented here for the first time with gcc as its required to stop certain things from crashing.

Being able to remove UB is helpful for anyone using the codebase for anything more than just reading it. It is only really a good thing imo, and not a subjective change, like say, including enhancements in the repo. It either is UB or it isnt. Theres not really a slippery slope for things to escalate here

Makefile
```diff
@@ -45,8 +62,18 @@ ifneq ($(shell type $(MIPS_BINUTILS_PREFIX)ld >/dev/null 2>/dev/null; echo $$?),
   $(error Please install or build $(MIPS_BINUTILS_PREFIX))
 endif
 
-CC       := tools/ido_recomp/$(DETECTED_OS)/7.1/cc
-CC_OLD   := tools/ido_recomp/$(DETECTED_OS)/5.3/cc
+# Detect compiler and set variables appropriately.
+ifeq ($(COMPILER),gcc)
+  CC       := $(MIPS_BINUTILS_PREFIX)gcc
+  CC_OLD   := $(CC)
```
Dragorn421, line 68:
```suggestion
```

spec
```diff
@@ -109,6 +109,9 @@ beginseg
     include "build/src/libultra/os/gethwintrroutine.o"
     include "build/asm/__osSetWatchLo.o"
     include "build/data/rsp_boot.text.o"
+#ifdef COMPILER_GCC
+    include "build/src/gcc-fix/missing_gcc_functions.o"
```
Dragorn421, line 113: huge nitpick here but this is inconsistent `-`/`_` usage, may as well fix it before it's too late

src/boot/z_std_dma.c
```diff
@@ -210,6 +210,9 @@ const char* DmaMgr_GetFileNameImpl(u32 vrom) {
     }
     //! @bug Since there is no return, in case the file isn't found, the return value will be a pointer to the end
     // of gDmaDataTable
+#ifdef AVOID_UB
+    return (const char*)iter;
```
Dragorn421, line 214:
```suggestion
    return NULL;
```
we may as well have code that makes sense, since I wouldn't call casting iter to a string well-defined behavior

src/code/z_player_lib.c
```diff
@@ -654,7 +654,22 @@ u8 sEyeMouthIndexes[][2] = {
  * This allows the array to only contain the symbols for one file and have it apply to both. This is a problem for
  * shiftability, and changes will need to be made in the code to account for this in a modding scenario. The symbols
  * from adult Link's object are used here.
+ *
+ * Using AVOID_UB here will use a 2D array instead and properly use the child link pointers to allow for shifting.
  */
+#ifdef AVOID_UB
```
Dragorn421, line 660: Could we put avoid_ub things inside the if(n)defs, so hypothetical processing of the source could "get rid" of all avoid_ub related things
```suggestion
 */
#ifdef AVOID_UB
// Using AVOID_UB here will use a 2D array instead and properly use the child link pointers to allow for shifting.
```

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -452,9 +452,15 @@ s16 EnZf_FindNextPlatformAwayFromPlayer(Vec3f* pos, s16 curPlatform, s16 arg2, G
         }
     }
 
-    // These functions have no side effects, so these two calls do nothing
+    //! @bug `altNextPlatform` can be -1 in certain conditions and cause an out of bounds access.
+    //! Under normal conditions, this doesn't cause problems because the data before `sPlatformPositions`
+    //! is section padding between .text and .data, so 0 gets read as a float.
+    //! Other compilers can shift things and result in a crash. Defining `AVOID_UB` will remove these function calls
+    //! entirely as they don't have any effect (and return values are not used).
+#ifndef AVOID_UB
```
Dragorn421, line 460: as in the previous comment,
```suggestion
    //! Other compilers can shift things and result in a crash.
#ifndef AVOID_UB
    // Defining `AVOID_UB` will remove these function calls entirely as they don't have any effect (and return values are not used).
```
(may need a line break)

tools/mkldscript.c
Dragorn421: So to summarize your detailed explanation into the comment, what about

Note: \*SECTION ALIGNMENT\* Object files built with a compiler such as GCC can, by default, use narrower alignment for sections size, compared to IDO padding sections to a 0x10-aligned size.
To properly generate relocations relative to section starts, sections currently need to be aligned explicitly (to 0x10 currently, a narrower alignment might work), otherwise the linker does implicit alignment and inserts padding between the address indicated by section start symbols (such as `*SegmentRoDataStart`) and the actual aligned start of the section.
With IDO, the padding of sections to an aligned size makes the section start at aligned addresses out of the box, so the explicit alignment has no further effect.

AngheloAlf: Yeah, it sounds good to me. :+1:

AngheloAlf (commented):
I was wondering if it would be possible to disable `CC_CHECK` when building with gcc to avoid running two gcc per file

Makefile
```diff
@@ -172,6 +215,10 @@ build/src/code/%.o: CC := python3 tools/asm_processor/build.py $(CC) -- $(AS) $(
 build/src/overlays/%.o: CC := python3 tools/asm_processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
 
 build/assets/%.o: CC := python3 tools/asm_processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
+else
+build/src/libultra/libc/ll.o: OPTFLAGS := -Ofast
```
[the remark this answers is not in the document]
Thar0: `-Ofast` was just to avoid having to host even more functions in missing_gcc_functions.c

Makefile
```diff
@@ -98,7 +97,7 @@ INC        := -Iinclude -Isrc -Iassets -Ibuild -I.
 
 # Check code syntax with host compiler
 CHECK_WARNINGS := -Wall -Wextra -Wno-format-security -Wno-unknown-pragmas -Wno-unused-parameter -Wno-unused-variable -Wno-missing-braces -Wno-int-conversion
-CC_CHECK   := gcc -fno-builtin -fsyntax-only -fsigned-char -std=gnu90 -D _LANGUAGE_C -D NON_MATCHING $(INC) $(CHECK_WARNINGS)
+CC_CHECK   := gcc -fno-builtin -fsyntax-only -fsigned-char -std=gnu90 -D_LANGUAGE_C -DNON_MATCHING $(INC) $(CHECK_WARNINGS)
```
EllipticEllipsis, line 100:
```suggestion
```
Duplicated below.

fig02: whoops

AngheloAlf (approved):

EllipticEllipsis (approved):

src/boot/z_std_dma.c
EllipticEllipsis: Recasting a non-function pointer is pretty much always okay unless the alignment requirements are not the same: https://stackoverflow.com/questions/4810417/when-is-casting-between-pointer-types-not-undefined-behavior-in-c

*Dereferencing*, on the other hand...

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -452,9 +452,15 @@ s16 EnZf_FindNextPlatformAwayFromPlayer(Vec3f* pos, s16 curPlatform, s16 arg2, G
...
-    // These functions have no side effects, so these two calls do nothing
+    //! @bug `altNextPlatform` can be -1 in certain conditions and cause an out of bounds access.
+    //! Under normal conditions, this doesn't cause problems because the data before `sPlatformPositions`
+    //! is section padding between .text and .data, so 0 gets read as a float.
+#ifndef AVOID_UB
+    // Defining `AVOID_UB` will remove these function calls entirely as they don't 
+    // have any effect (and return values are not used).
     Math_Vec3f_DistXYZ(&player->actor.world.pos, &sPlatformPositions[nextPlatform]);
     Math_Vec3f_DistXYZ(&player->actor.world.pos, &sPlatformPositions[altNextPlatform]);
+#endif
```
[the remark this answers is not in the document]
fig02: Agreed on vanilla code going first.
Not sure on an else just for the comment though. I think the way it is now its fine, if you go through the trouble of removing all AVOID_UB blocks, you dont care about it anyway and dont need the explanation of it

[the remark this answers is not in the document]
Dragorn421: To be clear if one was to process the source to remove avoid_ub blocks, hypothetically
```c
#ifndef AVOID_UB
    // Defining `AVOID_UB` will remove these function calls entirely as they don't 
    // have any effect (and return values are not used).
    Math_Vec3f_DistXYZ(&player->actor.world.pos, &sPlatformPositions[nextPlatform]);
    Math_Vec3f_DistXYZ(&player->actor.world.pos, &sPlatformPositions[altNextPlatform]);
#endif
```
would become
```c
    // Defining `AVOID_UB` will remove these function calls entirely as they don't 
    // have any effect (and return values are not used).
    Math_Vec3f_DistXYZ(&player->actor.world.pos, &sPlatformPositions[nextPlatform]);
    Math_Vec3f_DistXYZ(&player->actor.world.pos, &sPlatformPositions[altNextPlatform]);
```
and indeed since in this scenario you wouldn't care about the avoid_ub code you would want the comments getting rid of, hence why I suggested instead having
```c
#ifndef AVOID_UB
    Math_Vec3f_DistXYZ(&player->actor.world.pos, &sPlatformPositions[nextPlatform]);
    Math_Vec3f_DistXYZ(&player->actor.world.pos, &sPlatformPositions[altNextPlatform]);
#else
    // Defining `AVOID_UB` will remove these function calls entirely as they don't 
    // have any effect (and return values are not used).
#endif
```
which would become
```c
    Math_Vec3f_DistXYZ(&player->actor.world.pos, &sPlatformPositions[nextPlatform]);
    Math_Vec3f_DistXYZ(&player->actor.world.pos, &sPlatformPositions[altNextPlatform]);
```
without the avoid_ub comments

in this case it's a bit weird because the avoid_ub block is empty code-wise, but that's also exactly the "fix" (having no code) so it kinda makes sense imo

it also clearly separate avoid_ub comments from the vanilla code, keeping the vanilla code sticked to the `@bug`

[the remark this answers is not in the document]
fig02: oh yea I see now, thanks. will change

src/overlays/actors/ovl_En_Tana/z_en_tana.c
```diff
@@ -33,6 +33,11 @@ const ActorInit En_Tana_InitVars = {
 static const char* sShelfTypes[] = {
     "木の棚", // "Wooden Shelves"
     "石の棚", // "Stone Shelves"
+#ifndef AVOID_UB
+// missing string for second stone type
+#else
```
fig02, line 38: before someone comments on this, i tried indenting it but the almighty formatter said no

src/boot/z_std_dma.c
Dragorn421: This is the only discussion left I'd like to see resolved, so the question is,
Does `return (const char*)iter;` lead to UB?

EllipticEllipsis: Hmm, it will read whatever is after the end of the nonzero stuff in the dma table. But the iteration terminating relies on the next thing being zero anyway. So the function actually forces that the dma table end in a row of zeros to avoid UB in the first place, so it will just point to bytes which will be interpreted as `'\0'`. So provided that we end the dma table in a row of zeros, the behaviour is well-defined. (You can always dereference non-null pointers to read chars, according to the C standard, according to that link I posted earlier.)

EllipticEllipsis: And it looks like we do not currently terminate the dma table with 0s, so there's UB elsewhere in this function already.

EllipticEllipsis: Oh, yes we do, I didn't spot it in the rather Byzantine way it's set up now
```c

DmaEntry gDmaDataTable[] = {
#include "tables/dmadata_table.h"
    { 0 },
};```

Comment (fig02) [reply src/boot/z_std_dma.c:215]: So if this list were to go all the way through without finding a match, it would end with `iter` pointing to this padding right? and would dereference to a string with only a null terminator? 
if so then I think we're fine

Comment (Dragorn421) [reply src/boot/z_std_dma.c:215]: it's sketch but I don't see a compiler doing something weird enough that it wouldn't be fine, so ok sure

but it's pretty weird to resort to "oh nice there's zeros here so ido's regalloc does something accidentally sensible" to justify this choice
the game does expect a zero vromEnd, but it's still reading vromStart as a string (the first (null) byte of it at least)

**Review by Dragorn421 — APPROVED** [review]

**Review by louist103 — APPROVED** [review]

File: src/boot/z_std_dma.c (hunk 22 above)
Comment (Thar0) [reply src/boot/z_std_dma.c:215]: Thinking about it, it's probably more appropriate to `return "";` to ensure it's always an empty string, preserving the behavior even more accurately.

**Review by fig02 — APPROVED** [review]

File: Makefile [hunk 29]
```diff
@@ -8,14 +8,31 @@ COMPARE ?= 1
...
+
+# ORIG_COMPILER cannot be combined with a non-IDO compiler. Check for this case and error out if found.
+ifneq ($(COMPILER),ido)
+  ifeq ($(ORIG_COMPILER),1)
+    $(error ORIG_COMPILER can only be used with the IDO compiler. Please check your Makefile variables and try again)
+  endif
+endif
+
+ifeq ($(COMPILER),gcc)
+  CPPFLAGS += -DCOMPILER_GCC
```
Roman971, line 25: Just a minor suggestion but `CFLAGS` could probably be set to define COMPILER_GCC here as well, for consistency and to avoid confusion in case someone tries to use it in the codebase

Roman971 (commented):
Sorry for the late review. I have a couple comments but otherwise this looks mostly good to me

src/boot/z_std_dma.c
```diff
@@ -208,8 +208,12 @@ const char* DmaMgr_GetFileNameImpl(u32 vrom) {
         iter++;
         name++;
     }
+#ifndef AVOID_UB
     //! @bug Since there is no return, in case the file isn't found, the return value will be a pointer to the end
     // of gDmaDataTable
+#else
+    return "";
+#endif
```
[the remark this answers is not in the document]
fig02: the intent was to keep the vanilla code always first, and in some cases that means an empty block goes first. Keeping the comments in the block also means that if someone is not interested in ub fixes, they can preprocess the codebase and remove the UB related explanations they arent interested in with it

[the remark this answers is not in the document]
fig02: I think this example helps show that intent more
```
#ifndef AVOID_UB
    Math_Vec3f_DistXYZ(&player->actor.world.pos, &sPlatformPositions[nextPlatform]);
    Math_Vec3f_DistXYZ(&player->actor.world.pos, &sPlatformPositions[altNextPlatform]);
#else
    // Defining `AVOID_UB` will remove these function calls entirely as they don't
    // have any effect (and return values are not used).
#endif
```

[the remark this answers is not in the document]
fig02: the suggestion was mainly prompted by dragorn, who is coming from the perspective of not even really wanting to introduce UB fixes into the codebase at all. 
Which while i dont agree with entirely, I do understand his perspective. And the point was to be able to remove UB related explanations that someone who is not interested in the fixes wouldnt want to see.

I would personally be ok with making these changes to only use `ifdef` or `ifndef` when it makes sense, but think if theres a situation that needs two blocks that the vanilla code should always be prioritized first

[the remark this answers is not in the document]
Dragorn421: about readability I do think using consistently
```c
#ifndef AVOID_UB

#else

#endif
```
would be more readable

then you always get vanilla code first (including nothing, no code)

and you less have to pay close attention to the n in `ifndef` (`ifdef` and `ifndef` being very similar at first glance...)

Currently it's fine as whatever because the code is not littered with avoid_ub (yet), but eventually there may end up being a lot more, and I do think it would be bothersome to have to pay attention to ifdef/ifndef when just reading some code

If I'm the only one on this issue though I can give in

src/overlays/actors/ovl_En_Zf/z_en_zf.c
```diff
@@ -452,9 +452,13 @@ s16 EnZf_FindNextPlatformAwayFromPlayer(Vec3f* pos, s16 curPlatform, s16 arg2, G
...
     }
 
-    // These functions have no side effects, so these two calls do nothing
+    //! @bug `altNextPlatform` can be -1 in certain conditions and cause an out of bounds access.
+    //! Under normal conditions, this doesn't cause problems because the data before `sPlatformPositions`
+    //! is section padding between .text and .data, so 0 gets read as a float.
+#ifndef AVOID_UB
     Math_Vec3f_DistXYZ(&player->actor.world.pos, &sPlatformPositions[nextPlatform]);
     Math_Vec3f_DistXYZ(&player->actor.world.pos, &sPlatformPositions[altNextPlatform]);
+#endif
```
Dragorn421, line 461: did you mean to get rid of the comment on this code being essentially useless

fig02: I did mean to get rid of the comment that speaks directly to the AVOID_UB stuff, but yeah I can add that other part of it back to the overall bug explanation

src/boot/z_std_dma.c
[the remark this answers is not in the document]
Roman971: I agree with prioritizing the original code first, but yeah I think we should stick to 1 block when possible, at least for now

Roman971 (approved):

PR #1059, opened 2021-12-04
Cleanup around some generic animation info structs

Dragorn421:
This PR focuses on 4 generic structures used to hold animation parameters, like the animation header, play speed, play mode...

One of them, the most complete, `AnimationInfo` has a helper for it in `code` `Animation_ChangeInfo`. Other structures usually use a helper that is inside each actor's overlay. Those helpers are often similar.

Two of the animation info structures were already in z64.h, I added two and named the 4 of them.
I ensured the helpers use a consistent prototype across actors, and named them similarly (eg `EnCs_ChangeAnim`).

The helpers work by indexing into an array of animation info: I also introduced enums (eg `ENDU_ANIM_2`) when there was none to use as indices instead of numeric constants.


src/code/z_actor.c
```diff
@@ -3856,19 +3856,19 @@ s16 func_80034DD4(Actor* actor, GlobalContext* globalCtx, s16 arg2, f32 arg3) {
     return arg2;
 }
 
-void func_80034EC0(SkelAnime* skelAnime, struct_80034EC0_Entry* animations, s32 index) {
+void Animation_ChangeInfo(SkelAnime* skelAnime, AnimationInfo* animationInfo, s32 index) {
```
fig02, line 3859: this name implies to me that its changing something about the info and not the animation itself. I dont have any better suggestions right now but will think about it more

AngheloAlf: What about `Animation_ChangeByInfo`?

EllipticEllipsis: `Using` or `From`'d probabably be better

Dragorn421: I went with `Animation_ChangeByInfo` for its shorter length

fig02 (approved):

louist103 (approved):

Roman971 (approved):

PR #1065, opened 2021-12-06
Name `ACTOR_FLAG_11` as `ACTOR_FLAG_ENKUSA_CUT`

Dragorn421:
My favourite actor flag

### DIFF
--- a/include/z64actor.h
+++ b/include/z64actor.h
@@ -110,7 +110,7 @@ typedef struct {
 #define ACTOR_FLAG_8 (1 << 8)
 #define ACTOR_FLAG_9 (1 << 9)
 #define ACTOR_FLAG_10 (1 << 10)
-#define ACTOR_FLAG_11 (1 << 11)
+#define ACTOR_FLAG_ENKUSA_CUT (1 << 11)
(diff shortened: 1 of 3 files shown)


fig02:
lol nice flag
we can probably go with bush instead of enkusa, at least imo

Dragorn421:
I was thinking it was best to wait until we give the actors proper names, since there are other "bush" actors like [Wood02](https://github.com/zeldaret/oot/blob/master/src/overlays/actors/ovl_En_Wood02/z_en_wood02.c#L4) or [Hana](https://github.com/zeldaret/oot/blob/master/src/overlays/actors/ovl_Obj_Hana/z_obj_hana.c#L62)

fig02:
I guess, but I think the "cut" gives you the context you need to know. I would also be ok with "grass", given thats what kusa translates to.
I also think this flag was probably not specifically meant for this one actor and probably had a larger meaning. But of course we cant really tell from the information we have

Roman971 (approved):
Either BUSH or the original actor name seem fine to me

fig02 (approved):
would prefer bush, but if the change will be made when we change actor names anyway then I wont push the issue

PR #1069, opened 2021-12-06
Added information about reg. 8011D3F0

svenar-nl:
This PR documents information about register 8011D3F0.
If 8011D3F0 is greater than 0, it counts back to 0 with 10ms interval
Vertical camera borders are visible & HUD elements are hidden when 8011D3F0 is greater than 0.

OOTDC-8011D3F0

### DIFF
--- a/src/code/z_camera.c
+++ b/src/code/z_camera.c
@@ -6851,7 +6851,7 @@ void Camera_Init(Camera* camera, View* view, CollisionContext* colCtx, GlobalCon
     camera->atLERPStepScale = 1;
     sCameraInterfaceFlags = 0xFF00;
     sDbgModeIdx = -1;
-    D_8011D3F0 = 3;
+    sCameraVBorderTimerCountdown = 3;
     osSyncPrintf(VT_FGCOL(BLUE) "camera: initialize --- " VT_RST " UID %d\n", camera->uid);
 }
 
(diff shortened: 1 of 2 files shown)


fig02:
I think we'll probably want to give this a name that closer fits what its intended for. Just looking at its usage, its only set in camera_init and not really used again. So it seems like its meant to be a initial timer sort of thing as the camera is starting up.

contextual documentation like this is why its best for someone documenting the system as a whole name these things, as opposed to just naming one thing by what it does from observation. imo at least

Dragorn421:
The reason for doing batch symbol documentation (like, whole-system) is that sometimes symbols can have nearly identical effects and their names should reflect the small differences

So when naming a single symbol (especially one that isn't core to the system and widely used), it may be that the name chosen for it is actually more appropriate for a different symbol

The resulting situation can be annoying in a further documentation pass based on a larger understanding of the system, which may need to rename the symbol after maybe figuring out it isn't quite accurate

In particular with camera it's very hard to gauge visually what is happening and I don't think this is the "root control" for enabling this behavior of no-hud + borders.

fig02:
Im going to close this for now, as this variable will get named in a larger camera documentation pass.
If you want to get more involved with the documentation process, I do encourage you to join our discord. It would be for the best if documentation was done at the system-level instead of variable/function level. So we can discuss those kinds of details in the discord

PR #1073, opened 2021-12-07
Cleanup symbols 7

Dragorn421:
#1071 continued and last of the series

Also handles objects with unaligned textures


fig02 (approved):
we really just need to fix the texture problem, but this can work for now

Roman971 (approved):

PR #1074, opened 2021-12-08
Extract z_select_static

louist103:


AngheloAlf (approved):

include/ultra64/gbi.h
```diff
@@ -10,7 +10,9 @@
 /* Private macro to wrap other macros in do {...} while (0) */
 #define _DW(macro) do {macro} while (0)
 
-#define F3DEX_GBI_2
+#ifndef F3DEX_GBI
+ #define F3DEX_GBI_2
+#endif
 
 #ifdef    F3DEX_GBI_2
 # ifndef  F3DEX_GBI
```
Dragorn421, line 18: you can remove this part defining `F3DEX_GBI` by default since you added `F3DEX_GBI_2` defined by default

assets/xml/misc/z_select_static.xml
```diff
@@ -0,0 +1,42 @@
+<Root>
+    <!-- OOTSW97 is currently being used here because ZAPD can't manually select f3dex. A fix for this is in a ZAPD PR. When it is merged into OOT this updated XML can be used here
+    https://gist.github.com/louist103/75d7a58f9890652d02feb6702bc4b2e0-->
```
Dragorn421, line 3: should we just wait for ZAPD to merge the fix and be updated in oot then, since this PR has no other content

fig02: zap has been updated in a recent pr so this is probably fixable now? didnt keep up if the fix was in that round of changes though

louist103: ZAPD has been updated but its not in master yet.  I would like to see a few more PRs for it get in before bringing it in since this PR isn't too important.

fig02: elliptics mac pr brought in updates, which is what im thinking of. that isnt merged yet but it will be soon

louist103: Looks like his PR has that.  When it gets merged ill update this.

include/ultra64/gbi.h
Dragorn421: ^

louist103: I cant get it to even build without this added here. If anyone else has a solution that is cleaner I would be willing to use it instead.

Dragorn421: my bad, I always assumed it was one or the other but it's actually "incremental" I guess, like `F3DEX_GBI_2` implies `F3DEX_GBI`

louist103: I mean there probably is a better way. I just dont know what it is.

louist103:
Waiting for #1086

fig02:
with zap updates in i believe we can make the changes we wanted with this

louist103:
> with zap updates in i believe we can make the changes we wanted with this

Updated

fig02 (approved):

Roman971 (approved):

PR #1075, opened 2021-12-08
Changed progress.py csv output to be compatible with new website

EllipticEllipsis:
For the new website, we want project progress csvs to be in the same format as far as possible, with pairs of columns indicating current decompiled total and current decompilable total in each category tracked. For OoT the change is very minor: just add `src` and `total` after the hash.

PR marked as draft for now since the csv itself will need to be changed, which will be done in sync with deploying the new site.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -39,9 +39,9 @@ pipeline {
             }
             steps {
                 sh 'mkdir reports'
-                sh 'python3 progress.py csv >> reports/progress.csv'
-                sh 'python3 progress.py csv -m >> reports/progress_matching.csv'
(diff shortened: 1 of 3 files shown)


ethteck:
Version should be incremented to 2 for this. The idea behind that column was that you could write separate parsing rules for each version

EllipticEllipsis:
Oh, it's the csv version, not the game version? I'll change that.

fig02 (approved):

Dragorn421 (approved):

ethteck (approved):

ethteck:
bypassing @Roman971 's review on this to ensure a smooth site transition to the new one

PR #1079, opened 2021-12-11
Document the lens system

louist103:
Not complete yet but I wanted to have eyes on it.
Thanks to Tharo and Dragorn for helping with the RCP commands.


Dragorn421 (approved):

src/code/z_actor.c
```diff
@@ -2258,27 +2275,44 @@ void func_8003115C(GlobalContext* globalCtx, s32 numInvisibleActors, Actor** inv
...
                             G_TD_CLAMP | G_TP_NONE | G_CYC_1CYCLE | G_PM_NPRIMITIVE,
                         G_AC_THRESHOLD | G_ZS_PRIM | Z_UPD | G_RM_CLD_SURF | G_RM_CLD_SURF2);
+        // makes the drawn image the color of the primcolor, red
         gDPSetCombineMode(POLY_XLU_DISP++, G_CC_MODULATEIA_PRIM, G_CC_MODULATEIA_PRIM);
         gDPSetPrimColor(POLY_XLU_DISP++, 0, 0, 255, 0, 0, 255);
+        // the z-buffer will later only allow drawing inside the lens circle
     } else {
+        // setup to update the zbuffer (Z_UPD)
+        // but leave the framebuffer untouched:
+        // `G_BL_CLR_BL, G_BL_0, G_BL_CLR_MEM, G_BL_1MA` sets the blender to leave the framebuffer untouched
```
fig02, line 2292: Im still not very much a fan of the very verbose docs of what the gfx stuff is doing. feels like something better explained outside of the code. Or if we must have it in the code, something a bit more brief.

I understand that this is complex and that the detail is probably desired. Which is why external documentation seems more appropriate in this case to me

fig02: everything else in the pr seems fine to me though

louist103: I think it's good to keep until we get those more extensive docs in either the GBI file or another central place. Cause yeah the SDK docs are nice but there aren't really examples in them.

fig02: maybe to some extent, but we can scale back the verbosity a bit. For example i dont think its necessary to comment on a prim color macro saying its red.

Dragorn421: https://github.com/louist103/oot/pull/8

fig02: with the changes in the pr this would be ready to approve from me

fig02 (approved):

src/code/z_actor.c
```diff
@@ -2245,20 +2245,37 @@ void func_80030ED8(Actor* actor) {
...
     gDPPipeSync(POLY_XLU_DISP++);
 
     CLOSE_DISPS(gfxCtx, "../z_actor.c", 6183);
 }
 
-void func_8003115C(GlobalContext* globalCtx, s32 numInvisibleActors, Actor** invisibleActors) {
+#undef LENS_MASK_WIDTH
+#undef LENS_MASK_HEIGHT
+#undef LENS_MASK_OFFSET_S
+#undef LENS_MASK_OFFSET_T
```
Roman971, line 2276: These undefs are unnecessary imo, and I think we could just omit them in general

Roman971 (commented):
Looks good to me overall, I just have a small comment

fig02 (approved):

Roman971 (approved):

PR #1080, opened 2021-12-12
Remove unnecessary recomp error files and update tools/.gitignore

MegaMech:
Recomp/ contained err.english.cc files. As far as I know these serve no purpose.


ethteck:
These aren't useless. They contain error message strings that are used by the compiler. Otherwise, you get error codes with no words

PR #1081, opened 2021-12-12
Defines for player state flags

Dragorn421:
72 new unnamed constants!

These defines are used with the bitmasks in the `Player` struct `u32 stateFlags1`, `u32 stateFlags2` and `u8 stateFlags3`

Every single bit is used, unlike actor flags.

--------

The prefix is up for debate. The flag names are long as `PLAYER_STATE_FLAG_1_0` and it would be even worse when naming the flags `PLAYER_STATE_FLAG_1_WHATEVER_THIS_DOES`

One suggestion is `PL_STATE1_WHATEVER_THIS_DOES`, but all player-related constants have been named `PLAYER_*` up until now.

Why not `PLAYER_SF1_WHATEVER_THIS_DOES`? (`SF` = State Flag)
We already have `PLAYER_AP_*` for `PlayerActionParam` so acronyms wouldn't be new

EDIT: changed the PR to `PLAYER_STATE1_WHATEVER_THIS_DOES`


fig02:
After sitting on this some more, Im kinda thinking I would be ok with just `PLAYER_STATE_`
Have a feeling we will quickly realize that there isnt any distinction between 1/2/3, but dont have much definitive proof yet. And dont think we really have to say in the name of the define that they are flags. The context of the variable name its being compared too + the fact that they are being operated on with & or | will tell you that much.

Only downside I see against it currently is there could end up being other things within player that are called "state". But with my current knowledge of player I dont think that will be an issue. Who knows what will happen as we document more tho

Im also much more open to the idea of an acronym now, SF could work if people dont agree with the above. But I feel like `PLAYER_STATE_` will just be cleaner and easier to understand in general

fig02 (approved):

fig02:
(conversation was continued in discord, im fine with what this pr has currently)

Roman971 (approved):

fig02:
will be good to merge after conflict is resolved

PR #1082, opened 2021-12-13
Document Obj_Switch subtypes

Yanis42:
- ~~Normal~~ ~~Default~~ Once = the default subtype of a type of switch, it's the yellow switches that doesn't do anything special
- Toggle = can be toggled on or off, without restrictions
- "~~Reset~~ Release when not stood on" = you have to stay on the switch in order to keep it activated
- ~~Tied~~ Sync (crystal only) = when you trigger another switch with the same switch flag it'll sync switches that have this subtype, ~~hence the name "tied" because it's tied to the switch flag (I wasn't sure how to name this one)~~
- There's a subtype called ``OBJSWITCH_SUBTYPE_RESET_INVERTED``, its behavior is inverted compared to the other reset subtype; when it's down it'll clear the switch flag, when it's up it'll set the switch flag

~~"Reset" subtype has two default states, subtype 2 is a blue switch being off by default meanwhile subtype 3 is on by default~~

~~Also I updated the comment for ``unk_17F``, it seems to be used to check if the player hit the switch, I didn't named it because I'm not 100% sure for this one so I'll leave it to AC experts~~

Regarding how I named the subtypes, for most of them I took a obj_switch parameter from the game and I did the maths using the code, I read more of the it too to make sure I wasn't going the wrong way (and I used the practice rom to try subtypes I had not figured out yet)

I'm not really good at naming things so, if you have any suggestions go on. ~~I'm not sure that keeping ``OBJSWITCH_`` is a good thing though, having ``SUBTYPE_FLOOR_NORMAL`` is enough in my opinion~~ (with every enums merged into one I'm ok with those names)

### DIFF
--- a/src/overlays/actors/ovl_Obj_Switch/z_obj_switch.c
+++ b/src/overlays/actors/ovl_Obj_Switch/z_obj_switch.c
@@ -97,7 +97,7 @@ static ColliderTrisInit sRustyFloorTrisInit = {
         OC2_NONE,
         COLSHAPE_TRIS,
     },
-    2,
+    ARRAY_COUNT(D_80B9EC34),
     D_80B9EC34,
 };
 
@@ -135,7 +135,7 @@ static ColliderTrisInit trisColliderEye = {
         OC2_NONE,
         COLSHAPE_TRIS,
     },
-    2,
+    ARRAY_COUNT(D_80B9ECBC),
     D_80B9ECBC,
 };
 
@@ -162,7 +162,7 @@ static ColliderJntSphInit sCyrstalJntSphereInit = {
         OC2_TYPE_2,
         COLSHAPE_JNTSPH,
     },
-    1,
+    ARRAY_COUNT(D_80B9ED44),
     D_80B9ED44,
 };
 
@@ -382,25 +382,25 @@ void ObjSwitch_FloorUp(ObjSwitch* this, GlobalContext* globalCtx) {
         }
     } else {
         switch ((this->dyna.actor.params >> 4 & 7)) {
-            case OBJSWITCH_SUBTYPE_FLOOR_NORMAL:
+            case OBJSWITCH_SUBTYPE_DEFAULT:
                 if (func_8004356C(&this->dyna)) {
                     ObjSwitch_FloorPressInit(this);
                     ObjSwitch_SetOn(this, globalCtx);
                 }
                 break;
-            case OBJSWITCH_SUBTYPE_FLOOR_TOGGLE:
+            case OBJSWITCH_SUBTYPE_TOGGLE:
                 if ((this->dyna.unk_160 & 2) && !(this->unk_17F & 2)) {
                     ObjSwitch_FloorPressInit(this);
                     ObjSwitch_SetOn(this, globalCtx);
                 }
                 break;
-            case OBJSWITCH_SUBTYPE_FLOOR_RESET_OFF:
+            case OBJSWITCH_SUBTYPE_RESET_OFF:
                 if (func_800435B4(&this->dyna)) {
                     ObjSwitch_FloorPressInit(this);
                     ObjSwitch_SetOn(this, globalCtx);
                 }
                 break;
-            case OBJSWITCH_SUBTYPE_FLOOR_RESET_ON:
+            case OBJSWITCH_SUBTYPE_RESET_ON:
                 if (func_800435B4(&this->dyna)) {
                     ObjSwitch_FloorPressInit(this);
                     ObjSwitch_SetOff(this, globalCtx);
@@ -416,7 +416,7 @@ void ObjSwitch_FloorPressInit(ObjSwitch* this) {
 }
 
 void ObjSwitch_FloorPress(ObjSwitch* this, GlobalContext* globalCtx) {
-    if ((this->dyna.actor.params >> 4 & 7) == OBJSWITCH_SUBTYPE_FLOOR_RESET_ON || !this->cooldownOn ||
+    if ((this->dyna.actor.params >> 4 & 7) == OBJSWITCH_SUBTYPE_RESET_ON || !this->cooldownOn ||
         func_8005B198() == this->dyna.actor.category || this->cooldownTimer <= 0) {
         this->dyna.actor.scale.y -= 99.0f / 2000.0f;
         if (this->dyna.actor.scale.y <= 33.0f / 2000.0f) {
@@ -435,23 +435,23 @@ void ObjSwitch_FloorDownInit(ObjSwitch* this) {
 
 void ObjSwitch_FloorDown(ObjSwitch* this, GlobalContext* globalCtx) {
     switch ((this->dyna.actor.params >> 4 & 7)) {
-        case OBJSWITCH_SUBTYPE_FLOOR_NORMAL:
+        case OBJSWITCH_SUBTYPE_DEFAULT:
             if (!Flags_GetSwitch(globalCtx, (this->dyna.actor.params >> 8 & 0x3F))) {
                 ObjSwitch_FloorReleaseInit(this);
             }
             break;
-        case OBJSWITCH_SUBTYPE_FLOOR_TOGGLE:
+        case OBJSWITCH_SUBTYPE_TOGGLE:
             if ((this->dyna.unk_160 & 2) && !(this->unk_17F & 2)) {
                 ObjSwitch_FloorReleaseInit(this);
                 ObjSwitch_SetOff(this, globalCtx);
             }
             break;
-        case OBJSWITCH_SUBTYPE_FLOOR_RESET_OFF:
-        case OBJSWITCH_SUBTYPE_FLOOR_RESET_ON:
+        case OBJSWITCH_SUBTYPE_RESET_OFF:
+        case OBJSWITCH_SUBTYPE_RESET_ON:
             if (!func_800435B4(&this->dyna) && !Player_InCsMode(globalCtx)) {
                 if (this->releaseTimer <= 0) {
                     ObjSwitch_FloorReleaseInit(this);
-                    if ((this->dyna.actor.params >> 4 & 7) == OBJSWITCH_SUBTYPE_FLOOR_RESET_OFF) {
+                    if ((this->dyna.actor.params >> 4 & 7) == OBJSWITCH_SUBTYPE_RESET_OFF) {
                         ObjSwitch_SetOff(this, globalCtx);
                     } else {
                         ObjSwitch_SetOn(this, globalCtx);
@@ -472,13 +472,13 @@ void ObjSwitch_FloorReleaseInit(ObjSwitch* this) {
 void ObjSwitch_FloorRelease(ObjSwitch* this, GlobalContext* globalCtx) {
     s16 subType = (this->dyna.actor.params >> 4 & 7);
 
-    if (((subType != OBJSWITCH_SUBTYPE_FLOOR_TOGGLE) && (subType != OBJSWITCH_SUBTYPE_FLOOR_RESET_ON)) || !this->cooldownOn ||
+    if (((subType != OBJSWITCH_SUBTYPE_TOGGLE) && (subType != OBJSWITCH_SUBTYPE_RESET_ON)) || !this->cooldownOn ||
         func_8005B198() == this->dyna.actor.category || this->cooldownTimer <= 0) {
         this->dyna.actor.scale.y += 99.0f / 2000.0f;
         if (this->dyna.actor.scale.y >= 33.0f / 200.0f) {
             ObjSwitch_FloorUpInit(this);
             Audio_PlayActorSound2(&this->dyna.actor, NA_SE_EV_FOOT_SWITCH);
-            if (subType == OBJSWITCH_SUBTYPE_FLOOR_TOGGLE) {
+            if (subType == OBJSWITCH_SUBTYPE_TOGGLE) {
                 func_800AA000(this->dyna.actor.xyzDistToPlayerSq, 120, 20, 10);
             }
         }
@@ -548,13 +548,13 @@ void ObjSwitch_EyeClosedInit(ObjSwitch* this) {
 
 void ObjSwitch_EyeClosed(ObjSwitch* this, GlobalContext* globalCtx) {
     switch ((this->dyna.actor.params >> 4 & 7)) {
-        case OBJSWITCH_SUBTYPE_EYE_NORMAL:
+        case OBJSWITCH_SUBTYPE_DEFAULT:
             if (!Flags_GetSwitch(globalCtx, (this->dyna.actor.params >> 8 & 0x3F))) {
                 ObjSwitch_EyeOpeningInit(this);
                 this->dyna.actor.params &= ~0x80;
             }
             break;
-        case OBJSWITCH_SUBTYPE_EYE_TOGGLE:
+        case OBJSWITCH_SUBTYPE_TOGGLE:
             if (ObjSwitch_EyeIsHit(this) || (this->dyna.actor.params >> 7 & 1)) {
                 ObjSwitch_EyeOpeningInit(this);
                 ObjSwitch_SetOff(this, globalCtx);
@@ -570,7 +570,7 @@ void ObjSwitch_EyeOpeningInit(ObjSwitch* this) {
 }
 
 void ObjSwitch_EyeOpening(ObjSwitch* this, GlobalContext* globalCtx) {
-    if ((this->dyna.actor.params >> 4 & 7) != OBJSWITCH_SUBTYPE_EYE_TOGGLE || !this->cooldownOn ||
+    if ((this->dyna.actor.params >> 4 & 7) != OBJSWITCH_SUBTYPE_TOGGLE || !this->cooldownOn ||
         func_8005B198() == this->dyna.actor.category || this->cooldownTimer <= 0) {
         this->eyeTexIndex--;
         if (this->eyeTexIndex <= 0) {
@@ -590,14 +590,14 @@ void ObjSwitch_CrystalOffInit(ObjSwitch* this) {
 
 void ObjSwitch_CrystalOff(ObjSwitch* this, GlobalContext* globalCtx) {
     switch ((this->dyna.actor.params >> 4 & 7)) {
-        case OBJSWITCH_SUBTYPE_CRYSTAL_NORMAL:
+        case OBJSWITCH_SUBTYPE_DEFAULT:
             if ((this->jntSph.col.base.acFlags & AC_HIT) && this->disableAcTimer <= 0) {
                 this->disableAcTimer = 10;
                 ObjSwitch_SetOn(this, globalCtx);
                 ObjSwitch_CrystalTurnOnInit(this);
             }
             break;
-        case OBJSWITCH_SUBTYPE_CRYSTAL_TIED:
+        case OBJSWITCH_SUBTYPE_SYNC:
             if (((this->jntSph.col.base.acFlags & AC_HIT) && this->disableAcTimer <= 0) ||
                 Flags_GetSwitch(globalCtx, (this->dyna.actor.params >> 8 & 0x3F))) {
                 this->disableAcTimer = 10;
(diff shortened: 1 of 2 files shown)


src/overlays/actors/ovl_Obj_Switch/z_obj_switch.h
```diff
@@ -17,21 +17,21 @@ typedef enum {
...
 
 typedef enum {
-    /* 0 */ OBJSWITCH_SUBTYPE_FLOOR_0,
-    /* 1 */ OBJSWITCH_SUBTYPE_FLOOR_1,
-    /* 2 */ OBJSWITCH_SUBTYPE_FLOOR_2,
-    /* 3 */ OBJSWITCH_SUBTYPE_FLOOR_3
+    /* 0 */ OBJSWITCH_SUBTYPE_FLOOR_NORMAL,     // Normal Yellow Switch
+    /* 1 */ OBJSWITCH_SUBTYPE_FLOOR_TOGGLE,     // On/Off Red Switch
+    /* 2 */ OBJSWITCH_SUBTYPE_FLOOR_RESET_OFF,  // Reset when not stood on, blue switch (default state: off)
+    /* 3 */ OBJSWITCH_SUBTYPE_FLOOR_RESET_ON    // Reset when not stood on, blue switch (default state: on)
```
Dragorn421, line 23: does the default state being off/on mean "is the switch up or down", or "the switch flag being cleared or set"?

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.h
```diff
@@ -17,21 +17,21 @@ typedef enum {
...
+    /* 1 */ OBJSWITCH_SUBTYPE_EYE_TOGGLE        // On/Off Grey Eye
 } ObjSwitchSubTypeEye;
 
 typedef enum {
-    /* 0 */ OBJSWITCH_SUBTYPE_CRYSTAL_0,
-    /* 1 */ OBJSWITCH_SUBTYPE_CRYSTAL_1,
-    /* 4 */ OBJSWITCH_SUBTYPE_CRYSTAL_4 = 4
+    /* 0 */ OBJSWITCH_SUBTYPE_CRYSTAL_NORMAL,   // Normal
+    /* 1 */ OBJSWITCH_SUBTYPE_CRYSTAL_TOGGLE,   // On/Off Blue/Red Crystal
+    /* 4 */ OBJSWITCH_SUBTYPE_CRYSTAL_TIED = 4  // Tied to Switch Flag's State
```
Dragorn421, line 34: this subtype's behavior is pretty weird and I think it should be explained more
for example, what does hitting such a switch even do

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.h
```diff
@@ -58,7 +58,7 @@ typedef struct ObjSwitch {
     /* 0x017A */ u8 x2TexScroll;
     /* 0x017B */ u8 y2TexScroll;
     /* 0x017C */ Color_RGB8 crystalColor;
-    /* 0x017F */ u8 unk_17F; // used for different purposes between floor and eye switch
+    /* 0x017F */ u8 unk_17F; // seems to be used to check if a switch has been hit by an AC collider
```
Dragorn421, line 61: AC is Attack Checker so something's wrong here
I would probably just get rid of the comment entirely tbh

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.h
```diff
@@ -17,21 +17,21 @@ typedef enum {
 } ObjSwitchType;
 
 typedef enum {
-    /* 0 */ OBJSWITCH_SUBTYPE_FLOOR_0,
-    /* 1 */ OBJSWITCH_SUBTYPE_FLOOR_1,
-    /* 2 */ OBJSWITCH_SUBTYPE_FLOOR_2,
-    /* 3 */ OBJSWITCH_SUBTYPE_FLOOR_3
+    /* 0 */ OBJSWITCH_SUBTYPE_FLOOR_NORMAL,     // Normal Yellow Switch
```
Dragorn421, line 20: all "NORMAL" subtypes should probably be more explicit about the behavior, both in the name and comment
(what is "normal" supposed to be?)

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.h
Yanis42: It means "is the switch up or down", but if you use "On" by default it'll go back to off if there's nothing on the switch

One use case I see is if you want to have a floor switch that resets and it has a small crate on it by default to avoid a cutscene when you load the room

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.h
Yanis42: - hitting this subtype of switch will act as normal
- if you hit another switch that shares the same switch flag (only the flag matters) it will update with the switch flag (so flag set = crystal on, flag unset = crystal off)

However toggle switches won't be updated with the switch flag, if you trigger a switch that's not updated it will do nothing at all, not even a cutscene, the state will update though

One use case I see is using it with the toggle subtype, which can prevent the player from using the tied subtype but you could use it later, can be used with a gate for example

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.h
Yanis42: From what I understand this is used to detect when the player wants to trigger the switch, hence the AC check

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.h
Yanis42: Normal is the classic behavior of a switch, you turn it on and it'll set the switch flag, but that's just what it does

I wrote it that way because I thought people would understand what's each subtype easier, that's also why I mentioned the colors

Yanis42:
I thought I mentioned it but apparently I didn't, floor rusty switches shares the same subtypes as the normal floor switch

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.h
fig02: right, but you phrased your comment as "hit by an AC collider"
and AC collider (blue in gz) gets hit by AT colliders (red), so its not the AC collider doing the hitting

at a quick glance its also used for more than that though, it stores the value of a dynapoly variable we havent named. 
agreed on just removing the comment, it will get a proper name when we can name it

Yanis42: oh ok, I'll remove it then

mzxrules:
Not directly related to the PR, but I'm working on the MM variant of this actor and I discovered an error in the struct definition for this actor.

sCyrstalJntSphereInit only defines 1 ColliderJntSphElementInit, but ObjSwitchJntSph defines 2 items. Additionally, the JntSphereInit and TrisInit structs should be changed to use ARRAY_COUNT to define the number of elements they have.

mzxrules:
So I have a bit of an issue with this pr. When looking through the pr I came across line 250, within ObjSwitch_SetOn, which is called by all switch types:

    if (subType == 0 || subType == 4) {
        OnePointCutscene_AttentionSetSfx(globalCtx, &this->dyna.actor, NA_SE_SY_CORRECT_CHIME);
    } else {
        OnePointCutscene_AttentionSetSfx(globalCtx, &this->dyna.actor, NA_SE_SY_TRE_BOX_APPEAR);
    }

To me, this code combined with the gap in ObjSwitchSubTypeCrystal, and the fact that subtypes 0 and 1 are functionally the same suggests that the original subtype was and should still be a singular enumeration, with types 2, 3 being named to express they only work with floor type switches, and 4 only works with crystal switches.

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.h
mzxrules: Looking at the spawn data, there are only two instances of subtype 4 being used in vanilla and master quest. 

In vanilla, there's a room in the water temple where you have to jump into a pool with a fast moving current, equip irons, and shoot a crystal switch that's underwater inside a dragon's mouth to open up a timed gate to reach a key. A second crystal switch on the opposite side of the gate lets you open the gate again and swim back. 

In master quest, there is a room in spirit temple with two crystal switches that opens up a gate meant to drop a like like on your head. The second crystal allows you to open the gate if you've entered from above.

So the general purpose is to sync up the state of two crystal switches, rather than have them act independently.

Yanis42:
I think I figured out everything mzx said, tell me if there's anything wrong (I wasn't sure for the ARRAY_COUNT but I think it's correct)

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.h
```diff
@@ -17,26 +17,16 @@ typedef enum {
...
-
-typedef enum {
-    /* 0 */ OBJSWITCH_SUBTYPE_CRYSTAL_0,
-    /* 1 */ OBJSWITCH_SUBTYPE_CRYSTAL_1,
-    /* 4 */ OBJSWITCH_SUBTYPE_CRYSTAL_4 = 4
-} ObjSwitchSubTypeCrystal;
+    /* 0 */ OBJSWITCH_SUBTYPE_DEFAULT,      // Normal Yellow Switches
+    /* 1 */ OBJSWITCH_SUBTYPE_TOGGLE,       // Switches that can be turned on and off
+    /* 2 */ OBJSWITCH_SUBTYPE_RESET_OFF,    // Floor Types only, reset itself when not stood on (off by default)
+    /* 3 */ OBJSWITCH_SUBTYPE_RESET_ON,     // Floor Types only, reset itself when not stood on (on by default)
```
Dragorn421, line 23:
```suggestion
    /* 2 */ OBJSWITCH_SUBTYPE_RESET_OFF, // Floor Types only, released when not stood on (flag cleared when released)
    /* 3 */ OBJSWITCH_SUBTYPE_RESET_ON,  // Floor Types only, released when not stood on (flag set when released)
```

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.h
```diff
@@ -17,26 +17,16 @@ typedef enum {
...
-    /* 0 */ OBJSWITCH_SUBTYPE_EYE_0,
-    /* 1 */ OBJSWITCH_SUBTYPE_EYE_1
-} ObjSwitchSubTypeEye;
-
-typedef enum {
-    /* 0 */ OBJSWITCH_SUBTYPE_CRYSTAL_0,
-    /* 1 */ OBJSWITCH_SUBTYPE_CRYSTAL_1,
-    /* 4 */ OBJSWITCH_SUBTYPE_CRYSTAL_4 = 4
-} ObjSwitchSubTypeCrystal;
+    /* 0 */ OBJSWITCH_SUBTYPE_DEFAULT,      // Normal Yellow Switches
```
Dragorn421, line 20:
```suggestion
    /* 0 */ OBJSWITCH_SUBTYPE_ONCE, // Switches that can only be turned on
```
not sure about "once" tbh, but I really don't like "default"

Dragorn421 (changes requested):
`ObjSwitch_SetOn` and `ObjSwitch_SetOff` can also use the subtype enum now that it has been made common to all types

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.h
Yanis42: The flag is cleared for both reset subtypes when it's released, the only difference between them is the default state of the switch when you load the room

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.h
Yanis42: I wasn't sure about default too, I can't find the best word for this, I'll take "once" for now and if someone got a better name then I'll update it

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.h
Dragorn421: are you sure? I don't think the switch flag is set on init, and either `ObjSwitch_SetOn`/Off is called in `ObjSwitch_FloorUp`/Down depending on the subtype

Yanis42: The switch flag doesn't change with that subtype, only the state of the switch itself is affected

It's probably meant to be used to avoid watching a cutscene when you load a room that has a switch with the reset subtype with something on it like a small crate, I don't think it's used in the game though

Dragorn421: `ObjSwitch_SetOn` sets the switch flag and `ObjSwitch_SetOff` clears it

Yanis42: (Resolved in https://github.com/zeldaret/oot/pull/1082/commits/ab983bf2c1966fe506f3230aaf692cfd0d096a03)

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.h
```diff
@@ -47,7 +37,7 @@ typedef struct {
 typedef struct ObjSwitch {
     /* 0x0000 */ DynaPolyActor dyna;
     /* 0x0164 */ ObjSwitchActionFunc actionFunc;
-    /* 0x0168 */ s16 releaseTimer; // used for SUBTYPE_FLOOR_2 and SUBTYPE_FLOOR_3
+    /* 0x0168 */ s16 releaseTimer; // used for SUBTYPE_RESET_OFF and SUBTYPE_RESET_ON
```
Dragorn421, line 40: :eyes:

Yanis42: FailFish

Dragorn421 (approved):

fig02:
apologies for the review being very delayed-- holidays were busy and getting through other prs as well.
Will finally take a look at this soon

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.c
```diff
@@ -322,6 +322,8 @@ void ObjSwitch_Init(Actor* thisx, GlobalContext* globalCtx) {
     if (this->dyna.actor.params >> 7 & 1) {
         ObjSwitch_EyeFrozenInit(this);
     } else if (type == OBJSWITCH_TYPE_FLOOR || type == OBJSWITCH_TYPE_FLOOR_RUSTY) {
+        //! @bug if the subtype is OBJSWITCH_SUBTYPE_RESET_INVERTED, the switch should be Down or Up according to
+        //! `!switchFlagSet` instead of `switchFlagSet`
```
fig02, line 326: I dont understand this, wouldnt you want the switch to be down (pressed) if the switch flag is set? So isnt this right?

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.c
```diff
@@ -382,25 +384,25 @@ void ObjSwitch_FloorUp(ObjSwitch* this, GlobalContext* globalCtx) {
...
-            case OBJSWITCH_SUBTYPE_FLOOR_1:
+            case OBJSWITCH_SUBTYPE_TOGGLE:
                 if ((this->dyna.unk_160 & 2) && !(this->unk_17F & 2)) {
                     ObjSwitch_FloorPressInit(this);
                     ObjSwitch_SetOn(this, globalCtx);
                 }
                 break;
-            case OBJSWITCH_SUBTYPE_FLOOR_2:
+            case OBJSWITCH_SUBTYPE_RESET:
                 if (func_800435B4(&this->dyna)) {
```
fig02, line 400: "reset" is blue switches right? I think naming it HOLD, after fact that something needs to hold the switch down for it to work would possibly be more clear. but not sure if that works in all contexts

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.h
```diff
@@ -17,26 +17,21 @@ typedef enum {
...
+            // Switches that can only be turned on (On -> Flag Set)
+    /* 0 */ OBJSWITCH_SUBTYPE_ONCE,
+            // Switches that can be turned on and off (On -> Flag Set, Off -> Flag Cleared)
+    /* 1 */ OBJSWITCH_SUBTYPE_TOGGLE,
+            // Floor Types only, released when not stood on (Down -> Flag Set, Up -> Flag Cleared)
+    /* 2 */ OBJSWITCH_SUBTYPE_RESET,
+            // Floor Types only, unused, inverted Switch Flag behavior (Down -> Flag Cleared, Up -> Flag Set)
+    /* 3 */ OBJSWITCH_SUBTYPE_RESET_INVERTED,
+            // Crystal Types only, syncs with the Switch Flag (On -> Flag Set, Off -> Flag Cleared)
+    /* 4 */ OBJSWITCH_SUBTYPE_SYNC
```
fig02, line 29: I would put these comments beside the enum value
example
` /* 0 */ OBJSWITCH_SUBTYPE_ONCE,  // Switches that can only be turned on (On -> Flag Set)`

fig02 (commented):
again, sorry for the very late review

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.c
Dragorn421: The `OBJSWITCH_SUBTYPE_RESET_INVERTED` subtype specifically actually has the switch flag set when it's released and cleared when it's pressed, is the documentation on that in the actor header unclear?

This bug is specific to `OBJSWITCH_SUBTYPE_RESET_INVERTED` which isn't used in game iirc

fig02: i completely glazed over it being specific to `OBJSWITCH_SUBTYPE_RESET_INVERTED`. But I think we can still improve on the wording of the bug description to make it more clear

maybe something like "this condition does not account for `OBJSWITCH_SUBTYPE_RESET_INVERTED` which expects the switch flag to be set for the up position (and vice versa)"

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.c
Yanis42: yeah it's the blue switches, and yes HOLD seems better than RESET (and I don't think there's a context where it doesn't work since you actually need to have something to keep it held)

fig02: ok great. this works cause im currently documenting actor flags, and theres this one 
`#define ACTOR_FLAG_CAN_HOLD_SWITCHES (1 << 26)`
given to ruto, armos etc

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.h
Yanis42: I'm not 100% sure the formatting is good in the [commit](https://github.com/Yanis42/oot/commit/f8c15dab3225f89891591105825617571363c1e4) I did

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.c
Yanis42: That wording is better imo, I'll wait to have @Dragorn421's input on that idea though

Maybe something like that to be more precise? "This condition does not account for the subtype, that makes ``OBJSWITCH_SUBTYPE_HOLD_INVERTED`` have the wrong init since the switch flag is supposed to be set when the switch is released"

fig02: I think overall a high level explanation of the problem that gets the point across quickly is good enough. editing my previous idea even further:
"This condition does not account for `OBJSWITCH_SUBTYPE_HOLD_INVERTED` which expects the relationship between the switch position and the switch flag to be inverted"
that way we arent confuysing the reader about down vs up and set vs not set

Dragorn421 (approved):

fig02 (approved):

Roman971 (approved):

PR #1083, opened 2021-12-13
Finish documenting the fire arrow

jw2476:
Building on the work done by Fig previously, I named the last few unknowns in the fire arrow overlay. Mainly stuff to do with the trail that the arrow leaves behind

### DIFF
--- a/src/overlays/actors/ovl_Arrow_Fire/z_arrow_fire.c
+++ b/src/overlays/actors/ovl_Arrow_Fire/z_arrow_fire.c
@@ -45,12 +45,12 @@ void ArrowFire_Init(Actor* thisx, GlobalContext* globalCtx) {
 
     Actor_ProcessInitChain(&this->actor, sInitChain);
     this->radius = 0;
-    this->unk_158 = 1.0f;
+    this->trailLen = 1.0f;
     ArrowFire_SetupAction(this, ArrowFire_Charge);
     Actor_SetScale(&this->actor, 0.01f);
     this->alpha = 160;
     this->timer = 0;
-    this->unk_15C = 0.0f;
+    this->hitAnimItensity = 0.0f;
 }
 
 void ArrowFire_Destroy(Actor* thisx, GlobalContext* globalCtx) {
@@ -62,6 +62,8 @@ void ArrowFire_Charge(ArrowFire* this, GlobalContext* globalCtx) {
     EnArrow* arrow;
 
     arrow = (EnArrow*)this->actor.parent;
+
+    // if arrow has been put away, kill the fire effect
     if ((arrow == NULL) || (arrow->actor.update == NULL)) {
         Actor_Kill(&this->actor);
         return;
@@ -78,17 +80,18 @@ void ArrowFire_Charge(ArrowFire* this, GlobalContext* globalCtx) {
 
     // if arrow has no parent, player has fired the arrow
     if (arrow->actor.parent == NULL) {
-        this->unkPos = this->actor.world.pos;
+        this->trailEnd = this->actor.world.pos;
         this->radius = 10;
         ArrowFire_SetupAction(this, ArrowFire_Fly);
         this->alpha = 255;
     }
 }
 
-void func_80865ECC(Vec3f* unkPos, Vec3f* firePos, f32 scale) {
-    unkPos->x += ((firePos->x - unkPos->x) * scale);
-    unkPos->y += ((firePos->y - unkPos->y) * scale);
-    unkPos->z += ((firePos->z - unkPos->z) * scale);
+void shrinkTrail(Vec3f* trailEnd, Vec3f* currentPos, f32 scale) {
+    // shrinks trail by 5% (scale) of the difference between the current position and trail end
+    trailEnd->x += ((currentPos->x - trailEnd->x) * scale);
+    trailEnd->y += ((currentPos->y - trailEnd->y) * scale);
+    trailEnd->z += ((currentPos->z - trailEnd->z) * scale);
 }
 
 void ArrowFire_Hit(ArrowFire* this, GlobalContext* globalCtx) {
@@ -115,7 +118,7 @@ void ArrowFire_Hit(ArrowFire* this, GlobalContext* globalCtx) {
             offset = ((this->timer - 8) * (1.0f / 24.0f));
             offset = SQ(offset);
             this->radius = (((1.0f - offset) * scale) + 10.0f);
-            this->unk_158 += ((2.0f - this->unk_158) * 0.1f);
+            this->trailLen += ((2.0f - this->trailLen) * 0.1f);
             if (this->timer < 16) {
                 if (1) {}
                 this->alpha = ((this->timer * 0x23) - 0x118);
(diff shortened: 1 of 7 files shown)


jw2476:
Did the same for the ice arrow since the unknowns were the same

src/overlays/actors/ovl_Arrow_Fire/z_arrow_fire.c
```diff
@@ -212,10 +215,10 @@ void ArrowFire_Draw(Actor* thisx, GlobalContext* globalCtx2) {
         Matrix_Scale(0.01f, 0.01f, 0.01f, MTXMODE_APPLY);
 
         // Draw red effect over the screen when arrow hits
-        if (this->unk_15C > 0) {
+        if (this->hitAnimItensity > 0) {
```
hatal175, line 218: typo in member name

src/overlays/actors/ovl_Arrow_Fire/z_arrow_fire.c
```diff
@@ -78,17 +80,18 @@ void ArrowFire_Charge(ArrowFire* this, GlobalContext* globalCtx) {
...
         ArrowFire_SetupAction(this, ArrowFire_Fly);
         this->alpha = 255;
     }
 }
 
-void func_80865ECC(Vec3f* unkPos, Vec3f* firePos, f32 scale) {
-    unkPos->x += ((firePos->x - unkPos->x) * scale);
-    unkPos->y += ((firePos->y - unkPos->y) * scale);
-    unkPos->z += ((firePos->z - unkPos->z) * scale);
+void shrinkFireTrail(Vec3f* trailEnd, Vec3f* currentPos, f32 scale) {
```
Dragorn421, line 90: The way functions are named is typically `SystemName_FunctionName`

Here the "system name" is `ArrowFire` (this name comes from the strings inside the debug rom, the overlay file was named "ovl_arrow_fire" originally)
For example `ArrowFire_Hit` below

src/overlays/actors/ovl_Arrow_Fire/z_arrow_fire.c
```diff
@@ -78,17 +80,18 @@ void ArrowFire_Charge(ArrowFire* this, GlobalContext* globalCtx) {
...
+    // shrinks trail by 5% (scale) of the difference between the current position and trail end
```
Dragorn421, line 91: I don't think the comment should tell the typical value of `scale` since that comes from outside the scope of the function
(if you keep modding in mind, functions may be reused for whatever purpose)

I also feel like (may just be me though) that the function in general should be named after what it does (move vector A towards vector B by a fraction of the AB distance), instead of what it is used for
`Math_ApproachF` looks similar apart from the `step` parameter so you may want to reuse "Approach" here

src/overlays/actors/ovl_Arrow_Fire/z_arrow_fire.c
```diff
@@ -62,6 +62,8 @@ void ArrowFire_Charge(ArrowFire* this, GlobalContext* globalCtx) {
     EnArrow* arrow;
 
     arrow = (EnArrow*)this->actor.parent;
+
+    // if arrow has been put away, kill the fire effect
```
Dragorn421, line 66: The word "effect" usually refers to the effects system (particle system) of the game (like various kinds of dust, sparks...) (files in `src/overlays/effects/`, kinda similar to actors), so we try to avoid using it in other contexts

I would suggest if I may
```suggestion
    // if arrow has been put away, also remove the fire part
```

src/overlays/actors/ovl_Arrow_Fire/z_arrow_fire.c
```diff
@@ -156,12 +159,12 @@ void ArrowFire_Fly(ArrowFire* this, GlobalContext* globalCtx) {
     // copy position and rotation from arrow
     this->actor.world.pos = arrow->actor.world.pos;
     this->actor.shape.rot = arrow->actor.shape.rot;
-    distanceScaled = Math_Vec3f_DistXYZ(&this->unkPos, &this->actor.world.pos) * (1.0f / 24.0f);
-    this->unk_158 = distanceScaled;
+    distanceScaled = Math_Vec3f_DistXYZ(&this->trailEnd, &this->actor.world.pos) * (1.0f / 24.0f);
+    this->trailLen = distanceScaled;
```
Dragorn421, line 163: I think we can afford the full word :p
```suggestion
    this->trailLength = distanceScaled;
```

src/overlays/actors/ovl_Arrow_Fire/z_arrow_fire.c
```diff
@@ -212,10 +215,10 @@ void ArrowFire_Draw(Actor* thisx, GlobalContext* globalCtx2) {
         Matrix_Scale(0.01f, 0.01f, 0.01f, MTXMODE_APPLY);
 
         // Draw red effect over the screen when arrow hits
-        if (this->unk_15C > 0) {
+        if (this->hitAnimIntensity > 0) {
```
Dragorn421, line 218: Usually we use "anim" when there's an animated mesh involved (typically using the SkelAnime system)

It is very common that values change every frame (so are animated), so we just don't include that information in the name usually. (as an example among many, the `BgGanonOtyuka` struct has members which change every frame so are "animated", but none of them mentions it)

So at least I would remove "anim" so `hitIntensity`

src/overlays/actors/ovl_Arrow_Fire/z_arrow_fire.c
```diff
@@ -231,7 +234,7 @@ void ArrowFire_Draw(Actor* thisx, GlobalContext* globalCtx2) {
         } else {
             Matrix_Translate(0.0f, 1500.0f, 0.0f, MTXMODE_APPLY);
         }
-        Matrix_Scale(this->radius * 0.2f, this->unk_158 * 4.0f, this->radius * 0.2f, MTXMODE_APPLY);
+        Matrix_Scale(this->radius * 0.2f, this->trailLen * 4.0f, this->radius * 0.2f, MTXMODE_APPLY);
```
Dragorn421, line 237: Is this really a "trail" length? In my understanding it stretches the fire halo in the direction of shooting the arrow, and calling that a trail may bring confusion with the arrow's own trail (the white line, iirc)

src/overlays/actors/ovl_En_Arrow/z_en_arrow.h
```diff
@@ -40,5 +40,4 @@ typedef enum {
     /*  9  */ ARROW_SEED,
     /*  10 */ ARROW_NUT
 } ArrowType;
-
```
Dragorn421, line 43: I would leave that empty line in for consistency with all other headers (you may have not intended to commit it lol)

jw2476: No I didn't, vscode being nosy again

fig02:
will be closing this due to inactivity on the PR, as some suggestions have been pending for a while now.
No worries about it though, and if you decide you want to come back to work on this feel free to reopen it!

PR #1086, opened 2021-12-14
Building on Macs

EllipticEllipsis:
This updates ZAPD to build properly and warning-free on Macs
- zeldaret/ZAPD#210
- zeldaret/ZAPD#217

Also fixed the multithreading in `extract_assets.py`.

This *should* be enough for the repo to build correctly on a Mac.


tools/ZAPD/README.md
```diff
@@ -16,6 +16,14 @@ In a Debian/Ubuntu based environment, those could be installed with the followin
 sudo apt install libpng-dev
 ```
 
+On a Mac, you will need to install libpng with Homebrew or MacPorts; we currently only support Homebrew. You can run
```
jpburnett, line 19:
```suggestion
On a Mac, you will need to install libpng with Homebrew. You can run
```

Since we don't support MacPorts I don't think mentioning it is necessary?

jpburnett (approved):
Verified that this works on my ancient mac. If it could be beneficial, I installed binutils in a separate folder to not make it global, I could add instructions for that if it could be beneficial. 

Thanks for working on this!

extract_assets.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 import argparse, json, os, signal, time
-from multiprocessing import Pool, cpu_count, Event, Manager, ProcessError
+from multiprocessing import *
```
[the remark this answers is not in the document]
Roman971: It's not a major issue but I agree with avoiding wildcard imports if we can

tools/ZAPD/README.md
EllipticEllipsis: This would have to be changed in the ZAPD repo.

EllipticEllipsis:
Updated ZAPD to current. Includes the following:
- zeldaret/ZAPD#218
- zeldaret/ZAPD#214
- zeldaret/ZAPD#226
- zeldaret/ZAPD#220
- zeldaret/ZAPD#228
- zeldaret/ZAPD#223
- zeldaret/ZAPD#216

fig02 (approved):

extract_assets.py
```diff
@@ -16,13 +17,13 @@ def ExtractFile(xmlPath, outputPath, outputSourcePath):
         # Don't extract if another file wasn't extracted properly.
         return
 
-    execStr = "tools/ZAPD/ZAPD.out e -eh -i %s -b baserom/ -o %s -osf %s -gsf 1 -rconf tools/ZAPDConfigs/MqDbg/Config.xml" % (xmlPath, outputPath, outputSourcePath)
-    
+    execStr = f"tools/ZAPD/ZAPD.out e -eh -i {xmlPath} -b baserom/ -o {outputPath} -osf {outputSourcePath} -gsf 1 -rconf tools/ZAPDConfigs/MqDbg/Config.xml {ZAPDArgs}"
```
Dragorn421, line 20: I thought we were avoiding f-strings for mac compatibility (some macs do be having very old python)

extract_assets.py
```diff
@@ -71,12 +72,32 @@ def main():
     parser = argparse.ArgumentParser(description="baserom asset extractor")
     parser.add_argument("-s", "--single", help="asset path relative to assets/, e.g. objects/gameplay_keep")
     parser.add_argument("-f", "--force", help="Force the extraction of every xml instead of checking the touched ones.", action="store_true")
+    parser.add_argument("-j", "--jobs", help="Number of cpu cores to extract with.")
     parser.add_argument("-u", "--unaccounted", help="Enables ZAPD unaccounted detector warning system.", action="store_true")
+    parser.add_argument("-Z", help="Pass the argument on to ZAPD, e.g. `-ZWunaccounted` to warn about unaccounted blocks in XMLs. Each argument should be passed separately, *without* the leading dash.", metavar="ZAPD_ARG", action="append")
     args = parser.parse_args()
 
+    global ZAPDArgs;
```
Dragorn421, line 80: get those `;` out of here

Dragorn421 (changes requested):

extract_assets.py
EllipticEllipsis: ?

AngheloAlf: Python doesn't require the use of `;` at the end of the line (is actually bad practice to use them), so yeah, remove it

jpburnett: It's python, you don't need semicolons

extract_assets.py
jpburnett: I think that all macs have python 2.7 by default. But to build the repo one would need to get python 3.X anyways

EllipticEllipsis: And, this script uses python3 anyway.

Dragorn421: f-strings were added in python 3.6, I don't know anything about macs but it seems like some people with macs often have very old python versions
I'm fine to keep them, I don't plan on using outdated macs myself :p

EllipticEllipsis: ZAPD doesn't build on a Mac with pre C++17 support anyway.

extract_assets.py
```diff
@@ -1,7 +1,8 @@
 #!/usr/bin/env python3
 
-import argparse, json, os, signal, time
-from multiprocessing import Pool, cpu_count, Event, Manager, ProcessError
+import argparse, json, os, signal, time, colorama, multiprocessing
```
Dragorn421, line 3: colorama is currently a diff.py requirement only (so optional like diff.py usage), I don't see a point making it a requirement for building

(you probably didn't realize it needs to be installed separately, it isn't in the standard python library)

AngheloAlf: Then add the package as a dependency.
It isn't a big package and colored output is always nice

EllipticEllipsis: Yes, I'd argue that we should encourage colour output in our Python scripts where appropriate. And Windows (as usual) doesn't support the ANSI stuff except in a proper terminal, which is what colorama is supposed to get around.

Dragorn421: then you should add instructions in the readme on installing it

extract_assets.py
louist103, line 20: The unaccounted static flag should be used now since this PR brings that in.

louist103 (changes requested):

extract_assets.py
AngheloAlf: Did we end up on an agreement on that?

Dragorn421: There is the `--unaccounted` option already, should we make it the default?

louist103: nothing official but I personally dont see any reason to not have them be static. if someone really wants to look for them in the map file they can just turn the flag off.

AngheloAlf: In that case, I think `extract_assets.py` should have a flag to toggle that behaviour instead of hardcodding it on by default

EllipticEllipsis: Or, extract_assets.py in the makefile can pass the ZAPD flag directly since that's what the new `-Z` flag does.

requirements.txt
```diff
@@ -0,0 +1,2 @@
+libyaz0>=0.5
```
Dragorn421, line 1: I think we should only add the dependency when it actually gets used

Dragorn421 (changes requested):
Could you also update ZAPD again :sweat_smile: 
Latest has https://github.com/zeldaret/ZAPD/pull/215 (which I guess I'm waiting on in #1073)

README.md
```diff
@@ -90,6 +70,12 @@ sudo apt-get update
 sudo apt-get install git build-essential binutils-mips-linux-gnu python3 libpng-dev
 ```
 
+To install the Python dependencies simply run in a terminal:
+
+```bash
+python3 -m pip install -r requirements.txt
```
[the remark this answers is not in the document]
fig02: also not too much a fan of a new file in root for something like this. people are going to copy paste the command anyway regardless of if its all the packages or the requirements.txt line

extract_assets.py
```diff
@@ -71,12 +72,32 @@ def main():
...
+                print(f"{colorama.Fore.LIGHTRED_EX}error{colorama.Fore.RESET}: argument \"{z}\" starts with \"-\", which is not supported.", file=os.sys.stderr)
+                badZAPDArg = True
+            else:
+                args.Z[i] = "-" + z
+
+        if badZAPDArg:
+            exit(1)
+
+        ZAPDArgs = " ".join(args.Z)
+        print("Using extra ZAPD arguments: " + ZAPDArgs)
```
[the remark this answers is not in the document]
EllipticEllipsis: I think it's clearer to give each one a separate error message, but I might incorporate some other stuff from here.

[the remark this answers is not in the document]
EllipticEllipsis: Actually I think I find this harder to follow than what I had: those inline for things that Python does are dreadful for readability.

EllipticEllipsis:
Would we like to put the Docker stuff in a separate file too, and just have the WSL way in the main readme?

EllipticEllipsis:
Pulled latest ZAPD update, zeldaret/ZAPD#215

fig02 (approved):

README.md
```diff
@@ -39,37 +39,24 @@ It builds the following ROM:
 
 ## Installation
 
+We recommend using WSL on Windows, or native Linux, which the rest of this readme describes. We currently have instructions for
+
+* [Windows](#Windows), with and without WSL
+* [macOS](docs/BUILDING_MACOS.md)
+* [Linux](#Linux-Native-or-under-WSL--VM)), natively or using WSL / VM
```
Dragorn421, line 46:
```suggestion
* [Linux](#Linux-Native-or-under-WSL--VM), natively or using WSL / VM
```

Dragorn421 (changes requested):
installing python dependencies (colorama) isn't mentioned in macOS and cygwin build instructions
the dockerfile thankfully already installs colorama

docs/BUILDING_CYGWIN.md
```diff
@@ -40,3 +53,34 @@ Add the new binutils binaries to your system PATH:
...
+## 4. Install required Python packages
+
+To install the Python dependencies simply run in a terminal:
+
+```bash
+python3 -m pip install colorama
+```
+
+
+## 4. Continue with Linux instructions
```
Dragorn421, line 77:
```suggestion
## 5. Continue with Linux instructions
```

Dragorn421 (approved):

fig02 (approved):

louist103 (approved):

NEstelami (approved):

Roman971 (commented):
Sorry for the late review. I just have a few comments but this looks good to me overall

extract_assets.py
```diff
@@ -67,16 +68,35 @@ def initializeWorker(abort, unaccounted: bool, extractedAssetsTracker: dict, man
     globalExtractedAssetsTracker = extractedAssetsTracker
     globalManager = manager
 
+def processZAPDArgs(argsZ):
+    badZAPDArg = False
+    for z in argsZ:
+        if z[0] == '-':
+            print(f"{colorama.Fore.LIGHTRED_EX}error{colorama.Fore.RESET}: argument \"{z}\" starts with \"-\", which is not supported.", file=os.sys.stderr)
```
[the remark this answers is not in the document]
EllipticEllipsis: I think this is a change worth making: it's good to make errors and warnings stand out, and this is likely the most portable way to do so in Python. I would encourage changing the other scripts to do this too, although I'm not sure it's in scope here necessarily.

colorama is also required for the diff script already, so most of our current contributors will have it already anyway. We will also have to add other dependencies once we move on to compressed roms: I know @Thar0 would prefer to use libyaz0 over invoking our current yaz0 C script, for example.

docs/BUILDING_CYGWIN.md
```diff
@@ -1,4 +1,17 @@
-# Building mips-linux-binutils on Windows using Cygwin
+# Building with Cygwin
+
+**N.B.** These have not been tested recently, you may find the requirements have changed a bit.
```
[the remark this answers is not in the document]
EllipticEllipsis: I agree with the other other comment, but this one I feel makes more sense with "not [...] recently", in that if someone does test them, hopefully they will update the docs and alter this comment appropriately. I think "not [...] recently" is useful here firstly because it will remain true "by default" (unlike the "currently" one), and secondly because it conveys that at some point they did work, but you shouldn't necessarily expect it to work "out-of-the-box" now.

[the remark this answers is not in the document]
Dragorn421: You could reword it as something like "not been tested since spring 2021" (or whenever idk)
Like on Wikipedia, it's bad practice to use relative time for something that may last a "long" time so they encourage refering to actual dates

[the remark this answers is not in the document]
fig02: I think this is good in principle, but in this specific case i honestly have no idea when cygwin was last tested lol
when it was added to the repo maybe?

I dont have an opinion much here either way. I think "recently" here is more of a general statement and not so much an indicator of a specific time. at least not in a way that would be confusing. but yeah if theres a better way to word it then sure

fig02 (approved):

docs/BUILDING_CYGWIN.md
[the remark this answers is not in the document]
EllipticEllipsis: I understand the point, but I don't think it's actually that helpful to be more specific: it indicates that something may be broken, but whether it's a little time or a long time usually doesn't have a bearing on how difficult it might be to get it to work.

Roman971 (approved):

PR #1087, opened 2021-12-15
Document RCP functions

hatal175:
Documented 2 RCP functions:
1. The first prints the status for both RSP and RDP.
2. The second sets flags for both which I *think* is a reset state. Considering it is called at the main function end (probably never used) and from Sched_HandleReset, this makes sense to me.

First pull request, so comments are welcome, especially regarding naming.


fig02:
Hi, thanks for this.
This file doesnt look like it has too much going on and is mainly for logging. So Im wondering if rcp.c is too broad.
Granted im not familar with how much other code in the codebase, if any, is strictly dealing with the rcp. But its something to think about.

If we do decide that rcp is good enough, our function names follow the pattern of SystemName_FunctionName. So you would want to do something like `Rcp_Reset`

hatal175:
Renamed. I tend to agree regarding the broad description, maybe rcp_utils.c?
Either way, thanks for the review :)

Roman971:
I think `rcp.c` could easily be confused with `z_rcp.c`, so a different name is definitely needed imo. That being said, I don't know much about the rsp/rdp status stuff, so I can't say much about these names. The first function looks fine, but it seems like the second function is simply about halting/interrupting and not resetting completely, but that's just based on the status names themselves so I'm not sure

fig02 (approved):

Roman971 (approved):

PR #1088, opened 2021-12-15
Document local position/rotate constant

hatal175:


src/libultra/gu/position.c
```diff
@@ -5,13 +5,13 @@
  * Creates a rotation/parallel translation modeling matrix (floating point)
  */
 void guPositionF(f32 mf[4][4], f32 rot, f32 pitch, f32 yaw, f32 scale, f32 x, f32 y, f32 z) {
-    static f32 D_80134D00 = M_PI / 180.0;
+    static f32 degToGrad = M_PI / 180.0;
```
EllipticEllipsis, line 8: Do you mean `Rad`? Gradians are a completely different angle unit. It might also possibly be better to use `DEGF_TO_RADF(1)` from the macro in `z64math.h`, although it depends on how we decide to treat libultra.

Thar0:
I've already been going through the libultra directory as a whole cleaning up these kinds of unnamed locals, among other things such as introducing missed defines/enums. I think it would be best to avoid the libultra directory until that's done.

PR #1091, opened 2021-12-16
Document some z_actor.c functions

hatal175:


src/code/z_rcp.c
```diff
@@ -921,7 +921,7 @@ Gfx* func_800937E4(Gfx* gfx) {
     return gfx;
 }
 
-Gfx* func_80093808(Gfx* gfx) {
+Gfx* Gfx_SetupTitleCardDL(Gfx* gfx) {
     gSPDisplayList(gfx++, sSetupDL[0x34]);
     gDPSetColorDither(gfx++, G_CD_DISABLE);
```
fig02, line 926: with these generic setup display lists it will be better (but very hard) to name them after what the preset actually means/what its achieving. In this case its only used by one thing, which is the title card dl, but that wont always necessarily be the case (in a modding context). Its designed to be a generic dl that can be used for many things

putting all of what these setup dls do into words is going to be very hard though, and why all of them are still unnamed right now

fig02:
also, are you in the discord? Youre doing some good documentation work here, and it would be good to stay in contact with other people working on the project. As documentation picks up we'll probably want to coordinate on what we are working on so we dont duplicate work

src/code/z_rcp.c
hatal175: Figured as much, but thought the same since it was used exclusively there.

hatal175:
I am on the discord. Although it's not really the best format to keep track of who is working on what...

src/code/z_rcp.c
fig02: yeah it is. but i dont think its a display list that will work exclusively for title card textures. its just what the final game ended up using it for in this case

fig02: another way to phrase my thoughts: if this were really exclusive to title cards it probably wouldnt be a function in z_rcp.c and would just be inlined in the title card itself

fig02:
its not, and we will need to organize documentation better as this effort picks up. but we talk about important things in there and it would be good to have easier access to talk to contributors

src/code/z_rcp.c
hatal175: Removed

include/functions.h
```diff
@@ -369,7 +369,7 @@ void TitleCard_InitBossName(GlobalContext* globalCtx, TitleCardContext* titleCtx
                             u8 height);
 void TitleCard_InitPlaceName(GlobalContext* globalCtx, TitleCardContext* titleCtx, void* texture, s32 x, s32 y,
                              s32 width, s32 height, s32 delay);
-s32 func_8002D53C(GlobalContext* globalCtx, TitleCardContext* titleCtx);
+s32 TitleCard_ResetTimers(GlobalContext* globalCtx, TitleCardContext* titleCtx);
 void Actor_Kill(Actor* actor);
 void Actor_SetFocus(Actor* actor, f32 offset);
```
fig02, line 374: from looking at its usage, I think a more appropriate name is probably something like `TitleCard_Clear`, or something that makes it clear that it is making the title card go away. it does reset timers, but I think its a bit too literal

its specifically used whenever player tries to do anything involving a get item. the title cards and get items use the same memory for their assets, so player clears/hides the title cards out of view before pulling out something that uses a get item object

src/code/z_actor.c
```diff
@@ -655,7 +655,7 @@ void Flags_SetCollectible(GlobalContext* globalCtx, s32 flag) {
     }
 }
 
-void func_8002CDE4(GlobalContext* globalCtx, TitleCardContext* titleCtx) {
+void TitleCard_Reset(GlobalContext* globalCtx, TitleCardContext* titleCtx) {
     titleCtx->durationTimer = titleCtx->delayTimer = titleCtx->intensity = titleCtx->alpha = 0;
 }
```
fig02, line 660: the other function mentioned above is probably more deserving of reset-- this gets called once in a function that initlizes the actor system. TitleCard_Init is probably more appropriate

fig02: contextual stuff like this is also why we try to tackle documentation as more of a system wide thing. You have to be familiar with the actor system itself to see these nuances to properly name them. Just something to consider as you find stuff you want to work on

src/code/z_actor.c
```diff
@@ -213,9 +213,9 @@ void Actor_SetFeetPos(Actor* actor, s32 limbIndex, s32 leftFootIndex, Vec3f* lef
     }
 }
 
-void func_8002BE04(GlobalContext* globalCtx, Vec3f* arg1, Vec3f* arg2, f32* arg3) {
-    SkinMatrix_Vec3fMtxFMultXYZW(&globalCtx->viewProjectionMtxF, arg1, arg2, arg3);
-    *arg3 = (*arg3 < 1.0f) ? 1.0f : (1.0f / *arg3);
+void Actor_ProjectPos(GlobalContext* globalCtx, Vec3f* src, Vec3f* xyzDest, f32* wDest) {
+    SkinMatrix_Vec3fMtxFMultXYZW(&globalCtx->viewProjectionMtxF, src, xyzDest, wDest);
+    *wDest = (*wDest < 1.0f) ? 1.0f : (1.0f / *wDest);
```
Dragorn421, line 218: It should be properly documented that the last argument is *not* a "w"
It's not even an inverted w since it caps it at 1

I'm personally not familiar enough with [homogeneous coordinates](https://en.wikipedia.org/wiki/Homogeneous_coordinates) to explain the part where they check ` w < 1`.
I would understand a `IS_ZERO`-like check, to avoid dividing by almost-0 and eventually making garbage float operations, but a threshold of 1 looks large so that must not be the intent

so I would leave `wDest` unnamed for now (and probably forever)

src/code/z_actor.c
```diff
@@ -334,22 +334,22 @@ void func_8002C124(TargetContext* targetCtx, GlobalContext* globalCtx) {
             spCE = targetCtx->unk_48;
         }
 
-        func_8002BE04(globalCtx, &targetCtx->targetCenterPos, &spBC, &spB4);
+        Actor_ProjectPos(globalCtx, &targetCtx->targetCenterPos, &projTargetCenter, &projTargetW);
 
-        spBC.x = (160 * (spBC.x * spB4)) * var1;
-        spBC.x = CLAMP(spBC.x, -320.0f, 320.0f);
+        projTargetCenter.x = (160 * (projTargetCenter.x * projTargetW)) * var1;
```
Dragorn421, line 339:
```suggestion
        targetScreenPos.x = (160 * (targetScreenPos.x * somethingNotW)) * var1;
```

src/code/z_actor.c
```diff
@@ -701,56 +701,57 @@ void TitleCard_Update(GlobalContext* globalCtx, TitleCardContext* titleCtx) {
...
+            gSPTextureRectangle(OVERLAY_DISP++, titleX, titleSecondY, ((doubleWidth * 2) + titleX) - 4, titleSecondY + (height * 4) - 1,
                                 G_TX_RENDERTILE, 0, 0, 1 << 10, 1 << 10);
         }
 
         CLOSE_DISPS(globalCtx->state.gfxCtx, "../z_actor.c", 2880);
     }
 }
 
-s32 func_8002D53C(GlobalContext* globalCtx, TitleCardContext* titleCtx) {
+s32 TitleCard_Clear(GlobalContext* globalCtx, TitleCardContext* titleCtx) {
```
Dragorn421, line 754: this name makes me feel like it's a cleanup function, but looking at its usage it looks more like an "interrupt title card" function (eg when talking to an NPC apparently?), it hides the title card early

this may just be me, and I'm not sure what name I'd like more anyway (anyone would like `Abort` maybe?)

src/code/z_actor.c
hatal175: While I don't quite understand the capping, it's obviously an inverted w like you said. It's used like 1/w to multiply the x and y coordinates to get the screen coordinates.
It should really be named thus.

fig02: also is ProjectPos a good name for this? that sounds very generic like its just projecting the pos onto something. when this is specifically relevant to the `viewProjectionMtx`
not smart enough with view things and matrix stuff to know personally, so its a genuine question

Dragorn421: `viewProjectionMtx` is also used to compute `Actor#projectedPos`:
https://github.com/zeldaret/oot/blob/master/src/code/z_actor.c#L2364
```c
SkinMatrix_Vec3fMtxFMultXYZW(&globalCtx->viewProjectionMtxF, &actor->world.pos, &actor->projectedPos,
                             &actor->projectedW);
```
so yeah I think the name is good

------

about naming wDest I really wouldn't just dismiss the cap on the grounds of not knowing its purpose lol

src/code/z_actor.c
fig02: if it were a cleanup function in the traditional sense we would use Destroy like we would everywhere else in the codebase, imo
Clear as in "clear it off the screen" is good enough for me, but more opinions would be good

fig02: just a side note, but where do you see it happening when talking? afaict both uses are related to using a get item (showing a trade item in hand or receiving an item)
And this is cause the title cards use the gi space allocated by link

src/code/z_actor.c
hatal175: How about cappedInvWDest? It does encapsulate all the issues we have with it? Same for its uses in the code.

Dragorn421 (approved):

fig02 (approved):
apologies for late review on this (and your other PRs), holidays were busy and been catching up on PRs in general.

This seems mostly fine to me, but I am slightly worried about `Actor_SetNaviToActor`. at a glance it seems fine, but Target Context stuff is very complex, and im worried about the possibility of conflicting names when that gets documented more thoroughly. Its probably fine though

Roman971 (approved):

PR #1092, opened 2021-12-17
lifemeter_z.c documentation

hatal175:
I've renamed HealthMeter_HandleCriticalAlarm to HealthMeter_HandleBeatingHeart since the alarm sound looks a bit tacked on this function to be honest. It's mostly about ticking the oscillator.

Other than that, I think I documented the leftover members and variables.


include/z64.h
```diff
@@ -654,10 +654,10 @@ typedef struct {
...
     /* 0x021E */ s16    heartsEnvG[2];
     /* 0x0222 */ s16    heartsEnvB[2];
-    /* 0x0226 */ s16    unk_226;
-    /* 0x0228 */ s16    unk_228;
-    /* 0x022A */ s16    unk_22A;
-    /* 0x022C */ s16    unk_22C;
+    /* 0x0226 */ s16    unk_226; // Used only in uncalled functions
+    /* 0x0228 */ s16    unk_228; // Used only in uncalled functions
+    /* 0x022A */ s16    beatingHeartOscillator;
+    /* 0x022C */ s16    beatingHeartOscillatorDirection; // 1 for decreasing, 0 for increasing
```
Dragorn421, line 660: alternatively instead of `Direction` you could consider `beatingHeartOscillatorDirection` as a boolean and name it `beatingHeartOscillatorDecrease`, then use true/false to set it and remove `!= 0` in the conditionals.

Removing `!= 0` in conditionals usually doesn't break matching (doesn't change the mips the compiler produces) but sometimes it does

Either way I don't have a strong preference for 0/1-Direction or boolean-Decreasing, just wanted to show that possibility

src/code/z_lifemeter.c
```diff
@@ -442,18 +445,18 @@ void HealthMeter_Draw(GlobalContext* globalCtx) {
...
-            temp1 *= 0.68f;
-            gSPTextureRectangle(OVERLAY_DISP++, (s32)((temp2 - temp1) * 4), (s32)((temp3 - temp1) * 4),
-                                (s32)((temp2 + temp1) * 4), (s32)((temp3 + temp1) * 4), G_TX_RENDERTILE, 0, 0,
-                                (s32)temp4, (s32)temp4);
+            heartCenterY = 26.0f + offsetY;
+            heartCenterX = 30.0f + offsetX;
+            heartTexCoord = 1.0f;
+            heartTexCoord /= 0.68f;
+            heartTexCoord *= 1 << 10;
+            halfHeartLen = 8.0f;
```
Dragorn421, line 453: I think we can afford the full word "length"
```suggestion
            halfHeartLength = 8.0f;
```

I also am not a fan of "length" but can't think of a better name

src/code/z_lifemeter.c
```diff
@@ -442,18 +445,18 @@ void HealthMeter_Draw(GlobalContext* globalCtx) {
...
+            halfHeartLen *= 0.68f;
+            gSPTextureRectangle(OVERLAY_DISP++, (s32)((heartCenterX - halfHeartLen) * 4), (s32)((heartCenterY - halfHeartLen) * 4),
+                                (s32)((heartCenterX + halfHeartLen) * 4), (s32)((heartCenterY + halfHeartLen) * 4), G_TX_RENDERTILE, 0, 0,
+                                (s32)heartTexCoord, (s32)heartTexCoord);
```
Dragorn421, line 457:
```suggestion
                                (s32)heartTexCoordPerPixel, (s32)heartTexCoordPerPixel);
```

A botched job of an explanation:

[`gSPTextureRectangle`](http://n64devkit.square7.ch/n64man/gsp/gSPTextureRectangle.htm)

`1 / 0.68 * (1 << 10)` represents `1 / 0.68` in the fixed-point s5.10 format
that means every screen pixel, the texture coordinate advances by `1 / 0.68 = 1.47` image pixel

so by drawing over a `halfHeartLen*2`-wide screen pixels area the texture coordinate changes by `halfHeartLen*2 * 1/0.68`
(with `halfHeartLen = 8 * 0.68`) `halfHeartLen*2 * 1/0.68 = 16` image pixels, which is indeed the width of eg `gHeartFullTex` (as can be seen in `assets/xml/textures/parameter_static.xml`)

src/code/z_lifemeter.c
```diff
@@ -489,42 +495,42 @@ void HealthMeter_Draw(GlobalContext* globalCtx) {
     CLOSE_DISPS(gfxCtx, "../z_lifemeter.c", 606);
 }
 
-void HealthMeter_HandleCriticalAlarm(GlobalContext* globalCtx) {
+void HealthMeter_HandleBeatingHeart(GlobalContext* globalCtx) {
```
Dragorn421, line 498: (I know "handle" comes from current master)

I feel like `Update` instead of `Handle` makes more sense, since it's part of `Interface_Update` and indeed updates stuff
It's also not uncommon for functions dubbed "update" to play sound
```suggestion
void HealthMeter_UpdateBeatingHeart(GlobalContext* globalCtx) {
```

Dragorn421 (approved):

include/z64.h
fig02: I think I would prefer direction and not treat it as a bool, personally
dont think the word oscillator needs to be in the var though, `beatingHeartDir` is sufficient for me. with the comment explaining which number is which direction

hatal175: I kinda liked @Dragorn421 approach. It makes the code more readable.

include/z64.h
```diff
@@ -654,10 +654,10 @@ typedef struct {
...
     /* 0x0222 */ s16    heartsEnvB[2];
-    /* 0x0226 */ s16    unk_226;
-    /* 0x0228 */ s16    unk_228;
-    /* 0x022A */ s16    unk_22A;
-    /* 0x022C */ s16    unk_22C;
+    /* 0x0226 */ s16    unk_226; // Used only in unused functions
+    /* 0x0228 */ s16    unk_228; // Used only in unused functions
+    /* 0x022A */ s16    beatingHeartOscillator;
+    /* 0x022C */ s16    beatingHeartOscillatorDecrease;
     /* 0x022E */ s16    unk_22E;
```
fig02, line 661: (sorry again for the late review on your things)
I would probably call this a direction, and same for `heartColorOscillatorDecrease`, and not treat it as a bool. I think there was some conversation about this initially that I may have missed. But this pattern is pretty common throughout interface related things and so far I think we've pretty consistently been treating it as a pulsing direction

louist103 (approved):

src/code/z_lifemeter.c
```diff
@@ -156,17 +156,17 @@ void HealthMeter_Update(GlobalContext* globalCtx) {
 
     if (interfaceCtx) {}
 
-    if (interfaceCtx->unk_200 != 0) {
-        interfaceCtx->unk_1FE--;
-        if (interfaceCtx->unk_1FE <= 0) {
-            interfaceCtx->unk_1FE = 0;
-            interfaceCtx->unk_200 = 0;
+    if (interfaceCtx->beatingHeartChangeDirection) {
+        interfaceCtx->beatingHeartChangeFactor--;
```
fig02, line 160: now that its direction we should add != 0 and change the true/falses. after that this is good from me

include/z64.h
```diff
@@ -645,8 +645,8 @@ typedef struct {
     /* 0x01F8 */ s16    naviCalling;
     /* 0x01FA */ s16    unk_1FA;
     /* 0x01FC */ s16    unk_1FC;
-    /* 0x01FE */ s16    unk_1FE;
-    /* 0x0200 */ s16    unk_200;
+    /* 0x01FE */ s16    beatingHeartColorFactor;
+    /* 0x0200 */ s16    beatingHeartColorFactorDirection;
```
Dragorn421, line 649: why "beating heart" if this is about their color

include/z64.h
```diff
@@ -655,10 +655,10 @@ typedef struct {
...
     /* 0x021E */ s16    heartsEnvG[2];
     /* 0x0222 */ s16    heartsEnvB[2];
-    /* 0x0226 */ s16    unk_226;
-    /* 0x0228 */ s16    unk_228;
-    /* 0x022A */ s16    unk_22A;
-    /* 0x022C */ s16    unk_22C;
+    /* 0x0226 */ s16    unk_226; // Used only in unused functions
+    /* 0x0228 */ s16    unk_228; // Used only in unused functions
+    /* 0x022A */ s16    beatingHeartOscillator;
+    /* 0x022C */ s16    beatingHeartPulsingDirection;
```
Dragorn421, line 661: these two are now named differently for some reason?

include/z64.h
Dragorn421: Also, I think the word "oscillator" itself was fine

Dragorn421 (approved):

fig02 (approved):

Roman971 (approved):

PR #1093, opened 2021-12-23
Misc. doc/cleanup 4

Dragorn421:


AngheloAlf (approved):

src/boot/z_std_dma.c
```diff
@@ -11,7 +11,7 @@ s32 sDmaMgrCurFileLine;
 
 u32 D_80009460 = 0;
 u32 gDmaMgrDmaBuffSize = 0x2000;
-u32 sDmaMgrDataExistError = 0;
+u32 sDmaMgrIsRomCompressed = false;
```
louist103, line 14: Since this seems to update per file it should probably be `IsFileCompressed`

louist103 (changes requested):

src/boot/z_std_dma.c
Dragorn421: Usually we say a rom is decompressed if no files are compressed, so a rom is "compressed" if at least one file is compressed (also in compressed roms most files are typically compressed but not all)

What matters anyway is that at the end of `DmaMgr_ProcessMsg` this boolean is used if no relevant entry is found in the dma table containing the target vrom offset,
either to error if the rom is compressed (`sDmaMgrIsRomCompressed` is true),
or to do dma anyway assuming vrom = rom

and vrom = rom needs the rom to be "fully" decompressed to have a chance of being true (as a piece of trivia, vrom = rom isn't required otherwise though)

louist103 (approved):

Makefile
```diff
@@ -239,6 +239,9 @@ build/dmadata_table_spec.h: build/$(SPEC)
 
 build/src/boot/z_std_dma.o: build/dmadata_table_spec.h
 build/src/dmadata/dmadata.o: build/dmadata_table_spec.h
+build/src/code/z_actor_dlftbls.o: include/tables/actor_table.h
+build/src/code/z_scene.o: include/tables/object_table.h
+build/src/code/z_effect_soft_sprite_dlftbls.o: include/tables/effect_ss_table.h
```
Roman971, line 244: I really don't think we should start manually declaring dependencies like this.
Imo these files should be treated like any other include, ie. if you modify them you have to touch the corresponding files or do a a full build for the changes to apply, at least until we restructure headers and add automatic include dependencies to the makefile. (Hopefully this should happen sooner rather than later, tho maybe not before retail support)

src/overlays/actors/ovl_En_Ex_Item/z_en_ex_item.c
```diff
@@ -488,10 +488,10 @@ void EnExItem_DrawHeartPiece(EnExItem* this, GlobalContext* globalCtx) {
 }
 
 void EnExItem_DrawMagic(EnExItem* this, GlobalContext* globalCtx, s16 magicIndex) {
-    static s16 sgiDrawIds[] = { GID_DINS_FIRE, GID_FARORES_WIND, GID_NAYRUS_LOVE };
+    static s16 giDrawIds[] = { GID_DINS_FIRE, GID_FARORES_WIND, GID_NAYRUS_LOVE };
```
Roman971, line 491: The previous name wasn't good either so good catch, but technically this should have the `s` prefix since it's static.
It's something we need to fix in general because it wasn't always applied consistently over time, so I'll probably do a full cleanup soon if no one gets to it before me.

fig02: we havent been doing it for in function static vars on purpose though

Dragorn421: in-function static should have the `s` prefix too? that's more the exception than the norm overall :sweat_smile:

Makefile
Dragorn421: I understand the concerns of manually defined dependencies, but these tables are such a thing that modding will definitely touch that I would really like to have this in the repo now instead of in several months

These tables aren't typical includes either

src/overlays/actors/ovl_En_Ex_Item/z_en_ex_item.c
Roman971: In-function static variables are still very different from local variables so I think it's important to have the prefix. Also, whether a static variable is inside or outside function doesn't change anything besides its scope, so to me I doesn't really make sense to name it differently.

(Relevant discussion we had on discord about this, where I brought up why naming without the prefix could be very misleading: https://discord.com/channels/688807550715560050/688851317593997489/887738378525749268)

Dragorn421: I don't really care about prefixing in-function static with `s` (I do think it's slightly better)
But regardless of going with that or not, what should this PR do about it? (I think nothing, there's enough stuff in there already)

fig02 (approved):

src/overlays/actors/ovl_En_Ex_Item/z_en_ex_item.c
fig02: agreed on doing nothing btw, and roman said he'll clean up the variable names at a later time

Im still a bit confused on the whole `s` thing for in function static variables, we made it a point in reviews thus far to remove them for in function static. But I do also understand the reasoning of keeping them to make them stand out from normal variables.

Either way its a conversation that needs to happen separately outside of this pr

Roman971 (approved):

PR #1094, opened 2021-12-24
Defines for `SaveContext` flags (`eventChkInf`, `itemGetInf`, `infTable`, `eventInf`)

Dragorn421:
The goal of this PR is to improve how `eventChkInf` `itemGetInf` `infTable` `eventInf` (all from `SaveContext gSaveContext`) are accessed.
All four of them are `u16` arrays where each bit of each `u16` corresponds to a "flag".

For `eventChkInf` and `infTable`, there are helper functions like `Flags_GetEventChkInf`, `Flags_SetEventChkInf` ( https://github.com/zeldaret/oot/blob/6213fe3bbd0bfcad695640d30d0b0bc8f9faffbe/src/code/z_actor.c#L4152 ) to get/set those flags by a "flag" value where the 4 lower bits set the target bit in the u16 value and the higher bits form the index of the u16 into the array

I reused this way of representing flags for `itemGetInf`, the macros introduced `GET_EVENTCHKINF`, `GET_ITEMGETINF`, `GET_INFTABLE`, `GET_EVENTINF` and `SET_EVENTCHKINF`, `SET_ITEMGETINF`, `SET_INFTABLE`, `SET_EVENTINF` are basically copypaste of those helper functions.

I also introduced `CLEAR_EVENTCHKINF`, `CLEAR_INFTABLE`, `CLEAR_EVENTINF`

I put the actual flags defines in `z64save.h` right before `SaveContext`

----------------

The vast majority of flags can use the macros, but not all. A few spots need to use the index/shift/mask "manually"
Such cases can be found by looking at the defines in z64save.h that aren't simply (for example) `#define EVENTCHKINF_3C 0x3C` but look like:
```c
// 0x40
#define EVENTCHKINF_40_INDEX 4
#define EVENTCHKINF_40_SHIFT 0
#define EVENTCHKINF_40_MASK (1 << EVENTCHKINF_40_SHIFT)
#define EVENTCHKINF_40 ((EVENTCHKINF_40_INDEX << 4) | EVENTCHKINF_40_SHIFT)
```
this is usually because the code does something weird or special for that flag. With `EVENTCHKINF_40` there is this:
https://github.com/Dragorn421/oot/blob/532ca5cca65d0fbddcb415340ea425568c7a0862/src/code/z_sram.c#L412-L413
```c
        i = gSaveContext.eventChkInf[EVENTCHKINF_40_INDEX] & ~EVENTCHKINF_40_MASK;
        gSaveContext.eventChkInf[EVENTCHKINF_40_INDEX] = i;
```

There are also flag groups like with `EVENTCHKINF_90_91_92_93_INDEX` (freed carpenters flags) where the code assumes using the same index (same u16 value) for the flags.

More macros should probably be introduced to deal with some flag groups (especially `EVENTINF_0X_INDEX` is a complete mess) but I think that's out of scope for this PR which is large enough being just refactoring with no documentation.


include/macros.h
```diff
@@ -64,6 +64,9 @@
 
 #define HIGH_SCORE(score) (gSaveContext.highScores[score])
 
+#define GET_ITEMGETINF(flag) (gSaveContext.itemGetInf[flag >> 4] & (1 << (flag & 0xF)))
+#define SET_ITEMGETINF(flag) (gSaveContext.itemGetInf[flag >> 4] |= (1 << (flag & 0xF)))
```
AngheloAlf, line 68: No macro to unset?
```c
#define UNSET_ITEMGETINF(flag) (gSaveContext.itemGetInf[flag >> 4] &= ~(1 << (flag & 0xF)))
```

Dragorn421: I think we would call that "clear"?

The game does clear (few) flags from the other three arrays, but not flags from `itemGetInf`
This may have to do with the purpose of `itemGetInf` to be permanent maybe? since it's for items supposedly

So on one hand it would be more consistent with the other arrays to have a clear macro,
and on the other hand `itemGetInf` flags are probably not meant to be cleared ever

Dragorn421:
per this debug string https://github.com/zeldaret/oot/blob/0b8edc21c03a719d8f7874bc20204612d6d1c107/src/code/z_parameter.c#L1876 fig mentioned,

the `infTable` array would have originally been accessed through a `Get_Inf_Table(index, mask)` macro

include/z64save.h
```diff
@@ -60,6 +60,456 @@ typedef struct {
     /* 0x24 */ s32 tempCollectFlags;
 } FaroresWindData; // size = 0x28
 
+
+// eventChkInf
+
+#define EVENTCHKINF_02 0x02
```
Roman971, line 66: I suggest putting all this after the structs/enums in this file, so it's not too broken up and a bit easier to navigate (similarly to how we leave the enums at the end of `z64actor.h` for instance)

Roman971 (commented):
Looks good to me overall, I only have a few suggestions.

Also the special index/shift defines aren't ideal but I guess there no real way around it. I'll probably experiment with some alternatives in the future but for now it seems fine to me at least

Roman971 (approved):

fig02 (approved):

Dragorn421:
Finally merged 👏 

I would like to dedicate this PR to another kind of flags, the flags around the world
🎏     🇩🇪 🇦🇺🇫🇷🇹🇼🇨🇦  🇲🇽  🇧🇪🇬🇧 🇺🇦🇲🇦🇦🇷 🇨🇿 🇷🇺  🇪🇺 🇦🇹 🇱🇹 🇿🇦 🇺🇸 🇩🇿  🇧🇷 🇨🇳 🏳️‍🌈 🇯🇵🇪🇸
please peace for everyone between countries and communities :)

PR #1101, opened 2022-01-01
Documentation of Horse System

tram98:
First commit, please leave feedback so I can improve.

My commits include: 

1. a function rename (from func_ signature)
2. insertion of enums using defines (in order to prevent mismatch)
   -> This might be the wrong way to do things, are there better approaches?
3. This is my first merge. Is that "merge branch" commit intended to be there or did I break something??

you can contact me on discord: tram98#4269

### DIFF
--- a/src/code/z_horse.c
+++ b/src/code/z_horse.c
@@ -1,6 +1,14 @@
 #include "global.h"
 #include "vt.h"
 
+#define HORSETYPE_1 1
+#define HORSETYPE_2 2
+#define HORSETYPE_5 5
+#define HORSETYPE_6 6
+#define HORSETYPE_7 7
+#define HORSETYPE_8 8
+#define HORSETYPE_9 9
+
 s32 Horse_CheckSceneValid(s32 scene) {
     s32 validScenes[] = { SCENE_SPOT00, SCENE_SPOT06, SCENE_SPOT09, SCENE_SPOT12, SCENE_SPOT20 };
     s32 i;
@@ -42,15 +50,15 @@ typedef struct {
 void func_8006D0EC(GlobalContext* globalCtx, Player* player) {
     s32 i;
     HorseSpawn horseSpawns[] = {
-        { SCENE_SPOT00, -460, 100, 6640, 0, 2 }, { SCENE_SPOT06, -1929, -1025, 768, 0, 2 },
-        { SCENE_SPOT09, 2566, -259, 767, 0, 2 }, { SCENE_SPOT12, -328, 10, 953, 0, 2 },
-        { SCENE_SPOT20, 928, 0, -2280, 0, 2 },
+        { SCENE_SPOT00, -460, 100, 6640, 0, HORSETYPE_2 }, { SCENE_SPOT06, -1929, -1025, 768, 0, HORSETYPE_2 },
+        { SCENE_SPOT09, 2566, -259, 767, 0, HORSETYPE_2 }, { SCENE_SPOT12, -328, 10, 953, 0, HORSETYPE_2 },
+        { SCENE_SPOT20, 928, 0, -2280, 0, HORSETYPE_2 },
     };
 
     if ((AREG(6) != 0) && (Flags_GetEventChkInf(0x18) || (DREG(1) != 0))) {
         player->rideActor = Actor_Spawn(&globalCtx->actorCtx, globalCtx, ACTOR_EN_HORSE, player->actor.world.pos.x,
                                         player->actor.world.pos.y, player->actor.world.pos.z, player->actor.shape.rot.x,
-                                        player->actor.shape.rot.y, player->actor.shape.rot.z, 9);
+                                        player->actor.shape.rot.y, player->actor.shape.rot.z, HORSETYPE_9);
 
         ASSERT(player->rideActor != NULL, "player->ride.actor != NULL", "../z_horse.c", 343);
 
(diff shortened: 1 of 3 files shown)


src/code/z_horse.c
```diff
@@ -1,7 +1,15 @@
...
 #include "vt.h"
 
-s32 func_8006CFC0(s32 scene) {
+#define HORSETYPE_1 1
+#define HORSETYPE_2 2
+#define HORSETYPE_5 5
+#define HORSETYPE_6 6
+#define HORSETYPE_7 7
+#define HORSETYPE_8 8
+#define HORSETYPE_9 9
```
Dragorn421, line 10: These should probably be an enum, maybe inside z_en_horse.h since they're used for the EnHorse actor's params (`Actor#params`, a s16 value that is usually used to set actor properties, eg speed for a moving platform, or the contents of a chest)

usually we name enum types `SystemNameWhatever` like for example `DemoKekkaiType` https://github.com/zeldaret/oot/blob/master/src/overlays/actors/ovl_Demo_Kekkai/z_demo_kekkai.h#L32
I would name the enum values themselves like `ENHORSE_TYPE_*` but if you feel like `HORSE_TYPE_*` or `HORSE_*` or something else is better I personally don't mind

also, you can find in z_en_horse by looking at params usage that there's also `HORSETYPE_3` (used in `EnHorseGameCheck_InitIngoRace` afaict from searching for where `ACTOR_EN_HORSE` is used)

also z_en_horse.c should use these values as well when checking params

src/code/z_horse.c
```diff
@@ -1,7 +1,15 @@
...
+
+s32 Horse_CheckSceneValid(s32 scene) {
```
Dragorn421, line 12: I think this function name isn't specific enough, from this name I expected the function to be used as "can a horse spawn in this scene" but it doesn't look like that's it

It looks like the scenes where you can leave Epona and find her at the same spot any time maybe? not sure

tram98: I will look into it and create a new commit as soon as I know more.
According to z_horse.c:256, this function was once called Horse_Set_Check(). 
That name could be better imho. What do you think?
```c
//Line 255...
void func_8006DC68(GlobalContext* globalCtx, Player* player) {
    if (LINK_IS_ADULT) {
        if (!Horse_CheckSceneValid(gSaveContext.horseData.scene)) {
            osSyncPrintf(VT_COL(RED, WHITE));
            // "Horse_Set_Check():%d set spot is no good."
            osSyncPrintf("Horse_Set_Check():%d セットスポットまずいです。\n", gSaveContext.horseData.scene);
            osSyncPrintf(VT_RST);
            func_8006D074(globalCtx);
        }
//Line 265...
```

src/code/z_horse.c
tram98: These horse types are used in multiple files. 
How do we handle this, eg. where would we put such an enum? 
Maybe put that enum inside of `z_horse.h` and `#include "z_horse.h"` in `z_en_horse.c`??
Or maybe create a completely separate header for that enum? (I am relatively new to c programming, I have done mostly oop in c++ and java)

src/code/z_horse.c
Dragorn421: We very rarely name things based on names from debug strings, since those names usually weren't very good (it's pretty safe to say many devs didn't master english)

In this case I don't think `Horse_Set_Check` is any good, as the name probably should mention being a scene check (which `Horse_CheckSceneValid` does) and also the check's purpose as I said above

src/code/z_horse.c
fig02: the header file for the horse actor (z_en_horse) would make most sense. 
You picked an unusual file as your first to document, in that its a file in the "engine" (so to speak) with code relating to a specific actor. So even though this code isnt the actor itself, it helps facilitate things for that actor.
Player is the other main example of this, with the actor being z_player.c and the helper stuff being z_player_lib.c

Dragorn421: As fig said, anything used by a specific actor should go in that actor's header file

Then files needing to interact with that actor include the header

For example z_camera includes z_en_horse.h https://github.com/zeldaret/oot/blob/0b8edc21c03a719d8f7874bc20204612d6d1c107/src/code/z_camera.c#L4
to use for example the `EN_HORSE_CHECK_JUMPING` macro defined in the EnHorse actor's header

src/code/z_horse.c
tram98: I thought a little more, how about `Horse_CanSpawnInScene`?

src/overlays/actors/ovl_En_Horse/z_en_horse.h
```diff
@@ -86,7 +86,13 @@ typedef enum {
...
     /* 0 */ HORSE_EPONA,
-    /* 1 */ HORSE_HNI
+    /* 1 */ HORSE_HNI,
+    /* 2 */ HORSE_TYPE_2,
+    /* 5 */ HORSE_TYPE_5 = 5,
+    /* 6 */ HORSE_TYPE_6,
+    /* 7 */ HORSE_TYPE_7,
+    /* 8 */ HORSE_TYPE_8,
+    /* 9 */ HORSE_TYPE_9
 } HorseType;
```
Dragorn421, line 96: it looks like this enum is meant for EnHorse#type, not as the value passed by params

the `horse_type_` should be another enum, idk if current HorseType should be renamed or what

src/code/z_horse.c
Dragorn421: I don't think that's accurate, there are plenty `Actor_Spawn` calls to spawn Epona that get called without this function playing a role

src/overlays/actors/ovl_En_Horse/z_en_horse.h
tram98: I'll look into it, ty

PR #1111, opened 2022-01-19
inline overlay flags definitions

playerskel:
Inline flag definitions for actors, as discussed.

### DIFF
--- a/src/overlays/actors/ovl_Bg_Jya_Bombchuiwa/z_bg_jya_bombchuiwa.c
+++ b/src/overlays/actors/ovl_Bg_Jya_Bombchuiwa/z_bg_jya_bombchuiwa.c
@@ -1,8 +1,6 @@
 #include "z_bg_jya_bombchuiwa.h"
 #include "overlays/effects/ovl_Effect_Ss_Kakera/z_eff_ss_kakera.h"
 #include "objects/object_jya_obj/object_jya_obj.h"
-#define FLAGS ACTOR_FLAG_0
-
 void BgJyaBombchuiwa_Init(Actor* thisx, GlobalContext* globalCtx);
 void BgJyaBombchuiwa_Destroy(Actor* thisx, GlobalContext* globalCtx);
 void BgJyaBombchuiwa_Update(Actor* thisx, GlobalContext* globalCtx);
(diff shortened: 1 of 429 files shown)


fig02 (commented):
dont believe we discussed making this change, but nows a good time for the conversation to start

I personally dont care too much for the FLAGS define and would probably be fine with it going. But it does help make things look cleaner in the case where there are alot of flags. I scrolled through here and the largest i saw was 6 flags being set at once 
`ACTOR_FLAG_0 | ACTOR_FLAG_2 | ACTOR_FLAG_4 | ACTOR_FLAG_9 | ACTOR_FLAG_12 | ACTOR_FLAG_14,`

for now this fits on one line, but that may change as the flags get named. so its something to consider

src/overlays/actors/ovl_Bg_Jya_Bombchuiwa/z_bg_jya_bombchuiwa.c
```diff
@@ -1,8 +1,6 @@
 #include "z_bg_jya_bombchuiwa.h"
 #include "overlays/effects/ovl_Effect_Ss_Kakera/z_eff_ss_kakera.h"
 #include "objects/object_jya_obj/object_jya_obj.h"
-#define FLAGS ACTOR_FLAG_0
-
 void BgJyaBombchuiwa_Init(Actor* thisx, GlobalContext* globalCtx);
```
Dragorn421, line 4: missing empty line

src/overlays/actors/ovl_En_Arow_Trap/z_en_arow_trap.c
```diff
@@ -6,16 +6,14 @@
 
 #include "z_en_arow_trap.h"
 #include "overlays/actors/ovl_En_Arrow/z_en_arrow.h"
-#define FLAGS ACTOR_FLAG_4
-
 void EnArowTrap_Init(Actor* thisx, GlobalContext* globalCtx);
```
Dragorn421, line 9: missing empty line

playerskel:
Exactly, yeah. When that becomes the case, one could decide that formatting such as

Dragorn421 (commented):
On the plus side of formatting if `ACTOR_FLAG_5` becomes a long enough name `Obj_Makeoshihiki_InitVars` may be formatted consistently with all other initvars!

fig02:
am currently leaning toward waiting on doing this until the flags are named and we can see how it will look like in context.
as mentioned in discord, we dont really want to format things by hand because we use the formatter

fig02:
closing for reason mentioned above. we can revisit this at a later time when the flags are named and we have a better idea of what it will look like when finalized

PR #1113, opened 2022-01-19
Naming1

playerskel:
Couple of small naming changes that I think are improvements. Happy to revert where needed of course.


src/code/z_message_PAL.c
```diff
@@ -114,7 +114,8 @@ void Message_ResetOcarinaNoteState(void) {
     R_OCARINA_NOTES_YPOS(2) = 179;
     R_OCARINA_NOTES_YPOS(3) = 174;
     R_OCARINA_NOTES_YPOS(4) = 169;
-    sOcarinaNoteBuf[0] = 0xFF;
+    sOcarinaNoteBuf[0] = OCARINA_NOTE_INVALID;
```
Dragorn421, line 117: this is being handled in #1026 as well as the comment on `sLastPlayedSong` and changes in `code_800EC960.c`

src/boot/z_std_dma.c
```diff
@@ -9,7 +9,7 @@ u8 sDmaMgrStack[0x500];
 const char* sDmaMgrCurFileName;
 s32 sDmaMgrCurFileLine;
 
-u32 D_80009460 = 0;
+u32 gDmaVerbose = 0;
```
Dragorn421, line 12:
```suggestion
u32 gDmaMgrVerbose = 0;
```

src/code/z_debug.c
```diff
@@ -33,8 +33,7 @@ InputCombo inputCombos[REG_GROUPS] = {
 
 char regChar[] = " SOPQMYDUIZCNKXcsiWAVHGmnBdkb";
 
-// initialize GameInfo
-void func_800636C0(void) {
+void GameInfo_Init(void) {
```
Dragorn421, line 36: Not sure about the `GameInfo` prefix

The way functions are named is `System_Name`

This would introduce the `GameInfo` system then, but that would be the only function in it.
We would need to know whether other functions would fit inside that system, to avoid this prefix getting a single use

Dragorn421 (changes requested):

src/boot/z_std_dma.c
playerskel: Done.

src/code/z_message_PAL.c
playerskel: Reverted.

src/code/z_debug.c
fig02: its probably ok, but would honestly prefer this be handled with the whole z_debug.c being handled at once to see if its appropriate. if youre not going to do that id leave it unnamed for now

playerskel: I may take a look at that later, but think I'll revert for now for simplicity sake.

fig02:
you may have forgotten to push the changes mentioned above. besides those I think this is good to go for me

playerskel:
Reverted commit.

src/code/code_800EC960.c
```diff
@@ -1575,11 +1575,11 @@ void func_800ED458(s32 arg0) {
...
             osSyncPrintf("Presss NA_KEY_B4 %08x\n", 2);
-            sCurOcarinaBtnVal = 0xB;
+            sCurOcarinaBtnVal = 11;
             sCurOcarinaBtnIdx = 3;
-        } else if (D_8016BA18 & sOcarinaCUPBtnMap) {
-            osSyncPrintf("Presss NA_KEY_D5 %08x\n", sOcarinaCUPBtnMap);
-            sCurOcarinaBtnVal = 0xE;
+        } else if (D_8016BA18 & sOcarinaCUpBtnMap) {
+            osSyncPrintf("Presss NA_KEY_D5 %08x\n", sOcarinaCUpBtnMap);
+            sCurOcarinaBtnVal = 14;
```
Dragorn421, line 1582: Can be reverted too, all these `code_800EC960.c` changes are handled in #1026 :
https://github.com/zeldaret/oot/blob/0e72cc864c722f51bf7f679ab81cc8ec7a9ec12b/src/code/code_800EC960.c#L1718-L1724

playerskel: Done.

playerskel:
Threw in two more commits. Not sure if you mind.

Dragorn421 (approved):
Usually we avoid adding stuff to PRs after they started getting reviews or it never ends 😅 
Merging and formatting changes are fine of course

src/boot/z_std_dma.c
```diff
@@ -9,7 +9,7 @@ u8 sDmaMgrStack[0x500];
 const char* sDmaMgrCurFileName;
 s32 sDmaMgrCurFileLine;
 
-u32 D_80009460 = 0;
+u32 gDmaMgrVerbose = 0;
```
fig02, line 12: while longer, it might be worth being a longer name like `gDmaMgrPrintfVerbosity`
i see the existing name and think "why is the dmamgr system being verbose, for what"

fig02: conversely, Overlay_Relocate has `gOverlayLogSeverity` which i kinda like

fig02 (approved):
I would prefer `gDmaMgrPrintfVerbosity` -> `gDmaMgrLogSeverity` to be consistent with overlay log severity, but can do that later

Roman971 (approved):

PR #1114, opened 2022-01-19
Entranceindex

playerskel:
Document most, but not all, entranceIndex usages. Surely there's plenty to improve here still, but I think it greatly improves the ability to understand the game flow in the code.


fig02:
Think we'll probably handle this a bit differently than what youve started here. For one, we are looking to introduce new handling of the scene table, which imo is a prerequisite to doing entrances. And from there we will make a similar generated table using the same mechanism found in `/include/tables.h`

The wip youve got going here is a good reference, but we will probably go with our planned method and do everything at once so the whole table is complete and shiftable

fig02:
closing for reason mentioned above. if you have any questions feel free to ask in discord

PR #1115, opened 2022-01-20
Environment documentation pass

fig02:
What started as only naming a couple things snowballed into documenting most of the reamining parts of the system...

- all env context struct members named
- all but one function named (someone feel free to check out the weird zbuffer/depth stuff going on there)
- enums introduced for various environment members
- CLOCK_TIME macro implemented (same from mm). This shows the time as 24 clock time ( (18, 0) is 6pm, for example).
admittedly this probably should have been its own pr, but it was very helpful having the times readable to document other parts of the environment system.
There are also some unfortunate cases where the exact value cant be used, and you need to add 1 or 2 units to the macro. These are pretty rare in code, but needed alot in the skybox and light time entries in z_kankyo. Even with the cases where it isnt perfect, I still think it is a massive improvement over keeping them in hex.

TODO
-name the rest of the temp variables in functions
-name the remaining function args
-figure out how the zbuffer stuff works and name `func_8006F0A0` / `Struct_8011FAF0`
-macro-ize accesses to `sTimeBasedLightConfigs` and `gTimeBasedSkyboxConfigs`.
The existing macros for these were removed as they were kind of a bandaid for something that needs a larger solution. The portions of code in `Environment_Update` that use these arrays are practically unreadable at the moment with nested array accesses and lerp macro usages. This needs to be improved alot but Im not too great at making macros for things like this.


src/overlays/actors/ovl_En_Weather_Tag/z_en_weather_tag.c
```diff
@@ -127,82 +127,86 @@ void EnWeatherTag_Init(Actor* thisx, GlobalContext* globalCtx) {
...
+                    if (globalCtx->envCtx.stormRequest == STORM_REQUEST_NONE) {
+                        globalCtx->envCtx.weatherChgSkyState = WEATHER_CHANGE_SKY_REQUESTED;
+                        globalCtx->envCtx.skyboxConfig = arg2;
+                        globalCtx->envCtx.nextSkyboxConfig = arg3;
+                        globalCtx->envCtx.weatherChgSkyTimer = arg6;
+                        globalCtx->envCtx.weatherChgLights = true;
+                        globalCtx->envCtx.lightConfig = arg4;
+                        globalCtx->envCtx.nextLightConfig = nextLightConfig;
+                        gSavedNextLightConfig = nextLightConfig;
+                        globalCtx->envCtx.weatherChgDuration = arg6;
```
louist103, line 156: Can probably name the args with this.

src/overlays/actors/ovl_En_Weather_Tag/z_en_weather_tag.c
```diff
@@ -127,82 +127,86 @@ void EnWeatherTag_Init(Actor* thisx, GlobalContext* globalCtx) {
...
+                gWeatherMode = WEATHER_MODE_CLEAR;
+                globalCtx->envCtx.weatherChgSkyState = WEATHER_CHANGE_SKY_REQUESTED;
+                globalCtx->envCtx.skyboxConfig = arg2;
+                globalCtx->envCtx.nextSkyboxConfig = arg3;
+                globalCtx->envCtx.weatherChgSkyTimer = arg6;
+                globalCtx->envCtx.weatherChgLights = true;
+                globalCtx->envCtx.lightConfig = arg4;
+                globalCtx->envCtx.nextLightConfig = nextLightConfig;
+                gSavedNextLightConfig = nextLightConfig;
+                globalCtx->envCtx.weatherChgDuration = arg6;
```
louist103, line 196: Same here

louist103 (changes requested):
Nice job. Just a few things.

src/overlays/actors/ovl_Bg_Po_Syokudai/z_bg_po_syokudai.c
```diff
@@ -102,7 +102,7 @@ void BgPoSyokudai_Init(Actor* thisx, GlobalContext* globalCtx) {
 
         Actor_Spawn(&globalCtx->actorCtx, globalCtx, ACTOR_EN_PO_SISTERS, 119.0f, 225.0f, -1566.0f, 0, 0, 0,
                     thisx->params);
-        globalCtx->envCtx.unk_BF = 0x4;
+        globalCtx->envCtx.lightSettingOverride = 4;
```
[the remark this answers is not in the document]
fig02: a light setting override is a light setting, which is not possible to make an enum for. they vary per scene.
the enum youre seeing is `LIGHT_SETTING_OVERRIDE_NONE`, a magic number for 0xFF

include/z64environment.h
```diff
@@ -72,26 +131,26 @@ typedef struct {
...
-    /* 0x21 */ u8 unk_21;
-    /* 0x22 */ u16 unk_22;
-    /* 0x24 */ u16 unk_24;
+    /* 0x1E */ u8 indoors; // when true, day time has no effect on lighting
+    /* 0x1F */ u8 lightConfig; // only used outdoors
+    /* 0x20 */ u8 nextLightConfig; // only used outdoors
+    /* 0x21 */ u8 weatherChgLights;
+    /* 0x22 */ u16 weatherChgLightTimer; // time left to transition lights for weather change
+    /* 0x24 */ u16 weatherChgDuration; // total time to change weather. used for lights and sky lerp
     /* 0x26 */ char unk_26[0x02];
```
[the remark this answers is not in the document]
fig02: its not, no. LightInfo will start at 26 if its removed

src/overlays/actors/ovl_En_Weather_Tag/z_en_weather_tag.c
fig02: could, but I also see it as something that could be cleaned up with the actor in general. I may do the rest of them though

src/code/z_play.c
```diff
@@ -1488,18 +1488,18 @@ void Gameplay_SpawnScene(GlobalContext* globalCtx, s32 sceneNum, s32 spawn) {
...
-           (globalCtx->viewProjectionMtxF.wx * src->x + globalCtx->viewProjectionMtxF.wy * src->y +
-            globalCtx->viewProjectionMtxF.wz * src->z);
+    w = globalCtx->viewProjectionMtxF.ww +
+        (globalCtx->viewProjectionMtxF.wx * src->x + globalCtx->viewProjectionMtxF.wy * src->y +
+         globalCtx->viewProjectionMtxF.wz * src->z);
 
-    dest->x = 160.0f + ((dest->x / temp) * 160.0f);
-    dest->y = 120.0f - ((dest->y / temp) * 120.0f);
+    dest->x = 160.0f + ((dest->x / w) * 160.0f);
+    dest->y = 120.0f - ((dest->y / w) * 120.0f);
```
Thar0, line 1502:
```suggestion
    dest->x = (SCREEN_WIDTH / 2) + ((dest->x / w) * (SCREEN_WIDTH / 2));
    dest->y = (SCREEN_HEIGHT / 2) - ((dest->y / w) * (SCREEN_HEIGHT / 2));
```
Should still match, the integers should be promoted to float since the variables are floats

src/code/z_kankyo.c
```diff
@@ -190,27 +174,40 @@ SkyboxFile gSkyboxFiles[] = {
...
 LightNode* sNGameOverLightNode;
 LightInfo sNGameOverLightInfo;
 LightNode* sSGameOverLightNode;
 LightInfo sSGameOverLightInfo;
 u8 sGameOverLightsIntensity;
-u16 D_8015FDB0;
+u16 sSandstormScroll;
 
 s32 func_8006F0A0(s32 a0) {
     s32 ret = ((a0 >> 4 & 0x7FF) << D_8011FAF0[a0 >> 15 & 7].unk0) + D_8011FAF0[a0 >> 15 & 7].unk1;
```
Thar0, line 213: Not so much a suggestion as much as me leaving my rambling thoughts here for anyone else interested in figuring out exactly what's going on here.
At a distance, this function is responsible for converting a z-buffer depth value (more specifically, a depth value `* 4` or `<< 2`) to a screen depth value (as in, `z` in screen space)
The contents of the z-buffer is an interesting format. It's 16 bits in 2 parts, `z` and `dz`. `z` is a 3.11 fixed point number so takes up 14 bits, the remaining 2 bits are for `dz`. The maximum depth value is encapsulated in the constant `G_MAXFBZ` (`0x3FFF`). On the other hand, the "screen depth" value is a floating point number with a different scale.
The `unk1` member of `D_8011FAF0` follows a sequence where the difference between each value is half the previous difference.

Dragorn421: This from where `func_8006F0A0` is called
```c
wZ = (s32)((multDest.z * wDest) * 16352.0f) + 16352;
```
is straight out of the manual (12.7.6 Depth Source)
>     `screenZ=32*((z/w)*Viewport.vscale[2]+Viewport.vtrans[2])`
> Note: `Viewport.vscale` and `Viewport.vtrans[2]` are usually both `G_MAXZ/2 = 0x1ff`, which makes the formula: `screenZ=(z/w)*0x3fe0 + 0x3fe0`. Since `(z/w)` ranges from -1.0 to +1.0 the result will range from 0x0 to 0x7fc0. 

In the next line `zBuf = gZBuffer[(s32)((wY * -120.0f) + 120.0f)][(s32)((wX * 160.0f) + 160.0f)] * 4` I guess the `* 4` (`<< 2`) accounts for the actual zbuffer value being 18 bits (two 9-bits bytes), this is backed up by this value then being passed to `func_8006F0A0` which seems to consider it as 18 bits

In `func_8006F0A0`
the dz value (`a0 & 0xF`) seems to be ditched
`a0 >> 4 & 0x7FF` : 11 lower bits of the z value
`a0 >> 15 & 7` : 3 higher bits of the z value

for `unk1` in `D_8011FAF0` looking at them in binary is what I think reveals structure but I have no idea what the function is doing with it, doesn't make sense to me

include/z64environment.h
```diff
@@ -21,14 +48,47 @@ typedef enum {
...
-    /* 12 */ SKYBOX_DMA_FILE2_DONE,
-    /* 13 */ SKYBOX_DMA_PAL2_START
-} SkyboxDmaState;
+    /* 0 */ WEATHER_MODE_CLEAR,
+    /* 1 */ WEATHER_MODE_HYRULE_CLOUDY,
+    /* 2 */ WEATHER_MODE_LONLON_DMT_CLOUDY,
+    /* 3 */ WEATHER_MODE_ZORA_SNOW,
+    /* 4 */ WEATHER_MODE_LAKE_HYLIA_RAIN,
+    /* 5 */ WEATHER_MODE_KAK_RAIN
+} WeatherMode;
```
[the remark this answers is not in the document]
fig02: yes unfortunately it does, the light setting numbers are specific to the location they are in. if you tried to use a kak setting in hyrule field for example, you lights will be all green and messed up (speedrunning community calls this fog glitch lol)

src/code/z_kankyo.c
```diff
@@ -1438,37 +1528,37 @@ void Environment_DrawLensFlare(GlobalContext* globalCtx, EnvironmentContext* env
...
-            func_800C016C(globalCtx, &pos, &screenPos);
-            D_8015FD7E = (s16)screenPos.x;
-            D_8015FD80 = (s16)screenPos.y - 5.0f;
-            if (D_8011FB44 != 0xFFFC || screenPos.x < 0.0f || screenPos.y < 0.0f || screenPos.x > SCREEN_WIDTH ||
-                screenPos.y > SCREEN_HEIGHT) {
+        if (isSun) {
+            Gameplay_GetScreenPos(globalCtx, &pos, &screenPos);
+            sLensFlareDepthTestX = (s16)screenPos.x;
+            sLensFlareDepthTestY = (s16)screenPos.y - 5.0f;
+            if (sLensFlareScreenDepth != GPACK_ZDZ(G_MAXFBZ, 0) || screenPos.x < 0.0f || screenPos.y < 0.0f ||
```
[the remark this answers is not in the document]
fig02: the name DepthTextX/Y was chosen cause the -5 offset makes it not really the suns position anymore. the depth check is offset a bit from the sun itself, so i dont think its fair to say sun position.
I will change lens flare to sun though

include/z64environment.h
```diff
@@ -72,26 +131,26 @@ typedef struct {
     /* 0x14 */ char unk_14[0x01];
     /* 0x15 */ u8 skyboxDisabled;
     /* 0x16 */ u8 sunMoonDisabled;
-    /* 0x17 */ u8 unk_17; // currentWeatherMode for skybox? (prev called gloomySky)
-    /* 0x18 */ u8 unk_18; // nextWeatherMode for skybox?
-    /* 0x19 */ u8 unk_19;
-    /* 0x1A */ u16 unk_1A;
+    /* 0x17 */ u8 skyboxConfig; // only used outdoors
+    /* 0x18 */ u8 nextSkyboxConfig; // only used outdoors
```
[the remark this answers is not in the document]
fig02: would a comment next to nextSkyboxConfig help? I think adding weather change to the name will make the name terribly long, but thats just me personally

I dont follow what you mean by that last point, do you mean removing all "weatherChange" from all the weather changing variables?

src/code/z_kankyo.c
```diff
@@ -18,140 +12,144 @@ typedef enum {
...
-u16 D_8011FB44 = 0xFFFC;
+typedef struct {
+    /* 0x00 */ u16 startTime;
+    /* 0x02 */ u16 endTime;
+    /* 0x04 */ u8 lightSetting;
+    /* 0x05 */ u8 nextLightSetting;
+} TimeBasedLightEntry; // size = 0x6
 
-struct_8011FB48 D_8011FB48[][7] = {
+TimeBasedLightEntry sTimeBasedLightConfigs[][7] = {
```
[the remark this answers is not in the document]
fig02: I think the tables themselves are ok if you were to manually line the fields up so its easier to see. different entries have the additional time units added which makes them all shifted around and its impossible to read.
this goes against our current trend of not disabling the formatter unless needed though. so not really sure what to do.

not a fan of the macro currently cause i feel it hides too much, but it also may be for the best. not entirely sure yet.

what needs to be macro'd for sure is the accesses to them in the code. those long nested accesses are very dumb

include/z64environment.h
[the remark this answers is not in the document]
fig02: resolved in discord, will move to a scene independent naming scheme

src/code/z_kankyo.c
```diff
@@ -557,38 +583,39 @@ f32 Environment_LerpWeightAccelDecel(u16 endFrame, u16 startFrame, u16 curFrame,
...
-                    envCtx->unk_DE++;
+void Environment_UpdateStorm(EnvironmentContext* envCtx, u8 unused) {
+    if (envCtx->stormRequest != STORM_REQUEST_NONE) {
+        switch (envCtx->stormState) {
+            case STORM_STATE_OFF:
+                if ((envCtx->stormRequest == STORM_REQUEST_START) && !gSkyboxIsChanging) {
+                    envCtx->changeSkyState = CHANGE_SKY_REQUESTED;
+                    envCtx->skyboxConfig = 0;
+                    envCtx->nextSkyboxConfig = 1;
+                    envCtx->changeSkyTimer = 100;
```
Dragorn421, line 594: this is getting complicated, there are two kinds of "change" things

`gSkyboxIsChanging` refers to changing between skybox indices (for an entry inside the config);
the other `change*` things are refering to a config change

This could be addressed by adding config to change* stuff and `gSkyboxIsChanging` -> `gSkyboxIndexIsChanging` I guess

include/z64environment.h
```diff
@@ -72,26 +136,26 @@ typedef struct {
...
-    /* 0x20 */ u8 unk_20; // prev outdoor light index?
-    /* 0x21 */ u8 unk_21;
-    /* 0x22 */ u16 unk_22;
-    /* 0x24 */ u16 unk_24;
+    /* 0x1E */ u8 indoors; // when true, day time has no effect on lighting
+    /* 0x1F */ u8 lightConfig; // only used outdoors
+    /* 0x20 */ u8 nextLightConfig; // only used outdoors
+    /* 0x21 */ u8 changeLights;
+    /* 0x22 */ u16 changeLightTimer;
+    /* 0x24 */ u16 outdoorChangeDuration; // total time to change skybox and light configs
```
Dragorn421, line 149: I'm not sure why this one specifically is named `outdoorChange`
I don't think it's useful to put that in the name, `lightConfig` is the main thing about this light stuff and doesn't have it

include/z64environment.h
```diff
@@ -72,26 +136,26 @@ typedef struct {
...
     /* 0x15 */ u8 skyboxDisabled;
     /* 0x16 */ u8 sunMoonDisabled;
-    /* 0x17 */ u8 unk_17; // currentWeatherMode for skybox? (prev called gloomySky)
-    /* 0x18 */ u8 unk_18; // nextWeatherMode for skybox?
-    /* 0x19 */ u8 unk_19;
-    /* 0x1A */ u16 unk_1A;
+    /* 0x17 */ u8 skyboxConfig; // only used outdoors
+    /* 0x18 */ u8 nextSkyboxConfig; // only used outdoors
+    /* 0x19 */ u8 changeSkyState;
+    /* 0x1A */ u16 changeSkyTimer;
```
Dragorn421, line 142: I think it's not obvious enough the "change" stuff is about changing to the "next" configs, here's an idea
```suggestion
    /* 0x18 */ u8 changeSkyboxNextConfig; // only used outdoors
    /* 0x19 */ u8 changeSkyboxState;
    /* 0x1A */ u16 changeSkyboxTimer;
```
see also next comment

include/z64environment.h
Dragorn421, line 149:
```suggestion
    /* 0x20 */ u8 changeLightNextConfig; // only used outdoors
    /* 0x21 */ u8 changeLightEnabled;
    /* 0x22 */ u16 changeLightTimer;
    /* 0x24 */ u16 changeDuration; // total time to change skybox and light configs
```

fig02: i need to describe whats changing somehow, changeDuration is not enough imo. if you have any ideas feel free

Dragorn421: I was thinking the `change` prefix is enough

if it's really a worry `changeSkyboxAndLightDuration` I guess

Dragorn421: `changeConfigsDuration`?

src/code/z_kankyo.c
fig02: imo `gSkyboxIsChanging` is appropriate as is, it signifies the skybox itself is currently changing instead of the config. with the others now mentioning that the config is whats changing I think its clear that its separate

src/code/z_kankyo.c
```diff
@@ -1941,107 +2050,132 @@ void Environment_PlaySceneSequence(GlobalContext* globalCtx) {
...
+
+        case TIMESEQ_NIGHT_CRITTERS:
             Audio_SetNatureAmbienceChannelIO(NATURE_CHANNEL_CRITTER_0, CHANNEL_IO_PORT_1, 0);
-            if (globalCtx->envCtx.unk_EE[0] == 0 && globalCtx->envCtx.unk_F2[0] == 0) {
+
+            if (globalCtx->envCtx.precipitation[PRECIP_RAIN_MAX] == 0 &&
+                globalCtx->envCtx.precipitation[PRECIP_SOS_MAX] == 0) {
                 Audio_SetNatureAmbienceChannelIO(NATURE_CHANNEL_CRITTER_1 << 4 | NATURE_CHANNEL_CRITTER_3,
                                                  CHANNEL_IO_PORT_1, 1);
             }
```
engineer124, line 2141: I wonder if it's worth some kind of comment to describe what's going on with these `Audio_SetNatureAmbienceChannelIO` function. Or maybe there's some kind of local macro that could be used? It's not immediately obvious from reading this code that setting port 1 to a val of 0 turns off that critter channel, while setting port 1 to a val of 1 turns it on.

What the value of a specific channel does is entirely sequence and channel dependent, so I don't know if there's a general purpose macro that's worth it or not.

src/code/z_kankyo.c
Dragorn421: Well with a config change the skybox itself changes too right?
Both changing config and changing setting is done by lerping to the next thing (and changing config checks `gSkyboxIsChanging` is false to make sure both don't change at the same time)

So since the effect in game sounds similar (fading to a different skybox) I figured this could use clarification

I'm fine with this now that I understand the code, but if I was to discover it again, I wouldn't rule out `gSkyboxIsChanging` being about the config change being active

At the very least `gSkyboxIsChanging` may deserve a comment like
```c
// Indicates whether the skybox is changing to a different index of the same config (based on time)
u8 gSkyboxIsChanging = false;
```
idk not sure about my wording there. Don't want to explain the whole thing either

src/code/z_kankyo.c
Dragorn421: Is there hope that more audio documentation could help figure out a solution, including having enough documentation that it wouldn't be too hard for someone to figure it out themselves?

fig02: I was also expecting this to be an improvement made on audios end. But I dont know enough about audio to say that its appropriate

engineer124: No worries, yeah I'm not sure exactly how to tackle these ambiences yet but it'll likely be after sequence 1 is decompiled. So we can leave it to the audio side of things for now then I suppose

src/code/z_kankyo.c
```diff
@@ -1668,16 +1777,16 @@ void Environment_DrawRain(GlobalContext* globalCtx, View* view, GraphicsContext*
...
 
-void func_80074CE8(GlobalContext* globalCtx, u32 arg1) {
-    if ((globalCtx->envCtx.unk_BD != arg1) && (globalCtx->envCtx.unk_D8 >= 1.0f) &&
-        (globalCtx->envCtx.unk_BF == 0xFF)) {
-        if (arg1 > 30) {
-            arg1 = 0;
+void Environment_ChangeLightSetting(GlobalContext* globalCtx, u32 lightSetting) {
+    if ((globalCtx->envCtx.lightSetting != lightSetting) && (globalCtx->envCtx.lightBlend >= 1.0f) &&
+        (globalCtx->envCtx.lightSettingOverride == LIGHT_SETTING_OVERRIDE_NONE)) {
+        if (lightSetting > LIGHT_SETTING_MAX) {
```
Dragorn421, line 1784: I feel like this should be
```suggestion
        if (lightSetting >= LIGHT_SETTING_MAX) {
```
with `LIGHT_SETTING_MAX` being 31

it would also ditch the +1 in `lightSettingOverride <= LIGHT_SETTING_MAX + 1`

I'm not sure what is this `LIGHT_SETTING_MAX` value about though, `sTimeBasedLightConfigs` references 0 to 15 and 20 to 23 boundaries included, would be 24 settings, and 7 bonus settings (24-30)? looks kinda random
(`ganontika_sceneLightSettings0x000300` has 26 settings, most in the game afaict)

If anything this should use `envCtx->numLightSettings`
I also don't see why the check in `Environment_Update` would allow 31 but this doesn't

I know you can't fix the jank code but I'm questioning if the define makes sense to have. Maybe it does and the check in `Environment_Update` is a typo that should be `<` instead of `<=`, or they just used two different max values that happened to be 1 off

src/code/z_kankyo.c
```diff
@@ -1538,27 +1648,27 @@ void Environment_DrawLensFlare(GlobalContext* globalCtx, EnvironmentContext* env
...
+                    Math_SmoothStepToF(&envCtx->lensFlareFillAlpha, 0.0f, 0.5f, 50.0f, 0.1f);
                 }
 
                 temp = colorIntensity / 120.0f;
                 temp = CLAMP_MIN(temp, 0.0f);
 
                 gDPSetPrimColor(POLY_XLU_DISP++, 0, 0, 255, (u8)(temp * 75.0f) + 180, (u8)(temp * 155.0f) + 100,
-                                (u8)envCtx->unk_84);
+                                (u8)envCtx->lensFlareFillAlpha);
                 gDPFillRectangle(POLY_XLU_DISP++, 0, 0, SCREEN_WIDTH - 1, SCREEN_HEIGHT - 1);
```
Dragorn421, line 1661: Pretty sure this is glare implementation
(in my understanding "lens flare" is the localized spots of lights when looking at the sun,
"glare" is the overall blindness induced)

Would suggest
`lensFlareFillAlpha` -> `glareAlpha`
`screenFillAlpha` -> `glareStrength`
`alphaScale` -> `glareAlphaScale`
also `alphaScaleTarget` -> `lensFlareAlphaScaleTarget`

fig02:
Looking at `gWeatherMode` from mzx's perspective, it's fair to say that the weather modes are internal to weather tag, and they had to be exposed to the larger environment system strictly for the purposes of allowing changes across scenes. In which case naming the modes after their location does make sense.

This is mainly a note to myself to think about this some more and decide what to do when I come back to implementing these review changes, but anyone else can weigh in on this if they like

src/code/z_kankyo.c
```diff
@@ -860,46 +892,53 @@ void Environment_Update(GlobalContext* globalCtx, EnvironmentContext* envCtx, Li
             }
         }
 
-        func_800766C4(globalCtx); // increments or decrements unk_EE[1] depending on some condition
-        func_80075B44(globalCtx); // updates bgm/sfx and other things as the day progresses
+        Environment_UpdateRain(globalCtx);
+        Environment_PlayTimeBasedSequence(globalCtx);
 
-        if (((void)0, gSaveContext.nextDayTime) >= 0xFF00 && ((void)0, gSaveContext.nextDayTime) != 0xFFFF) {
+        if (((void)0, gSaveContext.nextDayTime) >= 0xFF00 && ((void)0, gSaveContext.nextDayTime) != NEXT_TIME_NONE) {
```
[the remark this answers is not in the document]
fig02: been a minute since ive looked at any of this stuff

but at first glance, the magic number seems better to me than doing math that happens to end up at the number. it doesnt really make much sense. I see your point about being able to locate it, but I dont think thats enough for me

fig02:
reviving this PR from the dead

merged master and addressed most open comments. I still plan to take a second look at the CLOCK_TIME suggestion above and see if it can help improve things

include/z64environment.h
```diff
@@ -67,11 +133,9 @@ typedef struct {
     /* 0x14 */ s16 fogFar;
 } EnvLightSettings; // size = 0x16
 
-// 1.0: 801D8EC4
-// dbg: 80222A44
 typedef struct {
     /* 0x00 */ char unk_00[0x02];
-    /* 0x02 */ u16 timeIncrement; // how many units of time that pass every update
+    /* 0x02 */ u16 timeSpeed; // how many units of time that pass every update
```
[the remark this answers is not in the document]
fig02: it gets its value directly from the room command for time speed, so I think its more than an appropriate name. But yea I can move the comment to gTimeSpeed

src/code/z_kankyo.c
```diff
@@ -1668,16 +1779,16 @@ void Environment_DrawRain(GlobalContext* globalCtx, View* view, GraphicsContext*
     }
 }
 
-void func_80074CE8(GlobalContext* globalCtx, u32 arg1) {
-    if ((globalCtx->envCtx.unk_BD != arg1) && (globalCtx->envCtx.unk_D8 >= 1.0f) &&
-        (globalCtx->envCtx.unk_BF == 0xFF)) {
-        if (arg1 > 30) {
-            arg1 = 0;
+void Environment_ChangeLightSetting(GlobalContext* globalCtx, u32 lightSetting) {
```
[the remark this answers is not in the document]
fig02: Basically, I dont see it as a "try" or "reqeuest", rather just an implementation detail of changing the light setting based on the floor polygon. and one other actor happens to use it when they probably should have done it another way

src/code/z_kankyo.c
```diff
@@ -922,54 +962,68 @@ void Environment_Update(GlobalContext* globalCtx, EnvironmentContext* envCtx, Li
...
         }
 
-        if ((envCtx->unk_BF != 0xFF) && (envCtx->unk_DC != 2) && (envCtx->unk_BD != envCtx->unk_BF) &&
-            (envCtx->unk_D8 >= 1.0f) && (envCtx->unk_BF < 0x20)) {
-            envCtx->unk_BE = envCtx->unk_BD;
-            envCtx->unk_BD = envCtx->unk_BF;
-            envCtx->unk_D8 = 0.0f;
+        if ((envCtx->lightSettingOverride != LIGHT_SETTING_OVERRIDE_NONE) &&
+            (envCtx->lightBlendOverride != LIGHT_BLEND_OVERRIDE_FULL_CONTROL) &&
+            (envCtx->lightSetting != envCtx->lightSettingOverride) && (envCtx->lightBlend >= 1.0f) &&
```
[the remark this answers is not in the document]
fig02: the override is meant to convey that youre breaking away from lights that are being set in the floor poly, and handing you the control instead. 
For the same reason I think the "waiting for light blend thing" doesnt make it a request in `Environment_ChangeLightSetting`, I see it the same way here

src/code/z_kankyo.c
```diff
@@ -1743,7 +1854,7 @@ void Environment_DrawLightningFlash(GlobalContext* globalCtx, u8 red, u8 green,
 }
 
 void Environment_UpdateLightningStrike(GlobalContext* globalCtx) {
```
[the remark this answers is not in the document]
fig02: looking at the function the bulk of it is pretty clearly to update the logic of the lightning, and it gets drawn separately in `Environment_DrawLightning`. Just another case of updating in draw imo, and mixing the two with the flash call being in the update call. 
Its just a mess, but its more update oriented imo

include/z64environment.h
```diff
@@ -111,27 +175,26 @@ typedef struct {
...
     /* 0xE2 */ u8 screenFillColor[4];
     /* 0xE6 */ u8 sandstormState;
     /* 0xE7 */ u8 sandstormPrimA;
     /* 0xE8 */ u8 sandstormEnvA;
     /* 0xE9 */ u8 customSkyboxFilter;
     /* 0xEA */ u8 skyboxFilterColor[4];
-    /* 0xEE */ u8 unk_EE[4];
-    /* 0xF2 */ u8 unk_F2[4];
-    /* 0xF6 */ char unk_F6[0x06];
+    /* 0xEE */ u8 precipitation[5];
```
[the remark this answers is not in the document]
fig02: its probably ok? itll be really short compared to the rest, and someone reading this will probably be familiar with the enum_mzx construct to understand that its different than the rest

src/code/z_kankyo.c
```diff
@@ -190,30 +175,56 @@ SkyboxFile gSkyboxFiles[] = {
...
+    /* 0x10 */ Vec3f pos;
+    /* 0x1C */ s8 pitch;
+    /* 0x1D */ s8 roll;
+    /* 0x1E */ u8 textureIndex;
+    /* 0x1F */ u8 delayTimer;
+} LightningBolt; // size = 0x20
+
 LightningBolt sLightningBolts[3];
 LightningStrike gLightningStrike;
 s16 sLightningFlashAlpha;
```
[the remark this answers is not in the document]
fig02: yeah I hadnt considered that. will see how it looks without

src/code/z_kankyo.c
```diff
@@ -922,54 +962,68 @@ void Environment_Update(GlobalContext* globalCtx, EnvironmentContext* envCtx, Li
...
+            envCtx->lightSetting = envCtx->lightSettingOverride;
+            envCtx->lightBlend = 0.0f;
         }
 
-        if (envCtx->unk_BF != 0xFE) {
-            if (!envCtx->indoors && (envCtx->unk_BF == 0xFF)) {
-                for (i = 0; i < ARRAY_COUNT(D_8011FB48[envCtx->unk_1F]); i++) {
-                    if ((gSaveContext.skyboxTime >= TIME_ENTRY_1F.startTime) &&
-                        ((gSaveContext.skyboxTime < TIME_ENTRY_1F.endTime) || TIME_ENTRY_1F.endTime == 0xFFFF)) {
+        if (envCtx->lightSettingOverride != LIGHT_SETTING_OVERRIDE_NONE - 1) {
```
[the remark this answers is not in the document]
fig02: the -1 is because cutscene commands subtract 1 from the index
`globalCtx->envCtx.lightSettingOverride = cmd->setting - 1;`

I have to refresh myself on the override stuff in general to have more thoughtful opinions, but this is where the -1 comes from at least 

presumably a cutscene would pass 0xFF when it wants to use the light setting from the floor poly, and the cutscene system will subtract 1 to make that 0xFE

include/functions.h
```diff
@@ -884,9 +884,9 @@ void Environment_DrawSunLensFlare(GlobalContext* globalCtx, EnvironmentContext*
                                   GraphicsContext* gfxCtx, Vec3f pos, s32 unused);
 void Environment_DrawLensFlare(GlobalContext* globalCtx, EnvironmentContext* envCtx, View* view,
                                GraphicsContext* gfxCtx, Vec3f pos, s32 unused, s16 scale, f32 colorIntensity,
-                               s16 screenFillAlpha, u8 arg9);
+                               s16 screenFillAlpha, u8 isSun);
```
Dragorn421, line 887:
```suggestion
                               s16 glareStrength, u8 isSun);
```

include/macros.h
```diff
@@ -38,6 +38,8 @@
 #define YEARS_ADULT 17
 #define LINK_AGE_IN_YEARS (!LINK_IS_ADULT ? YEARS_CHILD : YEARS_ADULT)
 
+#define CLOCK_TIME(hr, min) ((s32)(((hr) * 60 + (min)) * 0x10000 / (24 * 60)))
```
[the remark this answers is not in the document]
EllipticEllipsis: The other possibility, which I don't know if anyone's tested, is to put the multiply and divide the other way round: some files in MM seem to do this (z_schedule in particular comes to mind).

src/code/z_kankyo.c
```diff
@@ -18,140 +12,132 @@ typedef enum {
...
-u8 gWeatherMode = 0; // "E_wether_flg"
+u8 gWeatherMode = WEATHER_MODE_CLEAR; // "E_wether_flg"
+
+u8 gSavedNextLightConfig = 0;
 
-u8 D_8011FB34 = 0;
+u8 gInterruptSongOfStorms = false;
 
-u8 D_8011FB38 = 0;
+// Indicates whether the skybox is changing to a different index of the same config (based on time)
```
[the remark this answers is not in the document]
fig02: skyboxes are differentiated by their index into gNormalSkyFiles.
In both `EnvironmentContext` and `TimeBasedSkyboxEntry`
```c
    /* 0x05 */ u8 skybox1Index;
    /* 0x06 */ u8 skybox2Index;
```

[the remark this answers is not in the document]
Dragorn421: My brain is fried oops 😅

include/z64environment.h
```diff
@@ -80,26 +154,26 @@ typedef struct {
...
+    /* 0x19 */ u8 changeSkyboxState;
+    /* 0x1A */ u16 changeSkyboxTimer;
     /* 0x1C */ char unk_1C[0x02];
-    /* 0x1E */ u8 indoors; // when set, day time has no effect on lighting
-    /* 0x1F */ u8 unk_1F; // outdoor light index
-    /* 0x20 */ u8 unk_20; // prev outdoor light index?
-    /* 0x21 */ u8 unk_21;
-    /* 0x22 */ u16 unk_22;
-    /* 0x24 */ u16 unk_24;
+    /* 0x1E */ u8 lightMode; // when true, day time has no effect on lighting
```
Dragorn421, line 162:
```suggestion
    /* 0x1E */ u8 lightMode;
```

include/z64environment.h
```diff
@@ -110,28 +184,27 @@ typedef struct {
...
-    /* 0xDE */ u8 unk_DE; // gloomy sky state
-    /* 0xDF */ u8 lightningMode;
-    /* 0xE0 */ u8 unk_E0; // env sounds state
+    /* 0xBC */ u8 lightBlendEnabled; // only used with `LIGHT_MODE_SETTINGS` or on override
+    /* 0xBD */ u8 lightSetting; // only used with `LIGHT_MODE_SETTINGS` or on override
+    /* 0xBE */ u8 prevLightSetting;
+    /* 0xBF */ u8 lightSettingOverride;
+    /* 0xC0 */ EnvLightSettings lightSettings; // settings for the currently "live" lights
+    /* 0xD6 */ u16 lightBlendRateOverride;
+    /* 0xD8 */ f32 lightBlend;
```
Dragorn421, line 193: if I understand how you have placed the "only used with" comments it's on the main variables?
```suggestion
    /* 0xD8 */ f32 lightBlend; // only used with `LIGHT_MODE_SETTINGS` or on setting override
```

src/code/z_kankyo.c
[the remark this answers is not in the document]
Dragorn421: Narrator: *it didn't match*

include/z64environment.h
fig02: that was my intention yeah, it got kind of spammy putting it on every one. Hoping its obvious which variables go with each other, tho it might be tough.
But yea ill make this change

src/code/z_kankyo.c
```diff
@@ -1538,27 +1654,27 @@ void Environment_DrawLensFlare(GlobalContext* globalCtx, EnvironmentContext* env
...
+                envCtx->glareAlpha = 0.0f;
             }
         }
     }
 
     CLOSE_DISPS(gfxCtx, "../z_kankyo.c", 2750);
 }
 
-f32 func_800746DC(void) {
+f32 Environment_RandCentred(void) {
```
Roman971, line 1677:
```suggestion
f32 Environment_RandCentered(void) {
```

src/code/z_scene.c
```diff
@@ -362,30 +362,31 @@ void func_80098D80(GlobalContext* globalCtx, SceneCmd* cmd) {
...
-        globalCtx->envCtx.timeIncrement = cmd->timeSettings.unk_06;
+        globalCtx->envCtx.timeSpeed = cmd->timeSettings.unk_06;
     } else {
-        globalCtx->envCtx.timeIncrement = 0;
+        globalCtx->envCtx.timeSpeed = 0;
     }
 
     if (gSaveContext.sunsSongState == SUNSSONG_INACTIVE) {
-        gTimeIncrement = globalCtx->envCtx.timeIncrement;
+        gTimeSpeed = globalCtx->envCtx.timeSpeed;
```
Roman971, line 371: I suggest naming the context field in a more specific way, so it doesn't get mistaken as the actual time speed. And based on the usage here, I think it always end up holding the time speed defined in the scene, so maybe we could name it as the "scene time speed" or something along those lines.
```suggestion
        gTimeSpeed = globalCtx->envCtx.sceneTimeSpeed;
```

Similarly I'd suggest naming `R_ENV_TIME_SPEED` around the fact that it doesn't actually get updated or used beyond init, so maybe `R_ENV_TIME_SPEED_INIT`. Or we could just assume it's an old register and name it something like `R_ENV_TIME_SPEED_OLD` to make it less likely to be confused as the real thing

Roman971 (approved):

Dragorn421 (approved):

PR #1117, opened 2022-01-23
Backgrounds

louist103:
Draft until ZAPD can use a palette from another file.


fig02:
can we just open these when zap supports it? we dont know how long that will take, and jenkins still runs on drafts afaik. these are short enough that they can be reviewed when its ready

PR #1119, opened 2022-01-24
proposal for link age macros

playerskel:
Another small macro change that makes reading the code just a bit less magic.

### DIFF
--- a/include/macros.h
+++ b/include/macros.h
@@ -31,8 +31,11 @@
 
 #define GET_ACTIVE_CAM(globalCtx) ((globalCtx)->cameraPtrs[(globalCtx)->activeCamera])
 
-#define LINK_IS_ADULT (gSaveContext.linkAge == 0)
-#define LINK_IS_CHILD (gSaveContext.linkAge == 1)
+#define LINK_ADULT 0
+#define LINK_CHILD 1
+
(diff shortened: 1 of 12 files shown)


include/macros.h
```diff
@@ -31,8 +31,11 @@
 
 #define GET_ACTIVE_CAM(globalCtx) ((globalCtx)->cameraPtrs[(globalCtx)->activeCamera])
 
-#define LINK_IS_ADULT (gSaveContext.linkAge == 0)
-#define LINK_IS_CHILD (gSaveContext.linkAge == 1)
+#define LINK_ADULT 0
+#define LINK_CHILD 1
```
Dragorn421, line 35: I would put this is z64save.h (where `linkAge` is defined), and make it a `LinkAge` enum. Also prefix should probably be `LINK_AGE_`

The comment on `SaveContext#linkAge` could then be removed entirely (enums for specific fields often have the same name as the field)

playerskel: Makes sense. Let me know if this is better.

Dragorn421 (approved):
Looks good to me, unless there's a reason this hasn't been done before

fig02 (approved):
i feel like theres possibly a different approach for age stuff that we were going to take, but i cant remember it if it ever existed or if im confusing it with something else. 
im fine with this either way

Roman971 (approved):

PR #1121, opened 2022-01-26
Document `kanji`-offset-getting file

EllipticEllipsis:
Took a deep dive into `kanji` and worked out what this handwritten function + giant data table in `code` is for. Tested the offsets in NTSC's `kanji` itself, also checked and found it's used by NTSC kanfont to DMA stuff, same idea as `Font_LoadChar`. I renamed the file to be clearer, although open to suggestions for improvement.

I took the new asm macros and documentation style from #865 , so it should fit nicely with that. I de-incbin'd the data since anyone editing `kanji` will also need to edit it, I don't see any particular reason to inbin it, it's easier for multi-version support, and it's rather easier to understand that way. It's currently formatted the same way as the shift-jis codepoints it represents are laid out... I suppose we could end up with `(gKanjiSymbolXYZ - _kanji_segmentStart) / FONT_CHAR_TEX_SIZE` to make it mostly shiftable, but that's miles off.

I included a C interpretation (that is equivalent, but obviously won't match) because there's nothing particularly low-level or fancy about this function, presumably it's just handwritten for speed, and I'm sure most people would rather read a C version than the assembly itself.


asm/sjis.s
```diff
@@ -0,0 +1,206 @@
...
+#             byte2--;
+#         }
+#
+#         return kanjiNonKanjiIndices[byte2 + byte1 * 0xBC] * 0x80;
+#     }
+# }
+
+BEGINDATA kanjiNonKanjiIndices
+    #       ___0    ___1    ___2    ___3    ___4    ___5    ___6    ___7    ___8    ___9    ___A    ___B    ___C    ___D    ___E    ___F
+    .half 0x0000, 0x0001, 0x0002, 0x0003, 0x0004, 0x0005, 0x0006, 0x0007, 0x0008, 0x0009, 0x000A, 0x000B, 0x000C, 0x000D, 0x000E, 0x000F # 814_
```
engineer124, line 116:
```suggestion
    #       __0    __1    __2    __3    __4    __5    __6    __7    __8    __9    __A    __B    __C    __D    __E   __F
    .half 0x000, 0x001, 0x002, 0x003, 0x004, 0x005, 0x006, 0x007, 0x008, 0x009, 0x00A, 0x00B, 0x00C, 0x00D, 0x00E, 0x00F # 814_
```
I think you could remove at least 1 leading zero as there isn't any number in this data that uses 4 digits. I would even suggest removing all leading 0's if the formatter doesn't break the nice alignment you've got going (or you can leave it otherwise)

asm/sjis.s
```diff
@@ -0,0 +1,206 @@
...
+# assembler directives
+.set noat      # allow manual use of $at
+.set noreorder # don't insert nops after branches
+.set gp=64     # allow use of 64-bit general purpose registers
+
+.section .text
+
+.balign 16
+
+# s32 Kanji_OffsetFromShiftJIS(s32 shiftJISCodepoint)
```
engineer124, line 12:
```suggestion
# s32 Kanji_OffsetFromShiftJis(s32 shiftJisCodepoint)
```
This one is probably up for more discussion. I think that abbreviations still use the camel case for names (think Dma instead of DMA, Id instead of ID), but I'm not sure how standardized that is

asm/sjis.s
```diff
@@ -0,0 +1,206 @@
...
+    addiu   $a2, %lo(kanjiNonKanjiIndices)
+    add     $a3, $a3, $a2
+    lh      $a2, ($a3)
+    jr      $ra
+     sll    $v0, $a2, 7
+     # returns kanjiNonKanjiIndices[(adjusted byte2) + (adjusted byte1) * 0xBC] * FONT_CHAR_TEX_SIZE
+END Kanji_OffsetFromShiftJIS
+
+# C pseudocode:
+# s32 Kanji_OffsetFromShiftJIS(s32 arg0) {
```
engineer124, line 85:
```suggestion
# s32 Kanji_OffsetFromShiftJIS(s32 shiftJISCodepoint) {
```
You can match the argument name to the one given above

asm/sjis.s
```diff
@@ -0,0 +1,206 @@
...
+#     s32 byte1;
+#     s32 byte2;
```
engineer124, line 87:
```suggestion
#     s32 upperByte;
#     s32 lowerByte;
```
These names make more sense to me at least on how they graph the upper and lower bytes of `shiftJISCodepoint`

include/macro.inc
```diff
@@ -1,4 +1,65 @@
...
+.endm
+
+# cop0 register aliases
+
+.set Index,         $0
+.set Random,        $1
+.set EntryLo0,      $2
+.set EntryLo1,      $3
+.set Context,       $4
+.set PageMask,      $5
```
engineer124, line 35: Could you explain the context of these variables that are set? Are these generic names or unique ones to `code_800D71F0.s`

spec
```diff
@@ -405,7 +405,7 @@ beginseg
     include "build/src/code/fault.o"
     include "build/data/fault.bss.o"
     include "build/src/code/fault_drawer.o"
-    include "build/asm/code_800D71F0.o"
+    include "build/asm/sjis.o"
```
engineer124, line 408: I'm not sure about the filename `sjis`, it seems to abbreviated. That's suppose to stand for `Shift-JIS` right? Perhaps `shift_jis.s`. Alternatively, you use the `Kanji_` prefix in the function name, so what about `kanji.s`?

engineer124 (commented):
Nice, some comments/ideas

asm/sjis.s
fig02: dont believe formatter is touching .s files so dont think thats a concern (someone please correct me if wrong though)

include/macro.inc
AngheloAlf: Those are the register names of the coprocessor 0 (a.k.a. Cop0).

From what I understand, this file does not need them, but I think Elliptic just took the whole file from Tharo's PR, probably to avoid conflicts.

spec
AngheloAlf: Personally I dont like just `kanji.s` because this function is  specific for the Shift-JIS encoding.

Kanji symbols (glyphs? characters?) are be encoded  differently in different encodings. 
For example, the original debug strings in the repo are encoded in EUC-JP (but we save the C files in UTF-8 and re-encode them during build), but the in-game japanese text is encoded in a modified Shift-JIS encoding (i know they added hylian symbols, replacing something else among other changes)

In summary (i didn't expect to write so much...), I think this file should somehow explicitly mention the Shift-JIS encoding.

include/macro.inc
EllipticEllipsis: Yeah, I needed the macros so I took the whole file to mitigate conflicts as far as possible.

louist103:
I'll review it more later but for now the function prototype can go in `functions.h`

spec
EllipticEllipsis: Yeah, this file is a sort of Shift-JIS++ text -> kanji offsets converter, and can only be used like that (if you want to convert EUC-JP, you'd have to convert it to JIS X 0208 and then Shift-JIS first, for example.

`kanji` is not precisely Shift-JIS, it has some symbols missing and, as @AngheloAlf says, some extra stuff like the Hylian (which I assume is not used at all, since `kanji` is not present on PAL), but it's it's probably about 90% the same as the particular section of Shift-JIS that it covers.

EllipticEllipsis:
> I'll review it more later but for now the function prototype can go in `functions.h`

I'm not sure how worth it this is given that it's unused in PAL (and only used in a couple of files in NTSC, from the looks of it).

asm/sjis.s
EllipticEllipsis: > dont believe formatter is touching .s files so dont think thats a concern (someone please correct me if wrong though)

It shouldn't be, and I doubt it would produce anything meaningful for non-C code (would probably add a final newline).

> I think you could remove at least 1 leading zero as there isn't any number in this data that uses 4 digits. I would even suggest removing all leading 0's if the formatter doesn't break the nice alignment you've got going (or you can leave it otherwise)

I left them like this mainly because they're accessed as halfs. I've thought for a while about how to format this data, and I think this is probably the least offensive way for now. In theory one could label all of them with the symbols, although there are some curious repeats, and holes, and to do so makes the file over 1000 lines long (there are 0x524 entries in this table, I believe).

asm/sjis.s
EllipticEllipsis: I haven't found a particular pattern for how these are usually referred to for Japanese encodings: unlike Unicode, however you store it is always meant to be "big-endian", according to the standard: https://www.sljfaq.org/afaq/encodings.html , but I think as a communication standard they usually refer to first, second, third, etc. bytes (only the first 2 mattering), because while this function only uses 2-byte characters, technically Shift-JIS should cover the 1-byte ones as well (rather like how Unicode can use as many bytes as it wants for a character, but much simpler). Hence the way you identify characters is theoretically supposed to be from reading the stream of bytes rather than wider types (and yes, since this function doesn't do it, it is big-endian only without appropriate wrapping, although that's inevitable since it's handwritten MIPS).

asm/sjis.s
```diff
@@ -0,0 +1,206 @@
...
+#         byte2 -= 0x40;
+#         if (byte2 >= 0x40) {
+#             byte2--;
+#         }
+#
+#         return kanjiNonKanjiIndices[byte2 + byte1 * 0xBC] * 0x80;
+#     }
+# }
+
+BEGINDATA kanjiNonKanjiIndices
```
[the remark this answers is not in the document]
EllipticEllipsis: I don't think this makes sense, since you can't limit the scope of symbols in assembly, right?

asm/sjis.s
```diff
@@ -0,0 +1,206 @@
...
+.set noreorder # don't insert nops after branches
+.set gp=64     # allow use of 64-bit general purpose registers
+
+.section .text
+
+.balign 16
+
+# s32 Kanji_OffsetFromShiftJIS(s32 shiftJISCodepoint)
+# 
+# Returns the offset of the glyph texture data in `kanji` corresponding to a
```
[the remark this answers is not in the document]
EllipticEllipsis: `kanji` is the file that holds the font textures that Japanese uses. TBH I thought PAL had it too, but apparently it's NTSC-only 😅

asm/sjis.s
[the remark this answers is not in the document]
Dragorn421: afaik the `s`/`g` prefix is up to whether the data is used locally or not, not whether it is declared static or not (though it's usually tied)

asm/sjis.s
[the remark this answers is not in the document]
Dragorn421: so this code is present in PAL without some of the data it is meant to be used in conjunction with? Nice

Also I'd suggest making it clear it's a file you're refering to:
```suggestion
# Returns the offset of the glyph texture data in the `kanji` file corresponding to a
```
maybe also worth adding in the other spots mentioning `kanji`

asm/sjis.s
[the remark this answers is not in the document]
Thar0: You can restrict visibility of symbols in asm files:
- Symbols are by default local
- Symbols aren't global unless you say so with `.global` (`BEGIN` does this)
- Symbols starting with `.L` don't appear in the object's symbol table at all

But yeah it's not actually important here, there are plenty of `s`-prefixed variables even in C that aren't marked static, it's just to indicate it isn't used outside of the file it's in.

spec
EllipticEllipsis: Renamed it based on the original name of the n64dd library file that contains this function.

asm/kanread.s
```diff
@@ -0,0 +1,206 @@
...
+    srl     $a1, $a0, 8
+    addi    $a1, $a1, -0x81
+    multu   $a2, $a1
+    # Get byte2 and adjust so starts at 0
+    andi    $a3, $a0, 0xff
+    addi    $a3, $a3, -0x40
+    slti    $at, $a3, 0x40
+    mflo    $a2
+    # 0xXX7F is always empty and elided in the file, so decrement
+    bnezl   $at, .not_kanji_lower_block
```
Dragorn421, line 68: I initially misread this label as "not kanji-lower-block" instead of "not-kanji lower-block". But I don't have a suggestion to prevent that confusion

asm/kanread.s
```diff
@@ -0,0 +1,206 @@
...
+#             byte2--;
+#         }
+#
+#         return sKanjiNonKanjiIndices[byte2 + byte1 * 0xBC] * 0x80;
+#     }
+# }
+
+BEGINDATA sKanjiNonKanjiIndices
+    #       ___0    ___1    ___2    ___3    ___4    ___5    ___6    ___7    ___8    ___9    ___A    ___B    ___C    ___D    ___E    ___F
+    .half 0x0000, 0x0001, 0x0002, 0x0003, 0x0004, 0x0005, 0x0006, 0x0007, 0x0008, 0x0009, 0x000A, 0x000B, 0x000C, 0x000D, 0x000E, 0x000F # 814_
```
Dragorn421, line 116: hexadecimal purist suggestion
```suggestion
    #     0x___0  0x___1  0x___2  0x___3  0x___4  0x___5  0x___6  0x___7  0x___8  0x___9  0x___A  0x___B  0x___C  0x___D  0x___E  0x___F
    .half 0x0000, 0x0001, 0x0002, 0x0003, 0x0004, 0x0005, 0x0006, 0x0007, 0x0008, 0x0009, 0x000A, 0x000B, 0x000C, 0x000D, 0x000E, 0x000F # 0x814_
```

Dragorn421 (commented):
What is the reasoning to name the file `kanread`? Imo spelling out `kanji` in full would be good, and what about the `read` part?

Also imo it would make more sense to merge Tharo's #865 first even if that means this PR gets to stand around for a bit, for better context of how/why macro.inc gets a revamp and the asm style. Tharo also introduced using cpp on asm files, so this PR could also then use C style comments.

Also, I'm not a big fan of including C equivalents of assembly functions. One reason is, it's kinda hard to make sure they are equivalent, and it feels like advanced enough doc that it would fit in a wiki more than in the repo imo.
The idea of having custom code in the repo under src/ also just isn't appealing to me at all, even commented or ifdef'ed out

asm/kanread.s
EllipticEllipsis: Yeah, there's not really anything that ties this section together apart from the glyphs not being kanji. I suppose `non_kanji` or `nonkanji` would be an option, but I don't think it's really an improvement.

asm/kanread.s
EllipticEllipsis: I suppose, although does make the table even wider.

EllipticEllipsis:
> What is the reasoning to name the file `kanread`? Imo spelling out `kanji` in full would be good, and what about the `read` part?

This is the name of the file containing the function on the oldest version of the n64dd library we have, which is the only one I'm currently aware of that uses this form of the function, without range checks (we have some newer versions that do have range checks, and are in their own file, called `getkadr`, which is probably worse). Presumably it's called `kanread` because it'll be used for reading the image data?

> Also imo it would make more sense to merge Tharo's #865 first even if that means this PR gets to stand around for a bit, for better context of how/why macro.inc gets a revamp and the asm style. Tharo also introduced using cpp on asm files, so this PR could also then use C style comments.

I'm happy to wait for #865 : this PR is pretty non-urgent since it's dealing with a system that doesn't even work in the only version we currently support. I'm not sure we get a huge amount of benefit from C-style comments, though, since asm already has line comments, and it's not often we want block comments.

> Also, I'm not a big fan of including C equivalents of assembly functions. One reason is, it's kinda hard to make sure they are equivalent, and it feels like advanced enough doc that it would fit in a wiki more than in the repo imo. The idea of having custom code in the repo under src/ also just isn't appealing to me at all, even commented or ifdef'ed out

I would agree if it wasn't possible to write C that was sufficiently close to equivalent, but this is not some low-level weird OS function like a lot of the handwritten ones in libultra. It's quite possible to write a C function that's provably equivalent in this case (proof: I compiled this and tested it on every valid input. Non-valid inputs are UB anyway since the surroundings of the `kanji` file can easily be different in different roms, although they probably actually aren't.) The only wrinkle is the use of `add`, but being out-of-bounds in the array is going to bite you with UB well before the overflow exceptions come into play.

Also, this is basically a literal translation of the assembly: I translated it line-by-line and removed the unnecessary temps. It's obvious that it's easy to write a function that's shorter and easier to read, which is not what I did; I'm not really sure there's been much interpretation on my part.

Of course if people really don't want it, I can remove it, but I do think for the typical user of the repo going forward, who most likely does not speak MIPS, this just makes this file much more approachable.

asm/kanread.s
Dragorn421: It's only two characters wider with the `0x` in the `0x814_`, sounds reasonable

asm/kanread.s
Dragorn421: maybe `not_kanji__lower_block` (notice the double underscore in the middle)

asm/kanread.s
Dragorn421: to be clear
```
    #       ___0    ___1    ___2    ___3    ___4    ___5    ___6    ___7    ___8    ___9    ___A    ___B    ___C    ___D    ___E    ___F
```
->
```
    #     0x___0  0x___1  0x___2  0x___3  0x___4  0x___5  0x___6  0x___7  0x___8  0x___9  0x___A  0x___B  0x___C  0x___D  0x___E  0x___F
```
is the exact same amount of space

asm/kanread.s
```diff
@@ -0,0 +1,209 @@
+.include "macro.inc"
```
Thar0, line 1:
```suggestion
```

asm/kanread.s
```diff
@@ -0,0 +1,209 @@
...
+#include "ultra64/asm.h"
+
+// assembler directives
+.set noat
+.set noreorder
+.set gp=64
```
Thar0, line 7:
```suggestion
```

asm/kanread.s
```diff
@@ -0,0 +1,209 @@
...
+    // so convert index directly.
+    li      $at, 0x8800
+    slt     $at, $a0, $at
+    bnez    $at, .nonkanji
+     // 0xBC is number of glyphs in one block in the `kanji` file:
+     // 0x100 possible codepoints with the same byte1
+     // - 0x40 unused at beginning 
+     // - 1 unused at 0x7F 
+     // - 3 unused at 0xFD, 0xFE, 0xFF
+     li     $a2, 0xBC
```
Thar0, line 37: Would it help modding and/or readability if defines were made for this number? I'd also want to suggest a define for the texture size but since it's encoded in a shift I recognise that may not be as useful. Might still be nice to have the shift amount in one place rather than two though.

Thar0 (commented):
Some little things

asm/kanread.s
```diff
@@ -0,0 +1,209 @@
...
+    // Get byte1 and adjust so starts at 0
+    srl     $a1, $a0, 8
+    addi    $a1, $a1, -0x88
+    multu   $a2, $a1
+    // Get byte2 and adjust so starts at 0
+    andi    $a3, $a0, 0xff
```
Dragorn421, line 43:
```suggestion
    andi    $a3, $a0, 0xFF
```
I think this is the only lowercase hex in the file

asm/kanread.s
```diff
@@ -0,0 +1,209 @@
...
+    addi    $a3, $a3, -0x40
+    slti    $at, $a3, 0x40
```
Dragorn421, line 45: This could be
```suggestion
    slti    $at, $a3, (0x7F + 1 - 0x40)
```
to highlight that it's checking against 0x7F but after the `- 0x40`
but idk, no big deal

asm/kanread.s
```diff
@@ -0,0 +1,209 @@
...
+    add     $a3, $a3, $a2
+    lh      $a2, ($a3)
+    jr      $ra
+     sll    $v0, $a2, 7
+     // returns sKanjiNonKanjiIndices[(adjusted byte2) + (adjusted byte1) * 0xBC] * FONT_CHAR_TEX_SIZE
+END(Kanji_OffsetFromShiftJIS)
+
+// Nearly equivalent C code (Equivalent for all valid input, will behave differently on overflow due to the use of `add` in the original):
+// s32 Kanji_OffsetFromShiftJIS(s32 shiftJISCodepoint) {
+//     s32 byte1;
```
Dragorn421, line 89: While we're at it why not write it more naturally, like idk:
```c
extern s16 sKanjiNonKanjiIndices[];

s32 Kanji_OffsetFromShiftJIS(s32 shiftJISCodepoint) {
    s32 byte1 = shiftJISCodepoint >> 8;
    s32 byte2 = shiftJISCodepoint & 0xFF;

    if (shiftJISCodepoint >= 0x8800) {
        byte1 -= 0x88;
        if (byte2 > 0x7F) {
            byte2--;
        }
        byte2 -= 0x40;

        return (0x30A + byte1 * 0xBC + byte2) * FONT_CHAR_TEX_SIZE;
    } else {
        byte1 -= 0x81;
        if (byte2 > 0x7F) {
            byte2--;
        }
        byte2 -= 0x40;

        return sKanjiNonKanjiIndices[byte1 * 0xBC + byte2] * FONT_CHAR_TEX_SIZE;
    }
}
```
(take or leave anything you want from this)

asm/kanread.s
```diff
@@ -0,0 +1,209 @@
...
+//         if (byte2 >= 0x40) {
+//             byte2--;
+//         }
+//
+//         return sKanjiNonKanjiIndices[byte2 + byte1 * 0xBC] * 0x80;
+//     }
+// }
+
+DATA(sKanjiNonKanjiIndices)
+//                 0x___0  0x___1  0x___2  0x___3  0x___4  0x___5  0x___6  0x___7  0x___8  0x___9  0x___A  0x___B  0x___C  0x___D  0x___E  0x___F
```
Dragorn421, line 118: I would add a comment about the (good) way this is formatted, something along the lines of
```suggestion

// Formatting follows the mapping from Shift-JIS codepoints,
// and the data is divided into blocks by high byte of the codepoint
DATA(sKanjiNonKanjiIndices)
//                 0x___0  0x___1  0x___2  0x___3  0x___4  0x___5  0x___6  0x___7  0x___8  0x___9  0x___A  0x___B  0x___C  0x___D  0x___E  0x___F
```
I'm not convinced by my own comment but something like that

asm/kanread.s
EllipticEllipsis: I feel like `stli ... (0x80 - 0x40)` might be a bit clearer since it's a < rather than <=, but I suppose it's constrained by the available instructions. Of course it never checks for 0x7F itself anyway.

asm/kanread.s
EllipticEllipsis: I mainly wrote it the way that I did to make it clear how it corresponds to the assembly (since it does a lot of redundant stuff). If I were writing it as "good" C I'd move the byte2 stuff out of the if to avoid repeating it anyway, I think it's better to go one way or the other rather than only alter some of it.

fig02: if our reasoning for including it is to make it easier to understand, Id say that keeping its resemblance to the assembly is not that important. It would help if someone wants to read the assembly along with the c code to double check your work i guess, but the reasoning seems to mostly be accessibility and ease of use/understanding

Dragorn421 (approved):

fig02 (approved):

asm/kanread.s
```diff
@@ -0,0 +1,217 @@
...
+// assembler directives
+.set noat
+.set noreorder
+
+.section .text
+
+.balign 16
+
+/**
+ * s32 Kanji_OffsetFromShiftJIS(s32 sjis);
```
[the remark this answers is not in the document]
EllipticEllipsis: I weigh this up carefully, decided that u32 would make sense, then looked in the libleo headers and discovered it had this prototype.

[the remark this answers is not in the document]
Roman971: I see, then I guess we should probably keep this until we match code using it in other versions, but I would still support changing them to u32 eventually if they end up working (and maybe even size_t for the return type?)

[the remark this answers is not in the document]
EllipticEllipsis: `size_t` is probably the correct type for the return. But it is worth bearing in mind that `n64dd` on N64 versions has another version of this function: https://github.com/Roman971/oot-disassembly/blob/master/system/n64dd.s#L9577-L9797 , apparently straight from `libleo.a`, for which we'll want to use the headers anyway, so it might be better to stay consistent (assuming that all this matches anyway).

fig02 (approved):

Roman971 (approved):

PR #1122, opened 2022-01-26
Document Actor Flags

fig02:
This gives a name to every actor flag.

Some minor things were changed along with the flags, for example skyboxOffset -> quakeOffset which was discovered while figuring out how the flag worked


include/z64actor.h
```diff
@@ -100,34 +100,34 @@ typedef struct {
...
-#define ACTOR_FLAG_20 (1 << 20)
-#define ACTOR_FLAG_21 (1 << 21)
-#define ACTOR_FLAG_22 (1 << 22)
-#define ACTOR_FLAG_23 (1 << 23)
-#define ACTOR_FLAG_24 (1 << 24)
-#define ACTOR_FLAG_25 (1 << 25)
-#define ACTOR_FLAG_26 (1 << 26)
-#define ACTOR_FLAG_27 (1 << 27)
-#define ACTOR_FLAG_28 (1 << 28)
+#define ACTOR_FLAG_TARGETABLE (1 << 0) // navi will fly over the actor and it can be Z targeted
```
EllipticEllipsis, line 103:
```suggestion
#define ACTOR_FLAG_TARGETABLE (1 << 0) // Navi will fly over the actor and it can be Z targeted
```

include/z64actor.h
```diff
@@ -100,34 +100,34 @@ typedef struct {
...
+#define ACTOR_FLAG_UNFRIENDLY (1 << 2) // changes targeting behavior for unfriendly actors (sound, links stance)
```
EllipticEllipsis, line 104:
```suggestion
#define ACTOR_FLAG_UNFRIENDLY (1 << 2) // changes targeting behavior for unfriendly actors (sound, Player's stance)
```
Possibly?

include/z64actor.h
```diff
@@ -100,34 +100,34 @@ typedef struct {
...
+#define ACTOR_FLAG_FRIENDLY (1 << 3) // opposite of the unfriendly flag. flag is not checked against in the original game
+#define ACTOR_FLAG_NO_UPDATE_CULLING (1 << 4) // actor will not stop updating even if outside of the uncull zone
+#define ACTOR_FLAG_NO_DRAW_CULLING (1 << 5) // actor will not stop drawing even if outside of the uncull zone
+#define ACTOR_FLAG_NOT_CULLED (1 << 6) // actor is in the uncull zone. flag still updates regardless of the update/draw cull exclusion flags
+#define ACTOR_FLAG_REACT_TO_LENS (1 << 7) // hidden or revealed by lens of truth (depending on room flags)
```
EllipticEllipsis, line 109:
```suggestion
#define ACTOR_FLAG_REACT_TO_LENS (1 << 7) // hidden or revealed by Lens of Truth (depending on room flags)
```

include/z64actor.h
```diff
@@ -100,34 +100,34 @@ typedef struct {
...
+#define ACTOR_FLAG_TALK_REQUESTED (1 << 8) // player has requested to talk to the actor
+#define ACTOR_FLAG_HOOK_CAN_CARRY (1 << 9) // brings the actor back to player if hookshoted
+#define ACTOR_FLAG_HOOK_BRING_PLAYER (1 << 10) // brings player to the actor if hookshoted
```
EllipticEllipsis, line 112:
```suggestion
#define ACTOR_FLAG_HOOK_CAN_CARRY (1 << 9) // brings the actor back to player if Hookshoted
#define ACTOR_FLAG_HOOK_BRING_PLAYER (1 << 10) // brings player to the actor if Hookshoted
```
I also don't like abbreviating Hookshot much, but I'm sure most people will disagree.

include/z64actor.h
```diff
@@ -100,34 +100,34 @@ typedef struct {
...
+#define ACTOR_FLAG_ENKUSA_CUT (1 << 11) // grass actor has been cut
+#define ACTOR_FLAG_IGNORE_QUAKE (1 << 12) // actor will not shake when a quake occurs
+#define ACTOR_FLAG_HOOK_ATTACHED (1 << 13) // hookshot has attached to the actor (either a collider or surface)
+#define ACTOR_FLAG_ARROW_CAN_CARRY (1 << 14) // when an arrow hits the actor it will attach to the actor and carry it
```
EllipticEllipsis, line 116: What attaches to what here? Name sounds like it would be the other way round from the description.

include/z64actor.h
```diff
@@ -100,34 +100,34 @@ typedef struct {
...
+#define ACTOR_FLAG_ARROW_IS_CARRYING (1 << 15) // an arrow is currently carrying this actor
+#define ACTOR_FLAG_IMMEDIATE_TALK (1 << 16) // forces player to talk when in range. needs to be unset manually to avoid infinite talking
+#define ACTOR_FLAG_CARRY_SET_X_ROT (1 << 17) // copies links x rotation instead of y when holding the actor
```
EllipticEllipsis, line 119:
```suggestion
#define ACTOR_FLAG_CARRY_SET_X_ROT (1 << 17) // copies Player's x rotation instead of y when holding the actor
```

include/z64actor.h
```diff
@@ -100,34 +100,34 @@ typedef struct {
...
+#define ACTOR_FLAG_CHECK_WITH_NAVI (1 << 18) // navi can be used to trigger dialouge when targeting the actor
```
EllipticEllipsis, line 120:
```suggestion
#define ACTOR_FLAG_CHECK_WITH_NAVI (1 << 18) // Navi can be used to trigger dialogue when targeting the actor
```

include/z64actor.h
```diff
@@ -100,34 +100,34 @@ typedef struct {
...
+#define ACTOR_FLAG_SFX_AT_POS (1 << 19) // play sound from sfx field at the actor's position
+#define ACTOR_FLAG_SFX_CENTERED2 (1 << 20) // same as ACTOR_FLAG_SFX_CENTERED, unused in the original game
+#define ACTOR_FLAG_SFX_CENTERED (1 << 21) // play sound from sfx field at the center of the screen
+#define ACTOR_FLAG_IGNORE_POINT_LIGHTS (1 << 22) // ignores point lights but can still be lit by environment lights
+#define ACTOR_FLAG_ALWAYS_THROW (1 << 23) // player throws held actor even if standing still
+#define ACTOR_FLAG_PLAY_BODYHIT_SFX (1 << 24) // when actor hits players body, a thump sfx plays
```
EllipticEllipsis, line 126:
```suggestion
#define ACTOR_FLAG_PLAY_BODYHIT_SFX (1 << 24) // when actor hits Player's body, a thump sfx plays
```

EllipticEllipsis (commented):
Just a lot of grammar in the new comments in z64actor.h for now. It seems to me that it should be `Player` over `player` everywhere, too, since it's referring to the actual actor (and not the person at the controls).

include/z64actor.h
fig02: its just a generic "hook is attached to the actor" and doesnt say anything about the direction of who is getting pulled where

include/z64actor.h
fig02: i dont think this needs to be capitalized here, its not a name of a thing at this point

include/z64actor.h
```diff
@@ -100,34 +100,34 @@ typedef struct {
...
-#define ACTOR_FLAG_25 (1 << 25)
-#define ACTOR_FLAG_26 (1 << 26)
-#define ACTOR_FLAG_27 (1 << 27)
-#define ACTOR_FLAG_28 (1 << 28)
+#define ACTOR_FLAG_TARGETABLE (1 << 0) // Navi will fly over the actor and it can be Z targeted
+#define ACTOR_FLAG_UNFRIENDLY (1 << 2) // changes targeting behavior for unfriendly actors (sound, links stance)
+#define ACTOR_FLAG_FRIENDLY (1 << 3) // opposite of the unfriendly flag. flag is not checked against in the original game
+#define ACTOR_FLAG_NO_UPDATE_CULLING (1 << 4) // actor will keep updating even if outside of the uncull zone
+#define ACTOR_FLAG_NO_DRAW_CULLING (1 << 5) // actor will keep drawing even if outside of the uncull zone
+#define ACTOR_FLAG_NOT_CULLED (1 << 6) // actor is in the uncull zone. flag still updates regardless of the update/draw cull exclusion flags
```
fig02, line 108: I would be fine with changing "uncull zone" to "active zone" in general, it may help with all the negative words going on here that make this system confusing. uncull zone is a name i was never happy with since the beginning.
in that case I would also change `ACTOR_FLAG_NOT_CULLED` -> `ACTOR_FLAG_IN_ACTIVE_ZONE`

fig02: also, long conversation regarding these flags in general for anyone else who wants to talk about these https://discord.com/channels/688807550715560050/688851317593997489/935729171529887754

imo i would rather keep the first two as is, and rename the last one according to my suggestion above

Dragorn421: A summary of how these three flags work:

An actor is "culled" when it is outside the "uncull zone" defined by the three `uncullZone*` members in the `Actor` struct.

An actor being culled or not is indicated by `ACTOR_FLAG_6`/`ACTOR_FLAG_IN_UNCULL_ZONE` being set (actor is not culled) or cleared (actor is culled).

The set/clear state of this flag is updated every frame between actor update and actor draw (right before draw):
https://github.com/fig02/oot/blob/7774a67ea79076099820a6386a6a5b21c2ee5949/src/code/z_actor.c#L2421-L2425 :
```c
if (func_800314B0(globalCtx, actor)) {
    actor->flags |= ACTOR_FLAG_IN_UNCULL_ZONE;
} else {
    actor->flags &= ~ACTOR_FLAG_IN_UNCULL_ZONE;
}
```

When an actor is culled, by default its update and draw functions don't run.
This behavior can be overriden by setting the `ACTOR_FLAG_4`/`ACTOR_FLAG_NO_UPDATE_CULLING` and/or `ACTOR_FLAG_5`/`ACTOR_FLAG_NO_DRAW_CULLING` to make update and draw run regardless of the actor being culled.

-----------

Comments:

`ACTOR_FLAG_NO_UPDATE_CULLING`/draw one can't be "always update"/draw because frozen (`freezeTimer`) actors still don't update. This set of flag is "culling"-specific

"no X culling" means "do not apply culling for x" if that's clearer. Wording is hard, suggestions appreciated

Suggestions also appreciated on the "cull", "uncull" words if they can be replaced

src/overlays/actors/ovl_Arms_Hook/z_arms_hook.c
```diff
@@ -171,10 +172,11 @@ void ArmsHook_Shoot(ArmsHook* this, GlobalContext* globalCtx) {
...
         (this->collider.info.atHitInfo->elemType != ELEMTYPE_UNK4)) {
         touchedActor = this->collider.base.at;
-        if ((touchedActor->update != NULL) && (touchedActor->flags & (ACTOR_FLAG_9 | ACTOR_FLAG_10))) {
+        if ((touchedActor->update != NULL) &&
+            (touchedActor->flags & (ACTOR_FLAG_HOOK_CAN_CARRY | ACTOR_FLAG_HOOK_BRING_PLAYER))) {
             if (this->collider.info.atHitInfo->bumperFlags & BUMP_HOOKABLE) {
                 ArmsHook_AttachHookToActor(this, touchedActor);
-                if (CHECK_FLAG_ALL(touchedActor->flags, ACTOR_FLAG_10)) {
+                if (CHECK_FLAG_ALL(touchedActor->flags, ACTOR_FLAG_HOOK_BRING_PLAYER)) {
                     func_80865044(this);
```
Dragorn421, line 180: Again, relying on undocumented code makes it hard to understand the flags

But I can't say I understand even the documented parts of Hookshot anyway

`ArmsHook#child != NULL` seems to indicate pulling player to wherever hookshot is

For pulling something to player I can't figure out how it works, seems to pull in the direction `grabbedDistDiff` holds, but that holds the hookshot to hooked thing difference, which sounds like it would be different from the player to hookshot direction in many cases (for example hitting the side of an object with hookshot)

src/code/z_actor.c
```diff
@@ -2044,7 +2047,7 @@ void Actor_UpdateAll(GlobalContext* globalCtx, ActorContext* actorCtx) {
     sp80 = &D_80116068[0];
 
     if (player->stateFlags2 & PLAYER_STATE2_27) {
-        unkFlag = ACTOR_FLAG_25;
+        unkFlag = ACTOR_FLAG_OCARINA_NO_FREEZE;
```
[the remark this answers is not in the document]
fig02: where does 6 come from? `PLAYER_STATE2_27` here is the important one and this gets set both in the player function that interprets the ocarina action parameter to then actually play the ocarina, and again at the relevant code where demo_kankyo (warp song sparkles) are spawned. 

in general I think im more comfortable naming these things cause ive crawled around player alot, even though its not technically documented in the codebase yet

src/code/z_actor.c
```diff
@@ -2932,11 +2938,13 @@ void func_800328D4(GlobalContext* globalCtx, ActorContext* actorCtx, Player* pla
     sp84 = player->unk_664;
 
     while (actor != NULL) {
-        if ((actor->update != NULL) && ((Player*)actor != player) && CHECK_FLAG_ALL(actor->flags, ACTOR_FLAG_0)) {
+        if ((actor->update != NULL) && ((Player*)actor != player) &&
+            CHECK_FLAG_ALL(actor->flags, ACTOR_FLAG_TARGETABLE)) {
```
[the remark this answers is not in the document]
fig02: targetable encompasses the whole mechanism of navi flying over something and the player being able to target it. 
the CANT_TARGET flag makes it so you cant target it, but navi can still fly over it. I agree that its confusing but didnt have any other way to phrase it
a better way to phrase it might be CANT_LOCK_ON or something, but people would generally call that targeting

src/code/z_player_lib.c
```diff
@@ -1277,23 +1277,23 @@ void func_80090D20(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3s*
...
                 } else if (this->stateFlags1 & PLAYER_STATE1_11) {
                     Vec3s spB8;
 
                     Matrix_Get(&sp14C);
                     Matrix_MtxFToYXZRotS(&sp14C, &spB8, 0);
 
-                    if (hookedActor->flags & ACTOR_FLAG_17) {
-                        hookedActor->world.rot.x = hookedActor->shape.rot.x = spB8.x - this->unk_3BC.x;
+                    if (heldActor->flags & ACTOR_FLAG_CARRY_SET_X_ROT) {
+                        heldActor->world.rot.x = heldActor->shape.rot.x = spB8.x - this->unk_3BC.x;
```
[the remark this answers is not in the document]
fig02: it is specifically for heavy block yes. it gives the illusion of the block being top heavy, tipping back and forth as he picks it up. TOP_HEAVY was my first name for it, cause it could in theory be used for similar scenarios
but yeah I wouldnt be against naming it after the golden gauntlets block

src/code/z_actor.c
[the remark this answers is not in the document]
Dragorn421: the ones in `D_80116068`
I guess it's not directly related but it's in the alternative conditional so I wouldn't immediately dismiss it as unimportant

> this gets set both in the player function that interprets the ocarina action parameter to then actually play the ocarina, and again at the relevant code where demo_kankyo (warp song sparkles) are spawned.

thanks for the details, all I am seeing is `unk`s and `func_`s xP
In particular with the one I assume is ocarina action parameter, I have no clue how to tell that's what is going on

src/code/z_actor.c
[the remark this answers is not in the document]
Dragorn421: maybe "target_focus_only"? something that makes it sound like it's taking a specific thing away compared to the targeting flag alone

src/code/z_actor.c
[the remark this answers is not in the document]
fig02: `D_80116068` are some other conditions where actors can potentially freeze (skip updating this cycle), but its not related to the actor flag/checked against it. the actor flag is specifically for player using the ocarina or warping away 

interestingly `D_80116068` has entries for all categories but only the first one is used. so i guess different actor categories were going to freeze under certain conditions and they ended up just going with the first one, idk

src/code/z_actor.c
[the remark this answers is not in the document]
fig02: i would argue focus sounds like the targeting/locking on part, which is what is being disabled here

src/code/z_actor.c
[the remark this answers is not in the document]
fig02: elaborating a bit more, the first part in my image here is my current interpretation of that conditional.
the second is the flags contained in `D_80116068` according to the speedrunning communities current state flags documentation (and some quick research of my own). While it should not be taken as fact, these are pretty clearly all the player state conditions you would want an actor to be frozen

so essentially, this block in general is dealing with freezing actors. and being frozen from ocarina is the only one that is configurable via this actor flag. the rest are baked into the system
image

src/overlays/actors/ovl_Arms_Hook/z_arms_hook.c
fig02: yeah im not sure on the exact implementation details of how the hookshot pulls the player. but based on context and usage i think its safe to say that this is correct. `func_80865044` which does the child/parent handling gets called if `SurfaceType_IsHookshotSurface` passes, which is how general hookshottable collision that pulls player is handled. and if an actor has this flag set it also calls `func_80865044` 

plenty of other context shows that this is its usage, ganondorf enables it when hes vulnerable and the player can hookshot over to him for example. The scarecrows enable this by default etc

fig02: and ofc tested in game
https://user-images.githubusercontent.com/22045047/151274339-4daaeba5-c949-4bd0-9000-f5db3a0226d0.mp4

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -1151,7 +1151,7 @@ s32 func_80832224(Player* this) {
 s32 func_8083224C(GlobalContext* globalCtx) {
     Player* this = GET_PLAYER(globalCtx);
 
-    return CHECK_FLAG_ALL(this->actor.flags, ACTOR_FLAG_8);
+    return CHECK_FLAG_ALL(this->actor.flags, ACTOR_FLAG_TALK_REQUESTED);
```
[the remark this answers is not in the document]
fig02: Players usage of the flag was considered while naming, but decided it wasnt worth generalizing the flag name for it. Its specific to player and feels like more of a hack than anything. Teh request name makes sense for every other actor who will actually be doing the talking

src/overlays/actors/ovl_En_Si/z_en_si.c
```diff
@@ -80,7 +80,7 @@ s32 func_80AFB748(EnSi* this, GlobalContext* globalCtx) {
 void func_80AFB768(EnSi* this, GlobalContext* globalCtx) {
     Player* player = GET_PLAYER(globalCtx);
 
-    if (CHECK_FLAG_ALL(this->actor.flags, ACTOR_FLAG_13)) {
+    if (CHECK_FLAG_ALL(this->actor.flags, ACTOR_FLAG_HOOK_ATTACHED)) {
```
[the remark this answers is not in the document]
fig02: imo en_si doesnt really care about that, the hack is the boomerang using it

include/z64actor.h
```diff
@@ -100,34 +100,34 @@ typedef struct {
...
+#define ACTOR_FLAG_HOOK_ATTACHED (1 << 13) // hookshot has attached to the actor (either a collider or surface)
+#define ACTOR_FLAG_ARROW_CAN_CARRY (1 << 14) // when an arrow hits the actor it will attach to the actor and carry it
+#define ACTOR_FLAG_ARROW_IS_CARRYING (1 << 15) // an arrow is currently carrying this actor
+#define ACTOR_FLAG_IMMEDIATE_TALK (1 << 16) // forces Player to talk when in range. needs to be unset manually to avoid infinite talking
+#define ACTOR_FLAG_HEAVYBLOCK (1 << 17) // changes actor carrying behavior specifically for the golden gauntlets block actor
+#define ACTOR_FLAG_CHECK_WITH_NAVI (1 << 18) // Navi can be used to trigger dialogue when targeting the actor
+#define ACTOR_FLAG_SFX_AT_POS (1 << 19) // play sound from sfx field at the actor's position
+#define ACTOR_FLAG_SFX_CENTERED2 (1 << 20) // same as ACTOR_FLAG_SFX_CENTERED, unused in the original game
+#define ACTOR_FLAG_SFX_CENTERED (1 << 21) // play sound from sfx field at the center of the screen
+#define ACTOR_FLAG_IGNORE_POINT_LIGHTS (1 << 22) // ignores point lights but can still be lit by environment lights
```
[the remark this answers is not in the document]
fig02: sure, but I think the point is clearly to make it use environment lights only when looking at it contextually
and theres also the fact that the only two directional lights used in the game are the environment lights. but yeah obv that may not be the case in other contexts

src/overlays/actors/ovl_player_actor/z_player.c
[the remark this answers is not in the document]
fig02: added it to the comment though to make that difference more visible

include/z64actor.h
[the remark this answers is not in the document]
Dragorn421: I should have elaborated sorry, my point was the flag does not purposefully not ignore environment lights in how it's implemented, it's just that environment lights happen to be directional lights.

I get the purpose is to make the actor fit seamlessly in the room geometry, so this is kind of a hack (it would break the intent to use a point light to light the room geometry)

src/overlays/actors/ovl_En_Si/z_en_si.c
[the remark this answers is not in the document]
Dragorn421: I feel like someone reading ensi code would appreciate knowing this also handles boomerang. I think it's worth documenting since the code works with more than hookshot (kind of unintendedly so?)

src/overlays/actors/ovl_En_Bb/z_en_bb.c
```diff
@@ -1232,7 +1233,7 @@ void EnBb_Update(Actor* thisx, GlobalContext* globalCtx2) {
     if (this->actor.colChkInfo.damageEffect != 0xD) {
         this->actionFunc(this, globalCtx);
         if ((this->actor.params <= ENBB_BLUE) && (this->actor.speedXZ >= -6.0f) &&
-            ((this->actor.flags & ACTOR_FLAG_15) == 0)) {
+            ((this->actor.flags & ACTOR_FLAG_ARROW_IS_CARRYING) == 0)) {
```
Dragorn421, line 1236:
```suggestion
            !(this->actor.flags & ACTOR_FLAG_ARROW_IS_CARRYING)) {
```

src/overlays/actors/ovl_Elf_Msg2/z_elf_msg2.c
```diff
@@ -134,7 +135,8 @@ void ElfMsg2_WaitUntilActivated(ElfMsg2* this, GlobalContext* globalCtx) {
     if ((this->actor.world.rot.y >= 0x41) && (this->actor.world.rot.y <= 0x80) &&
         (Flags_GetSwitch(globalCtx, (this->actor.world.rot.y - 0x41)))) {
         ElfMsg2_SetupAction(this, ElfMsg2_WaitForTextRead);
-        this->actor.flags |= ACTOR_FLAG_0 | ACTOR_FLAG_18; // Make actor targetable and Navi-checkable
+        this->actor.flags |=
+            ACTOR_FLAG_TARGETABLE | ACTOR_FLAG_CHECK_WITH_NAVI; // Make actor targetable and Navi-checkable
```
Dragorn421, line 139:
```suggestion
            ACTOR_FLAG_TARGETABLE | ACTOR_FLAG_CHECK_WITH_NAVI;
```

include/z64actor.h
```diff
@@ -100,34 +100,34 @@ typedef struct {
...
+#define ACTOR_FLAG_HOOK_CAN_CARRY (1 << 9) // brings the actor back to Player if hookshoted
+#define ACTOR_FLAG_HOOK_BRING_PLAYER (1 << 10) // brings Player to the actor if hookshoted
+#define ACTOR_FLAG_ENKUSA_CUT (1 << 11) // grass actor has been cut
+#define ACTOR_FLAG_IGNORE_QUAKE (1 << 12) // actor will not shake when a quake occurs
+#define ACTOR_FLAG_HOOK_ATTACHED (1 << 13) // hookshot has attached to the actor (either a collider or surface)
+#define ACTOR_FLAG_ARROW_CAN_CARRY (1 << 14) // when an arrow hits the actor it will attach to the actor and carry it
+#define ACTOR_FLAG_ARROW_IS_CARRYING (1 << 15) // an arrow is currently carrying this actor
+#define ACTOR_FLAG_IMMEDIATE_TALK (1 << 16) // forces Player to talk when in range. needs to be unset manually to avoid infinite talking
+#define ACTOR_FLAG_HEAVYBLOCK (1 << 17) // changes actor carrying behavior specifically for the golden gauntlets block actor
+#define ACTOR_FLAG_CHECK_WITH_NAVI (1 << 18) // Navi can be used to trigger dialogue when targeting the actor
```
Dragorn421, line 120: Does "targeting" here refers to Navi hovering at the actor's focus pos or locking onto the actor?

src/overlays/actors/ovl_En_Bom_Chu/z_en_bom_chu.c
```diff
@@ -234,7 +234,7 @@ void EnBomChu_WaitForRelease(EnBomChu* this, GlobalContext* globalCtx) {
         //! @bug there is no NULL check on the floor poly.  If the player is out of bounds the floor poly will be NULL
         //! and will cause a crash inside this function.
         EnBomChu_UpdateFloorPoly(this, this->actor.floorPoly, globalCtx);
-        this->actor.flags |= ACTOR_FLAG_0; // make chu targetable
+        this->actor.flags |= ACTOR_FLAG_TARGETABLE; // make chu targetable
```
Dragorn421, line 237:
```suggestion
        this->actor.flags |= ACTOR_FLAG_TARGETABLE;
```

include/z64actor.h
```diff
@@ -100,34 +100,34 @@ typedef struct {
...
+#define ACTOR_FLAG_SFX_AT_POS (1 << 19) // play sound from sfx field at the actor's position
+#define ACTOR_FLAG_SFX_CENTERED2 (1 << 20) // same as ACTOR_FLAG_SFX_CENTERED, unused in the original game
+#define ACTOR_FLAG_SFX_CENTERED (1 << 21) // play sound from sfx field at the center of the screen
+#define ACTOR_FLAG_IGNORE_POINT_LIGHTS (1 << 22) // ignores point lights but not directional lights (such as environment lights)
+#define ACTOR_FLAG_ALWAYS_THROW (1 << 23) // Player throws held actor even if standing still
+#define ACTOR_FLAG_PLAY_BODYHIT_SFX (1 << 24) // when actor hits Player's body, a thump sfx plays
```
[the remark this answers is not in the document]
fig02: the name comes from the original sfx name which is bodyhit
I think it does help, there are many sounds you could take to mean a hit sfx. but bodyhit is the thump noise against his body

include/z64actor.h
```diff
@@ -100,34 +100,34 @@ typedef struct {
...
+#define ACTOR_FLAG_REACT_TO_LENS (1 << 7) // hidden or revealed by Lens of Truth (depending on room flags)
+#define ACTOR_FLAG_TALK_REQUESTED (1 << 8) // Player has requested to talk to the actor; Player uses this flag differently than every other actor
+#define ACTOR_FLAG_HOOK_CAN_CARRY (1 << 9) // brings the actor back to Player if hookshoted
+#define ACTOR_FLAG_HOOK_BRING_PLAYER (1 << 10) // brings Player to the actor if hookshoted
+#define ACTOR_FLAG_ENKUSA_CUT (1 << 11) // grass actor has been cut
+#define ACTOR_FLAG_IGNORE_QUAKE (1 << 12) // actor will not shake when a quake occurs
+#define ACTOR_FLAG_HOOK_ATTACHED (1 << 13) // hookshot has attached to the actor (either a collider or surface)
+#define ACTOR_FLAG_ARROW_CAN_CARRY (1 << 14) // when an arrow hits the actor it will attach to the actor and carry it
+#define ACTOR_FLAG_ARROW_IS_CARRYING (1 << 15) // an arrow is currently carrying this actor
+#define ACTOR_FLAG_IMMEDIATE_TALK (1 << 16) // forces Player to talk when in range. needs to be unset manually to avoid infinite talking
```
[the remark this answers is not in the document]
fig02: actors broadcast their ability to talk with `func_8002F2CC` and other related functions. they specify their own range, its not fixed

include/z64actor.h
```diff
@@ -100,34 +100,34 @@ typedef struct {
...
-#define ACTOR_FLAG_26 (1 << 26)
-#define ACTOR_FLAG_27 (1 << 27)
-#define ACTOR_FLAG_28 (1 << 28)
+#define ACTOR_FLAG_TARGETABLE (1 << 0) // Navi will fly over the actor and it can be Z targeted
+#define ACTOR_FLAG_UNFRIENDLY (1 << 2) // changes targeting behavior for unfriendly actors (sound, Link's stance)
+#define ACTOR_FLAG_FRIENDLY (1 << 3) // opposite of the unfriendly flag. flag is not checked against in the original game
+#define ACTOR_FLAG_NO_UPDATE_CULLING (1 << 4) // actor will keep updating even if outside of the uncull zone i.e. "update-culling" does not occur
+#define ACTOR_FLAG_NO_DRAW_CULLING (1 << 5) // actor will keep drawing even if outside of the uncull zone i.e. "draw-culling" does not occur
+#define ACTOR_FLAG_IN_UNCULL_ZONE (1 << 6) // actor is currently in the uncull zone
+#define ACTOR_FLAG_REACT_TO_LENS (1 << 7) // hidden or revealed by Lens of Truth (depending on room flags)
```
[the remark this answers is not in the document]
fig02: imo its kinda a given that the reaction an actor will have to lens is toggling visibility, cause that is its purpose

The more general wording of the flag is on purpose, cause whether or not it becomes visible or hidden is based on the room behavior. I think I would understand `VISIBLE_BASED_ON_LENS` to mean that it becomes visible when on, when thats not always the case

engineer124, line 109:
```suggestion
#define ACTOR_FLAG_REACT_TO_LENS (1 << 7) // hidden or revealed by Lens of Truth (depending on the room's lensMode)
```
Regardless of the name chosen, might be worth referencing exactly what determines this behavior instead of leaving it generic

PR #1123, opened 2022-01-27
Give code_*.c files more descriptive names

milotier:
The purpose of these files is very clear, so I think it is logical to give them more descriptive names. I don't know what the requirements are for giving a file the 'z_' prefix, but these files seem generic enough to omit it, like with sleep.c. 
The functions defined in mem.c are copies of ones from libc <string.h>, so string.c might also be a good name.
Also, maybe this is a stupid question, but why do the rand function names have underscores, but the mem function names do not?
I'm not too familiar with this project yet, so I hope this is a useful change.


milotier:
src/code/code_8006C3A0.c also seems like a candidate for renaming, however I'm not quite sure what it is for. Would envflags.c be a good name?

milotier:
There is also src/code/code_800FCE80.c with contains math functions, however there are already the sys_math*.c files so what should this be called?

fig02:
youre starting to see why some of them are still unnamed :D 
naming them isnt the most important thing at this stage, because currently we mostly use the original names for files (when we have them). This is going to change at some point and we will use our own names for everything. So until that process gets started for all files, naming some of these smaller ones that conflict with the purpose of other files isnt really a priority yet

as for what youve renamed in your pr they seem fine, but will probably be changed, or at the very least reconsidered at a later stage when we start a larger batch of renaming.

milotier:
Ok, if this is going to happen again at a later stage it is probably best to either close or merge this now, and I will look into other things.

Dragorn421 (commented):
`code_80069420.c` with `MemCopy`, `MemSet` -> `mem.c`

There's also `memcpy` from `string.c` and `func_801068B0` from `code_801068B0.c`
and `Lib_MemSet` from `z_lib.c` and `func_80106860` from `code_80106860.c`

I would rather see someone decide for all of this mess at once tbh

--------

`code_800FD970.c` -> `rand.c`

There are other `Rand_` functions but they are only wrappers for functions in `rand.c` so renaming that file sounds acceptable
Not sure if it should be rand, random, z_rand, z_random or whatever else, but I don't think it's worth debating now anyway

--------

All in all +1 towards closing this

milotier:
In that case, I will close this PR now

PR #1124, opened 2022-01-28
Replace literals with `ARRAY_COUNT`

milotier:
I've noticed that a lot of the for loops still use integer literals when iterating over arrays, where it is probably better to use `ARRAY_COUNT` to show that you are iterating over all elements.
Currently there is still a lot of code I have to go through, but i just wanted to make sure this is a useful change before spending a long time on this..

### DIFF
--- a/src/code/jpegdecoder.c
+++ b/src/code/jpegdecoder.c
@@ -130,7 +130,7 @@ s32 JpegDecoder_ParseNextSymbol(JpegHuffmanTable* hTable, s16* outCoeff, s8* out
     u16 codeOff = 0;
     u16 buff = JpegDecoder_ReadBits(16);
 
-    for (codeIdx = 0; codeIdx < 16; codeIdx++) {
+    for (codeIdx = 0; codeIdx < ARRAY_COUNT(hTable->codesB); codeIdx++) {
         if (hTable->codesB[codeIdx] == 0xFFFF) {
             continue;
         }
--- a/src/code/jpegdecoder.c
+++ b/src/code/jpegdecoder.c
@@ -130,7 +130,7 @@ s32 JpegDecoder_ParseNextSymbol(JpegHuffmanTable* hTable, s16* outCoeff, s8* out
     u16 codeOff = 0;
     u16 buff = JpegDecoder_ReadBits(16);
 
-    for (codeIdx = 0; codeIdx < ARRAY_COUNT(hTable->codesB); codeIdx++) {
+    for (codeIdx = 0; codeIdx < 16; codeIdx++) {
         if (hTable->codesB[codeIdx] == 0xFFFF) {
             continue;
         }
--- a/src/code/jpegutils.c
+++ b/src/code/jpegutils.c
@@ -7,7 +7,7 @@ void JpegUtils_ProcessQuantizationTable(u8* dqt, JpegQuantizationTable* qt, u8 c
         u8 j;
 
         dqt++;
-        for (j = 0; j < ARRAY_COUNT(qt[i].table); j++) {
+        for (j = 0; j < 64; j++) {
             qt[i].table[j] = *dqt++;
         }
     }
--- a/src/code/jpegutils.c
+++ b/src/code/jpegutils.c
@@ -7,7 +7,7 @@ void JpegUtils_ProcessQuantizationTable(u8* dqt, JpegQuantizationTable* qt, u8 c
         u8 j;
 
         dqt++;
-        for (j = 0; j < 64; j++) {
+        for (j = 0; j < ARRAY_COUNT(qt[i].table); j++) {
             qt[i].table[j] = *dqt++;
         }
     }
--- a/src/code/mtxuty-cvt.c
+++ b/src/code/mtxuty-cvt.c
@@ -7,8 +7,8 @@ void MtxConv_F2L(Mtx* m1, MtxF* m2) {
     LogUtils_CheckNullPointer("m1", m1, "../mtxuty-cvt.c", 31);
     LogUtils_CheckNullPointer("m2", m2, "../mtxuty-cvt.c", 32);
 
-    for (i = 0; i < 4; i++) {
-        for (j = 0; j < 4; j++) {
+    for (i = 0; i < ARRAY_COUNT(m2->mf); i++) {
+        for (j = 0; j < ARRAY_COUNT(m2->mf[i]); j++) {
             s32 value = (m2->mf[i][j] * 0x10000);
 
             m1->intPart[i][j] = value >> 16;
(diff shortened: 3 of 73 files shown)


Dragorn421:
In general when this helps shifting I think this is a good idea 👍

However if changing the array length doesn't make sense (e.g. matrices or rgba colors), or would involve updating the code in some other way (e.g. effect buffers for some bosses), then I don't think it's useful to use array_length and that it actually harms readability.

So it's highly situational imo, I will pinpoint examples later

----------

Also audio is on ongoing documentation, and Tharo voiced planning to do "pad mgr" stuff. So the people involved with those should say if they're ok with touching those files (if it will conflict, whether they're fine with it)

milotier:
I have now gone through the rest of the code (although I've probably missed some cases). Let me know if there is anything that should be reverted.

src/overlays/actors/ovl_En_Anubice/z_en_anubice.c
```diff
@@ -358,6 +358,7 @@ void EnAnubice_Update(Actor* thisx, GlobalContext* globalCtx) {
         this->actor.shape.yOffset == 0.0f) {
         EnAnubice_Hover(this, globalCtx);
         for (i = 0; i < 5; i++) {
```
milotier, line 360: This seems like a bug, because this->flameCircles only has length 4, or the loop must somehow always be broken before it runs off the end of the array. Or is the declaration just wrong?

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
```diff
@@ -2612,7 +2612,7 @@ void func_80904D88(BossGanon2* this, GlobalContext* globalCtx) {
         }
         gSPDisplayList(POLY_XLU_DISP++, ovl_Boss_Ganon2_DL_00B308);
 
-        for (i = 0; i < 15; i++) {
+        for (i = 0; i < ARRAY_COUNT(this->unk_234) - 1; i++) {
```
mzxrules, line 2615: I thought this looked weird, so I took a look at all instances where unk_234 is used within z_boss_ganon2. It appears as if this array should only be 15 elements long and not 16, as the limb index lookup tables only use 0x0E as a max index.

src/overlays/actors/ovl_En_Anubice/z_en_anubice.c
milotier: That makes sense. As far as I can see, the code never uses more than 5 elements of flameCircle. That still leaves 4 bytes of unused space after it, though. Maybe in the original code, it was an array of length 6, where the last element was unused?

src/overlays/actors/ovl_Boss_Ganon2/z_boss_ganon2.c
milotier: I think you are right. I've now changed the length.

src/code/jpegdecoder.c
```diff
@@ -130,7 +130,7 @@ s32 JpegDecoder_ParseNextSymbol(JpegHuffmanTable* hTable, s16* outCoeff, s8* out
     u16 codeOff = 0;
     u16 buff = JpegDecoder_ReadBits(16);
 
-    for (codeIdx = 0; codeIdx < 16; codeIdx++) {
+    for (codeIdx = 0; codeIdx < ARRAY_COUNT(hTable->codesB); codeIdx++) {
```
Dragorn421, line 133: index is used like `codeOff = buff >> (15 - codeIdx);`

src/code/jpegutils.c
```diff
@@ -7,7 +7,7 @@ void JpegUtils_ProcessQuantizationTable(u8* dqt, JpegQuantizationTable* qt, u8 c
         u8 j;
 
         dqt++;
-        for (j = 0; j < 64; j++) {
+        for (j = 0; j < ARRAY_COUNT(qt[i].table); j++) {
```
Dragorn421, line 10: this also relies on the size of the `dqt` array

src/code/mtxuty-cvt.c
```diff
@@ -7,8 +7,8 @@ void MtxConv_F2L(Mtx* m1, MtxF* m2) {
     LogUtils_CheckNullPointer("m1", m1, "../mtxuty-cvt.c", 31);
     LogUtils_CheckNullPointer("m2", m2, "../mtxuty-cvt.c", 32);
 
-    for (i = 0; i < 4; i++) {
-        for (j = 0; j < 4; j++) {
+    for (i = 0; i < ARRAY_COUNT(m2->mf); i++) {
+        for (j = 0; j < ARRAY_COUNT(m2->mf[i]); j++) {
```
Dragorn421, line 11: Matrices in oot and in graphics rendering in general are 4x4 and anyone trying to change this is doing very *very* wild things

using array_count for this would make it confusing and harm readability more than anything

src/code/sys_math3d.c
```diff
@@ -1784,7 +1784,7 @@ s32 Math3D_CylVsLineSeg(Cylinder16* cyl, Vec3f* linePointA, Vec3f* linePointB, V
     /**
      * Places the found intersection points into intersectA and intersectB. IntersectA is always closer to point A
      */
-    for (count = 0, i = 0; i < 4; i++) {
+    for (count = 0, i = 0; i < ARRAY_COUNT(intPts); i++) {
```
Dragorn421, line 1787: this whole function very much just assumes intPts to hold 4 positions

src/overlays/actors/ovl_En_Anubice/z_en_anubice.c
```diff
@@ -357,7 +357,7 @@ void EnAnubice_Update(Actor* thisx, GlobalContext* globalCtx) {
     if (this->actionFunc != EnAnubice_SetupDie && this->actionFunc != EnAnubice_Die &&
         this->actor.shape.yOffset == 0.0f) {
         EnAnubice_Hover(this, globalCtx);
-        for (i = 0; i < 5; i++) {
+        for (i = 0; i < ARRAY_COUNT(this->flameCircles); i++) {
             // Shouldn't flameCircles have length 5?
```
mzxrules, line 361:
```suggestion
        for (i = 0; i < ARRAY_COUNT(this->flameCircles); i++) {
```

Dragorn421 (changes requested):
In general I think using array_count in for statements makes sense to use where it would make sense to use a `for each` construct, if that was a thing in C

milotier:
@Dragorn421 Most of your suggestions have now been applied. I did leave `src/code/code_800D2E30.c` as it was, since it only contains simple for loops that zero out arrays.

Dragorn421:
I would appreciate if you could apply the logic I've been describing about using ARRAY_COUNT yourself, to what this PR touches

Or come up with a different perspective of why use/not use array_count, but at least have a perspective to not make it kind of feel like find-and-replace

fig02:
I think we probably need to talk a bit more about the details of what should be using array count and plan a bit more before doing a pass of changes like this. if we agree we can close this for now and revisit it

fig02:
As talked about above, we are going to sort out or plans with array_count a bit more before making larger scale changes like this

PR #1125, opened 2022-01-30
Document EnInsect actor

milotier:
There is still quite a bit of things that need to be named:
* `INSECT_FLAG_7` which seems to say that a certain sound has been played.
* `D_80A7DEB0` which seems to have something to do with the progress towards the soft soil patch.
* Some of `EnInsect`'s members
* Some local variables.

I'm also not quite sure whether we want to keep the current names of the insect flags.


src/overlays/actors/ovl_En_Insect/z_en_insect.c
```diff
@@ -255,36 +274,36 @@ void func_80A7C3F4(EnInsect* this, GlobalContext* globalCtx) {
...
-    this->unk_314 |= 0x100;
+void EnInsect_SetupCrawlToSpawn(EnInsect* this) {
+    this->actionTimer = Rand_S16Offset(10, 45);
+    EnInsect_SetupAnimation(this);
+    this->actionFunc = EnInsect_CrawlToSpawn;
+    this->flags |= INSECT_FLAG_CRAWLING;
 }
 
-void func_80A7C5EC(EnInsect* this, GlobalContext* globalCtx) {
+void EnInsect_CrawlToSpawn(EnInsect* this, GlobalContext* globalCtx) {
```
Dragorn421, line 298: Just reading the code, I don't think the intent here is crawling to home (spawn), it's probably just to avoid the bugs running away too far

I mean, with this action function the bugs will hover around the home point, not reach home and stop there

In `EnInsect_Crawl` the line `Math_SmoothStepToF(&this->actor.speedXZ, 0.0f, 0.1f, 0.5f, 0.0f);` lowers the speed to 0
So I think the actual movement is handled by `EnInsect_CrawlToSpawn`, and `EnInsect_SetupCrawl` is more like "slow down"/"stop"

src/overlays/actors/ovl_En_Insect/z_en_insect.h
```diff
@@ -17,16 +27,25 @@ typedef struct EnInsect {
...
 } EnInsect; // size = 0x032C
 
+typedef enum {
+    /* 0 */ INSECT_PERMANENT,
+    /* 1 */ INSECT_SPAWNED,
+    /* 2 */ INSECT_FIRST_DROPPED,
+    /* 3 */ INSECT_EXTRA_DROPPED // An additional bug spawned by dropping one from a bottle.
+} EnInsectType;
+
+#define IS_DROPPED(type) ((type) == INSECT_FIRST_DROPPED || (type) == INSECT_EXTRA_DROPPED)
```
Dragorn421, line 49: I think this should go in the .c maybe, or if it stays in the .h it should have a less generic name

src/overlays/actors/ovl_En_Insect/z_en_insect.c
```diff
@@ -351,56 +370,56 @@ void func_80A7C86C(EnInsect* this, GlobalContext* globalCtx) {
...
     this->actor.draw = NULL;
     this->actor.speedXZ = 0.0f;
 
-    func_80A7BF58(this);
+    EnInsect_SetupAnimation(this);
 
     this->skelAnime.playSpeed = 0.3f;
-    this->actionFunc = func_80A7CAD0;
-    this->unk_314 &= ~0x100;
+    this->actionFunc = EnInsect_UpdateCaught;
```
Dragorn421, line 391:
```suggestion
    this->actionFunc = EnInsect_Caught;
```
same with `UpdateDropped`, can drop the `Update` part

Dragorn421 (commented):
Very nice job on this! 🎉

src/overlays/actors/ovl_En_Insect/z_en_insect.c
```diff
@@ -417,54 +437,54 @@ void func_80A7CC3C(EnInsect* this, GlobalContext* globalCtx) {
...
             this->soilActor->unk_152 = 1;
         }
         Actor_Kill(&this->actor);
     }
 }
 
-void func_80A7CE60(EnInsect* this) {
-    this->unk_31A = Rand_S16Offset(120, 50);
-    func_80A7BF58(this);
+void EnInsect_SetupSink(EnInsect* this) {
```
Dragorn421, line 457: the bugs are actually able to walk on water for a bit before drowning (`EnInsect_Drown`), in this action they actually stay afloat and move around trying to find land

I'm not sure what a good name for this action would be (not Sink since they stay afloat and the sinking is in `EnInsect_Drown`)

E.g. you can test this in game in zora's river (spot03) with this bit of code in `Gameplay_Update`
```c
    // drop bugs where player is when Dpad-up is pressed
    if (input[0].press.button & BTN_DUP) {
        Player* player = GET_PLAYER(globalCtx);
        Vec3f* pos = &player->actor.world.pos;

        Actor_Spawn(&globalCtx->actorCtx, globalCtx, ACTOR_EN_INSECT, pos->x, pos->y, pos->z, 0, 0, 0, 2);
    }
```

src/overlays/actors/ovl_En_Insect/z_en_insect.h
```diff
@@ -9,6 +9,16 @@ struct EnInsect;
 
 typedef void (*EnInsectActionFunc)(struct EnInsect*, GlobalContext*);
 
+#define INSECT_FLAG_TEMP_AND_ALIVE (1 << 0) // Set when the bug is temporary and is not drowning.
+#define INSECT_FLAG_ON_GROUND (1 << 1) // Set when the bug was dropped and is standing on the ground.
+#define INSECT_FLAG_TEMP (1 << 2) // Set when the bug disappears after a while.
+#define INSECT_FLAG_UNCATCHABLE (1 << 3) // Set when the bug is about to disappear and can no longer be caught.
+#define INSECT_FLAG_FOUND_SOIL (1 << 4)
+#define INSECT_FLAG_SOIL_CLOSE (1 << 5)
```
Dragorn421, line 17: I'm hesitant if this insect flag should be named by usage (the current `INSECT_FLAG_SOIL_CLOSE`) or by effect (disabling OC collision)

I guess naming by usage makes the intent clearer so it's probably better as is

src/overlays/actors/ovl_En_Insect/z_en_insect.h
```diff
@@ -9,6 +9,16 @@ struct EnInsect;
...
+#define INSECT_FLAG_LANDED (1 << 6) // Set when the landing sound has been played.
```
Dragorn421, line 18: It could be made clearer this is after dropping bugs

For example it could also be understood as the land falling from an edge and landing on some group below

Dragorn421 (changes requested):
Some more comments, mostly on insect flags
Also could you merge the master branch into yours?

src/overlays/actors/ovl_En_Insect/z_en_insect.c
```diff
@@ -255,37 +275,36 @@ void func_80A7C3F4(EnInsect* this, GlobalContext* globalCtx) {
...
     }
 
-    if (((this->unk_314 & 4) && this->unk_31C <= 0) ||
-        ((sp2E == 2 || sp2E == 3) && (this->unk_314 & 1) && (this->actor.bgCheckFlags & BGCHECKFLAG_GROUND) &&
-         D_80A7DEB8 >= 4)) {
-        func_80A7CBC8(this);
-    } else if ((this->unk_314 & 1) && (this->actor.bgCheckFlags & BGCHECKFLAG_WATER_TOUCH)) {
-        func_80A7CE60(this);
+    if (((this->insectFlags & INSECT_FLAG_TEMP) && this->lifeTimer <= 0) ||
+        (IS_DROPPED(type) && (this->insectFlags & INSECT_FLAG_TEMP_AND_ALIVE) && (this->actor.bgCheckFlags & BGCHECKFLAG_WATER_TOUCH) && sDroppedCount >= 4)) {
```
Dragorn421, line 283: merge issue here, I think that is the only issue
```suggestion
        (IS_DROPPED(type) && (this->insectFlags & INSECT_FLAG_TEMP_AND_ALIVE) && (this->actor.bgCheckFlags & BGCHECKFLAG_GROUND) && sDroppedCount >= 4)) {
```

Dragorn421 (commented):
Jenkins (which automatically checks if PRs build to a matching rom) failed due to a merge issue (I assume)

milotier:
On my end the problem isn't that the rom doesn't match, the build is failing in some completely unrelated file:
```
cfe: Error: src/overlays/actors/ovl_Demo_Im/z_demo_im_cutscene_data.c:26: Missing ')' in macro instantiation
cfe: Error: src/overlays/actors/ovl_Demo_Im/z_demo_im_cutscene_data.c:26: Missing ')' in macro instantiation
Failed to compile file src/overlays/actors/ovl_Demo_Im/z_demo_im.c. Command line:

tools/ido_recomp/linux/7.1/cc -c -G 0 -non_shared -Xfullwarn -Xcpluscomm -Iinclude -Isrc -Iassets -Ibuild -I. -Wab,-r4300_mul -woff 649,838,712 -mips2 -O2 -I /home/milo/projects/oot/src/overlays/actors/ovl_Demo_Im -o build/src/overlays/actors/ovl_Demo_Im/z_demo_im.o /tmp/preprocessed9zxyybi_.c
```
This also happens for me on the master branch. The weird thing is that that file doesn't seem to have been changed recently.

Dragorn421:
Can you get on discord (link in the readme) to discuss more efficiently
I'm not sure I can help myself but this isn't never-seen-before iirc

mzxrules:
`EnInsect_FoundNearbySoil` is past tense and confusing for a function name. It should probably be renamed to `EnInsect_TryFindNearbySoil`, and the `ret` should be changed to use `true` and `false`

src/overlays/actors/ovl_En_Insect/z_en_insect.h
```diff
@@ -17,16 +27,23 @@ typedef struct EnInsect {
...
     /* 0x032A */ u8 unk_32A;
 } EnInsect; // size = 0x032C
 
+typedef enum {
+    /* 0 */ INSECT_PERMANENT,
+    /* 1 */ INSECT_SPAWNED,
+    /* 2 */ INSECT_FIRST_DROPPED,
+    /* 3 */ INSECT_EXTRA_DROPPED // An additional bug spawned by dropping one from a bottle.
+} EnInsectType;
+
```
mzxrules, line 48: I think this can be done better. 

Type 0 is exclusive to obj_mure, which is an actor that not only spawns a few small creatures a child actors, it can directly manipulate the child spawns to achieve certain behaviors. As I also mentioned in another comment, there is more to Type 0 than not having a life timer. Type 0 should probably have a name tied to it's spawn source, since the unusual behavior is likely done to reduce bg collision calls.

Types 1-3 from what I can tell essentially exhibit the same overall behavior, with the exception being that Type 1 is missing `INSECT_FLAG_ON_GROUND` thus UPDBGCHECKINFO_FLAG_0 isn't set (not sure how this one difference affects things). `INSECT_SPAWNED` isn't as expressive of it's purpose as it could be (it's spawned by bushes and rocks when the player discovers them hiding underneath them), and I don't like `SPAWNED` as it's past tense (which doesn't make sense in the context of ActorSpawn calls). `FIRST_DROPPED` and `EXTRA_DROPPED` could just be named to reflect that these types are only spawned when dropping bugs from a bottle. A better name could eliminate the need for a comment on Type 3

Lastly, you probably should name give the types names like `INSECT_TYPE_*`.

src/overlays/actors/ovl_En_Insect/z_en_insect.h
```diff
@@ -9,6 +9,16 @@ struct EnInsect;
 
 typedef void (*EnInsectActionFunc)(struct EnInsect*, GlobalContext*);
 
+#define INSECT_FLAG_TEMP_AND_ALIVE (1 << 0) // Set when the bug is temporary and is not drowning.
+#define INSECT_FLAG_ON_GROUND (1 << 1) // Set when the bug was dropped and is standing on the ground.
+#define INSECT_FLAG_TEMP (1 << 2) // Set when the bug disappears after a while.
```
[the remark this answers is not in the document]
mzxrules: There's probably no need for a comment with a better name like
```suggestion
#define INSECT_FLAG_IS_SHORT_LIVED (1 << 2) // Set when the bug disappears after a while.
```

Dragorn421:
If you don't feel like maintaining this PR anymore, don't worry I (or whoever wishes to) can take it from here

fig02:
Closing due to inactivity.
sorry about your build issues, not sure what happened there. if you decide you want to try to fix it again, can come by the discord.
Someone else will take your pr from here, since you did some good work. Just needs to be a cleaned up a bit before its merged.

PR #1126, opened 2022-01-31
Defines for `Actor#bgCheckFlags`

Dragorn421:
I probably should name the flags while I'm at it, what about:

```c
#define BGCHECKFLAG_GROUND (1 << 0) // Standing on the ground
#define BGCHECKFLAG_GROUND_TOUCH (1 << 1) // Has touched the ground (only active for 1 frame)
#define BGCHECKFLAG_GROUND_LEAVE (1 << 2) // Has left the ground (only active for 1 frame)
#define BGCHECKFLAG_WALL (1 << 3) // Touching a wall
#define BGCHECKFLAG_CEILING (1 << 4) // Touching a ceiling
#define BGCHECKFLAG_WATER (1 << 5) // On or below water surface
#define BGCHECKFLAG_WATER_TOUCH (1 << 6) // Has touched water (actor is responsible for unsetting this the frame it touches the water)
#define BGCHECKFLAG_BELOW_GROUND (1 << 7) // Similar to BGCHECKFLAG_0 but with no velocity check and is cleared every frame
#define BGCHECKFLAG_CRUSHED (1 << 8) // Crushed between a floor and ceiling (triggers a void for player)
#define BGCHECKFLAG_9 (1 << 9) // Unknown, only set/used by player
```

### DIFF
--- a/include/z64actor.h
+++ b/include/z64actor.h
@@ -136,9 +136,9 @@ typedef struct {
 #define BGCHECKFLAG_4 (1 << 4) // Touching a ceiling
 #define BGCHECKFLAG_5 (1 << 5) // On or below water surface
 #define BGCHECKFLAG_6 (1 << 6) // Has touched water (actor is responsible for unsetting this the frame it touches the water)
-#define BGCHECKFLAG_7 (1 << 7) // Similar to & 0x1 but with no velocity check and is cleared every frame
+#define BGCHECKFLAG_7 (1 << 7) // Similar to BGCHECKFLAG_0 but with no velocity check and is cleared every frame
 #define BGCHECKFLAG_8 (1 << 8) // Crushed between a floor and ceiling (triggers a void for player)
-#define BGCHECKFLAG_9 (1 << 9) // Unknown (only set/used by player so far)
+#define BGCHECKFLAG_9 (1 << 9) // Unknown, only set/used by player
 
 typedef struct Actor {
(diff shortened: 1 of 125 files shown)


fig02:
suggested names are mostly good
`BGCHECKFLAG_BELOW_GROUND` does seem to work out in practice. believe this is because if you are below the ground, your floor height is either going to be the new floor below you, or -32000 for bottom of the map. so `(floorHeightDiff >= 0.0f)` wont pass and the flag doesnt get set

hard to tell because the flag never gets checked against (only unset in a couple scenarios).

it seems to be the exact same as standing on the ground, but im not sure of the significance of the y velocity check. when would you ever be on the ground but with a y velocity greater than 0?

fig02:
im fine with ground_strict for now. its not as important to me cause it doesnt get used

flag 9 can be `BGCHECKFLAG_PLAYER_WALL_INTERACT` i suppose. pretty confident thats its usage, a short list of when it gets used from me and mzx:
flag is set when link is a certain distance from a wall, and it signals the ability to:
-target a wall
-climb a wall
-jump over a wall
-grab ladder/vines
-enter crawlspace
-bonk a tree actor to shake it
-push a block
-grab heavy block

if youd rather not name it now though thats fine, but id be more than fine with it named now

fig02 (approved):

engineer124 (approved):

PR #1128, opened 2022-01-31
Implement the scene table with defines

Roman971:
This follows the same pattern used for other tables like actors, objects, etc.

One important note is it uses the same define for scenes with and without title cards, but the argument for the title card can be set to 'none' to disable it for that scene (which is used for a bunch of vanilla scenes). The implementation for that is a bit hacky but it was discussed on discord quite a bit and it's probably fine for now at least.


include/tables/scene_table.h
```diff
@@ -0,0 +1,122 @@
...
+/* 0x5C */ DEFINE_SCENE(spot11_scene, g_pn_55, SCENE_SPOT11, 0, 10, 0)
+/* 0x5D */ DEFINE_SCENE(spot12_scene, g_pn_53, SCENE_SPOT12, 0, 11, 0)
+/* 0x5E */ DEFINE_SCENE(spot13_scene, g_pn_37, SCENE_SPOT13, 0, 12, 0)
+/* 0x5F */ DEFINE_SCENE(spot15_scene, g_pn_38, SCENE_SPOT15, 0, 13, 0)
+/* 0x60 */ DEFINE_SCENE(spot16_scene, g_pn_39, SCENE_SPOT16, 0, 14, 0)
+/* 0x61 */ DEFINE_SCENE(spot17_scene, g_pn_40, SCENE_SPOT17, 0, 15, 0)
+/* 0x62 */ DEFINE_SCENE(spot18_scene, g_pn_41, SCENE_SPOT18, 0, 16, 0)
+/* 0x63 */ DEFINE_SCENE(spot20_scene, g_pn_42, SCENE_SPOT20, 0, 17, 0)
+/* 0x64 */ DEFINE_SCENE(ganon_tou_scene, g_pn_43, SCENE_GANON_TOU, 0, 36, 0)
+// Debug only scenes
```
Dragorn421, line 113:
```suggestion
// Debug-only scenes
```
not sure if gramatically correct but sounds better 😅

src/code/z_scene_table.c
```diff
@@ -810,128 +810,39 @@ EntranceInfo gEntranceTable[] = {
...
+#include "tables/scene_table.h"
+
+#undef DEFINE_SCENE
+
+// Scene Table definition
+#define DEFINE_SCENE(name, title, _2, unk_10, config, unk_12) \
+    { (u32)_##name##SegmentRomStart,                          \
+      (u32)_##name##SegmentRomEnd,                            \
+      (u32)_##title##SegmentRomStart,                         \
+      (u32)_##title##SegmentRomEnd,                           \
```
Dragorn421, line 827:
```suggestion
    { ROM_FILE(name),                                         \
      ROM_FILE(title),                                        \
```

after formatting it even looks like this:
```c
#define DEFINE_SCENE(name, title, _2, unk_10, config, unk_12) \
    { ROM_FILE(name), ROM_FILE(title), unk_10, config, unk_12, 0 },
```

Dragorn421 (commented):
I would also have liked a `DEFINE_SCENE_UNSET` variant for modding purposes, but as Roman pointed out `DEFINE_SCENE` can actually accidentally be used that way with `DEFINE_SCENE(none, none, ...)` so that's good.

--------------

In `#define DEFINE_SCENE(name, title, _2, unk_10, config, unk_12)`, can I suggest moving `unk_10` to after `config`? Like `#define DEFINE_SCENE(name, title, _2, config, unk_10, unk_12)`

Scene config gets used, the unks are not even read afaik so completely useless
It seems weird to "reorder" the `SceneTableEntry` struct this way but I think it's fine

include/tables/scene_table.h
fig02: imo not needed, but not a huge deal
if the word "exclusive" was used instead of only, for example, I dont think you'd need a dash. but i also am not sure what is technically grammatically correct

src/code/z_scene_table.c
fig02: +1 for this and for reordering the args like dragorn mentioned in his main post

fig02 (approved):

src/code/z_scene_table.c
```diff
@@ -810,128 +810,32 @@ EntranceInfo gEntranceTable[] = {
...
 
-#define UNTITLED_SCENE(name, unk_10, config, unk_12) \
-    { (u32) _##name##SegmentRomStart, (u32)_##name##SegmentRomEnd, 0, 0, unk_10, config, unk_12, 0 }
+// Scene Table definition
+#define DEFINE_SCENE(name, title, _2, config, unk_10, unk_12) \
+    { ROM_FILE(name), ROM_FILE(title), unk_10, config, unk_12, 0 },
+
+// Handle `none` as a special case for scenes without a title card
+#define _noneSegmentRomStart 0
+#define _noneSegmentRomEnd 0
```
Thar0, line 828:
```suggestion
#define _noneSegmentRomStart NULL
#define _noneSegmentRomEnd NULL
```
It seems slightly more "correct" to define these to null since it's a stand-in for pointer-typed segment symbols?

Thar0 (commented):
Very minor suggestion

fig02:
we can remove scene declarations from segment_symbols.h as well right?

src/code/z_scene_table.c
Dragorn421: #1089 changes vromStart and vromEnd to `uintptr_t`: https://github.com/zeldaret/oot/blob/b9c0124de42099a5b9c614155badc96ee311a949/include/z64scene.h#L6-L9
so yes, if we do see rom/vrom offsets as pointers (I guess we are forced to by using symbols)

Roman971:
We can remove them, yeah. I originally left them in there since I was hoping we would remove rooms as well soon, but I don't think ZAPD has been updated to extern the room symbols where they are needed yet, so it will have to be done separately.

fig02 (approved):

Thar0 (approved):

Dragorn421 (approved):

PR #1131, opened 2022-02-04
Document Transition System

fig02:
This documents the transition system in the game.
Note: this does NOT document the code for the individual "instance" type transitions, this focuses more on the user-facing system in z_play.

I wrote a high level overview about the transition system and how there are "instance" types and "non-instance" types: https://discord.com/channels/688807550715560050/688851317593997489/938730525428957204

There is an unused transition type that gets controlled by `gTrnsnUnkState`. Would love to know more about it, but I got lost trying to read the code.

This pr also updates the entrance table to use proper enum values for everything, including the new transition macros. The circle macros especially are quite long, and the formatter added some undesirable line breaks in the entries. These will stay for now, but will eventually be moved to a header for the `DEFINE_ENTRANCE` stuff that we plan to do.


include/z64.h
```diff
@@ -969,6 +969,77 @@ typedef struct {
...
+    /* 1 */ TC_WHITE,
+    /* 1 */ TC_GRAY
+} TransitionCircleColor;
+
+typedef enum {
+    /* 0 */ TC_FAST,
+    /* 1 */ TC_SLOW
+} TransitionCircleSpeed;
+
+#define TRANS_TYPE_CIRCLE(type, color, speed, unk_14) (((unk_14 & 1) << 5) | ((color & 3) << 3) | ((type & 3) << 1) | (speed & 1))
```
Dragorn421, line 1041: where does unk_14 come from? it just looks like what makes the values start at 32 (it's always 1 right?)

fig02: it is a property of the circle which I was unable to determine its purpose.
It does make sense that it makes them start at 32 though, in which case the code in the circles that depends on this value being 0 is probably dead code

fig02: 0 gets used in practice when leaving an area (transitioning in), so im clearly missing something. will look into it more

fig02: i see whats happening now and im mad at how dumb it is. 
for context, the relevant code in the circle happens in its setType callback, which gets called in z_play via `globalCtx->transitionCtx.setType()`

if a transition type is 32 or above, it will pass its type as the arg to the function 
```
if ((globalCtx->transitionCtx.transitionType >> 5) == 1) {
    globalCtx->transitionCtx.setType(&globalCtx->transitionCtx.data,
                                     globalCtx->transitionCtx.transitionType | 0x80);
}
```
this is where unk_14 would gets its value from the type. but as you mentioned, it has to be 1 anyway for this to be true.

Now, later on in z_play it calls the setType callback *again* (this time for all the other types that are below 32)
```
if (globalCtx->transitionTrigger == TRANS_TRIGGER_OUT) {
    globalCtx->transitionCtx.setType(&globalCtx->transitionCtx.data, 1);
} else {
    globalCtx->transitionCtx.setType(&globalCtx->transitionCtx.data, 2);
}
```
since it is not passing 0x80 to the circle type, it does not do the normal circle setup. if TRIGGER_IN (leaving), unk_14 is hardcoded to be 0 instead of 1
```
    if (type & 0x80) {
        this->unk_14 = (type >> 5) & 0x1;
        this->typeColor = (type >> 3) & 0x3;
        this->speedType = type & 0x1;
        this->effect = (type >> 1) & 0x3;
    } else if (type == 1) {
        this->unk_14 = 1;
    } else {
        this->unk_14 = 0;
    }
    ```

Comment (fig02) [reply include/z64.h:1041]: phrased another way, any transition with a type above 32 has its setType callback called twice in a row. And the circle only cares about what unk_14 is the first time its called. But in order for that first call to even happen it needs to be set...

Comment (fig02) [reply include/z64.h]: figured out and addressed, the parameter has been removed from the macro. unk_14 still has meaning, but not from the macro lol

File: include/z64.h [hunk 2]
```diff
@@ -969,6 +969,80 @@ typedef struct {
...
 
+#define TRANS_TRIGGER_OUT -20 // transition is going out
+#define TRANS_TRIGGER_OFF 0 // transition is not active
+#define TRANS_TRIGGER_IN 20 // activate transition to come in
+
+typedef enum {
+    /* 00 */ TRANS_MODE_OFF,
+    /* 01 */ TRANS_MODE_SETUP,
+    /* 02 */ TRANS_MODE_INSTANCE_INIT,
+    /* 03 */ TRANS_MODE_INSTANCE_RUNNING,
```
engineer124, line 980: What's the motivation for putting these transition macros & enums in `z64.h`? Could they go in `z64transition.h` instead?

fig02: transition.h is specifically for the "instance" type transitions. it didnt really feel right to, these are more about the global context facing variables

fig02: imo z64transition will get a different name at some point to be more specific about those kinds of transitions, and the TransitionContext + these enums will go in the actual transition.h

fig02: well actually no, cause the transition context is unique to the instance types... idk its complicated lol. but yeah we need to put more thought into that kinda thing in general. so z64.h is fine for now

include/functions.h
```diff
@@ -1487,7 +1487,7 @@ void KaleidoScopeCall_Draw(GlobalContext* globalCtx);
 void func_800BC490(GlobalContext* globalCtx, s16 point);
 s32 func_800BC56C(GlobalContext* globalCtx, s16 arg1);
 void func_800BC590(GlobalContext* globalCtx);
-void func_800BC5E0(GlobalContext* globalCtx, s32 arg1);
+void Gameplay_SetupTransition(GlobalContext* globalCtx, s32 arg1);
```
engineer124, line 1490:
```suggestion
void Gameplay_SetupTransition(GlobalContext* globalCtx, s32 transitionType);
```

src/code/z_scene_table.c
```diff
@@ -30,784 +30,1682 @@
...
+    ENTRANCE(SCENE_MORIBOSSROOM, 0, true, true, TRANS_TYPE_FADE_BLACK, TRANS_TYPE_FADE_BLACK),
+    ENTRANCE(SCENE_MORIBOSSROOM, 0, true, true, TRANS_TYPE_FADE_BLACK, TRANS_TYPE_FADE_BLACK),
+    ENTRANCE(SCENE_MIZUSIN, 0, false, true, TRANS_TYPE_FADE_BLACK, TRANS_TYPE_FADE_BLACK),
+    ENTRANCE(SCENE_MIZUSIN, 0, false, true, TRANS_TYPE_FADE_BLACK, TRANS_TYPE_FADE_BLACK),
+    ENTRANCE(SCENE_MIZUSIN, 0, false, true, TRANS_TYPE_FADE_BLACK, TRANS_TYPE_FADE_BLACK),
+    ENTRANCE(SCENE_MIZUSIN, 0, false, true, TRANS_TYPE_FADE_BLACK, TRANS_TYPE_FADE_BLACK),
+    ENTRANCE(0x6E, 0, false, true, TRANS_TYPE_FADE_BLACK, TRANS_TYPE_FADE_BLACK),
+    ENTRANCE(0x6E, 0, false, true, TRANS_TYPE_FADE_BLACK, TRANS_TYPE_FADE_BLACK),
+    ENTRANCE(0x6E, 0, false, true, TRANS_TYPE_FADE_BLACK, TRANS_TYPE_FADE_BLACK),
+    ENTRANCE(0x6E, 0, false, true, TRANS_TYPE_FADE_BLACK, TRANS_TYPE_FADE_BLACK),
```
engineer124: Perhaps others already already know, but do you know what's going on with scene `0x6E`? Does this enum value appear anywhere else in code?

Also as a tangent, just a random thought I had. Is there a reason `sceneNum` is used instead of `sceneId` throughout code? `Num` often means total with our style of naming afaik

include/z64environment.h
```diff
@@ -30,6 +30,14 @@ typedef enum {
     /* 13 */ SKYBOX_DMA_PAL2_START
 } SkyboxDmaState;
 
+typedef enum {
+    /* 0 */ SANDSTORM_OFF,
+    /* 1 */ SANDSTORM_FILL,
+    /* 2 */ SANDSTORM_UNFILL,
+    /* 2 */ SANDSTORM_ACTIVE,
+    /* 2 */ SANDSTORM_DISSIPATE
+} SandStormState;
```
engineer124, line 39:
```suggestion
} SandstormState;
```
Is sandstorm 1 word?

src/code/z_scene_table.c
fig02: its referencing a scene that doesnt exist in the game. everything about it got removed and all we have is this reference to its scene number. Not sure if we have any knowledge of that the scene actually was, but there isnt afaik.

I had this same thought recently. Pretty sure its just rom hacking carry overs and id would probably make more sense

include/z64environment.h
fig02: it is yeah, thanks

src/code/z_scene_table.c
engineer124: Do you think it's worth adding that removed scene to the enum list? I'm wondering if someone else comes in and sees the number `0x6E` and wonders the same thing. It would be easier to search the repo for `SCENE_REMOVED` to see that this is the only place it's used.

I guess the other thing to be mindful of is when retail support is incorporated and some of the debug scenes are also removed for those version. So maybe that's not the best name.

fig02: with how the scene table works (DEFINE_SCENE) an enum entry cant really be added without also defining other stuff that makes it not match anymore. at least as far as im aware

fig02: and i also dont see any benefit from having it anyway, as it references something that doesnt exist literally anywhere else

src/code/z_fbdemo_fade.c
```diff
@@ -49,16 +49,16 @@ void TransitionFade_Update(void* thisx, s32 updateRate) {
...
         case 1:
             this->fadeTimer += updateRate;
-            if (this->fadeTimer >= gSaveContext.fadeDuration) {
-                this->fadeTimer = gSaveContext.fadeDuration;
+            if (this->fadeTimer >= gSaveContext.transFadeDuration) {
+                this->fadeTimer = gSaveContext.transFadeDuration;
                 this->isDone = 1;
             }
-            if (!gSaveContext.fadeDuration) {
+            if (!gSaveContext.transFadeDuration) {
```
engineer124, line 56:
```suggestion
            if (gSaveContext.transFadeDuration == 0) {
```

src/code/z_play.c
```diff
@@ -590,175 +612,184 @@ void Gameplay_Update(GlobalContext* globalCtx) {
...
                     break;
 
-                case 5:
-                    globalCtx->envCtx.screenFillColor[3] = (D_801614C8 / 20.0f) * 255.0f;
-                    if (D_801614C8 >= 20 && 1) {
-                        globalCtx->state.running = 0;
+                case TRANS_MODE_FILL_IN:
+                    globalCtx->envCtx.screenFillColor[3] = (sTransitionFillTimer / 20.0f) * 255.0f;
+
+                    if (sTransitionFillTimer >= 20 && 1) {
```
engineer124, line 652:
```suggestion
                    if (sTransitionFillTimer >= 20 && true) {
```
Seems like a strange if-condition. The `true` is needed?

src/code/z_scene_table.c
mzxrules: not having it part of the enum technically breaks shiftability, since adding another scene will change where those entrances go.

src/code/z_fbdemo_fade.c
fig02: no match

src/code/z_play.c
```diff
@@ -467,19 +477,20 @@ void Gameplay_Update(GlobalContext* globalCtx) {
 
         if (globalCtx->transitionMode) {
```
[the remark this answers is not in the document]
fig02: no match

src/code/z_scene_table.c
fig02: realized that we do this to, for example, add a "max" entry for other enums that get generated by tables
```
typedef enum {
    #include "tables/actor_table.h"
    /* 0x0192 */ ACTOR_ID_MAX // originally "ACTOR_DLF_MAX"
} ActorID;
```
so we could do that here. just not sure on a name.
it could be named as MAX since its the max in every other sense, but happens to get used as an entry in the entrance table. or it could be named after the fact that its removed/not used

include/functions.h
```diff
@@ -1487,7 +1487,7 @@ void KaleidoScopeCall_Draw(GlobalContext* globalCtx);
 void func_800BC490(GlobalContext* globalCtx, s16 point);
 s32 func_800BC56C(GlobalContext* globalCtx, s16 arg1);
 void func_800BC590(GlobalContext* globalCtx);
-void func_800BC5E0(GlobalContext* globalCtx, s32 arg1);
+void Gameplay_SetupTransition(GlobalContext* globalCtx, s32 TransitionType);
```
Dragorn421, line 1490:
```suggestion
void Gameplay_SetupTransition(GlobalContext* globalCtx, s32 transitionType);
```

src/code/z_play.c
```diff
@@ -43,13 +43,14 @@ void func_800BC590(GlobalContext* globalCtx) {
...
-void func_800BC5E0(GlobalContext* globalCtx, s32 transitionType) {
+void Gameplay_SetupTransition(GlobalContext* globalCtx, s32 transitionType) {
     TransitionContext* transitionCtx = &globalCtx->transitionCtx;
 
     bzero(transitionCtx, sizeof(TransitionContext));
 
     transitionCtx->transitionType = transitionType;
 
+    // types above 32 are treated as circles
     if ((transitionCtx->transitionType >> 5) == 1) {
```
Dragorn421, line 54: I think this comment can be confusing, as the conditional isn't checking `>= 32`. It relies on knowing that the circle types 32-55 are the only types past 32 so idk it's weird

I would just remove the comment, only indicating it's checking for circle, like

```suggestion
    if ((transitionCtx->transitionType >> 5) == 1) { // circle types
```

Dragorn421 (commented):
Small things

Also, would the magic value `0xFF` used for `gSaveContext.nextTransition` warrant a define?

include/z64.h
```diff
@@ -969,6 +969,77 @@ typedef struct {
     /* 0x4C */ u32 unk_4C;
 } PreRender; // size = 0x50
 
+#define TRANS_TRIGGER_OUT -20 // transition is going out
+#define TRANS_TRIGGER_OFF 0 // transition is not active
+#define TRANS_TRIGGER_IN 20 // activate transition to come in
```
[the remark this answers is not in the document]
fig02: I agree on the distinction being confusing, but couldnt figure out how to address it. I went with the fadeOut and fadeIn distinction made by the entrance table 

the variable means way more to the transition than the actual load that occurs after the transition is done, which is why i found it more appropriate to name it after the transition

include/z64.h
```diff
@@ -969,6 +969,77 @@ typedef struct {
...
+    /* 13 */ TRANS_TYPE_FADE_WHITE_CS_DELAYED,
+    /* 14 */ TRANS_TYPE_SANDSTORM_PERSIST,
+    /* 15 */ TRANS_TYPE_SANDSTORM_END,
+    /* 16 */ TRANS_TYPE_CS_BLACK_FILL,
+    /* 17 */ TRANS_TYPE_FADE_WHITE_INSTANT,
+    /* 18 */ TRANS_TYPE_FADE_GREEN,
+    /* 19 */ TRANS_TYPE_FADE_BLUE,
+    // transition types 20 - 31 are unused
+    // transition types 32 - 55 are constructed using the TRANS_TYPE_CIRCLE macro
+    /* 56 */ TRANS_TYPE_MAX = 56
```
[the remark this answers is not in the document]
fig02: `TRANS_TYPE_MAX` is weird to begin with. 
I need to look at it closer but there are checks for being >= to 56, but in both cases it doesnt necessarily treat it as "invalid"

for the instance types, if the type is >= to 56 it will just end the transition and perform the scene load instantly. This is probably related to 
```
void func_800BC88C(GlobalContext* globalCtx) {
    globalCtx->transitionCtx.transitionType = -1;
}
```
pretty sure this is actually 0xFF not -1 since u8, and 255 would satisfy that condition of >= 56

(note that this does not have the same meaning as 0xFF for nextTransition. that should get a define of its own though, mainly wanted to figure out this usage first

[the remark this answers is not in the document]
fig02: oh nvm, got the transitionCtx and global context versions of the variable confused. this one is s32 lol
in that case i dont know how these tie in to each other

include/z64.h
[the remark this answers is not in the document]
Dragorn421: So if I get it right these mean:
```suggestion
#define TRANS_TRIGGER_OUT -20 // activate fading out transition (when entering scene)
#define TRANS_TRIGGER_OFF 0 // transition is not active
#define TRANS_TRIGGER_IN 20 // activate fading in transition (leads to loading next scene)
```
is that right, would that be better?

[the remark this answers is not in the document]
Dragorn421: If it's fine to name these after the fact they are used when changing scene, maybe `trigger_arrive` and `trigger_exit` would be more straightforward but idk.

That said, I'm fine with the current in/out refering to the visual transition instead of the scene transition, it's going to be kind of weird to wrap your head around it regardless, so as long as there are clear comments it's enough for me

So your call :) or if anyone has an opinion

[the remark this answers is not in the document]
fig02: arrive and exit do make it easier to wrap your head around when just reading the defines. I do think these variables mean more for the transistion itself than the scene triggering part, but they are closely tied together so its probably fine. more opinions would be good

src/code/z_scene_table.c
Dragorn421: -1 on adding this deleted (?) scene to the scene enum
if we really want to document it we could have a `#define` for that particular value (and I don't think we should care since it's only a set of entries in the entrance table), but it shouldn't be part of the actual enum imo

fig02: I agree to a point, but also think mzx has a point, theat when modders add a new scene to the end of the list this entrance would have an entry that uses its value. Itd be pretty confusing

include/z64.h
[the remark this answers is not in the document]
EllipticEllipsis: This is one of the first things I looked at in this PR, and I did wonder what these meant with IN/OUT, so I think I'd prefer if they're associated to exit/enter instead.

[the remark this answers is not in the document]
mzxrules: Dragorn's comments have made me realized that that the names are backwards to what the transitions are doing; `TRANS_TRIGGER_IN` initializes a fade out (fade to black/white/etc.) while `TRANS_TRIGGER_OUT` initializes a fade in (fade from black/white/etc. to the game).

I agree with fig in that the variables appear to be specific to the transition system; I almost feel like the weird values assigned might have been related to fade duration (fast fade black/white use a 20 updates/1 second? duration) at one point. But more to my point, I think we should name these variables based on what they do, not the coupling they have with initializing game state switches, and especially not directly tie them into scene switching since there are instances where you don't switch into a scene.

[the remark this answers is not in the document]
fig02: I wouldnt say its wrong, but it depends on perspective. the current names are based off of the fade itself (coming) in, for the fade (going) out. That perspective difference is why the current names are not good, cause you can see it either way

src/code/z_fbdemo_circle.c
```diff
@@ -37,51 +37,50 @@ Gfx sCircleDList[] = {
...
-        this->step = 0x28;
-        this->color.rgba = this->effect == 1 ? RGBA8(0, 0, 0, 255) : RGBA8(160, 160, 160, 255);
+        this->speed = 40;
+        this->color.rgba = this->appearanceType == TCA_WAVE ? RGBA8(0, 0, 0, 255) : RGBA8(160, 160, 160, 255);
     }
-    if (this->unk_14 != 0) {
+    if (this->direction != 0) {
         this->texY = 0;
-        if (this->typeColor == 3) {
+        if ((u32)this->colorType == TCC_SPECIAL) {
```
Dragorn421, line 78:
```suggestion
        if (this->colorType == TCC_SPECIAL) {
```

src/code/z_fbdemo_circle.c
```diff
@@ -182,15 +181,19 @@ s32 TransitionCircle_IsDone(void* thisx) {
...
-    if (type & 0x80) {
-        this->unk_14 = (type >> 5) & 0x1;
-        this->typeColor = (type >> 3) & 0x3;
-        this->speed = type & 0x1;
-        this->effect = (type >> 1) & 0x3;
+    if (type & TC_SET_PARAMS) {
+        // This bit is always going to be set under this context, this does nothing.
+        // Additionally SetType is called twice for circles, and the value will be overwritten on the
+        // second call.
+        this->direction = (type >> 5) & 0x1;
```
Dragorn421, line 188: It took me a while to realize these three lines of comments apply to `this->direction`. I thought the "bit" it refered to was the `TC_SET_PARAMS` bit, and "the value" refered to the four struct members

It may be clearer as
```c
        // The `(type >> 5) & 1` bit is always set in this context of being a circle type,
        // so it does not actually carry the information on the transition direction.
        // SetType is called twice for circles, the actual direction value will be set on the second call.
        this->direction = (type >> 5) & 0x1;
```

src/code/z_play.c
```diff
@@ -325,27 +335,27 @@ void Gameplay_Init(GameState* thisx) {
...
     globalCtx->unk_11E16 = 0xFF;
     globalCtx->unk_11E18 = 0;
     globalCtx->unk_11DE9 = 0;
 
     if (gSaveContext.gameMode != 1) {
         if (gSaveContext.nextTransition == 0xFF) {
-            globalCtx->fadeTransition =
-                (gEntranceTable[((void)0, gSaveContext.entranceIndex) + tempSetupIndex].field >> 7) & 0x7F; // Fade In
+            globalCtx->transitionType =
+                (gEntranceTable[((void)0, gSaveContext.entranceIndex) + tempSetupIndex].field >> 7) & 0x7F;
```
Dragorn421, line 352: I wouldn't get rid of the "Fade In" comment, the mask access looks pretty magic otherwise

src/code/z_play.c
```diff
@@ -467,19 +477,21 @@ void Gameplay_Update(GlobalContext* globalCtx) {
...
-                        s16 sp6E = 0;
+                case TRANS_MODE_SETUP:
+                    if (globalCtx->transitionTrigger != TRANS_TRIGGER_OUT) {
+                        s16 csOffset = 0;
+                        
                         Interface_ChangeAlpha(1);
 
                         if (gSaveContext.cutsceneIndex >= 0xFFF0) {
-                            sp6E = (gSaveContext.cutsceneIndex & 0xF) + 4;
+                            csOffset = (gSaveContext.cutsceneIndex & 0xF) + 4;
```
Dragorn421, line 487:
```suggestion
                            sceneSetupIndex = (gSaveContext.cutsceneIndex & 0xF) + 4;
```
since it's used in a similar fashion to `gSaveContext.sceneSetupIndex`
https://github.com/zeldaret/oot/blob/8926b0858247a1861a045af9cf33ce616054945e/src/code/z_play.c#L290

src/code/z_play.c
```diff
@@ -467,19 +477,21 @@ void Gameplay_Update(GlobalContext* globalCtx) {
...
                         }
 
-                        if (!(gEntranceTable[globalCtx->nextEntranceIndex + sp6E].field & 0x8000)) { // Continue BGM Off
+                        if (!(gEntranceTable[globalCtx->nextEntranceIndex + csOffset].field & 0x8000)) {
```
Dragorn421, line 490: also wouldn't get rid of the "Continue BGM Off" comment
(but if you can word it better be my guest)

src/code/z_play.c
```diff
@@ -489,98 +501,109 @@ void Gameplay_Update(GlobalContext* globalCtx) {
...
+                    if (!R_TRANS_DBG_ENABLED) {
+                        Gameplay_SetupTransition(globalCtx, globalCtx->transitionType);
                     } else {
-                        func_800BC5E0(globalCtx, CREG(12));
+                        Gameplay_SetupTransition(globalCtx, R_TRANS_DBG_TYPE);
                     }
 
-                    if (globalCtx->transitionMode >= 4) {
+                    if (globalCtx->transitionMode >= TRANS_MODE_FILL_WHITE_INIT) {
+                        // special modes break out of this switch, instance types continue
```
Dragorn421, line 511:
```suggestion
                        // special modes break out of this switch, instance modes continue
```
also I don't think the "special modes" terminology is used anywhere else, so it may be better to refer to those as "non-instance modes"?

src/code/z_play.c
```diff
@@ -590,175 +613,184 @@ void Gameplay_Update(GlobalContext* globalCtx) {
...
-                        globalCtx->sceneLoadFlag = 0;
+
+                        globalCtx->transitionTrigger = TRANS_TRIGGER_OFF;
                     } else {
                         globalCtx->transitionCtx.update(&globalCtx->transitionCtx.data, R_UPDATE_RATE);
                     }
                     break;
             }
 
+            // update non-instance type transitions
```
Dragorn421, line 632:
```suggestion
            // update non-instance mode transitions
```
feels weird to write but at this point the type bears no role, right?
can't think of better wording atm

src/code/z_play.c
```diff
@@ -590,175 +613,184 @@ void Gameplay_Update(GlobalContext* globalCtx) {
...
                     break;
 
-                case 5:
-                    globalCtx->envCtx.screenFillColor[3] = (D_801614C8 / 20.0f) * 255.0f;
-                    if (D_801614C8 >= 20 && 1) {
-                        globalCtx->state.running = 0;
+                case TRANS_MODE_FILL_IN:
+                    globalCtx->envCtx.screenFillColor[3] = (sTransitionFillTimer / 20.0f) * 255.0f;
+
+                    if (sTransitionFillTimer >= 20 && 1) {
```
Dragorn421, line 653: engineer's comment on the `&& 1` was marked as resolved but no explanation, and it isn't required for matching, is there any reason to keep it?

Dragorn421 (changes requested):

include/z64environment.h
```diff
@@ -30,6 +30,14 @@ typedef enum {
     /* 13 */ SKYBOX_DMA_PAL2_START
 } SkyboxDmaState;
 
+typedef enum {
+    /* 0 */ SANDSTORM_OFF,
+    /* 1 */ SANDSTORM_FILL,
+    /* 2 */ SANDSTORM_UNFILL,
```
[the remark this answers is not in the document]
fig02: the fill part is referring specifically to the screen fill portion which blocks the whole screen. its not intended to mean the sandstorm as a whole filling the screen. Theres a solid part of the sandstorm that exits while the rest of the sandstorm stays around

fig02:
no idea, the initial instance-type transition documentation was done by krim years ago. im sure even he doesnt remember

src/code/z_play.c
fig02: youre right that at this point the "mode" is what is referenced here, but its still pretty clearly seperated into two seperate block to handle instance types and non instance types. I dont think the switch needs to be on that variable specifically to say that the block of code is handling certain "types"

Dragorn421: for completeness, the switch also handles this single instance related mode `TRANS_MODE_INSTANCE_WAIT` so it's 99% non-instance, but not 100% lol

and sure if you understand the system it's clear the modes correspond to types at this point, but for discovering the system it's not immediately obvious the types are used to initialize the modes, so it may be confusing imo to assimilate types and modes

fig02: imo wording it like this makes the separation sound more extreme than it actually is, handling "modes" just doesnt sound right to me personally. the types are what you interface with as the user and what youll be used to seeing them as

Dragorn421: then a compromise could be to not call them modes or types at all here but just
```suggestion
            // update non-instance transitions
```
lel

src/code/z_play.c
fig02: it may be removable in this specific case, but its required all over with the `1 && HREG(63)` checks. at least it was last i looked

src/code/z_play.c
fig02: scene setup index is used for both the age/time of day offset and cutscenes.
This is specifically the offset as designated by the cutscenes, which is why it does the + 4 to get passed the age/time stuff. So I think this warrants being named after cutscenes specifically

fig02: nvm I understand. its the cs offset + 4 which now makes it the scene setup index

include/z64.h
```diff
@@ -977,6 +977,80 @@ typedef struct {
...
+    /* 1 */ TCA_WAVE,
+    /* 2 */ TCA_RIPPLE,
+    /* 3 */ TCA_STARBURST
+} TransitionCircleAppearance;
+
+typedef enum {
+    /* 0 */ TCC_BLACK,
+    /* 1 */ TCC_WHITE,
+    /* 2 */ TCC_GRAY,
+    /* 3 */ TCC_SPECIAL // color varies depending on appearance. unused
```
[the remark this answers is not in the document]
fig02: think it should still be documented as a type cause it gets referenced as such. but yea ill mention that its broken

include/z64.h
```diff
@@ -977,6 +977,80 @@ typedef struct {
...
+    /* 06 */ TRANS_MODE_FILL_OUT,
+    /* 07 */ TRANS_MODE_FILL_BROWN_INIT,
+    /* 08 */ TRANS_MODE_08, // unused
+    /* 09 */ TRANS_MODE_09, // unused
+    /* 10 */ TRANS_MODE_INSTANT,
+    /* 11 */ TRANS_MODE_INSTANCE_WAIT,
+    /* 12 */ TRANS_MODE_SANDSTORM_INIT,
+    /* 13 */ TRANS_MODE_SANDSTORM,
+    /* 14 */ TRANS_MODE_SANDSTORM_END_INIT,
+    /* 15 */ TRANS_MODE_SANDSTORM_END,
```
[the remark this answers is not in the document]
fig02: for the reason you mentioned i dont think its fair to name it after start. The sandstorm is already started in the case you mentioned so it would be more confusing

include/z64environment.h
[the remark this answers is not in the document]
Dragorn421: It's fine as it is imo, afaict the sandstorm drawing works the same regardless of the sandstorm state (same dlist), it's just different prim and env rgba values.
The fancy combiner `gFieldSandstormDL` uses does the rest

The names are good as is already, and it will also be better eventually :tm: when the `Environment_DrawSandstorm` variables are better documented

fig02:
All changes have been made, including new names for the transition trigger defines

Dragorn421 (approved):
Great job 👍 this is a major improvement on the readability of a very important and not so simple system

Note to myself: investigate what I was talking about two months ago in the comments with a transition related crash in syateki_niw
EDIT: I was mistaken by looking at the wrong thing, nothing wrong here

include/z64scene.h
```diff
@@ -309,6 +309,8 @@ typedef enum {
     /* 0x6E */ SCENE_ID_MAX
 } SceneID;
 
+#define SCENE_UNUSED_6E SCENE_ID_MAX
```
Roman971, line 312: I think this could be worth a comment to explain what's going on

src/code/z_common_data.c
```diff
@@ -18,6 +18,6 @@ void SaveContext_Init(void) {
     gSaveContext.nextDayTime = 0xFFFF;
     gSaveContext.skyboxTime = 0;
     gSaveContext.dogIsLost = true;
-    gSaveContext.nextTransition = 0xFF;
+    gSaveContext.nextTransitionType = 0xFF;
```
Roman971, line 21: Would it make sense to have an enum value for 0xFF? Maybe _NONE since this seems to be the default for not having a specific transition type?

src/code/z_scene_table.c
```diff
@@ -30,784 +30,1738 @@
...
+             TRANS_TYPE_CIRCLE(TCA_NORMAL, TCC_BLACK, TCS_SLOW)),
+    ENTRANCE(SCENE_SPOT00, 0, false, true, TRANS_TYPE_CIRCLE(TCA_WAVE, TCC_BLACK, TCS_SLOW),
+             TRANS_TYPE_CIRCLE(TCA_WAVE, TCC_BLACK, TCS_SLOW)),
+    ENTRANCE(SCENE_SPOT00, 0, false, true, TRANS_TYPE_FADE_WHITE, TRANS_TYPE_FADE_WHITE),
+    ENTRANCE(SCENE_SPOT00, 0, true, true, TRANS_TYPE_FADE_BLACK, TRANS_TYPE_FADE_BLACK),
+    ENTRANCE(SCENE_SPOT00, 0, false, true, TRANS_TYPE_CIRCLE(TCA_WAVE, TCC_BLACK, TCS_SLOW),
+             TRANS_TYPE_CIRCLE(TCA_WAVE, TCC_BLACK, TCS_SLOW)),
+    ENTRANCE(SCENE_SPOT00, 0, false, true, TRANS_TYPE_FADE_WHITE, TRANS_TYPE_FADE_WHITE),
+    ENTRANCE(SCENE_SPOT00, 0, false, true, TRANS_TYPE_CIRCLE(TCA_NORMAL, TCC_WHITE, TCS_SLOW),
+             TRANS_TYPE_CIRCLE(TCA_NORMAL, TCC_WHITE, TCS_SLOW)),
```
Roman971, line 295: Mostly just a note but we will probably have to think of a shorter way to define these entries (maybe shortening the enums or working on the format), since these can get pretty long. Plus they would be even longer with the entrance enum and offsets if the table was made into its own include file

include/z64.h
```diff
@@ -1019,6 +1019,80 @@ typedef struct {
     /* 0x4C */ u32 unk_4C;
 } PreRender; // size = 0x50
 
+#define TRANS_TRIGGER_END -20 // transition is ending (arriving in a new area)
+#define TRANS_TRIGGER_OFF 0 // transition is not active
+#define TRANS_TRIGGER_START 20 // start transition (exiting an area)
```
Roman971, line 1024: Very minor but I feel like this order would make more sense, at least from a functional standpoint
```suggestion
#define TRANS_TRIGGER_OFF 0 // transition is not active
#define TRANS_TRIGGER_START 20 // start transition (exiting an area)
#define TRANS_TRIGGER_END -20 // transition is ending (arriving in a new area)
```

src/code/z_common_data.c
fig02: oh yeah, `nextTransitionType` is kinda better described as an override to take over from what is in the entrance table. I was thinking about naming the variable as such, but wasnt sure if I wanted to commit to it.

the distinction is important I think, cause 0xFF only means something to the "override" variable, it doesnt correspond to an actual transition type that the transition system can do anything about. 

so yea `transitionTypeOverride` and `TRANS_OVERRIDE_NONE` are a possibility, but im not 100% sure.
It can be left as `nextTransistionType` and have a define of `TRANS_TYPE_NONE`, but its slightly misleading cause its not the exact same kind of value as the other `TRANS_TYPE_`s

src/code/z_scene_table.c
fig02: agreed. I tried my best to make them short while also keeping prefixes on the enums. If there are any ideas on how to change them would be interested to hear

src/code/z_common_data.c
fig02: its also *only* for the case where a new scene is loading, so for the `TRANS_TRIGGER_END` case. maybe the name should reflect that as well. cause you end up with code like this where it sets the outgoing transition to one type, and the override for the other side of the transition to another
```
        globalCtx->transitionTrigger = TRANS_TRIGGER_START;
        globalCtx->transitionType = TRANS_TYPE_FADE_WHITE;
        gSaveContext.nextTransitionType = TRANS_TYPE_FADE_WHITE_SLOW;
```
but then again maybe it being called "next" is already getting that point across well, not sure

Dragorn421: Maybe `nextTransitionEndType`/`transitionEndType` ? I'm not a fan though :/

The purpose is to set the transition type for the second half of a scene transition, as you said overriding the "fade in" value of the new entrance

Setting `globalCtx->transitionType = 0xFF` will reach `Fault_AddHungupAndCrash("../z_play.c", 2290);`,
so indeed it may be a bad idea to make 0xFF a `TRANS_TYPE_`

I would see some random name working though, like `TRANS_END_ENTRANCE_TYPE` as in "for the transition end use the entrance's type". I don't think there's really an elegant name, but it's only used in 4 spots so a comment actually explaining what's going on, and a name that can't be confused with something else, is enough I think

src/code/z_scene_table.c
Dragorn421: Overall, I don't think there's a way around the verbosity, there are just many parameters.

Pushing one of my previous suggestions to the extreme:

<details>

```c
#define TCA_R TCA_RIPPLE
#define TCC_W TCC_WHITE
#define TCS_F TCS_FAST

#define TRANS_TYPE_C(a, c, s) TRANS_TYPE_CIRCLE(TCA_##a, TCC_##c, TCS_##s)

#define ENTRANCE(scene, spawn, continueBgm, displayTitleCard, fadeIn, fadeOut)                                   \
    {                                                                                                            \
        (scene), (spawn),                                                                                        \
            (((continueBgm)&1) << 15) | (((displayTitleCard)&1) << 14) | (((TRANS_TYPE_##fadeIn) & 0x7F) << 7) | \
                ((TRANS_TYPE_##fadeOut) & 0x7F)                                                                  \
    }

EntranceInfo gEntranceTable[] = {
    ENTRANCE(SCENE_YDAN, 0, false, true, FADE_BLACK, FADE_BLACK),
    ENTRANCE(SCENE_SPOT06, 7, false, true, C(R, W, F), C(R, W, F)),
};
```

</details>

Less extreme:

<details>

```c
#define TRANS_TYPE_TCIRCLE(a, c, s) TRANS_TYPE_CIRCLE(TCA_##a, TCC_##c, TCS_##s)

#define ENTRANCE(scene, spawn, continueBgm, displayTitleCard, fadeIn, fadeOut)                                   \
    {                                                                                                            \
        (scene), (spawn),                                                                                        \
            (((continueBgm)&1) << 15) | (((displayTitleCard)&1) << 14) | (((TRANS_TYPE_##fadeIn) & 0x7F) << 7) | \
                ((TRANS_TYPE_##fadeOut) & 0x7F)                                                                  \
    }

EntranceInfo gEntranceTable[] = {
    ENTRANCE(SCENE_YDAN, 0, false, true, FADE_BLACK, FADE_BLACK),
    ENTRANCE(SCENE_SPOT06, 7, false, true, TCIRCLE(RIPPLE, WHITE, FAST), TCIRCLE(RIPPLE, WHITE, FAST)),
};
```

</details>

Also, easy space gain, replace true/false with 0/1 :)

include/z64.h
Dragorn421: Looking at it with the play state lifecycle, end/off/start is actually the right order though

`Gameplay_Init`: `globalCtx->transitionTrigger = TRANS_TRIGGER_END;`
this leads to `transitionMode = TRANS_MODE_SETUP;`,
eventually `Gameplay_Update` does `globalCtx->transitionTrigger = TRANS_TRIGGER_OFF;`
and the transition fades out,

(play state runs...)

something sets `globalCtx->transitionTrigger = TRANS_TRIGGER_START;`,
leading to `transitionMode = TRANS_MODE_SETUP;`,
eventually the transition fades in
and then `SET_NEXT_GAMESTATE` puts an end to the current play state

Dragorn421: To be clear I think the end/off/start order should be prefered because that's always how the play state rolls (afaik, besides the debug combo for switching to map select)

off/start/end is the right order too but only if switching from play to play, and there are important cases where play switches to/from another state

fig02: i had it in that order originally just cause it was the order of the values from largest to highest. but yea i can see an argument for both

from a user standpoint youll be comparing against OFF sometimes and triggering it to start when you use it, so I can see why romans order makes sense from the users standpoint, as opposed to the order of things in the perspective of the play gamestate

I dont think i have a strong preference one way or another here but Ill put some more thought into it

src/code/z_common_data.c
fig02: with your explanation I understand what it means, but at first read `TRANS_END_USE_ENTRANCE_TYPE` was very hard to parse. You are right that its not used often, and not meant to be used by users of the system. But if we could have something a bit easier to read I think it would be nice

not that it affects the conversation much, but mm was given names for many save context fields, so the og name of this is `next_wipe`. So what we have currently was along the same lines as their intention. 
I feel like naming it after being an override still make things more clear, but Im not 100% set on that at the moment

But yea I do think naming it separate from `TRANS_TYPE_` is needed here cause it represents a different kind of number. Just need to decide on something to go with

Dragorn421: brainstorming here...

Maybe dropping the "END" makes it more legible `TRANS_USE_ENTRANCE_TYPE`

Or replace it with "NEXT"? `TRANS_NEXT_USE_ENTRANCE_TYPE`

`TRANS_NEXT_TYPE_USE_ENTRANCE` (also close to `nextTransitionType` 👀)

`TRANS_NEXT_TYPE_FROM_ENTRANCE`

Dragorn421: I feel like adding "override" in the name somehow would be sort of cramming it in there, and you would also have to negate it which usually doesn't look too good in names

I do like `TRANS_NEXT_TYPE_FROM_ENTRANCE` myself

I don't see the 0xFF value as "don't override" either but as the default behavior, since it's reset after being used if not 0xFF
https://github.com/fig02/oot/blob/01c573a3416aff4619b2446722aeb0cad7293792/src/code/z_play.c#L350-L357
```c
        if (gSaveContext.nextTransitionType == 0xFF) {
            // fade in
            globalCtx->transitionType =
                (gEntranceTable[((void)0, gSaveContext.entranceIndex) + tempSetupIndex].field >> 7) & 0x7F;
        } else {
            globalCtx->transitionType = gSaveContext.nextTransitionType;
            gSaveContext.nextTransitionType = 0xFF;
        }
```

Roman971: I also see as the default behavior, which is why I suggested `_NONE` originally. I would suggest `TRANS_NEXT_TYPE_DEFAULT` right now, but `TRANS_NEXT_TYPE_FROM_ENTRANCE` seems fine as well

fig02: I think i like default much more. a comment can explain that default in this context means that it will take from the entrance table. but that info being in the define name itself seems like a lot to try and parse, imo

Dragorn421: +1 on `TRANS_NEXT_TYPE_DEFAULT`, the name indeed doesn't need to say where the default comes from

include/z64.h
fig02: I went with changing the order just cause I think it does makes sense to see the different types listed in that way. I still agree that the other order makes sense in the context of the play life cycle, but guess I see that as less important over having it be in a more natural order

Roman971 (approved):

PR #1134, opened 2022-02-06
Document more PathwayLists for scenes

mrneo240:
- Rename Paths to recent naming convention
- Document more correctly the length of Path lists

### DIFF
--- a/assets/xml/scenes/dungeons/ganon_final.xml
+++ b/assets/xml/scenes/dungeons/ganon_final.xml
@@ -1,5 +1,6 @@
 <Root>
     <File Name="ganon_final_scene" Segment="2">
+        <Path Name="ganon_final_scenePathway_0001F4" Offset="0x01F4" NumPaths="6"/>
         <Scene Name="ganon_final_scene" Offset="0x0"/>
     </File>
     <File Name="ganon_final_room_0" Segment="3">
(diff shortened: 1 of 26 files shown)


fig02 (commented):
Thanks for this.
The term "pathway" is not used anywhere else in the codebase, "path list" seems more appropriate imo. I know you were just going off of the existing usage of that term, but it seems to only be used in scenes and is probably old

mrneo240:
> The term "pathway" is not used anywhere else in the codebase, "path list" seems more appropriate imo.
Unsure how collaboration and maintainers work with this repo, but if Pathlist is preferred over Pathway that's fine, I'll commit.

assets/xml/scenes/dungeons/ganon_final.xml
```diff
@@ -1,5 +1,6 @@
 <Root>
     <File Name="ganon_final_scene" Segment="2">
+        <Path Name="ganon_final_scenePathway_0001F4" Offset="0x01F4" NumPaths="6"/>
```
Dragorn421, line 3: I think there are only 4 here?

Dragorn421 (commented):
+1 for `PathList` like in `GlobalContext` or in the scene command:

https://github.com/zeldaret/oot/blob/eadc477187888e1ae078d021b4a00b1366f0c9a4/include/z64.h#L1186

https://github.com/zeldaret/oot/blob/eadc477187888e1ae078d021b4a00b1366f0c9a4/include/z64scene.h#L285

assets/xml/scenes/dungeons/ganon_final.xml
mrneo240: yes, my math was off. will correct when renaming to PathList for review

fig02:
I think we are still waiting on pathway -> path and fixing an error mentioned by dragorn above. 
if you dont want to maintain this pr anymore or arent active we can close this and do it in a future cleanup pass

fig02:
closing due to inactivity, feel free to open it again if you are around and want to revisit the feedback above

PR #1135, opened 2022-02-06
Force required alignment on DMA related stuff

AngheloAlf:
The address of any variable that's going to be DMA'd requires dword alignment (8 byte boundaries).

- Converts `OSPifRam` to an union and adds a `u64 force_structure_alignment;`.
- Add a `u64 force_structure_alignment` member to every buffer in `Font`.
- Add `ALIGNED8` to `gSaveContext` as a temporary fix until the main struct gets re-estructured.
- Use `ALIGN8(PLAYER_LIMB_MAX)` for `PLAYER_LIMB_BUF_COUNT`.
- Creates an `alignment.h` header and moves all the alignment related macros there.

### DIFF
--- a/include/z64player.h
+++ b/include/z64player.h
@@ -2,6 +2,7 @@
 #define Z64PLAYER_H
 
 #include "z64actor.h"
+#include "alignment.h"
 
 struct Player;
 
@@ -158,7 +159,7 @@ typedef enum {
 } PlayerDoorType;
 
 
-#define PLAYER_LIMB_BUF_COUNT PLAYER_LIMB_MAX + 2 // 2 extra entries in limb buffers?
+#define PLAYER_LIMB_BUF_COUNT ALIGN8(PLAYER_LIMB_MAX)
 
 typedef struct {
     /* 0x00 */ f32 unk_00;
--- a/include/z64player.h
+++ b/include/z64player.h
@@ -159,7 +159,7 @@ typedef enum {
 } PlayerDoorType;
 
 
-#define PLAYER_LIMB_BUF_COUNT ALIGN8(PLAYER_LIMB_MAX)
+#define LIMB_BUF_COUNT(limbCount) ((ALIGN16((limbCount) * sizeof(Vec3s)) + sizeof(Vec3s) - 1) / sizeof(Vec3s))
 
 typedef struct {
     /* 0x00 */ f32 unk_00;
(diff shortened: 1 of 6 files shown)


fig02:
theres probably a discussion to be had on if we want to maintain nintendos variable styling in `force_strcture_alignment` in our own stuff. I can see it both ways, the inconsistency isnt great but it makes it stand out from a variable that is actually supposed to be used

fig02 (approved):

src/code/z_common_data.c
```diff
@@ -1,6 +1,6 @@
 #include "global.h"
 
-SaveContext gSaveContext;
+ALIGNED8 SaveContext gSaveContext;
```
EllipticEllipsis, line 3: I'd put a note that this is temporary until SaveContext is restructured.

include/z64player.h
```diff
@@ -158,7 +159,7 @@ typedef enum {
 } PlayerDoorType;
 
 
-#define PLAYER_LIMB_BUF_COUNT PLAYER_LIMB_MAX + 2 // 2 extra entries in limb buffers?
+#define PLAYER_LIMB_BUF_COUNT ALIGN8(PLAYER_LIMB_MAX)
```
EllipticEllipsis, line 162: Do we want to turn the tables into buffers now, or save that for proper documentation?

Dragorn421, line 162: How is alignment on an array length supposed to help with alignment in memory?

fig02: this is assuming we keep them typed as vec3s right? I didnt really follow the conversation, but there was some talks about not even typing the buffers as vec3s anymore. not sure what the merit is in doing that and what the positives and negatives are, just parroting what i read in discord (and elliptics comments above)

fig02: I guess we go with the above suggestion and talk about the buffer thing later

AngheloAlf: Oh yeah, sorry, I haven't got time to update this PR. Will do now

fig02: yea no worries, I meant more so the discussion on the plain buffer vs Vec3s thing, but that should probably happen outside of this pr. so romans suggestion is fine

fig02 (approved):

EllipticEllipsis (approved):

include/z64player.h
```diff
@@ -158,7 +159,7 @@ typedef enum {
 } PlayerDoorType;
 
 
-#define PLAYER_LIMB_BUF_COUNT PLAYER_LIMB_MAX + 2 // 2 extra entries in limb buffers?
+#define LIMB_BUF_COUNT(limbCount) ((ALIGN16((limbCount) * sizeof(Vec3s)) + sizeof(Vec3s) - 1) / sizeof(Vec3s))
```
Roman971, line 162: It's pretty minor but I'd suggest keeping a `PLAYER_LIMB_BUF_COUNT` define to avoid duplication and keep things simpler in the struct

fig02: so that is, just define `PLAYER_LIMB_BUF_COUNT` as LIMB_BUF_COUNT(PLAYER_LIMB_MAX)`?

Roman971: Yeah basically

fig02: ill do that now, its a small change

PR #1136, opened 2022-02-07
Anubice fix, bit more documentation

EllipticEllipsis:
Fixed the out-of-bounds array access in EnAnubice, as originally pointed out by @mzxrules , and observed by @fig02 during GCC testing. Also took the opportunity to do a small amount of further documentation of the actor, limb enum, damage effect enum, rest of the struct variables, make function names consistent, etc., although nothing particularly significant.

Also also, added a top-of-file comment to BgJyaMegami, since I noticed it was missing during other research.

### DIFF
--- a/src/overlays/actors/ovl_En_Anubice/z_en_anubice.c
+++ b/src/overlays/actors/ovl_En_Anubice/z_en_anubice.c
@@ -21,8 +21,8 @@ void EnAnubice_FindFlameCircles(EnAnubice* this, GlobalContext* globalCtx);
 void EnAnubice_SetupIdle(EnAnubice* this, GlobalContext* globalCtx);
 void EnAnubice_Idle(EnAnubice* this, GlobalContext* globalCtx);
 void EnAnubice_GoToHome(EnAnubice* this, GlobalContext* globalCtx);
-void EnAnubis_SetupShootFireball(EnAnubice* this, GlobalContext* globalCtx);
-void EnAnubis_ShootFireball(EnAnubice* this, GlobalContext* globalCtx);
+void EnAnubice_SetupShootFireball(EnAnubice* this, GlobalContext* globalCtx);
+void EnAnubice_ShootFireball(EnAnubice* this, GlobalContext* globalCtx);
 void EnAnubice_Die(EnAnubice* this, GlobalContext* globalCtx);
 
 const ActorInit En_Anubice_InitVars = {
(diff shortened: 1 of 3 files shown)


src/overlays/actors/ovl_En_Anubice/z_en_anubice.h
```diff
@@ -28,15 +48,15 @@ typedef struct EnAnubice {
...
     /* 0x0280 */ Vec3f fireballPos;
     /* 0x028C */ Vec3f fireballRot;
     /* 0x0298 */ Vec3f home;
     /* 0x02A4 */ Vec3f knockbackRecoveryVelocity;
-    /* 0x02B0 */ BgHidanCurtain* flameCircles[4];
-    /* 0x02C0 */ char unk_2C0[0x8];
-    /* 0x02C8 */ ColliderCylinder col;
+    /* 0x02B0 */ BgHidanCurtain* flameCircles[5];
+    /* 0x02C0 */ char unk_2C4[0x4];
+    /* 0x02C8 */ ColliderCylinder collider;
```
fig02, line 59: update comment for 2C4. rest looks good

fig02 (approved):

src/overlays/actors/ovl_En_Anubice/z_en_anubice.c
```diff
@@ -354,15 +361,15 @@ void EnAnubice_Update(Actor* thisx, GlobalContext* globalCtx) {
...
-        for (i = 0; i < 5; i++) {
+        for (i = 0; i < ARRAY_COUNT(this->flameCircles); i++) {
             flameCircle = this->flameCircles[i];
 
-            if (flameCircle != NULL && fabsf(flameCircle->actor.world.pos.x - this->actor.world.pos.x) < 60.0f &&
-                fabsf(this->flameCircles[i]->actor.world.pos.z - this->actor.world.pos.z) < 60.0f &&
-                flameCircle->timer != 0) {
+            if ((flameCircle != NULL) && (fabsf(flameCircle->actor.world.pos.x - this->actor.world.pos.x) < 60.0f) &&
+                (fabsf(this->flameCircles[i]->actor.world.pos.z - this->actor.world.pos.z) < 60.0f) &&
+                (flameCircle->timer != 0)) {
```
mzxrules, line 372: why add back unnecessary parens?

fig02: everyone has their own preferences regarding this, and we dont enforce one or the other. this looks more readable to me personally, but understand not everyone shares that opinion 
until when and if we decide to standardize one over the other, theres not really much that can be said

jpburnett: I believe when it comes to warnings, if logic such as `&&` or `||` is not enclosed in parens, it will create a warning, so this is just preventing warnings. At least that is what I know in regards to how things are in MM.

fig02: this doesnt induce warnings at least in oot, its purely preference. but yea i guess mm may have different warnings enabled

Dragorn421: I prefer less parenthesis usage as they look really noisy in longer conditionals, but I think we should enforce explicit parenthesis everywhere for the sake of consistency.

src/overlays/actors/ovl_En_Anubice/z_en_anubice.c
```diff
@@ -292,52 +299,52 @@ void EnAnubice_SetupDie(EnAnubice* this, GlobalContext* globalCtx) {
...
 
     SkelAnime_Update(&this->skelAnime);
     Math_ApproachZeroF(&this->actor.shape.shadowScale, 0.4f, 0.25f);
 
-    if (this->unk_256) {
-        Math_SmoothStepToS(&this->actor.shape.rot.y, this->unk_258, 1, 10000, 0);
-        if (fabsf(this->actor.shape.rot.y - this->unk_258) < 100.0f) {
-            this->unk_256 = false;
+    if (this->isFallingOver) {
+        Math_SmoothStepToS(&this->actor.shape.rot.y, this->fallTargetPitch, 1, 10000, 0);
```
Dragorn421, line 327: So `fallTargetPitch` is initialized using rot.x but is used to change rot.y :thinking:

Are you sure about this falling over thing?

EllipticEllipsis: Hmm, I'm probably mixing up my x and y rotation. Either way it's a bit odd...

src/overlays/actors/ovl_En_Anubice/z_en_anubice.c
mzxrules: and I say we shouldn't enforce a bad pattern simply because it was more comfortable for mips2c to generate.

fig02: I mean, no lol. I just said above that I prefer explicit parenthesis, and clearly elliptic prefers it as well cause he added them here. And plenty more have said they prefer it. Its not just because of mips2c.
That said there are just as many that dont like them, its a pretty split issue.

src/overlays/actors/ovl_En_Anubice/z_en_anubice.c
```diff
@@ -292,52 +299,52 @@ void EnAnubice_SetupDie(EnAnubice* this, GlobalContext* globalCtx) {
...
+    this->fallTargetPitch = 0;
     this->deathTimer = 20;
     this->actor.velocity.x = this->actor.velocity.z = 0.0f;
     this->actor.gravity = -1.0f;
 
     if (BgCheck_SphVsFirstPoly(&globalCtx->colCtx, &this->fireballPos, 70.0f)) {
-        this->unk_256 = true;
-        this->unk_258 = this->actor.shape.rot.x - 0x7F00;
+        this->isFallingOver = true;
+        this->fallTargetPitch = this->actor.shape.rot.x - 0x7F00;
```
fig02, line 310: looking at this closer, im pretty sure that fireballPos is more correctly the headPos, looking at how it gets assigned in postLimbDraw.
it gets used to spawn the fireball initially, but its not the position of it as it travels. That makes this check here make alot more sense, if its head hits the wall (or any poly) as it falls, it tips over the other way to bounce off of it

fig02: assuming im right (can maybe test in game to double check), this can be more named after `ricochet` or something of the effect

src/overlays/actors/ovl_En_Anubice/z_en_anubice.c
```diff
@@ -274,11 +281,11 @@ void EnAnubis_ShootFireball(EnAnubice* this, GlobalContext* globalCtx) {
         Math_SmoothStepToS(&this->actor.shape.rot.y, this->actor.yawTowardsPlayer, 5, 3000, 0);
     }
 
-    EnAnubice_SetFireballRot(this, globalCtx);
+    EnAnubice_SetHeadRot(this, globalCtx);
```
fig02, line 284: we can revert the func name too
SetFireBallDirection or something along those lines might make more sense over just plainly setting the rotation. Or involve that its calculating where to send the fireball to hit link

EllipticEllipsis: AimFireball?

fig02: possibly. that makes it sound like more of an action to me than just a simple calculation + set, but its probably fine

src/overlays/actors/ovl_En_Anubice/z_en_anubice.c
```diff
@@ -292,52 +299,53 @@ void EnAnubice_SetupDie(EnAnubice* this, GlobalContext* globalCtx) {
...
     SkelAnime_Update(&this->skelAnime);
     Math_ApproachZeroF(&this->actor.shape.shadowScale, 0.4f, 0.25f);
 
-    if (this->unk_256) {
-        Math_SmoothStepToS(&this->actor.shape.rot.y, this->unk_258, 1, 10000, 0);
-        if (fabsf(this->actor.shape.rot.y - this->unk_258) < 100.0f) {
-            this->unk_256 = false;
+    // If near a wall, turn to avoid the tail falling through it during the death animation
+    if (this->isNearWall) {
+        Math_SmoothStepToS(&this->actor.shape.rot.y, this->fallTargetYaw, 1, 10000, 0);
```
Dragorn421, line 328: You think that's the explanation? 🤔 since `fallTargetYaw` is always set to the same thing instead of e.g. being relative to said wall (which btw doesn't have to be a wall, I think it could also be floor or ceiling if there was any nearby enough), it doesn't guarantee it will fall away from the wall

src/overlays/actors/ovl_En_Anubice/z_en_anubice.c
```diff
@@ -292,52 +299,53 @@ void EnAnubice_SetupDie(EnAnubice* this, GlobalContext* globalCtx) {
...
+    rotX = curFrame * -3000.0f;
+    rotX = CLAMP_MIN(rotX, -11000.0f);
 
     Matrix_RotateY(BINANG_TO_RAD(this->actor.shape.rot.y), MTXMODE_NEW);
-    Matrix_RotateX(BINANG_TO_RAD(phi_f2), MTXMODE_APPLY);
-    sp4C.y = Rand_CenteredFloat(10.0f) + 30.0f;
-    Matrix_MultVec3f(&sp4C, &fireEffectPos);
+    Matrix_RotateX(BINANG_TO_RAD(rotX), MTXMODE_APPLY);
+    fireEffectInitialPos.y = Rand_CenteredFloat(10.0f) + 30.0f;
+    Matrix_MultVec3f(&fireEffectInitialPos, &fireEffectPos);
```
Dragorn421, line 341: the source vector is the model space position of the effect, roughly (assuming the rotX matches the death animation)
```suggestion
    Matrix_MultVec3f(&fireEffectModelPos, &fireEffectPos);
```
at this point `fireEffectPos` is the offset from the actor in world coordinates, and on the next lines becomes a world position, so I guess it can just stay "pos" as is 🤷

src/overlays/actors/ovl_En_Anubice/z_en_anubice.c
```diff
@@ -387,20 +395,20 @@ void EnAnubice_Update(Actor* thisx, GlobalContext* globalCtx) {
...
+                baseKnockbackVelocity.x = 0.0f;
+                baseKnockbackVelocity.y = 0.0f;
+                baseKnockbackVelocity.z = -10.0f;
+                rotatedKnockbackVelocity.x = 0.0f;
+                rotatedKnockbackVelocity.y = 0.0f;
+                rotatedKnockbackVelocity.z = 0.0f;
 
                 Matrix_RotateY(BINANG_TO_RAD(this->actor.shape.rot.y), MTXMODE_NEW);
-                Matrix_MultVec3f(&sp48, &sp3C);
+                Matrix_MultVec3f(&baseKnockbackVelocity, &rotatedKnockbackVelocity);
```
Dragorn421, line 406: can use "model" and "world" here

src/overlays/actors/ovl_En_Anubice/z_en_anubice.h
```diff
@@ -9,34 +9,54 @@ struct EnAnubice;
...
     /* 0x014C */ SkelAnime skelAnime;
-    /* 0x0190 */ Vec3s jointTable[16];
-    /* 0x01F0 */ Vec3s morphTable[16];
+    /* 0x0190 */ Vec3s jointTable[ANUBICE_LIMB_MAX];
+    /* 0x01F0 */ Vec3s morphTable[ANUBICE_LIMB_MAX];
     /* 0x0250 */ EnAnubiceActionFunc actionFunc;
     /* 0x0254 */ s16 timeAlive;
-    /* 0x0256 */ s16 unk_256;
-    /* 0x0258 */ s16 unk_258;
+    /* 0x0256 */ s16 isNearWall; // Technically any bg poly, but in practice usually a wall
```
Dragorn421, line 39:
```suggestion
    /* 0x0256 */ s16 isNearWall; // Technically any bg poly, but in practice usually a wall
```
I would avoid using "bg" where we can, and use "collision" instead

src/overlays/actors/ovl_En_Anubice_Tag/z_en_anubice_tag.c
```diff
@@ -77,21 +77,16 @@ void EnAnubiceTag_ManageAnubis(EnAnubiceTag* this, GlobalContext* globalCtx) {
...
-    } else {
-        if (anubis->isMirroringLink) {
-            anubis->isLinkOutOfRange = true;
+    if (this->actor.xzDistToPlayer < (200.0f + this->extraTriggerRange)) {
+        if (!anubis->isPlayerOutOfRange && !anubis->isKnockedback) {
+            anubis->isMirroringPlayer = true;
+            offset.x = -Math_SinS(this->actor.yawTowardsPlayer) * this->actor.xzDistToPlayer;
+            offset.z = -Math_CosS(this->actor.yawTowardsPlayer) * this->actor.xzDistToPlayer;
+            Math_ApproachF(&anubis->actor.world.pos.x, (this->actor.world.pos.x + offset.x), 0.3f, 10.0f);
+            Math_ApproachF(&anubis->actor.world.pos.z, (this->actor.world.pos.z + offset.z), 0.3f, 10.0f);
```
Dragorn421, line 86: while you're at it
```suggestion
            Math_ApproachF(&anubis->actor.world.pos.x, this->actor.world.pos.x + offset.x, 0.3f, 10.0f);
            Math_ApproachF(&anubis->actor.world.pos.z, this->actor.world.pos.z + offset.z, 0.3f, 10.0f);
```

src/overlays/actors/ovl_En_Anubice/z_en_anubice.c
fig02: its the intent at least. since the "rebound" yaw target is always `0 - 0x7F00` (since it uses x rot for god knows why), theres prob a scenario where the wall is already at that angle and it will turn into it. we can try to clarify that in an additional comment 

I think it would be more correct to say it tries to avoid the head from falling through the wall(poly) instead of the tail.

src/overlays/actors/ovl_En_Anubice/z_en_anubice.h
fig02: until when and if we agree to use "collision" to only mean scene/dyna stuff, its prob for the best to be bg so people dont confuse it as including a collider

Dragorn421: well "collision poly" is non ambiguous imo, that's even the name of a struct involved in collision headers.

Also in some contexts "bg poly" only refers to "static"/"scene" collision iirc, excluding dynapoly, which was more my concern here

but yeah sure this can be cleaned up eventually

src/overlays/actors/ovl_En_Anubice/z_en_anubice.c
Dragorn421: It sounds like a sensible guess then, but should be worded as such to indicate being just a guess imo, like "this likely was intended to ... but is not functional"

fig02: its functional as long as the wall does not have an angle of -0x7F00. I think all the context leading up to it makes it very clear that its what its supposed to do, its just bugged.

Theres no other reason why you would want to check for a nearby poly when dying, and changing the angle to fall to if its true, imo. its just they didnt do a sufficent job calculating which angle it should go

fig02: I think a fair comment is 
```
// If near a wall, turn away from it while dying to avoid going through it.
// The implementation of this is bugged in the sense that the target angle is 
// hardcoded in practice. If the poly already has an angle of -0x7FFF, the expected behavior won't occur
```
you can choose to go into more detail about it being subtracted from x rotation which is always going to be 0, but its probably a bit too much to take in at once. dunno

Roman971 (approved):

src/overlays/actors/ovl_En_Anubice/z_en_anubice.h
fig02: yeah thats fair. i agree going with collision long term will be for the best, but plenty of people still refer to colliders as being under the "collision" umbrella. at least colloquially. and the file that handles it is called collision_check

we for sure need to clean this up in general at a later time

fig02 (approved):

PR #1138, opened 2022-02-10
Update Fado

EllipticEllipsis:
New stuff:
- Fix Mac support by 
  - using compiler defines instead of `endian.h`.
  - Extracting what we need from `elf.h` to our own header, `mips_elf.h`.
- Implement support for experimental options.
- Experimental support for alignments other than 0x10 if people want to test it (compile Fado with `make EXPERIMENTAL=1` to use).
- Bugfixes in `help.c` and `main.c`.
- Documentation.


i82orbom:
Checked in MacOS and it builds successfully, thanks for fixing it

PR #1140, opened 2022-02-12
Use `__builtin_offsetof` for non IDO builds

AngheloAlf:
The current way this repo handles `OFFSETOF` can be considered as UB. 
We already know it works fine with both IDO and GCC,  but some future version of GCC may change its behavior. So this PR changes the macro to use `__builtin_offsetof` on GCC builds.
I also renamed the macro to `offsetof`, because that is the name of this macro in the C89 standard, and it works exactly the same as the standard says.

### DIFF
--- a/include/ichain.h
+++ b/include/ichain.h
@@ -1,6 +1,8 @@
 #ifndef ICHAIN_H
 #define ICHAIN_H
 
+#include "libc/stddef.h"
+
 typedef struct {
     u32 cont:   1;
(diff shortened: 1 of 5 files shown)


Dragorn421 (approved):

fig02 (approved):

EllipticEllipsis (approved):
Would be curious to know how IDO actually defines this.

Roman971 (approved):

PR #1141, opened 2022-02-13
Document/fix `Player.bodyPartsPos` usage

Dragorn421:
Separated from #1098


fig02 (approved):

include/z64player.h
```diff
@@ -150,6 +150,28 @@ typedef enum {
     /* 0x16 */ PLAYER_LIMB_MAX
 } PlayerLimb;
 
+typedef enum {
+    /*  0 */ PLAYER_BODYPART_WAIST,      // PLAYER_LIMB_WAIST
```
EllipticEllipsis, line 154: Given that this is the same sort of thing as PlayerLimb, seems like it should use the same sort of indexing (hex/dec), right?

src/code/z_player_lib.c
```diff
@@ -858,7 +858,7 @@ s32 func_8008FCC8(GlobalContext* globalCtx, s32 limbIndex, Gfx** dList, Vec3f* p
     if (limbIndex == PLAYER_LIMB_ROOT) {
         D_80160014 = this->leftHandType;
         D_80160018 = this->rightHandType;
-        D_80160000 = &this->swordInfo[2].base;
+        sCurBodyPartPos = &this->bodyPartsPos[-1];
```
EllipticEllipsis, line 861: Hmm, does this lead to oob reads anywhere?

EllipticEllipsis (commented):
Nice job, couple of minor things.

src/code/z_player_lib.c
fig02: only way this will ever stay null is if player has no dlists at all for any limb in his skeleton (wont happen)
agree its sketch though

EllipticEllipsis: Might be worth noting that, then, especially given that #1142 has an example of an actual oob read in the same file.

include/z64player.h
Dragorn421: hex/dec debate hype

It corresponds to various limbs but i don't think that means it should necessarily use the same "unit" for indexing

That said, I think near everything should be in decimal unless it makes special sense in hexadecimal (like binangles for example), and imo limb indices don't need to use hexadecimal at all
So if I was to change something it would be the `PlayerLimb` indices

src/code/z_player_lib.c
Dragorn421: yeah it's incredibly sketch

this will work correctly as long as you don't touch link's skeleton basically, since the bodypart indices rely on both limb order and which limbs have dlists attached or not

In `func_8008FCC8` (the common overridelimbdraw):
- `sCurBodyPartPos = &this->bodyPartsPos[-1];` happens when the limb is root (first processed limb)
- `sCurBodyPartPos++` happens when the limb isn't root and the limb's dlist is non-NULL

In `func_80090D20` (postlimbdraw):
- `Matrix_MultVec3f(&sZeroVec, sCurBodyPartPos);` if the limb's dlist is non-NULL

so if the root limb had a dlist then yeah `this->bodyPartsPos[-1]` would be written to

It's easier to follow with the skelanime callbacks named from #1098, I don't think it needs an explanation, does it?

EllipticEllipsis: Ah, that's what this is about, I see, thanks. Well, it probably is worth pointing out that what looks like it might cause an oob read, actually it will not provided that ROOT doesn't have a dlist. But I do feel like at this point indexing with -1 kind of needs a comment either way.

include/z64player.h
Roman971: To be clear, the reason we generally use hex comments is to make it easier to identify and work with in hex editing tools, memory viewers, debuggers, mips2c output, etc... Not because it makes more sense as hex vs decimal.
External documentation typically uses hex as well, so having hex comments helps with the transfer of knowledge (both ways)

I think decimal comments are a lot less useful and we should stick to hex comments in most if not all cases, but yeah that's just my opinion.

Note: I agree almost every raw number should be in decimal when it comes to the code itself, but that's a separate matter imo

Dragorn421: I realize it's more convenient to have everything as hex for memory viewing purposes (and even then, I think the memory viewer used should have type conversion capabilities, like the Hxd or GHex hex editors do for example, so there's no hex reading to do)

But it's also more convenient to just use decimal because we humans have been trained to use base 10, and when reading code you're most likely going to need the decimal number to understand what it is. (e.g. to figure out the index into an array. Of course we can count in hex too but it's more difficult and why would we)

Thankfully we can rely on programs to do hex<->dec conversions for now, like the HexInspector extension for VS Code or just various calculators, so the debate doesn't need to be settled now or possibly ever.

Note some things like offsets of struct members as hex is perfectly fine, as the code doesn't care about it and it's purely for memory/assembly viewing

-------------

That said, do I change the body part indices to hex, the limb indices to dec, or leave as is?

The repo is inconsistent about it, it mixes hexadecimal `0x08`, decimal `8` and whatever the cursed `08` octal-notation-but-actually-decimal is. Not just limb enums but basically all enums

fig02: I think we need to think about if we want to even keep these offset comments long term. In my view they were meant to be temporary while we did decompiling work. That is, to make it easier to replace them while actually decompiling. 
In a modding context, all offset comments are just going to get in the way and they can become inaccurate at any moment if someone wants to shift a struct or enum

It may still be worth keeping them for retail decomp, but beyond that not sure if theyre worth staying

Dragorn421: There are a lot of cases where not having the offsets would lead to the formatter doing a very bad job. unless an enum being defined on a single line is fine
But surely we can improve how the formatter behaves so that shouldn't be a reason to keep the comments

Roman971: Regarding the formatter, I'm psure it should align things vertically if we have a trailing comma in the enum.

Otherwise I think the comments are useful at least until we support other versions, so we can easily identify stuff in mips2c output and asm diffs, and this is also a strong part in why I think hex is a lot better than decimal overall.

fig02: we currently warn for trailing commas in enums, but if we can disable that itd be cool

EllipticEllipsis: Yeah, MM already disables that warning, although we usually leave off trailing commas anyway. I think, certainly for the larger enums, it's useful to retain hex for the sake of looking in memory - are we really contemplating dropping them for scenes or items? If we did it would impoverish decomp as a reference for information about the binary: not everyone is going to build decomp and print stuff or whatever to understand how something works (quite apart from the memory-sensitive applications).

fig02 (approved):

Roman971 (approved):
Looks good to me tho I still think we should use hex comments for the enum

PR #1143, opened 2022-02-13
Doc (partially) player model and anim code data

Dragorn421:
Separated from #1098

`D_80853914` is now a 2d array instead, as the two functions linked below seem to indicate but it doesn't match with the proper 2d access in those two functions
[D_80853914](https://github.com/Dragorn421/oot/blob/b71ba439ea967a29b58d843bf57ae64eaca1b24b/src/overlays/actors/ovl_player_actor/z_player.c#L625)
[func_80833664](https://github.com/Dragorn421/oot/blob/b71ba439ea967a29b58d843bf57ae64eaca1b24b/src/overlays/actors/ovl_player_actor/z_player.c#L1611)
[func_80841860](https://github.com/Dragorn421/oot/blob/b71ba439ea967a29b58d843bf57ae64eaca1b24b/src/overlays/actors/ovl_player_actor/z_player.c#L7008)

I'm really not sure about how the sheath models are handled


src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -622,229 +622,142 @@ static GetItemEntry sGetItemTable[] = {
...
-static LinkAnimationHeader* D_80853D34[] = {
-    &gPlayerAnim_003468, &gPlayerAnim_003438, &gPlayerAnim_003438,
-    &gPlayerAnim_003468, &gPlayerAnim_003468, &gPlayerAnim_003468,
+static LinkAnimationHeader* D_80853914[PLAYER_ANIMGROUP_MAX][PLAYER_ANIMTYPE_MAX] = {
+    /* PLAYER_ANIMGROUP_0 */
+    { &gPlayerAnim_003240, &gPlayerAnim_003238, &gPlayerAnim_003238, &gPlayerAnim_002BE0, &gPlayerAnim_003240,
+      &gPlayerAnim_003240 },
+    /* PLAYER_ANIMGROUP_1 */
+    { &gPlayerAnim_003290, &gPlayerAnim_003268, &gPlayerAnim_003268, &gPlayerAnim_002BF8, &gPlayerAnim_003290,
+      &gPlayerAnim_003290 },
```
fig02, line 631: dont like what the formatter did here but its not in our control and will change when the anims are named. so w/e

fig02 (approved):

fig02 (approved):

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -7059,8 +7018,9 @@ void func_808417FC(Player* this, GlobalContext* globalCtx) {
 
 void func_80841860(GlobalContext* globalCtx, Player* this) {
     f32 frame;
-    LinkAnimationHeader* sp38 = D_80853914[this->modelAnimType + 144];
-    LinkAnimationHeader* sp34 = D_80853914[this->modelAnimType + 150];
+    // fake match? see func_80833664
+    LinkAnimationHeader* sp38 = D_80853914[0][this->modelAnimType + PLAYER_ANIMGROUP_24 * ARRAY_COUNT(D_80853914[0])];
+    LinkAnimationHeader* sp34 = D_80853914[0][this->modelAnimType + PLAYER_ANIMGROUP_25 * ARRAY_COUNT(D_80853914[0])];
```
Roman971, line 7023: Isn't that also UB?
I think we should probably make it a 1D array if we don't have a choice, and perhaps with a macro to simplify the code when accessing it (to avoid the * PLAYER_ANIMTYPE_MAX everywhere)

include/z64player.h
```diff
@@ -180,6 +180,123 @@ typedef enum {
...
+    /* 15 */ PLAYER_MODELTYPE_RH_HOOKSHOT, // holding hookshot (child: empty hand)
+    // sheath
+    /* 16 */ PLAYER_MODELTYPE_SHEATH_16, // sheathed kokiri/master sword?
+    /* 17 */ PLAYER_MODELTYPE_SHEATH_17, // empty sheath?
+    /* 18 */ PLAYER_MODELTYPE_SHEATH_18, // sword sheathed and shield on back?
+    /* 19 */ PLAYER_MODELTYPE_SHEATH_19, // empty sheath and shield on back?
+    // waist
+    /* 20 */ PLAYER_MODELTYPE_WAIST,
+    /* 21 */ PLAYER_MODELTYPE_MAX,
+    /* 0xFF */ PLAYER_MODELTYPE_RH_FF = 0xFF // disable shield collider, cutscene-specific
```
Roman971, line 239: I'd personally just go with hex for all these enums

src/overlays/actors/ovl_player_actor/z_player.c
Dragorn421: Yeah I agree. I guess I'll change it to a 1D array. I wasn't sure if this code is used at all or of any importance but still it should probably be a 1D array

---------

Unrelated to this PR but same issue, should the sSetupDL being 2D with this 1D cast also be reverted to 1D

https://github.com/zeldaret/oot/blob/054bac758306026b99358f97ac22062fbedff16b/src/code/z_rcp.c#L901

include/z64player.h
Dragorn421: I'll use hex from now on or change it without debating if commented on

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -622,142 +622,324 @@ static GetItemEntry sGetItemTable[] = {
     GET_ITEM_NONE,
 };
 
-static LinkAnimationHeader* D_80853914[PLAYER_ANIMGROUP_MAX][PLAYER_ANIMTYPE_MAX] = {
+#define PLAYER_ANIM_BY_GROUP_AND_TYPE(group, type) D_80853914[group * PLAYER_ANIMTYPE_MAX + type]
```
Roman971, line 625: This is a bit long for what it does imo, I think it could be shortened to something simpler like `GET_PLAYER_ANIM`. The macro itself contains names for the arguments already, and it would usually be called with enum values and/or variables named "group" or "animType", so I think it would be fine.

Also I believe this needs a formatter run (at least with the current name)

src/overlays/actors/ovl_player_actor/z_player.c
Roman971: Yeah you're right about z_rcp.c having the same issue. It was changed a while ago, but tbh I'm still not sure whether it's an improvement. I guess we should discuss it separately

src/overlays/actors/ovl_player_actor/z_player.c
Dragorn421: Surprisingly the PR is already correctly formatted (I already ran the formatter), so the name isn't that long I guess. it helped me going with this name even if it's quite long

I avoided giving it a generic name such as `get_player_anim` as it felt too generic and this isn't the only thing for handling link's animations, but maybe it's okay?

Roman971: I feel like it would be fine to name it with a more generic name now and rename it later if it ever starts conflicting or risks being confused with other macros. A better name will probably be easier to figure out at that point as well, since we'll know more about the other ways to define link's animations and how they differ from this one

fig02 (approved):

Roman971 (approved):

PR #1144, opened 2022-02-13
More documentation for IrqMgr

Thar0:
Also added a define for the ascii BEL character, I placed it in `vt.h` but maybe there's a more appropriate place?


src/code/irqmgr.c
```diff
@@ -1,122 +1,174 @@
...
 }
 
-void IrqMgr_JamMesgForClient(IrqMgr* this, OSMesg msg) {
-    IrqMgrClient* iter = this->clients;
-
-    while (iter != NULL) {
-        if (iter->queue->validCount >= iter->queue->msgCount) {
+/**
+ * Send `msg` to every registered client if the message queue is not full. This appears to be for
+ * high-priority messages that should be jammed to the front of the queue, however a bug prevents
```
EllipticEllipsis, line 123: Is this a technical sense of "jam"?

src/code/irqmgr.c
```diff
@@ -1,122 +1,174 @@
...
-    this->resetStatus = STATUS_PRENMI;
+    gIrqMgrResetStatus = preNmi;
+    this->resetStatus = IRQ_RESET_PRENMI;
     sIrqMgrResetTime = this->resetTime = osGetTime();
 
-    osSetTimer(&this->timer, OS_USEC_TO_CYCLES(450000), 0ull, &this->queue, (OSMesg)PRENMI450_MSG);
-    IrqMgr_JamMesgForClient(this, (OSMesg) & this->prenmiMsg);
+    // Schedule a PRENMI450 message to be handled in 450ms
+    osSetTimer(&this->timer, OS_MSEC_TO_CYCLES(450), 0, &this->queue, (OSMesg)PRENMI450_MSG);
+    IrqMgr_JamMesgToClients(this, (OSMesg) & this->prenmiMsg);
```
EllipticEllipsis, line 160:
```suggestion
    IrqMgr_JamMesgToClients(this, (OSMesg)&this->prenmiMsg);
```
This is surprisingly confusing-looking.

src/code/irqmgr.c
```diff
@@ -127,54 +179,64 @@ void IrqMgr_CheckStack() {
 }
 
 void IrqMgr_HandlePRENMI450(IrqMgr* this) {
-    u64 temp = STATUS_NMI; // required to match
-    gIrqMgrResetStatus = temp;
-    this->resetStatus = STATUS_NMI;
+    u64 nmi = IRQ_RESET_NMI; // required to match
+
+    gIrqMgrResetStatus = nmi;
+    this->resetStatus = IRQ_RESET_NMI;
```
EllipticEllipsis, line 185: This can't be a chained assignment, right?

src/code/irqmgr.c
```diff
@@ -127,54 +179,64 @@ void IrqMgr_CheckStack() {
...
 
-    osSetTimer(&this->timer, OS_USEC_TO_CYCLES(30000), 0ull, &this->queue, (OSMesg)PRENMI480_MSG);
-    IrqMgr_SendMesgForClient(this, (OSMesg) & this->nmiMsg);
+    // Schedule a PRENMI480 message to be handled in 30ms
+    osSetTimer(&this->timer, OS_MSEC_TO_CYCLES(30), 0, &this->queue, (OSMesg)PRENMI480_MSG);
+    // Send the NMI event to clients
+    IrqMgr_SendMesgToClients(this, (OSMesg) & this->nmiMsg);
```
EllipticEllipsis, line 190:
```suggestion
    IrqMgr_SendMesgToClients(this, (OSMesg)&this->nmiMsg);
```

src/code/irqmgr.c
```diff
@@ -127,54 +179,64 @@ void IrqMgr_CheckStack() {
...
 void IrqMgr_HandlePRENMI480(IrqMgr* this) {
     u32 ret;
 
-    osSetTimer(&this->timer, OS_USEC_TO_CYCLES(20000), 0ull, &this->queue, (OSMesg)PRENMI500_MSG);
+    // Schedule a PRENMI500 message to be handled in 20ms
+    osSetTimer(&this->timer, OS_MSEC_TO_CYCLES(20), 0, &this->queue, (OSMesg)PRENMI500_MSG);
+
     ret = osAfterPreNMI();
-    if (ret) {
+    if (ret != 0) {
```
EllipticEllipsis, line 200: `ret` doesn't look like a very good name for this since you'd normally expect it to be returned by this function. Would `preNMIComplete` be suitable?

src/code/irqmgr.c
```diff
@@ -127,54 +179,64 @@ void IrqMgr_CheckStack() {
...
     }
 }
 
 void IrqMgr_HandlePRENMI500(IrqMgr* this) {
-    IrqMgr_CheckStack();
+    IrqMgr_CheckStacks();
 }
 
+/**
+ * Ran on each VI retrace, measures the time elapsed between the first and second VI retrace
```
EllipticEllipsis, line 213:
```suggestion
 * Run on each VI retrace, measures the time elapsed between the first and second VI retrace
```

src/code/irqmgr.c
```diff
@@ -127,54 +179,64 @@ void IrqMgr_CheckStack() {
...
+    if (gIrqMgrRetraceTime == 0) {
         if (this->retraceTime == 0) {
             this->retraceTime = osGetTime();
         } else {
             gIrqMgrRetraceTime = osGetTime() - this->retraceTime;
         }
     }
     sIrqMgrRetraceCount++;
-    IrqMgr_SendMesgForClient(this, (OSMesg) & this->retraceMsg);
+    IrqMgr_SendMesgToClients(this, (OSMesg) & this->retraceMsg);
```
EllipticEllipsis, line 225:
```suggestion
    IrqMgr_SendMesgToClients(this, (OSMesg)&this->retraceMsg);
```

src/code/irqmgr.c
```diff
@@ -127,54 +179,64 @@ void IrqMgr_CheckStack() {
...
 }
 
-void IrqMgr_ThreadEntry(void* arg0) {
-    OSMesg msg;
-    IrqMgr* this = (IrqMgr*)arg0;
+void IrqMgr_ThreadEntry(void* arg) {
+    u32 msg = 0;
```
EllipticEllipsis, line 229: I suppose 0 has no special meaning apart from "no message"?

src/code/main.c
```diff
@@ -58,7 +58,7 @@ void Main(void* arg) {
     SystemHeap_Init(sysHeap, gSystemHeapSize); // initializes the system heap
     if (osMemSize >= 0x800000) {
         debugHeap = SysCfb_GetFbEnd();
-        debugHeapSize = (s32)(0x80600000 - debugHeap);
+        debugHeapSize = (s32)(K0BASE + 0x600000 - debugHeap);
```
EllipticEllipsis, line 61: This cast doesn't seem to be required (apart from this just being a generally weird calculation to do).

src/code/irqmgr.c
```diff
@@ -210,23 +272,25 @@ void IrqMgr_ThreadEntry(void* arg0) {
...
     this->nmiMsg.type = OS_SC_NMI_MSG;
-    this->resetStatus = STATUS_IDLE;
+    this->resetStatus = IRQ_RESET_IDLE;
     this->resetTime = 0;
 
     osCreateMesgQueue(&this->queue, this->msgBuf, ARRAY_COUNT(this->msgBuf));
     osSetEventMesg(OS_EVENT_PRENMI, &this->queue, (OSMesg)PRE_NMI_MSG);
     osViSetEvent(&this->queue, (OSMesg)RETRACE_MSG, retraceCount);
-    osCreateThread(&this->thread, 0x13, IrqMgr_ThreadEntry, this, stack, pri);
+    osCreateThread(&this->thread, 19, IrqMgr_ThreadEntry, this, stack, pri);
```
EllipticEllipsis, line 294: Will threads get defines for their ids, or is it not standard enough to be worth it?

src/code/main.c
```diff
@@ -69,33 +69,33 @@ void Main(void* arg) {
...
-    PadMgr_Init(&gPadMgr, &sSiIntMsgQ, &gIrqMgr, 7, Z_PRIORITY_PADMGR, &sIrqMgrStack);
+    PadMgr_Init(&gPadMgr, &sSiIntMsgQ, &gIrqMgr, 7, Z_PRIORITY_PADMGR, STACK_TOP(sPadMgrStack));
 
     AudioMgr_Unlock(&gAudioMgr);
 
     StackCheck_Init(&sGraphStackInfo, sGraphStack, STACK_TOP(sGraphStack), 0, 0x100, "graph");
     osCreateThread(&sGraphThread, 4, Graph_ThreadEntry, arg, STACK_TOP(sGraphStack), Z_PRIORITY_GRAPH);
     osStartThread(&sGraphThread);
-    osSetThreadPri(0, Z_PRIORITY_SCHED);
+    osSetThreadPri(NULL, 15);
```
EllipticEllipsis, line 98: So this is a coincidence rather than actually using the defines from `thread.h`?

EllipticEllipsis (commented):
Great job again on this, I once again have more questions than suggestions, I'm afraid.

include/ultra64/message.h
```diff
@@ -36,4 +36,11 @@ typedef struct OSMesgQueue {
     /* 0x14 */ OSMesg* msg;
 } OSMesgQueue; // size = 0x18
 
+/* Get count of valid messages in queue */
+#define MQ_GET_COUNT(mq)    ((mq)->validCount)
+
+/* Figure out if message queue is empty or full */
```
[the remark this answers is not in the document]
Thar0: Took the comments along with the macros from sdk headers 😂

include/irqmgr.h
```diff
@@ -3,6 +3,17 @@
 
 #include "ultra64.h"
 
+#define OS_SC_RETRACE_MSG       1
+#define OS_SC_DONE_MSG          2
+#define OS_SC_NMI_MSG           3 // name is made up, 3 is OS_SC_RDP_DONE_MSG in the original sched.c
```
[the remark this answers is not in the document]
Thar0: This refers to the libultra sched module, which was source-available so as I understand it devs would often make changes. In this case the deviation is appropriate since the event this message is associated with isn't related to the RDP

src/code/irqmgr.c
```diff
@@ -1,122 +1,174 @@
+/**
+ * @file irqmgr.c
+ *
+ * This file implements a manager for forwarding three key system interrupt events to
+ * registered clients. These interrupts are:
```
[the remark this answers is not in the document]
Thar0: IRQ usually means "interrupt request", although "interrupt queue" may make more sense.. issue is I'm not confident on choosing which I'd like to go with right now. I'll think about it some more.

src/code/irqmgr.c
```diff
@@ -1,122 +1,174 @@
...
+ *
+ *  - VI Retrace
+ *      This event is sent to the IRQ manager by the OS VI manager which only
+ *      supports the forwarding of VI events to a single message queue. The IRQ
+ *      manager will forward these events to every registered client. VI retrace
+ *      events are received when the Video Interface has reached vertical blank, happening
+ *      at approximately 60Hz on NTSC and 50Hz on PAL. Many threads sit idle until a
+ *      VI Retrace event wakes them up, at which point they will perform their task
+ *      and then return to idle to await the next retrace.
```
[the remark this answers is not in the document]
EllipticEllipsis: ... thanks for seemingly losing my comment, GitHub.

src/code/irqmgr.c
Thar0: In the libultra message queue sense, to "jam" a message means to add it to the front of the message queue rather than the back.

src/code/irqmgr.c
Thar0: Yeah.. the formatter has some weird ideas about these

src/code/irqmgr.c
Thar0: Tried it, doesn't seem to work

src/code/irqmgr.c
Thar0: Different threads may have the same thread id, since it's only relevant for debugging afaik. I think I will make thread id defines in the future anyway, but probably not in this PR

src/code/main.c
Thar0: Seems to be. This sets the priority of the current thread, main, so should be totally unrelated to the scheduler thread

src/code/irqmgr.c
EllipticEllipsis: I can sort of see its point, but still...

src/code/main.c
EllipticEllipsis: Right, so worth having another define, since the `Z_PRIORITY` stuff is all specific?

src/code/irqmgr.c
Thar0: That's correct, by the time `msg` is used it should be replaced by one of the internal message values

src/code/main.c
Thar0: Probably, although I'll probably have to review the priority stuff more generally I feel, similarly to thread ids

src/code/irqmgr.c
```diff
@@ -8,10 +8,10 @@
...
  *      the forwarding of VI events to a single message queue. The IRQ, manager will
  *      forward these events to every registered client. VI retrace, events are received
- *      when the Video Interface has reached vertical blank, happening, at approximately
- *      60Hz on NTSC and 50Hz on PAL. Many threads sit idle until a, VI Retrace event wakes
- *      them up, at which point they will perform their task, and then return to idle to
- *      await the next retrace.
+ *      when the Video Interface has reached the start of the vertical blanking interval,
+ *      happening, at approximately 60Hz on NTSC and 50Hz on PAL. Many threads sit idle
+ *      until a, VI Retrace event wakes them up, at which point they will perform their
+ *      task, and then return to idle to await the next retrace.
```
EllipticEllipsis, line 14: Think some extra commas have made their way in here.

include/ultra64/message.h
```diff
@@ -36,4 +36,11 @@ typedef struct OSMesgQueue {
     /* 0x14 */ OSMesg* msg;
 } OSMesgQueue; // size = 0x18
 
+/* Get count of valid messages in queue */
+#define MQ_GET_COUNT(mq)    ((mq)->validCount)
+
+/* Determine if message queue is empty or full */
+#define MQ_IS_EMPTY(mq)     (MQ_GET_COUNT(mq) == 0)
+#define MQ_IS_FULL(mq)      (MQ_GET_COUNT(mq) >= (mq)->msgCount)
+
```
engineer124, line 45: This might be fine as is due to context, but I worry if `MQ` might be conflated with "Master Quest" instead of "Message Queue". Is there any downside to having the prefix start with `MESQ_`?

Thar0: I don't think we should make any breaking changes to sdk headers. I'd imagine context is sufficient here

engineer124: If `MQ_` is an already preexisting standard already outside of decomp, then yeah I suppose it's fine

EllipticEllipsis: This did cross my mind, but I hope we can rely on people to tell the difference.

Dragorn421: I think context is enough, but I'm curious why we would want to keep the original headers intact?

Thar0: It seems simply better to be compatible than not, also means any sdk documentation stays immediately applicable.

EllipticEllipsis (approved):

src/code/irqmgr.c
```diff
@@ -1,122 +1,174 @@
...
-    this->clients = c;
+    client->queue = msgQ;
+    client->prev = this->clients;
+    this->clients = client;
 
     osSetIntMask(prevInt);
 
-    if (this->resetStatus > STATUS_IDLE) {
-        osSendMesg(c->queue, (OSMesg) & this->prenmiMsg, OS_MESG_NOBLOCK);
+    if (this->resetStatus > IRQ_RESET_IDLE) {
```
Dragorn421, line 65:
```suggestion
    if (this->resetStatus >= IRQ_RESET_PRENMI) {
```
for symmetry with the next if

src/code/irqmgr.c
```diff
@@ -1,122 +1,174 @@
...
+    IrqMgr_JamMesgToClients(this, (OSMesg) & this->prenmiMsg);
 }
 
-void IrqMgr_CheckStack() {
-    osSyncPrintf("irqmgr.c: PRENMIから0.5秒経過\n"); // "0.5 seconds after PRENMI"
-    if (StackCheck_Check(NULL) == 0) {
-        osSyncPrintf("スタックは大丈夫みたいです\n"); // "The stack looks ok"
+void IrqMgr_CheckStacks(void) {
+    // "0.5 seconds after PRENMI"
+    osSyncPrintf("irqmgr.c: PRENMIから0.5秒経過\n");
```
Dragorn421, line 165: based on this debug printf, this function is probably intended as `IrqMgr_HandlePRENMI500Impl` maybe, but it's not a big deal

What's the point of doing stack usage checks anyway? purely debugging/figuring out stack sizes?

src/code/irqmgr.c
```diff
@@ -127,54 +179,64 @@ void IrqMgr_CheckStack() {
...
-    ret = osAfterPreNMI();
-    if (ret) {
+    result = osAfterPreNMI();
+    if (result != 0) {
         // "osAfterPreNMI returned %d !?"
-        osSyncPrintf("osAfterPreNMIが %d を返しました！？\n", ret);
-        osSetTimer(&this->timer, OS_USEC_TO_CYCLES(1000), 0ull, &this->queue, (OSMesg)PRENMI480_MSG);
+        osSyncPrintf("osAfterPreNMIが %d を返しました！？\n", result);
+        // osAfterPreNMI failed, try again in 1ms
+        osSetTimer(&this->timer, OS_MSEC_TO_CYCLES(1), 0, &this->queue, (OSMesg)PRENMI480_MSG);
```
Dragorn421, line 204: this also means it would schedule more `PRENMI500_MSG` messages every millisecond if `osAfterPreNMI` keeps failing, I guess it's not an issue (?) but it was probably not intended

src/code/irqmgr.c
```diff
@@ -1,122 +1,174 @@
...
 volatile OSTime sIrqMgrResetTime = 0;
 volatile OSTime gIrqMgrRetraceTime = 0;
 u32 sIrqMgrRetraceCount = 0;
 
+// Internal messages
 #define RETRACE_MSG 666
 #define PRE_NMI_MSG 669
 #define PRENMI450_MSG 671
 #define PRENMI480_MSG 672
 #define PRENMI500_MSG 673
```
Dragorn421, line 39: To prevent confusion with `OS_SC_NMI_MSG` and the like, should these get some kind of prefix?

Also I know you want to avoid deviating from original headers but `OS_SC_` for the messages sent by the irq with the "scheduler" being something else, is confusing

Also the `OS_SC_` defines can be used in `Sched_ThreadEntry`

include/irqmgr.h
```diff
@@ -3,6 +3,17 @@
...
+#define OS_SC_PRE_NMI_MSG       4
+
+typedef enum {
+    IRQ_RESET_IDLE,
+    IRQ_RESET_PRENMI,
+    IRQ_RESET_NMI
+} IrqResetStatus;
```
Dragorn421, line 15: it's a bit more verbose but imo the prefix should be the full `IRQ_RESET_STATUS_` for clarity, and consistency with overall enum definitions

include/ultra64/message.h
Thar0: I've done this in libultra cleanup 3 ™ but figured I'd do it for this file while I was here. I may split it into it's own PR though.

src/code/irqmgr.c
Thar0: Stack checks are for debugging yeah, mostly checking whether a stack has overrun its maximum size

src/code/irqmgr.c
Thar0: Yeah the terminology is a bit messy right now, I'm going to document sched.c next, maybe the `OS_SC_` defines should be revisited in that when both systems are in view. I also plan to add the defines to sched.c in that, but I'll update the irqmgr-internal ones now.

src/code/irqmgr.c
```diff
@@ -1,122 +1,174 @@
...
+ *      task and then return to idle to await the next retrace.
+ *
+ *  - Pre-NMI
+ *      This event is sent to the IRQ manager by the OS Interrupt Handler when the reset
+ *      button on the N64 control deck is pressed. This event is forwarded to clients to
+ *      inform them that a reset will occur in 0.5s / 500ms and to begin any shutdown
+ *      procedures.
+ *
+ *  - NMI
+ *      This event is sent at 450ms into the Pre-NMI phase, informing clients that a
```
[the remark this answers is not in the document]
Dragorn421: Imo it should be made clearer here that the nmi event does not correspond to the "actual" hardware nmi.
It may also be nice to describe hardware nmi, like the other events have descriptions of retrace and prenmi, even if it doesn't correspond 1:1, to clear up confusion

The event may be more appropriately named as `NMI_SOON`, or something along those lines.
It sounds kind of redundant with PreNMI already being "nmi soon", but it may be a better tradeoff than confusion between NMI and PreNMI+450ms

src/code/irqmgr.c
```diff
@@ -127,76 +179,86 @@ void IrqMgr_CheckStack() {
...
-    ret = osAfterPreNMI();
-    if (ret) {
+    result = osAfterPreNMI();
+    if (result != 0) {
         // "osAfterPreNMI returned %d !?"
-        osSyncPrintf("osAfterPreNMIが %d を返しました！？\n", ret);
-        osSetTimer(&this->timer, OS_USEC_TO_CYCLES(1000), 0ull, &this->queue, (OSMesg)PRENMI480_MSG);
+        osSyncPrintf("osAfterPreNMIが %d を返しました！？\n", result);
+        // osAfterPreNMI failed, try again in 1ms
+        osSetTimer(&this->timer, OS_MSEC_TO_CYCLES(1), 0, &this->queue, (OSMesg)IRQ_PRENMI480_MSG);
```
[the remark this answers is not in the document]
Thar0: I think you're right, this looks like it breaks the linked list for any timers added after this one. It definitely invalidates the first `osSetTimer` but I believe the issue is indeed worse than that. This bug probably isn't visible in practice but I'll make a note of it.

src/code/main.c
```diff
@@ -58,7 +58,7 @@ void Main(void* arg) {
     SystemHeap_Init(sysHeap, gSystemHeapSize); // initializes the system heap
     if (osMemSize >= 0x800000) {
         debugHeap = SysCfb_GetFbEnd();
-        debugHeapSize = (s32)(0x80600000 - debugHeap);
+        debugHeapSize = K0BASE + 0x600000 - debugHeap;
```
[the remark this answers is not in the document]
Thar0: I've gone with `PHYS_TO_K0(0x600000)`
The second doesn't make sense since `debugHeap` is already in KSEG0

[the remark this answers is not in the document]
Dragorn421: right, ok better example
`K0BASE + 0x800000 - 0x100`
vs `PHYS_TO_K0(0x800000) - 0x100`
vs `PHYS_TO_K0(0x800000 - 0x100)`
(from `Fault_DrawMemDump`)
I was looking on documenting how memory is handled/allocated/partitioned and stumbled on this kind of thing

Dragorn421 (approved):

src/code/main.c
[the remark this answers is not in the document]
Thar0: Second or third are just as valid imo, since you can subtract constant offsets from either phys or kseg0

src/code/irqmgr.c
mzxrules: From my documentation, only the os internal threads (vi/pi) use the same thread id.

src/code/irqmgr.c
```diff
@@ -1,122 +1,179 @@
...
+    client->queue = msgQ;
+    client->prev = irqMgr->clients;
+    irqMgr->clients = client;
 
     osSetIntMask(prevInt);
 
-    if (this->resetStatus > STATUS_IDLE) {
-        osSendMesg(c->queue, (OSMesg) & this->prenmiMsg, OS_MESG_NOBLOCK);
+    if (irqMgr->resetStatus >= IRQ_RESET_STATUS_PRENMI) {
+        osSendMesg(client->queue, (OSMesg)&irqMgr->prenmiMsg, OS_MESG_NOBLOCK);
```
Roman971, line 71: Does it make sense to cast pointers to `OSMesg` since that's just a typedef for `void*`? I'm psure it wouldn't cause any warnings at least, but are you supposed to cast to OSMesg regardless of the type of the message or something like that?

Also noticed we have a lot of incorrect casts for osRecvMesg calls which cast to `OSMesg` even tho it takes a `OSMesg*`, but that's another issue

Thar0: You're correct it doesn't cause any warnings if the cast were omitted. It's mostly for clarity I believe? There's no necessary convention that you should cast to OSMesg afaik. If you think it's better off implicit then I can change that.

Roman971: I'm not sure tbh, and thinking about it more it's really a general thing so we don't have to change it with this PR

Roman971 (approved):
LGTM. I had a comment about OSMesg casts but we can probably leave that for a separate discussion/PR depending on what we want to do

src/code/irqmgr.c
Thar0: A separate PR sounds good, I have some things to clean up with message queues and threads in general in a separate PR already so we can probably deal with this more generally there.

PR #1150, opened 2022-02-19
Fix misc 8

Dragorn421:


include/z64.h
```diff
@@ -253,8 +253,8 @@ typedef struct {
     struct {
         /* 0x0104 */ u32    swch;
         /* 0x0108 */ u32    tempSwch;
-        /* 0x010C */ u32    unk0;
-        /* 0x0110 */ u32    unk1;
+        /* 0x010C */ u32    unk_0;
+        /* 0x0110 */ u32    unk_1;
```
Roman971, line 257: These aren't named based on the offset and literally mean "unknown flags 1" and "unknown flags 2" since we don't know what they were intended for, so it might be confusing to have them as `unk_X`.
That being said, they could probably be named `unk` and `tempUnk` since they follow the same pattern as other flags with a temp field (see `Flags_GetUnknown` and such)

include/z64actor.h
```diff
@@ -96,7 +96,7 @@ typedef struct {
     /* 0x0C */ ActorShadowFunc shadowDraw; // Shadow draw function
     /* 0x10 */ f32 shadowScale; // Changes the size of the shadow
     /* 0x14 */ u8 shadowAlpha; // Default is 255
-    /* 0x15 */ u8 feetFloorFlags; // Set if the actor's foot is clipped under the floor. & 1 is right foot, & 2 is left
+    /* 0x15 */ u8 feetFloorFlag; // 0 if actor or feet aren't on ground, or 1 or 2 depending on feet positions
```
Roman971, line 99: Since this technically stores 2 flags (& 1 and & 2), I feel like having it plural makes more sense

src/code/z_kankyo.c
```diff
@@ -35,8 +35,8 @@ typedef struct {
 } LightningBolt; // size = 0x20
 
 typedef struct {
-    /* 0x00 */ s32 unk0;
-    /* 0x04 */ s32 unk1;
+    /* 0x00 */ s32 unk_0;
+    /* 0x04 */ s32 unk_1;
```
Roman971, line 39:
```suggestion
    /* 0x00 */ s32 unk_00;
    /* 0x04 */ s32 unk_04;
```

src/code/z_scene.c
```diff
@@ -166,11 +166,11 @@ s32 Scene_ExecuteCommands(GlobalContext* globalCtx, SceneCmd* sceneCmd) {
         osSyncPrintf("*** Scene_Word = { code=%d, data1=%02x, data2=%04x } ***\n", cmdCode, sceneCmd->base.data1,
                      sceneCmd->base.data2);
 
-        if (cmdCode == 0x14) {
+        if (cmdCode == SCENE_CMD_ID_END) {
             break;
         }
 
-        if (cmdCode <= 0x19) {
+        if (cmdCode < ARRAY_COUNT(gSceneCmdHandlers)) {
```
Roman971, line 173: The size of the extern for `gSceneCmdHandlers` should also be updated to have `SCENE_CMD_ID_MAX` if we want it to shift properly. Or alternatively we could just use `SCENE_CMD_ID_MAX` here and leave the size unset in the extern(s)

include/z64animation.h
```diff
@@ -241,21 +241,23 @@ typedef s32 (*AnimUpdateFunc)();
...
+    /* 0x14 */ f32 animLength;    // Total number of frames in the current animation.
     /* 0x18 */ f32 curFrame;      // Current frame in the animation
     /* 0x1C */ f32 playSpeed;     // Multiplied by R_UPDATE_RATE / 3 to get the animation's frame rate.
     /* 0x20 */ Vec3s* jointTable; // Current translation of model and rotations of all limbs
     /* 0x24 */ Vec3s* morphTable; // Table of values used to morph between animations
     /* 0x28 */ f32 morphWeight;   // Weight of the current animation morph as a fraction in [0,1]
     /* 0x2C */ f32 morphRate;     // Reciprocal of the number of frames in the morph
-    /* 0x30 */ s32 (*update)();   // Can be Loop, Partial loop, Play once, Morph, or Tapered morph. Link only has Loop, Play once, and Morph
+    /* 0x30 */ s32 (*update)();   // Can be Loop, Partial loop, Play once, Morph, or Tapered morph.
+                                  // Link only has Loop, Play once, and Morph
```
fig02, line 260: i would personally have the comment line be longer than to have a line break in the struct. we dont run the formatter o headers so the line limit isnt a concern
just a personal opinion though, open to discussion from others

include/z64actor.h
fig02, line 99: since there are individual bits that have different meaning, flags seems more correct than flag to me no?

fig02: (whoops roman and i sent out review at the same time lol)

Dragorn421: It's actually just a value, it's always 0, 1 or 2 (except temporarily where the feet positions are computed), never 1 | 2

include/z64animation.h
Dragorn421: we need some agreed-on way to have longer comments on struct members

the issue with this style here is only the first line seems to appear in documentation when the mouse hovers on the struct member

is it really fine to have a very very long line? like 200 characters may be "ok", but 300 probably not

some options:

```c
    /* 0x30 */ s32 (*update)(); // Can be Loop, Partial loop, Play once, Morph, or Tapered morph. Link only has Loop, Play once, and Morph

    /* Can be Loop, Partial loop, Play once, Morph, or Tapered morph.
     * Link only has Loop, Play once, and Morph */
    /* 0x30 */ s32 (*update)();

    /*
     * Can be Loop, Partial loop, Play once, Morph, or Tapered morph.
     * Link only has Loop, Play once, and Morph
     */
    /* 0x30 */ s32 (*update)();
```

include/z64.h
Dragorn421: Neither are saved in the save context so I don't think they should go by unk/tempUnk even if the pattern does fit
I'll revert to unk0 and unk1
and also remove "current scene" from the `Flags_*Unknown` functions since the flags arent saved nor loaded so aren't per-scene

include/z64actor.h
Dragorn421: Also I can't make sense of the conditional that determines if the value should be 1 or 2 in the case where both feet touch the ground, and since the only use in the game is a single != 0 check in player it's hard to tell what they were going for

Roman971: I see the game expects it to be 3 at some point, so both flags are definitely intended to be set at the same time (at least temporarily) and it would seem wrong to treat it as a single value, but I'm not too familiar with this stuff so idk

Dragorn421: I'm not seeing this value as a bitfield but as

- 0 if actor or feet aren't on ground
- 1 if left feet is "more something" than right feet
- 2 if right feet is "more something" than left feet

the "more something" being unclear, especially due to the case where  `feetFloorFlag = 3` temporarily. I don't understand the conditional that tells apart 1 and 2:
https://github.com/Dragorn421/oot/blob/2e45e9ab8249df74cbe5a24287f4597fa81bad39/src/code/z_actor.c#L195-L203
```c
        } else if (actor->shape.feetFloorFlag == 3) {
            f32 footDistY = actor->shape.feetPos[FOOT_LEFT].y - actor->shape.feetPos[FOOT_RIGHT].y;

            if ((floorHeight[FOOT_LEFT] + footDistY) < (floorHeight[FOOT_RIGHT] - footDistY)) {
                actor->shape.feetFloorFlag = 2;
            } else {
                actor->shape.feetFloorFlag = 1;
            }
        }
```

mzxrules: Took a look at it.

Prior to that code a `for` loop that initializes actor->shape.feetFloorFlag. It sets the value to 0 if `(-1 <= distToFloor <= 0)` is false for both feet, 2 if only true for the left foot, 1 if only true for the right foot, and 3 if true for both feet. Thus, 3 means that both feet are planted on the ground.

Not fully sure what the `actor->shape.feetFloorFlag == 3` case is attempting to accomplish, but it seems to do this:
`footDistY` stores the diff between the left and right foot's shape.y position. if the left foot shape is lower than the right foot, then the condition effectively becomes

```python
# Left foot lower than right
if (left floor - abs(footDiff)) < (right floor + abs(footDiff))
    feetFloorFlag = LEFT_FOOT // 2
else
    feetFloorFlag = RIGHT_FOOT // 1
````

else if the right foot shape is lower than the left foot, the condition becomes

```python
# right foot lower than left
if (left floor + abs(footDiff)) < (right floor - abs(footDiff))
    feetFloorFlag = LEFT_FOOT // 2
else
    feetFloorFlag = RIGHT_FOOT // 1
````

fig02 (approved):
think the only thing holding us up here is feet floor flag stuff. maybe we can name it something that doesnt have the word flag in it, if it really isnt a bit packed field of flags?

either way its something that can be cleaned up later

Dragorn421:
I reverted the `feetFloorFlags` changes in this PR and opened #1180 separately for it

Roman971 (approved):

PR #1151, opened 2022-02-20
Teach msgenc to avoid comments

kemenaran:
Follow-up to #1145 (cc @Thar0) :

Since #1145 the C preprocessor is used to remove comments from the text files before encoding them.

However the option used (`-fpreprocessed`) is only supported by gcc – but not by clang, which breaks the macOS build.

By using Python to remove comments, we can remove the use of this gcc-specific option, and fix the macOS build.

## Implementation

The regex used to remove comments is the one used by mips_2_c.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -297,8 +297,7 @@ build/data/%.o: data/%.s
 	$(AS) $(ASFLAGS) $< -o $@
 
 build/assets/text/%.enc.h: assets/text/%.h assets/text/charmap.txt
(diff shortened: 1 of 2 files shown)


Dragorn421 (approved):

fig02 (approved):
if this solution works for tharo then it works for me as well

EllipticEllipsis (approved):
OK for me.

Thar0 (approved):

Roman971 (approved):

kemenaran:
Thanks to you all ❤️

PR #1153, opened 2022-02-22
Document Kak

louist103:
Named almost everything I could with where ZAPD is right now.


assets/xml/scenes/overworld/spot01.xml
```diff
@@ -1,12 +1,169 @@
...
+
+        <Array Name="gKakarikoVillageGrannysPotionSHopVtx" Count="43" Offset="0x6868">
+            <Vtx/>
+        </Array>
+
+        <Array Name="gKakarikoVillageHouseVtx" Count="25" Offset="0x65C8">
+            <Vtx/>
+        </Array>
+
+        <Array Name="gKakarikoVillageImpasHousePlatformAndboxesVtx" Count="60" Offset="0x5FE0">
```
Dragorn421, line 123:
```suggestion
        <Array Name="gKakarikoVillageImpasHousePlatformAndBoxesVtx" Count="60" Offset="0x5FE0">
```

Dragorn421 (commented):
What does "nw area" refer to?

louist103:
> What does "nw area" refer to?

Its from this DL
 
        <DList Name="gKakarikoVillageNorthWestAreaDL" Offset="0x48C8"/>

assets/xml/scenes/overworld/spot01.xml
```diff
@@ -1,12 +1,169 @@
...
+        <Texture Name="gKakarikoVillagePotionShopHouseWallTex" OutName="potion_shop_wall" Format="ci8" Width="32" Height="64" Offset="0x17B50" TlutOffset="0xA870"/>
+        <Scene Name="spot01_scene" Offset="0x0"/>
+    </File>
+    <File Name="spot01_room_0" Segment="3">
+        <!-- ZAPD doesn't support this yet
+        <ActorList Name="gKakarikoVillageChildDayActorList" Count="57" Offset="0x84"/>
+        <ActorList Name="gKakarikoVillageAdultDayActorList" Count="51" Offset="0x5CC"/>
+        <ActorList Name="gKakarikoVillageAdultNightActorList" Count="46" Offset="0x95C"/>
+        <ActorList Name="gKakarikoVillageChildtNightActorList" Count="50" Offset="0xC98"/>
+        -->
```
mzxrules, line 57: My understanding is that this data won't be extracted by ZAPD

louist103: There is a PR for them

fig02: his point is that the files will eventually be committed with only the assets being extracted. actor lists will just be in the file

louist103: Well until it does shouldn't they be extracted so they are at least named in the mean time.

Roman971: Imo we shouldn't start add those to the XML if they aren't going to be used in the future

For naming we can probably just do that whenever we commit them, since the names are pretty straightforward and will already be somewhat generated by ZAPD (or could even be automated to include child/adult/day/night based on the setup index)

louist103: was committing these ever discussed? I remember it came up once and we never actually went anywhere with it.

fig02: we are going to commit anything that isnt a controversial asset, ie vertices, textures, and possibly dlists. anything else is fair game (comparable to any other data you would see included in the repo currently)

louist103: Well even so until that happens why remove useful documentation?

fig02: its setting a precedent for something that we wont actually start doing. we are not going to have xml nodes for these,
we can very easily name these when we are at a stage that these can be committed

louist103: So then as a compromise can we at least leave the offsets in the XML? Its much easier than the trial and error method to determine which list is which.

fig02: the order is always consistent with age + time of day setups. you can easily search the alt headers list to see which one you need for any scene

louist103: Yes but someone who doesn't know will still have trouble with that.  Its nice to just be able to search for the offset from the XML.

fig02: I see where youre coming from, but its still doing something in one place that is going to be removed sooner than later regardless.
Its something that can be figured out easily without trial and error, and will be rectified soon when its committed and named. The documentation is good I agree, but the inconsistency and adding a need to clean things up later just isnt worth it in my opinion

louist103: My worry is that the same thing happens here that happens with player.  Docs like this are removed from PRs with the "larger doc pass" argument, then nothing happens for 3 months.

fig02: dunno what else to say other than the information will be added, and multiple people have expressed that they would rather the comments removed

louist103: I'll just ask straight out.  Will the PR be accepted with this information in the XMLs?

fig02: the comments here are the only thing holding up the pr

louist103: Because using this logic I should remove the path data no?

fig02: Explicitly defining path data helps zapd have more accurate extraction (which actor lists dont need). So no its a different story entirely 

there was already a pr merged in to address this on a wider scale and added all of the missing paths, including the one in this file https://github.com/zeldaret/oot/pull/1181/

fig02: By more accurate extraction, i mean that paths are floating data that arent 100% inferred by zapd via reference. The explicit xml definition is necessary

Dragorn421: To add onto what fig is saying, the actorlist resource is being added to zapd specifically to mark parts of actor lists that are unused, but still clearly at some point (part of) an actor list (this happens only in one debug scene iirc, see the relevant pr/issue in the zapd repo)
So in that sense it would be similar in use to the path resource that zapd doesn't know what amount of paths there is
It of course doesn't mean the actorlist resource is exclusively for unused stuff as far as zapd is concerned, but the motive for getting it in was unused stuff.

louist103: You are missing my point. My point was say there wasn't the issue with paths, or maybe use another example `SkeletonHeaders`. Say we determine those aren't "assets". Should we go removing them from all the XMLs? Or should we stop naming them in the future since "they will just be committed anyway"? Between now and when we actually commit these files having documentation is always a good thing and shouldn't be ignored because "we will do it soon".

Dragorn421: I indeed am missing your point. `SkeletonHeader`s can't be automatically extracted by zapd, zapd needs to know where they are
actor lists are references by room headers which zapd parses on its own

fig02: Yea ActorLists are really the odd one out here, which is why we have been saying this from the beginning. 
I understand where youre coming from like Ive been saying, but from the perspective of maintaining this repo and how we will do things later, it just makes more sense to remove the comments and add the information at a later time.

This is the only thing holding up this pr, and multiple people have expressed that this is not needed.

louist103:
Unable to reach agreement regarding extra documentation of `ActorLists`

fig02:
Dont think its worth closing the whole PR over one part of it that others didnt want, but if its an all or nothing situation for you then so be it I guess

PR #1157, opened 2022-02-26
Angle cleanup - `DEG_TO_RAD`

Dragorn421:


fig02 (approved):

include/z64math.h
```diff
@@ -94,15 +94,14 @@ typedef struct {
...
-#define RADF_TO_DEGF(radf) (radf * (180.0f / M_PI))
-#define DEGF_TO_RADF(degf) (degf * (M_PI / 180.0f))
-#define BINANG_ROT180(angle) ((s16)(angle - 0x7FFF))
-#define BINANG_SUB(a, b) ((s16)(a - b))
+#define DEGF_TO_BINANG(degreesf) (s16)((degreesf) * 182.04167f + .5f)
+#define RADF_TO_BINANG(radf) (s16)((radf) * (0x8000 / M_PI))
+#define RADF_TO_DEGF(radf) ((radf) * (180.0f / M_PI))
+#define BINANG_ROT180(angle) ((s16)((angle) - 0x7FFF))
+#define BINANG_SUB(a, b) ((s16)((a) - (b)))
 #define DEG_TO_RAD(degrees) ((degrees) * (M_PI / 180.0f))
```
Roman971, line 102: Shouldn't it be `DEG_TO_RADF`? Or am I just confused about the reasoning behind the F in these macros?

And I guess this question also applies to `BINANG_TO_RAD`

Dragorn421: That's a good question. I have no idea what the idea with the "F" is. Probably "float" but I feel like that's redundant with degrees/radians *always* being floats.

As you can see it's currently not consistent either to use that F or not. and `binang` does not use "S" either
In other spots `F`/`S` is used to distinguish radians/binangle (e.g. Math_CosF/CosS)

So I would move towards renaming the macros to not use the "F"

fig02: i would say remove as well. I guess it was intended to help people remember to put `.0f` on the numbers? but yea its implied imo

qwoodmansee:
I think I have a unique perspective as someone who recently pulled this codebase and started reading through it for the first time:

I'm curious on your thoughts on moving away from (even easy) unnecessary abbreviations? Some of the code is extremely readable, i.e `Matrix_Translate` and `Matrix_Rotate_ZYX` functions, while others are less - i.e. `gDPSetPrimColor`. I assume, to you guys, `gDP` is something you immediately know what's being abbreviated - but to me I have no idea. 

In this PR, this would result in using `DEGREE`, `RADIAN`, and `BINARY` as opposed to the abbreviated values. It's a small change, but I could see it long term resulting in a more approachable codebase as per the documentation goals.

fig02:
your specific example of `gDPSetPrimColor` is a macro named by nintendo/sgi and is standard for the n64 in general. we will unfortunately not be changing that.

For things we do have control over, we are always aiming to balance clarity while also keeping macro length in mind. Things would get pretty out of hand if nothing at all was abbreviated and things get super long

include/z64math.h
Roman971: I'm guessing BINANG is implied to be shorts, which is why the "S" is not included for those, and yeah I guess the F could be omitted from the RAD/DEG stuff as well if we follow the same logic

fig02 (approved):

Roman971 (approved):
Looks good to me aside from the "F" inconsistency discussed in my review comment, but I guess we could probably have a separate PR for that

PR #1158, opened 2022-02-26
Angle cleanup - `BINANG_TO_RAD`, `BINANG_TO_RAD_ALT`

Dragorn421:
- Rename `BINANG_TO_RAD` to `BINANG_TO_RAD_ALT`
- Add `BINANG_TO_RAD`

------

Before:
```c
#define BINANG_TO_RAD(binang) (((f32)(binang) / (f32)0x8000) * M_PI)
```

After:
```c
#define BINANG_TO_RAD(binang) ((f32)(binang) * (M_PI / 0x8000))
#define BINANG_TO_RAD_ALT(binang) (((f32)(binang) / (f32)0x8000) * M_PI)
```

------------

(old `BINANG_TO_RAD`) `BINANG_TO_RAD_ALT` does a divide and a multiply,
new `BINANG_TO_RAD` only does one multiply.

old `BINANG_TO_RAD` / new `BINANG_TO_RAD_ALT`: 144 matches in 42 files (count from #1156)
new `BINANG_TO_RAD`: 129 matches in 56 files


fig02:
if the alt gets used more, i think its safe to name that as the main one now. 
I dont agree with the above that it should depend on which is faster, but it being used more is enough for me personally

Dragorn421:
Is there any opposition to renaming the current `BINANG_TO_RAD` to `BINANG_TO_RAD_ALT`
and replacing `BINANG_TO_RAD` with the new macro `((f32)(binang) * (M_PI / 0x8000))` ?

The idea is the "new" `BINANG_TO_RAD` would be the more efficient variant over the _alt variant

See main PR comment at the top for more info

fig02 (approved):

qwoodmansee:
What mathematical difference does the division create (if any)? If there is a difference, could we somehow integrate that into the naming of the macro?

EllipticEllipsis:
Because floats are not a complete description of real numbers, float operations often involve implicit truncation, and that means that float multiplication is not associative.

Dragorn421:
Mathematically, it's of course the same. But computers and their finite repreresentations of real numbers are bad at pure math so there may be some imperceptible difference

The difference can be seen clearly when compiling the usage of each macro:

```c
f32 BinangToRad(s16 binang) {
    return BINANG_TO_RAD(binang);
}
```
compiles to
```asm
sw      a0,0(sp)      
sll     a0,a0,0x10    
sra     a0,a0,0x10    
mtc1    a0,$f4        
lui     at,0x8083     
lwc1    $f8,-0x8c0(at)
cvt.s.w $f6,$f4       
mul.s   $f0,$f6,$f8   
jr      ra            
nop                   
```
----------
```c
f32 BinangToRad(s16 binang) {
    return BINANG_TO_RAD_ALT(binang);
}
```
compiles to
```asm
sw      a0,0(sp)       
sll     a0,a0,0x10     
sra     a0,a0,0x10     
mtc1    a0,$f4         
lui     at,0x4700      
mtc1    at,$f8         
cvt.s.w $f6,$f4        
lui     at,0x8083      
lwc1    $f16,-0x8b0(at)
div.s   $f10,$f6,$f8   
mul.s   $f0,$f10,$f16  
jr      ra             
nop                    
```

-------

The reason for prefering `BINANG_TO_RAD` is it generates shorter (so usually faster) code

EllipticEllipsis:
Well, in this case it's clearly faster because the longer one also has an extra division in it, famously one of the slowest operations.

Roman971 (approved):

PR #1159, opened 2022-02-26
Angle cleanup - miscellaneous

Dragorn421:
Adds `BINANG_TO_RAD_ALT2` but idk if it's worth having since it only gets 9 uses

---------

If someone has an idea if this can use a macro let me know:
https://github.com/zeldaret/oot/blob/2306e094b28620f0e323bf4af61f8e00450e38fa/src/code/z_scene_table.c#L2316-L2318

---------

I think there are rodata memes (?) in z_boss_tw.c (twinrova) that are preventing using some angle macros, if someone wants to try understand it be my guest

---------

Some cleanup around angle literals, like
- `DEG_TO_RAD(100)` instead of `5.0 * M_PI / 9.0` (and type degree amount as int, it doesn't matter idk)
- `M_PI / 4` instead of `DEG_TO_RAD(45)`, just use radians directly for nice radians values, typically pi, pi/2, pi/4 (as it currently mostly is in the repo)

### DIFF
--- a/include/z64math.h
+++ b/include/z64math.h
@@ -103,6 +103,7 @@ typedef struct {
 #define BINANG_TO_DEGF(binang) ((f32)(binang) * (360.0001525f / 65535.0f))
 #define BINANG_TO_RAD(binang) ((f32)(binang) * (M_PI / 0x8000))
 #define BINANG_TO_RAD_ALT(binang) (((f32)(binang) / (f32)0x8000) * M_PI)
+#define BINANG_TO_RAD_ALT2(binang) (((f32)(binang) * M_PI) / 0x8000)
 
 // Vector macros
 #define SQXZ(vec) ((vec.x) * (vec.x) + (vec.z) * (vec.z))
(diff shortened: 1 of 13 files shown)


fig02 (approved):
feel like its probably not worth for so little uses, but could also probably still be useful for people who wouldnt recognize that pattern as being a binary angle to radians conversion.. so im probably ok with it

Dragorn421:
If you want, but I'd rather not drown the few manual changes in an ocean of replace-all changes, which may amount to a lot

Also `DEGF_TO_BINANG`/`BINANG_TO_DEGF` should probably be renamed to `_alt` anyway since they are mostly unused (only in camera, mostly, like other angle macros), and also they are slightly off, and proper macros may be added even if oot doesn't use them.
There's just a bunch more to tackle about these macros

fig02:
alright, yea i didnt realize there was more to do with these. thought this was the last round of changes lol

Roman971 (approved):
I don't really have an opinion on `BINANG_TO_RAD_ALT2`, but it's probably fine

PR #1160, opened 2022-02-26
Doc misc 1

Dragorn421:
Variables: gotta name 'em all

### DIFF
--- a/src/code/z_message_PAL.c
+++ b/src/code/z_message_PAL.c
@@ -1169,8 +1169,8 @@ void Message_LoadItemIcon(GlobalContext* globalCtx, u16 itemId, s16 y) {
 }
 
 void Message_Decode(GlobalContext* globalCtx) {
-    u8 curByte;
-    u8 curByte2;
+    u8 curChar;
+    u8 curChar2;
     u16 value;
     s32 loadChar;
     s32 charTexIdx = 0;
@@ -1188,10 +1188,10 @@ void Message_Decode(GlobalContext* globalCtx) {
     sTextFade = false;
 
     while (true) {
-        curByte2 = curByte = msgCtx->msgBufDecoded[decodedBufPos] = font->msgBuf[msgCtx->msgBufPos];
+        curChar2 = curChar = msgCtx->msgBufDecoded[decodedBufPos] = font->msgBuf[msgCtx->msgBufPos];
 
-        if (curByte == MESSAGE_BOX_BREAK || curByte == MESSAGE_TEXTID || curByte == MESSAGE_BOX_BREAK_DELAYED ||
-            curByte == MESSAGE_EVENT || curByte == MESSAGE_END) {
+        if (curChar == MESSAGE_BOX_BREAK || curChar == MESSAGE_TEXTID || curChar == MESSAGE_BOX_BREAK_DELAYED ||
+            curChar == MESSAGE_EVENT || curChar == MESSAGE_END) {
             // Textbox decoding ends with any of the above text control characters
             msgCtx->msgMode = MSGMODE_TEXT_DISPLAYING;
             msgCtx->textDrawPos = 1;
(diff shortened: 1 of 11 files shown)


src/overlays/actors/ovl_Bg_Dy_Yoseizo/z_bg_dy_yoseizo.c
```diff
@@ -1014,11 +1014,11 @@ void BgDyYoseizo_ParticleDraw(BgDyYoseizo* this, GlobalContext* globalCtx) {
...
     for (i = 0; i < 200; i++, particle++) {
         if (particle->alive == 1) {
-            if (phi_s3 == 0) {
-                gSPDisplayList(POLY_XLU_DISP++, SEGMENTED_TO_VIRTUAL(gGreatFairyParticleAppearDL));
+            if (!isMaterialSet) {
+                gSPDisplayList(POLY_XLU_DISP++, SEGMENTED_TO_VIRTUAL(gGreatFairyParticleMaterialDL));
                 gDPPipeSync(POLY_XLU_DISP++);
 
-                phi_s3++;
+                isMaterialSet++;
```
EllipticEllipsis, line 1021: Incrementing a bool?

Dragorn421: as long as it's not false

this pattern is common to near all draw effect functions, `BossFd_DrawEffects` also has it as an incremented boolean, many are just unnamed or half-used as bools somehow

EllipticEllipsis: Arguable that should disqualify them from being counted as bools, then... I'm sure C99 would complain.

src/code/z_message_PAL.c
```diff
@@ -1169,9 +1169,9 @@ void Message_LoadItemIcon(GlobalContext* globalCtx, u16 itemId, s16 y) {
 }
 
 void Message_Decode(GlobalContext* globalCtx) {
-    u8 temp_s2;
-    u8 phi_s1;
-    u16 phi_s0_3;
+    u8 curByte;
+    u8 curByte2;
```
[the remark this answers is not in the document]
Dragorn421: My initial idea was indeed ctrlChar but it doesn't line up at all with the weird usage like reading the message id high byte or the save file name.

curChar/curChar2 would be better?

(btw I put the 2 on the one variable that isn't used in the highest level ifs but it's arbitrary)

src/overlays/actors/ovl_Bg_Dy_Yoseizo/z_bg_dy_yoseizo.c
Dragorn421:
```suggestion
                isMaterialSet += true - false;
```
:uhhuh:

No but, I'd be ok with using `isMaterialSet = 0`, `isMaterialSet == 0` if it's that bad for C to increment "false", but the `isXXX` bool name should stay or what to name it? `isMaterialSetFlag`? `materialFlag`? `materialSet`?

fig02: i mean this C standard doesnt have bools, so trying to force ideals from the newer standard onto something that wasnt written for it seems weird to me. But in the end it doesnt matter to me too much. I would prefer it be treated as a bool all around cause it clearly is, regardless of how the value gets set

src/code/z_message_PAL.c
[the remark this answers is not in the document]
Thar0: Yeah when I did this file it was hard to think of particularly good names due to the reuse for unrelated data as is somewhat common in interface code. I think `curChar` is an improvement at least yeah.

src/overlays/actors/ovl_Bg_Dy_Yoseizo/z_bg_dy_yoseizo.c
```diff
@@ -1014,11 +1014,11 @@ void BgDyYoseizo_ParticleDraw(BgDyYoseizo* this, GlobalContext* globalCtx) {
 
     for (i = 0; i < 200; i++, particle++) {
         if (particle->alive == 1) {
```
mzxrules, line 1016:
```suggestion
        if (particle->alive == true) {
```
This could probably just be true/false, except there's one spot where you have `particle->alpha = particle->alive = 0;`

src/overlays/actors/ovl_Bg_Dy_Yoseizo/z_bg_dy_yoseizo.c
Dragorn421: I reverted this change to master, as I figured all these actor-effects systems should probably be named/documented at once.

src/code/z_room.c
```diff
@@ -119,44 +119,44 @@ void func_80095D04(GlobalContext* globalCtx, Room* room, u32 flags) {
...
+        SkinMatrix_Vec3fMtxFMultXYZW(&globalCtx->viewProjectionMtxF, &pos, &projectedPos, &projectedW);
+        if (-(f32)polygonDlist->unk_06 < projectedPos.z) {
+            temp_f2 = projectedPos.z - polygonDlist->unk_06;
             if (temp_f2 < globalCtx->lightCtx.fogFar) {
-                phi_v0 = spB4;
                 spA4->unk_00 = polygonDlist;
                 spA4->unk_04 = temp_f2;
-                if (phi_v0 == NULL) {
+                iter = spB4;
+                if (spB4 == NULL) {
```
Roman971, line 135: Pretty minor but I feel like this check is intended to be about `iter` since it's followed by a do while loop on iter != NULL. Is there a particular reason to change it?

Dragorn421: That's my bad, I didn't realize the do while did expect `iter` initialized. I thought the assignment was a fake match basically. I'll revert the conditional

Roman971 (approved):

Thar0 (approved):

fig02 (approved):

PR #1162, opened 2022-02-27
Name fmodf

Thar0:


data/unk_8012ABC0.data.s
```diff
@@ -11,7 +11,7 @@
 
 # Unused
 glabel D_8012ABC0
```
EllipticEllipsis, line 13: I'd love to know what this array was meant to be.

EllipticEllipsis (approved):

data/unk_8012ABC0.data.s
Dragorn421: there's also `data/unk_80009410.data.s` (in boot) and yeah I'd love to know too

Dragorn421 (approved):

data/unk_8012ABC0.data.s
AngheloAlf: If I had to guess, I would say it isn't an array, probably an struct. But who knows? :s

AngheloAlf (approved):

fig02 (approved):

Roman971 (approved):

PR #1164, opened 2022-02-28
Memstuff

EllipticEllipsis:
Name all the memory/string.h functions I could find, document them as consistently as possible. OoT has multiple memsets and memcpys, some of which are rather non-conformant, so I added fairly complete docs for them. It's not clear how to name the ones in `code_80069420`

I did not document the libultra `string.h` since it will likely be reworked in a significant way in future anyway.

Also added some more `sizeof`s to places that were missing them. I didn't touch `z_sram` since the whole thing needs a rework anyway, or `z_demo`, because I'm not sure which thing is appropriate to take the `sizeof` in that function. The most significant change was in BgCobra: given that this has a texture in its struct that it has to align, it has an oversized buffer, so I added some appropriate defines to be able to do this. There is a remaining question of whether to do
```c
Lib_MemSet(shadowTex, sizeof(this->shadowTextureBuffer) & ~0xF, 0)
```
or
```c
Lib_MemSet(shadowTex, COBRA_SHADOW_TEX_SIZE, 0);
```

Drafting until #1162 is merged... apparently I started this a while ago and documented fmodf too.


include/functions.h
```diff
@@ -831,7 +831,7 @@ void func_800645A0(GlobalContext* globalCtx, CutsceneContext* csCtx);
 void Cutscene_HandleEntranceTriggers(GlobalContext* globalCtx);
 void Cutscene_HandleConditionalTriggers(GlobalContext* globalCtx);
 void Cutscene_SetSegment(GlobalContext* globalCtx, void* segment);
-void* MemCopy(void* dest, void* src, s32 size);
+void* MemCpy(void* dest, const void* src, s32 size);
```
Dragorn421, line 834: since with the capitalization it's already not `memcpy`, is there really a point to removing the `o`
This may just bring more confusion between `memcpy` and `MemCpy`, not that the `o` would help much but it's still something (at least both functions are conformant `memcpy`s, but one is in code so may not be used in boot so it's not entirely irrelevant to prevent confusion)

EllipticEllipsis: Ideally these get a proper prefix anyway. Unfortunately there's very little context for how they're used: it's essentially "the one used outside `boot`", and the `MemSet` in here isn't used at all. It's totally gone in MM, and its duties reassigned to `bcopy` or `__osMemcpy`, so practically the advice should just be "don't bother using this one".

src/overlays/actors/ovl_Bg_Jya_Cobra/z_bg_jya_cobra.c
```diff
@@ -275,8 +275,8 @@ void BgJyaCobra_UpdateShadowFromSide(BgJyaCobra* this) {
     s32 l;
     s16 rotY;
 
-    temp_s2 = ALIGN16((s32)(&this->shadowTexture));
-    Lib_MemSet(temp_s2, 0x1000, 0);
+    shadowTex = COBRA_SHADOW_TEX_PTR(this);
+    Lib_MemSet(shadowTex, sizeof(this->shadowTextureBuffer) & ~0xF, 0);
```
Dragorn421, line 279: This should indeed use `COBRA_SHADOW_TEX_SIZE` imo, these zeros are meant as the default color (transparent black) afaict, it wouldn't care if the buffer happened to be oversized, if that's what using sizeof here is meant for

src/code/code_80069420.c
```diff
@@ -1,23 +1,48 @@
...
+/**
+ * @brief memset: sets @p len bytes to @p val starting at address @p dest .
+ *
+ * There are two other memsets in this codebase,
+ * @sa Lib_MemSet(), __osMemSet()
+ * This one is unused.
+ *
+ * @param[in,out] dest address to start at
+ * @param[in] val value to write (s32, but interpreted as u8)
+ * @param[in] len number of bytes to write. Has to be s32: it is not possible to set more than 2 GB of RAM at once.
```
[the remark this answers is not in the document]
EllipticEllipsis: Mm, I suppose it would probably be better to note that this cannot sensibly be a `size_t` (The function argument can, but the temp must be `s32`, and it feels disingenuous to allow the user to accidentally feed it a large integer that gets recast into a small negative one.

src/code/__osMemset.c
```diff
@@ -0,0 +1,23 @@
+#include "global.h"
+
+/**
+ * @brief memset: sets @p len bytes to @p val starting at address @p dest .
+ *
+ * There are two other memsets in this codebase,
+ * @sa Lib_MemSet(), MemSet()
+ * This one is used in __osMalloc, z_quake, z_view, and z_camera
```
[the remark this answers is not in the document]
EllipticEllipsis: This is probably a habit from looking on GitHub with its useless searching. I think it's useful to explicitly note patterns of usage (it makes them easier to compare when you have them explicitly written down, for example), but I can see that people would regard this an unnecessary in "final" documentation.

src/overlays/actors/ovl_Bg_Jya_Cobra/z_bg_jya_cobra.c
EllipticEllipsis: The point of this is to allow downsizing the buffer to the appropriate size, but it probably does make more sense to use the texture size directly. I'll change it.

src/code/__osMemset.c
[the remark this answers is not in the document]
fig02: I agree with dragorn that its not worth commenting on where it gets used

Dragorn421 (approved):

fig02 (approved):

Roman971 (commented):
Sorry for the very late review on this one.
The changes look pretty good to me, and I just have some comments on the documentation style

fig02 (approved):

Dragorn421 (approved):

src/code/fmodf.c
```diff
@@ -1,12 +1,27 @@
 #include "global.h"
 
+/**
+ * @brief Computes one @p x modulo @p y for floats.
```
Roman971, line 4: Looks like this one was missed in the changes?
```suggestion
 * Computes one `x` modulo `y` for floats.
```

fig02 (approved):

Roman971 (approved):

PR #1166, opened 2022-03-01
Name rsp ucodes and surrounding cleanup

Thar0:
- rspboot and aspMain are motivated by the programming manual
 - f3dzex is motivated by the AC map file
 - s2dex2d is from looking through sdk libraries, the only one with [Safe] in it's string is the d(ebug) version
 - njpgdspMain is also motivated by the programming manual but is more uncertain than the first two

Additional cleanup consists of
 - Removing uses of `gSPLoadUcodeEx` preferring `gSPLoadUcode` and `gSPLoadUcodeL`
 - `sys_ucode.c` improvements
 - `gfxbuffers.c` improvements
 - `JpegTaskData` struct improvements
 - OSTask struct assignments improvements
 - Split the audio assets tables into their own file


src/code/sys_ucode.c
```diff
@@ -1,20 +1,20 @@
...
-u32 SysUcode_GetUCodeBoot(void) {
-    return &D_80009320;
+u64* SysUcode_GetUCodeBoot(void) {
+    return rspbootTextStart;
 }
 
-u32 SysUcode_GetUCodeBootSize(void) {
-    return (u32)&D_800093F0 - (u32)&D_80009320;
+size_t SysUcode_GetUCodeBootSize(void) {
+    return (size_t)((u8*)&rspbootTextEnd - (u8*)&rspbootTextStart);
```
AngheloAlf: I think a `uintptr_t` may be a better type for this cast instead of a `u8*`, but I'm not completely sure. Any reason to prefer one over the other?

src/code/z_jpeg.c
```diff
@@ -23,9 +23,9 @@ void Jpeg_ScheduleDecoderTask(JpegContext* ctx) {
         0,                    // flags
         NULL,                 // ucode_boot
         0,                    // ucode_boot_size
-        gJpegUCode,           // ucode
+        njpgdspMainTextStart,           // ucode
         0x1000,               // ucode_size
-        gJpegUCodeData,       // ucode_data
+        njpgdspMainDataStart,       // ucode_data
```
AngheloAlf: Try to keep the same indentation

data/audio_tables.rodata.s
```diff
@@ -0,0 +1,22 @@
...
+    .incbin "baserom.z64", 0xBCC270, 0x270
+
+glabel gSequenceFontTable
+    .incbin "baserom.z64", 0xBCC4E0, 0x1C0
+
+glabel gSequenceTable
+    .incbin "baserom.z64", 0xBCC6A0, 0x6F0
+
+glabel gSampleBankTable
+    .incbin "baserom.z64", 0xBCCD90, 0x80
```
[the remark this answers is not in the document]
Thar0: I vaguely recall incbin from specific files to be worse for some reason. cc @Roman971 for possible clarifications

src/code/sys_ucode.c
Thar0: As much as possible I prefer to stick to pointers rather than moving straight to uintptr_t but maybe this is misguided? I'm not really sure.

src/code/z_jpeg.c
Thar0: Fixed this just before your review came through 😅

src/code/graph.c
```diff
@@ -12,15 +12,15 @@ FaultClient sGraphUcodeFaultClient;
 
 // clang-format off
 UCodeInfo D_8012D230[3] = {
-    { UCODE_F3DZEX, D_80155F50 },
+    { UCODE_F3DZEX, gspF3DZEX2_NoN_PosLight_fifoTextStart },
     { UCODE_UNK, NULL },
-    { UCODE_S2DEX, D_80113070 },
+    { UCODE_S2DEX, gspS2DEX2d_fifoTextStart },
```
Dragorn421, line 17: `gspS2DEX2d`? where is the last `d` from

Also in general, the microcode names here are based on the "vanilla" names (e.g. f3dex2 variants) I guess, and by their capacity (e.g. NoN no near clipping however that may have been established that this is what this ucode does)
Or are there strings inside the ucode that tell what version it is? I forgot, if I ever knew

Thar0: I explained where `d` and the other names come from in the PR description. The fact that this f3dex2 is a No-nearclipping, Point-Lighting-enabled ucode is clear from the string:
`RSP Gfx ucode F3DZEX.NoN  fifo 2.08J Yoshitaka Yasumoto/Kawasedo 1999` (`/Kawasedo` is the giveaway for point lights)

Dragorn421: Sorry for inability to read/remember

Is the choice/position of `PosLight` in `NoN_PosLight_fifo` arbitrary?

And then is `s2dex2d` made up from `s2dex2` + d(ebug) or actually called that anywhere

Thar0: I used the exact name from the AC map file, so it's the original order. `s2dex2d` is called exactly that in the unlinked sdk object

include/z64.h
```diff
@@ -1743,14 +1743,17 @@ typedef struct {
...
     /* 0x00 */ u32 address;
     /* 0x04 */ u32 mbCount;
     /* 0x08 */ u32 mode;
     /* 0x0C */ u32 qTableYPtr;
     /* 0x10 */ u32 qTableUPtr;
     /* 0x14 */ u32 qTableVPtr;
-    /* 0x18 */ char unk_18[0x8];
+    /* 0x18 */ u32 mbSize; // This field is used by the microcode to save the macroblock size during a yield
+    };
+    long long int force_structure_alignment;
```
Dragorn421, line 1756: then the 4 pad bytes at 0x1C come from `force_structure_alignment`, that may be a bit sneaky?

also this jpeg ucode was never released was it, the manual part on it sure isn't entirely accurate

src/code/code_800E4FE0.c
```diff
@@ -177,12 +177,12 @@ AudioTask* func_800E5000(void) {
...
     task->flags = 0;
-    task->ucode_boot = D_801120C0;
-    task->ucode_boot_size = 0x1000;
-    task->ucode_data_size = ((rspAspMainDataEnd - rspAspMainDataStart) * sizeof(u64)) - 1;
-    task->ucode = D_801120C0;
-    task->ucode_data = rspAspMainDataStart;
-    task->ucode_size = 0x1000;
+    task->ucode_boot = aspMainTextStart;
+    task->ucode_boot_size = SP_UCODE_SIZE;
+    task->ucode_data_size = (size_t)(aspMainDataEnd - aspMainDataStart) * sizeof(u64) - 1;
```
Dragorn421, line 182: unrelated to the PR but any idea of why there's a `- 1` ?

Also is this microcode different from the `naudio` ucode described in the manual

include/z64.h
Thar0: I don't see much of an issue with the struct size being padded out by `force_structure_alignment`, I think that's more so part of the point. It has to be a multiple of 8 bytes for the DMA to function properly after all.
The JPEG library described in the manual seems to be related but not the same. I don't really know much about the library described in the manual since afaik we don't have any unlinked version of it to investigate any debugging info.

Dragorn421: I mean are you sure that word is padding then, since up until now it's been documented as unk_
How did you figure out/learn about this `mbSize` field?

src/code/code_800E4FE0.c
Thar0: The audio microcode is weird in that it doesn't use `rspboot` as a bootstrap, it bootstraps itself. The `SP_RD_LEN` and `SP_WR_LEN` hardware registers require you to enter `len - 1` as the dma length (hardware optimization, since a transfer length of 0 doesn't make sense it can be omitted). Usually, `rspboot` will subtract 1 to hide this from the user, but since the audio microcode doesn't use `rspboot` it does it manually here to save instructions in the microcode itself.

This microcode is different from the public microcode associated with the `naudio` library, although iirc one is believed to be a fork of the other, if I'm remembering discussions about sm64 audio correctly.

include/z64.h
Thar0: I have an almost 100% documented disassembly of the jpeg microcode

src/code/sys_ucode.c
```diff
@@ -1,20 +1,20 @@
...
-u32 SysUcode_GetUCodeBootSize(void) {
-    return (u32)&D_800093F0 - (u32)&D_80009320;
+size_t SysUcode_GetUCodeBootSize(void) {
+    return (size_t)((u8*)rspbootTextEnd - (u8*)rspbootTextStart);
 }
 
-u32 SysUcode_GetUCode(void) {
-    return D_8012DBA0;
+u64* SysUcode_GetUCode(void) {
+    return sDefaultGSPUcodeText;
```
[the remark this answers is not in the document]
Thar0: The whole file is fairly geared towards graphics ucode, I think it may be best to revisit the function names when this file gets "complete" documentation. I can update the variable names however.

Dragorn421 (approved):

Dragorn421 (approved):

fig02 (approved):
thought i approved this a long time ago, guess i did not. apologies

include/ultra64/ucode.h
```diff
@@ -0,0 +1,27 @@
+#ifndef ULTRA64_UCODE_H
+#define ULTRA64_UCODE_H
+
+#include "types.h"
+
+#define SP_DRAM_STACK_SIZE8     (0x400)
+#define SP_DRAM_STACK_SIZE64    (SP_DRAM_STACK_SIZE8 >> 3)
```
Roman971, line 7: Is there a particular reason to use SIZE8/SIZE64 macros instead of handling it the same as `OS_YIELD_DATA_SIZE` with `/ sizeof(u64)` for example?

include/z64.h
```diff
@@ -1743,19 +1743,22 @@ typedef struct {
...
 } JpegTaskData; // size = 0x20
 
 typedef struct {
     /* 0x000 */ JpegTaskData taskData;
-    /* 0x020 */ char yieldData[0x200];
+    /* 0x020 */ u64 yieldData[0x200/sizeof(u64)];
```
Roman971, line 1761:
```suggestion
    /* 0x020 */ u64 yieldData[0x200 / sizeof(u64)];
```

data/unk_80009410.data.s
```diff
@@ -9,6 +9,18 @@
 
 .balign 16
 
+# Unused. This appears to be RSP code, however it is not part of
+#         rspboot as rspbootTextEnd is at this symbol
+glabel D_800093F0
```
Roman971, line 14: Since this appears to be RSP code, I feel like it would be better to keep it in `rsp_boot.text.s` even if it's not part of `rspbootText` itself. (at least I don't think this file is appropriate for it since it's .data and it's already weird on its own)

Roman971 (commented):
Looks good to me overall. I just have a few minor comments

include/ultra64/ucode.h
Thar0: The only reason is that's how they are in the original libultra headers

data/unk_80009410.data.s
Thar0: If anything I'd rather have another separate file, since eventually rspboot won't exist in it's current form and I'd rather not include this unrelated code in the eventual rspboot disassembly

include/ultra64/ucode.h
Roman971: I would personally go with `/ sizeof(u64)` since that fits the codebase a lot better, but I don't have a strong opinion on it

data/unk_80009410.data.s
Roman971: I personally feel like this belongs with the rest of the rsp code, even if it's unused/unidentified, but having another file would probably be okay as well

include/ultra64/ucode.h
Thar0: Since there's no `SP_DRAM_STACK_SIZE` macro I'd either have to use `SP_DRAM_STACK_SIZE8`, or make a new macro to use `/ sizeof(u64)`. I'm not sure if it's worth it over just using `SP_DRAM_STACK_SIZE64`

Roman971: I was thinking `SP_DRAM_STACK_SIZE8` could simply be `SP_DRAM_STACK_SIZE` and wherever the u64 size is needed we would have `SP_DRAM_STACK_SIZE / sizeof(u64)`, but I might be missing something

Thar0: That would break compatibility with the original header, which so far (afaik) we've managed not to do anywhere (other than having incomplete/jumbled libultra headers in general), and I think it would be good to strive to maintain such compatibility.
That is, adding our own stuff or making equivalent replacements (like the `unsigned` to `u32` in the handwritten asm PR) would be fine, but renaming or entirely removing something wouldn't be

Dragorn421 (approved):
Interesting that `unk_800093F0.data` can be either .data and at the start of the boot segment in the spec, or .text and at the end of the segment
We will never know which is "right" I guess? 😄

fig02 (approved):

Roman971 (approved):

PR #1167, opened 2022-03-03
Cleanup pass on in-actor effects systems

Dragorn421:
Some cleanup on the in-actor effects systems as I decided to call them.

A bunch of actors have similar (that is, copy-pasted) code to handle their own effects ("particles" but we call them effects) outside of the game's effects system. They are similar enough that the functions and variables involved can mostly be named consistently across the various actors.

Note: object_kankyo and en_viewer also have "effects" systems but they are very different from the ones handled in this PR (they look unique), so I left them untouched.

This is only a first pass, it turned out to be larger changes than I expected so I'm splitting it here

--------

- Name or rename stuff to use "effect". For example `UpdateEffects`, `DrawEffects`
- Name `materialFlag` the common "is material set up" "boolean" in DrawEffects functions
- `*_EFFECT_COUNT` defines for the effects buffers lengths
- Rename effects buffers to `effects`/`sEffects`
- Other cleanup, see commit messages for details


fig02:
feel like max is probably more appropriate than a count? its (usually) 200 slots possible for effects, but not all are necessarily used

Dragorn421:
Hm yeah `_effects_max` sounds better than `_effects_count`, but it may then be confused with the max enum value of effect types, for example `boss_fd_effects_max` for the effects buffer size, and for the max of `BossFdEffectType`, `boss_fd_effect_max` ? only the `s` would differ

This may not end up being an issue depending on the style of enum names for effect types, or not needing that max value at all. I haven't thoroughly checked effect types usage, I plan to do that in a next pass

It's not hard to change this, we can go with `_effects_max` now and change later if needed

fig02:
another option is BUFFER_SIZE, like we currently do for something like the lights buffer
`#define LIGHTS_BUFFER_SIZE 32` `LightNode buf[LIGHTS_BUFFER_SIZE];`
I think max makes just as much sense though. but yea that potential for confusion if unfortunate

Dragorn421:
Discussion continued on Discord https://discord.com/channels/688807550715560050/688851317593997489/951175437860630608

TLDR:
I want to eventually use the currently-`_effects_count` define in spots such as
https://github.com/Dragorn421/oot/blob/4785f642e91d9add18a630eac78c6a958849e577/src/overlays/actors/ovl_Boss_Tw/z_boss_tw.c#L747
```c
BossTw_AddDotEffect(globalCtx, &pos, &velocity, &accel, (s16)Rand_ZeroFloat(2.0f) + 8, blastType, 75);
```
the `75` could be `(s32)(0.5f * BOSS_TW_EFFECTS_COUNT)`

Imo it would look weird to have `(s32)(0.5f * BOSS_TW_EFFECTS_BUFFER_SIZE)` as I see "count" as some global parameter on effects count
Fig sees _effects_count as the number of active effects, which is not what the define is
Roman argued _effects_count just refers to the amount of effects, regardless of them being active or not

FWIW current master uses `BOSSFD_EFFECT_COUNT`, `EFFECT_COUNT` (fishing), `CLEAR_TAG_EFFECT_MAX_COUNT`

-------

All in all I don't have a strong opinion on `_effects_count` or other options, I'll keep `_effects_count` for now but I'm okay with changing to whatever

fig02 (approved):

Dragorn421:
I found out EnNiw, EnSyatekiNiw and EnFz also have this in-actor effect system (thanks regex `gSPDisplayList.*((\n|\s)*.*gDPPipeSync.*)?(\n|\s)*\n\s*[a-z]+\+\+`). Hopefully there aren't many more missing

src/overlays/actors/ovl_Bg_Dy_Yoseizo/z_bg_dy_yoseizo.h
```diff
@@ -10,6 +10,8 @@ struct BgDyYoseizo;
 
 typedef void (*BgDyYoseizoActionFunc)(struct BgDyYoseizo*, GlobalContext*);
 
+#define BG_DY_YOSEIZO_EFFECTS_COUNT 200
```
Roman971, line 13: Pretty minor but I think `_EFFECT_COUNT` (singular) would be more consistent with the rest of the codebase (with the added benefit of being shorter)

Roman971 (commented):
Sorry for the late review. I have a couple minor comments but this looks good to me otherwise

Roman971 (approved):

PR #1170, opened 2022-03-10
Renamed frameBuffer and frame buffer to framebuffer. Also zbuffer to zBuffer

MegaMech:
Modifying the tools might be an overstep. Let me know if there's any issues and I'll undo those changes.

This PR makes the framebuffer naming consistent in the repo and consistent across decomps (sm64 and mk64).
A small zBuffer change as well. Z and Buffer are considered two differing words.


fig02:
Yea if you could undo the ZAP and libgfxd changes

you can submit the fix to ZAP upstream here if you like, then we sync with that https://github.com/zeldaret/ZAPD
and libgfxd is glanks separate library. I imagine he probably doesnt care about that for his library so we can keep the changes just to oot and ZAPD

MegaMech:
You may want to double check. But the tools should be undone now.

fig02:
I think we will be closing for now as people weren't too excited about a one-word framebuffer. 
We'll revisit this and all the frame buffer related acronyms when we start renaming the original game's systems. Thanks for pointing this out to us tho

PR #1173, opened 2022-03-11
Fix DL exporting as binary instead of as a DL in market_alley_n.xml

MegaMech:
This PR results in:
```
Gfx market_alley_n_room_0DL_0756E0[] = {
    gsSPDisplayList(market_alley_n_room_0DL_075678),
    gsSPEndDisplayList(),
};
```
Instead of
```
u8 a[] = {
 0xDE, etc.
}
```

### DIFF
--- a/assets/xml/scenes/misc/market_alley_n.xml
+++ b/assets/xml/scenes/misc/market_alley_n.xml
@@ -4,5 +4,6 @@
     </File>
     <File Name="market_alley_n_room_0" Segment="3">
         <Room Name="market_alley_n_room_0" Offset="0x0"/>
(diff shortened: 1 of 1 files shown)


Dragorn421 (approved):
A quick search (`grep -l -r assets -e '0xDE, 0x00, 0x00, 0x00'`) seems to indicate the same thing happens in many rooms.
The few I've looked at were, like `market_alley_n_room_0DL_0756E0`, unused, so this is kinda weird. Probably an oversight with the original map tool?

MegaMech:
If you want I can fix all of them.

fig02 (approved):

Roman971 (approved):

PR #1174, opened 2022-03-12
Define gMorphaAndPoolWaterTex to remove unknown bytes.

MegaMech:
Note that this texture should be exported as an rgba16.

But it's rendered twice in-game.
Rgba16 for the pool water.
i8 for the morpha tentacle base.

### DIFF
--- a/assets/xml/objects/object_mo.xml
+++ b/assets/xml/objects/object_mo.xml
@@ -50,7 +50,7 @@
         <DList Name="gMorphaTentaclePart38DL" Offset="0x87A8"/>
         <DList Name="gMorphaTentaclePart39DL" Offset="0x87F0"/>
         <DList Name="gMorphaTentaclePart40DL" Offset="0x8838"/>
-        <Texture Name="gMorphaAndPoolWaterTex" Format="rgba16" Width="32" Height="32" Offset="0x8870"/>
+        <Texture Name="gMorphaWaterTex" Format="rgba16" Width="32" Height="32" Offset="0x8870"/>
 
         <!-- DLists for Morpha's particles -->
         <DList Name="gMorphaBubbleDL" Offset="0x140"/>
--- a/assets/xml/objects/object_mo.xml
+++ b/assets/xml/objects/object_mo.xml
@@ -50,6 +50,7 @@
         <DList Name="gMorphaTentaclePart38DL" Offset="0x87A8"/>
         <DList Name="gMorphaTentaclePart39DL" Offset="0x87F0"/>
         <DList Name="gMorphaTentaclePart40DL" Offset="0x8838"/>
+        <Texture Name="gMorphaAndPoolWaterTex" Format="rgba16" Width="32" Height="32" Offset="0x8870"/>
 
         <!-- DLists for Morpha's particles -->
         <DList Name="gMorphaBubbleDL" Offset="0x140"/>


assets/xml/objects/object_mo.xml
```diff
@@ -50,6 +50,7 @@
         <DList Name="gMorphaTentaclePart38DL" Offset="0x87A8"/>
         <DList Name="gMorphaTentaclePart39DL" Offset="0x87F0"/>
         <DList Name="gMorphaTentaclePart40DL" Offset="0x8838"/>
+        <Texture Name="gMorphaAndPoolWaterTex" Format="rgba16" Width="32" Height="32" Offset="0x8870"/>
```
fig02, line 54: we can prob just say gMorphaWaterTex, and it will be implied that it is used for both the pool and morpha itself because they both belong to the same body of water

MegaMech: I think:
`gMorphaBaseTentacleAndPoolWaterTex`
would be more suitable.

fig02: well, my point is that it probably doesnt need to be more specific cause its just a water texture. and morpha is made of water, so that is already a given

MegaMech: Are you sure it wouldn't be better to be like: "Hey, this texture is used twice" in the name?

MegaMech: Or perhaps a comment?

fig02: I dont think its needed personally, but we can see what other people think

louist103: It would be nice to indicate that it is used for both somewhere. Either a comment or the name is fine to me.

fig02: comment is probably preferred. Textures get reused pretty frequently, and it would be impossible to indicate all of the uses in its name in some cases. 
imo a comment is also not really needed cause if you want to see occurences you can just search for it. but i would compromise for a comment over the name

louist103: The only reason I would like to have the comment if we dont make the name specific is that if you global search for it VSCode (and I believe VS2022) wont look at the actual DLists by default but it will look at the XML.

fig02: if youre searching for anything in an xml its a given that you need to turn off the gitignore exclusion

fig02: its fair to not know that the some text editors work this way, of course. but you would know something is up when searching and not getting hits, and probably ask someone about it. This happened to mega the other day when i explained the feature to him.
It shouldnt be a justification for a naming scheme, is more my point

Dragorn421 (approved):

fig02 (approved):

Roman971 (approved):

assets/xml/objects/object_mo.xml
MegaMech: Yea, I've never noticed this git ignore issue. It was certainly breaking my brain.
It's my understanding that this has never been an issue for sm64 nor mk64.

fig02: my understanding is that dlists are not extracted and are included in the repo right?
the whole assets folder is gitignored for now, though the plan is to commit the files themselves and only ignore the files that contain the extracted assets that are deemed too sensitive

MegaMech: Also, keep in-mind. That technically the *second* use of this texture is not the same texture. It's rendered as an i8. So it's different. It's not the same.

Either way. I don't really care. Would it be possible to put a rush on this PR? Whatever name and comment or not is fine with me.

PR #1175, opened 2022-03-12
Add uintptr_t to main.c and sys_cfb.c

louist103:
I also added `stdint` from MM.


include/libc/stdint.h
```diff
@@ -0,0 +1,27 @@
...
+
+#define INT8_MIN    (-0x80)
+#define INT16_MIN   (-0x8000)
+#define INT32_MIN   (-0x80000000)
+#define INT64_MIN   (-0x8000000000000000)
+
+#define INT8_MAX    0x7f
+#define INT16_MAX   0x7fff
+#define INT32_MAX   0x7fffffff
+#define INT64_MAX   0x7fffffffffffffff
```
Dragorn421, line 15: the repo uses uppercase hexadecimal everywhere. idk if this should apply to a "standard" header

src/code/main.c
```diff
@@ -50,15 +50,15 @@ void Main(void* arg) {
     PreNmiBuff_Init(gAppNmiBufferPtr);
     Fault_Init();
     SysCfb_Init(0);
-    sysHeap = (u32)gSystemHeap;
+    sysHeap = gSystemHeap;
```
Dragorn421, line 53: Afaict `uintptr_t` is defined with `void*` pointers so does casting to `uintptr_t` first requires a cast to `void*`? Like:
```suggestion
    sysHeap = (void*)gSystemHeap;
```

src/code/sys_cfb.c
```diff
@@ -1,11 +1,11 @@
 #include "global.h"
 
-u32 sSysCfbFbPtr[2];
-u32 sSysCfbEnd;
+uintptr_t sSysCfbFbPtr[2];
+uintptr_t sSysCfbEnd;
 
 void SysCfb_Init(s32 n64dd) {
     u32 screenSize;
```
Dragorn421, line 7: `screenSize` may be `size_t`? it's the length of the 1D'ed color frame buffer array

src/code/main.c
```diff
@@ -3,13 +3,13 @@
...
 s32 gScreenHeight = SCREEN_HEIGHT;
-u32 gSystemHeapSize = 0;
+size_t gSystemHeapSize = 0;
 
 PreNmiBuff* gAppNmiBufferPtr;
 SchedContext gSchedContext;
 PadMgr gPadMgr;
 IrqMgr gIrqMgr;
-u32 gSegments[NUM_SEGMENTS];
+uintptr_t gSegments[NUM_SEGMENTS];
```
Dragorn421, line 12: there are many things to change associated with `gSegments`, like the `SEGMENTED_TO_VIRTUAL` and co macros, and various variables.
idk if it should be done in this PR or in another one

Dragorn421 (commented):
\+ Reminder we still need to figure out what to do about locations where we can't use the right types for matching reasons, which is not a rare occurrence (mainly with signed `size_t`).

src/code/main.c
louist103: I was going to do that in a different PR but I can do it here.

src/code/main.c
louist103: I dont think it needs it since that variable is an array. If anything it should have a `uintptr_t` cast.

Dragorn421: My point was `gSystemHeap` is a `u8*`, not a `void*`, and `uintptr_t` is for `void*` pointers afaict

src/code/main.c
```diff
@@ -37,10 +37,10 @@ void Main(void* arg) {
...
     OSMesgQueue irqMgrMsgQ;
     OSMesg irqMgrMsgBuf[60];
-    u32 systemHeapStart;
-    u32 fb;
-    s32 debugHeapStart;
-    s32 debugHeapSize;
+    uintptr_t systemHeapStart;
+    uintptr_t fb;
+    uintptr_t debugHeapStart;
+    uintptr_t debugHeapSize;
```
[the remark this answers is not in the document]
louist103: `debugHeapSize` is `size_t` on my local branch so I dont know why its `uintptr_t` here.

include/libc/stddef.h
```diff
@@ -3,9 +3,9 @@
 
 #define NULL ((void*)0)
 
-typedef unsigned long size_t;
+typedef unsigned int size_t;
 
-typedef unsigned int uintptr_t;
+typedef signed int ptrdiff_t;
```
Dragorn421, line 8: I'd rather see `ptrdiff_t` introduced when it will get used

Reading more stuff, it doesn't seem entirely like a "signed size_t", it's specifically for the difference between two pointers inside one array. and size_t is also an array index but not only that

with `s32 data[];` then `&data[5] - &data[3]` evaluates to the `ptrdiff_t` value 2 (5 - 3)

and I don't see that being useful anywhere?

louist103: I don't see any problem with having it there now. Its not like it hurts anything.

Dragorn421: To be clear with what I meant in my comment above, I think ptrdiff_t is useless because we shouldn't use it

Dragorn421:
Bump on the reminder above about what-about-types-that-cant-be-proper-types

Also may I ask again what the goal is here?

In the past the goal has been confusingly refered to and handled as being a combination of for-being-clean-best-practice / documentation-purposes / architecture-agnostic-no-fixed-pointer-size

For being clean and best practice : in line with the reminder above, oot's code was originally not clean so there's no way to achieve that. We need to figure out how to address this

For documentation purposes : this is in my understanding a wrong approach, what matters with proper types is compatibility, not documentation

For being architecture agnostic with no fixed pointer size : same as "being clean" I guess, I agree it would be better but again what about inconsistencies this would introduce when that goal clashes with the primary goal of having a matching decompilation

I like that last sentence as a summary of my whole feeling on the situation so,

TLDR: How do we resolve having both the goals of 1) using proper types everywhere, and 2) keeping the decomp matching

louist103:
> 

As discussed multiple times in the past in the discord the goal is the first two things you said. Compatibility with other architectures was never an explicit goal of this, something I have also made clear several times in the past.

If the problem is "this isn't how Nintendo originally did it, or Nintendo's code wasn't clean so whats the point?" we already do plenty of things they didn't do. Just see the ASM processor for example.

I don't see how this is the wrong approach for documentation. It is useful to know what variables are supposed to be treated as pointers and what are just integers that happen to need 32 bits.

If this isn't something that has a chance of getting I really want to know sooner rather than later so I can stop wasting my time and yours. I already wasted enough time with #1089

louist103:
See above comment:
https://github.com/zeldaret/oot/pull/1175#issuecomment-1099410013

PR #1176, opened 2022-03-14
Doc misc 3

Dragorn421:


src/overlays/actors/ovl_En_Door/z_en_door.c
```diff
@@ -299,27 +299,27 @@ void EnDoor_Update(Actor* thisx, GlobalContext* globalCtx) {
...
         transitionEntry = &globalCtx->transiActorCtx.list[(u16)this->actor.params >> 0xA];
         rot->z += this->actor.world.rot.y;
         if ((globalCtx->roomCtx.prevRoom.num >= 0) ||
             (transitionEntry->sides[0].room == transitionEntry->sides[1].room)) {
-            phi_v0_2 = ((this->actor.shape.rot.y + this->skelAnime.jointTable[3].z) + rot->z) -
-                       Math_Vec3f_Yaw(&globalCtx->view.eye, &this->actor.world.pos);
-            *dList = (ABS(phi_v0_2) < 0x4000) ? temp_a2[0] : temp_a2[1];
+            rotDiff = ((this->actor.shape.rot.y + this->skelAnime.jointTable[3].z) + rot->z) -
+                      Math_Vec3f_Yaw(&globalCtx->view.eye, &this->actor.world.pos);
+            *dList = (ABS(rotDiff) < 0x4000) ? doorDLists[0] : doorDLists[1];
```
fig02, line 316: how come you have yaw diff above in a different place and rotDiff here?
I guess a better question is what is this code doing

I would assume that rot.z from the skeleton stuff is actually the yaw, but is using z cause model space stuff. But that is my assumption without actually investigating

Dragorn421: I have no idea what this code is doing. I assume pick a "door side" but it's quite cryptic
I went with more generic rotDiff because I didn't actually look at the skeleton to check if the 3rd bone z rotation is actually a yaw
I'll take a look at improving `gDoorSkel` doc in this pr

Roman971: Are we waiting for more investigation on this?
`rotDiff` seems fine to me as a generic name either way, fwiw

fig02: yea i was waiting for a followup, but yea there will prob be more docs passes on this stuff later anyway. so its fine if not

Dragorn421: I didn't forget but there's a lack of tool capabilities to investigate this efficiently

I can revert this one variable if you want, but this code needs more than a named variable (e.g. comments) to be legible anyway

fig02: nah its more than fine

fig02 (approved):

Roman971 (approved):

PR #1178, opened 2022-03-14
Message Queues, Threads, and surroundings cleanup

Thar0:
- Gave all user threads id and priority defines
 - Used message queue macros everywhere appropriate
 - Used `ARRAY_COUNT` more in `osCreateMesgQueue` and fixed the size of several `OSMesg` arrays
 - Cleanup in `piacs.c` and `siacs.c`
 - Fixed the type for `__osThreadTail`
 - Used thread state defines (`OS_STATE_`) more
 - Moved 0x10 bytes of unknown out of `AudioMgr`, `GraphicsContext` and `JpegContext` and into `OSScTask`
 - Named message queue and message buffer variables in the `AudioMgr` struct
 - Fixed improper casts to `OSMesg*` although we need to discuss whether we are keeping these casts anyway

I also changed `format.sh` to prefer clang-format-11 for those of us who have both installed. Using clang-format when formatting this PR changed a lot of unrelated things.


glankk (approved):
I don't know if there's an established convention of naming messages queues `Q`, I would maybe consider a more descriptive `Queue`. But I'm fine with it either way.

fig02:
we may want to consider THREAD_ID, i read "thread did" a couple times until i realized what it was lol, as silly as that is

and is now a good time to get rid of the Z_ prefix on these? this is just a holdover from forever ago when we used to put it on everything afaik

also agreed with glank on the above on queue over q

fig02 (approved):
defines are much better thanks. PRIO instead of PRI may help people with understanding what the abbreviation is quicker, but i leave that up to you

glankk (approved):

src/code/z_ss_sram.c
```diff
@@ -32,20 +32,20 @@ void SsSram_Init(u32 addr, u8 handleType, u8 handleDomain, u8 handleLatency, u8
...
+    OSMesg msg;
 
-    osCreateMesgQueue(&sSsSramContext.mesgQ, &mesg, 1);
+    osCreateMesgQueue(&sSsSramContext.msgQueue, &msg, 1);
     sSsSramContext.ioMesg.dramAddr = dramAddr;
     sSsSramContext.ioMesg.size = size;
     osWritebackDCache(dramAddr, size);
     osEPiStartDma(&sSsSramContext.piHandle, &sSsSramContext.ioMesg, direction);
-    osRecvMesg(&sSsSramContext.mesgQ, &mesg, OS_MESG_BLOCK);
+    osRecvMesg(&sSsSramContext.msgQueue, &msg, OS_MESG_BLOCK);
```
Dragorn421, line 48: looks like a jank usage to store the message in the queue buffer (a=a here), thankfully it doesn't matter (`osContInit` also does it 👿, other similar queue usage pass rightfully NULL here)

src/code/z_msgevent.c
```diff
@@ -1,20 +1,19 @@
...
     task.next = NULL;
     task.flags = OS_SC_RCP_MASK;
-    task.msgQ = &queue;
+    task.msgQueue = &queue;
     task.msg = NULL;
     task.framebuffer = NULL;
     task.list.t.type = M_NULTASK;
-    osCreateMesgQueue(task.msgQ, &msg, 1);
-    osSendMesg(&gSchedContext.cmdQ, &task, OS_MESG_BLOCK);
+    osCreateMesgQueue(task.msgQueue, &msg, 1);
```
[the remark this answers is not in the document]
glankk: Both are equally valid, I personally prefer this when trying to receive exactly one message.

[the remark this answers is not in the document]
Thar0: Yeah there's no strict convention so I've opted to stick with not using arrays for single messages since that's what we already do. One notable exception is `siacs.c` and `piacs.c`, this is just to match the original libultra types.

include/padmgr.h
```diff
@@ -15,9 +15,9 @@ typedef struct PadMgr {
     /* 0x0010 */ OSMesg serialMsgBuf[1];
     /* 0x0014 */ OSMesg lockMsgBuf[1];
     /* 0x0018 */ OSMesg interruptMsgBuf[4];
-    /* 0x0028 */ OSMesgQueue serialMsgQ;
-    /* 0x0040 */ OSMesgQueue lockMsgQ;
-    /* 0x0058 */ OSMesgQueue interruptMsgQ;
+    /* 0x0028 */ OSMesgQueue serialMsgQueue;
+    /* 0x0040 */ OSMesgQueue lockQueue;
+    /* 0x0058 */ OSMesgQueue interruptQueue;
```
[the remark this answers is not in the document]
Thar0: I was indeed going to revisit this in the dedicated padmgr doc PR but I think I'll change it to `serialEventQueue` in this PR

include/ultra64/thread.h
```diff
@@ -30,6 +30,7 @@ typedef union {
         /* 0x00 */ f32 f_odd;
         /* 0x04 */ f32 f_even;
     } f;
+    f64 d;
```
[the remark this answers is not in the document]
Thar0: I just happened to notice it was missing (it's there in the original `os_thread.h`) so I figured I'd take the opportunity to add it.

src/boot/z_std_dma.c
```diff
@@ -351,16 +352,14 @@ s32 DmaMgr_SendRequestImpl(DmaRequest* req, u32 ram, u32 vrom, u32 size, u32 unk
...
-        if ((sDmaMgrQueueFullLogged == 0) && (sDmaMgrMsgQueue.validCount >= sDmaMgrMsgQueue.msgCount)) {
-            sDmaMgrQueueFullLogged++;
-            osSyncPrintf("%c", BEL);
-            osSyncPrintf(VT_FGCOL(RED));
-            osSyncPrintf("dmaEntryMsgQが一杯です。キューサイズの再検討をおすすめします。");
-            LOG_NUM("(sizeof(dmaEntryMsgBufs) / sizeof(dmaEntryMsgBufs[0]))", ARRAY_COUNT(sDmaMgrMsgs),
-                    "../z_std_dma.c", 952);
-            osSyncPrintf(VT_RST);
-        }
+    if (1 && (sDmaMgrQueueFullLogged == 0) && MQ_IS_FULL(&sDmaMgrMsgQueue)) {
```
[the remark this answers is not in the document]
Thar0: I think I'll stick with `1 &&` since it's also that way higher up in this function and I think that one has to be that way since it's in a nested condition? If so then assuming they come from the same thing it would make sense for them both to be this way.

src/code/padmgr.c
```diff
@@ -4,55 +4,55 @@
...
+void PadMgr_UnlockSerialMesgQueue(PadMgr* padMgr, OSMesgQueue* ctrlrQueue) {
     if (D_8012D280 > 2) {
         // "serialMsgQ Unlock"
         osSyncPrintf("%2d %d serialMsgQロック解除します   %08x %08x %08x\n", osGetThreadId(NULL),
-                     padMgr->serialMsgQ.validCount, padMgr, &padMgr->serialMsgQ, ctrlrQ);
+                     MQ_GET_COUNT(&padMgr->serialEventQueue), padMgr, &padMgr->serialEventQueue, ctrlrQueue);
     }
 
-    osSendMesg(&padMgr->serialMsgQ, ctrlrQ, OS_MESG_BLOCK);
+    osSendMesg(&padMgr->serialEventQueue, (OSMesg)ctrlrQueue, OS_MESG_BLOCK);
```
Dragorn421, line 33: Here `ctrlrQueue` is the `sSerialEventQueue` as created in Main() and passed to `PadMgr_Init` as `serialEventQueue`
`padMgr->serialEventQueue` only holds `serialEventQueue`, it isn't `serialEventQueue`

I'm not very inspired for names...
```suggestion
    osSendMesg(&padMgr->serialEventQueueLockQueue, (OSMesg)serialEventQueue, OS_MESG_BLOCK);
```

Thar0: I've attempted to make this less ambiguous for now, we can probably revisit in a more padmgr-focused setting

include/z64.h
```diff
@@ -42,13 +42,28 @@
...
 #define REGION_EU 3
 
-#define Z_PRIORITY_MAIN        10
-#define Z_PRIORITY_GRAPH       11
-#define Z_PRIORITY_AUDIOMGR    12
-#define Z_PRIORITY_PADMGR      14
-#define Z_PRIORITY_SCHED       15
-#define Z_PRIORITY_DMAMGR      16
-#define Z_PRIORITY_IRQMGR      17
+#define THREAD_PRI_IDLE_INIT    10
```
Roman971, line 45: The idea behind the Z_PRIORITY names was to follow the same scheme as thread.h defines while having Z_ instead of OS_ to differentiate them as being custom for the game(s).

I guess the intent here is to move away from that and the names seem good, but I find it a bit odd to have OS_PRIORITY_ and THREAD_PRI_ for the same group of values, so I'm not sure if it's a good idea or if there is a better solution

src/boot/z_std_dma.c
```diff
@@ -351,19 +352,17 @@ s32 DmaMgr_SendRequestImpl(DmaRequest* req, u32 ram, u32 vrom, u32 size, u32 unk
...
+        osSyncPrintf("%c", BEL);
+        osSyncPrintf(VT_FGCOL(RED));
+        osSyncPrintf("dmaEntryMsgQが一杯です。キューサイズの再検討をおすすめします。");
+        LOG_NUM("(sizeof(dmaEntryMsgBufs) / sizeof(dmaEntryMsgBufs[0]))", ARRAY_COUNT(sDmaMgrMsgBuf), "../z_std_dma.c",
+                952);
+        osSyncPrintf(VT_RST);
     }
 
-    osSendMesg(&sDmaMgrMsgQueue, req, OS_MESG_BLOCK);
+    osSendMesg(&sDmaMgrMsgQueue, (OSMesg)req, OS_MESG_BLOCK);
```
Roman971, line 365: Casting a OSMesg to OSMesg seems unnecessary, but I guess this goes back to the question about always casting when calling the function or not, and I'm still not sure what makes the most sense

Dragorn421: `req` is a `DmaRequest*` here, which indeed is trivially castable to `OSMesg` aka `void*`, but it's not super obvious either especially with all the weird casts to `OSMesg` there are in various places imo

the cast happens regardless, it's just that making it explicit seems like a better idea? to me at least it does

include/z64.h
Dragorn421: It does seem a bit weird to me too at first now that you pointed it out, but on second thought we also don't have z_ prefixes for every single thing that isn't libultra so it's probably fine to not have any here as well

the os_priority_ defines in libultra also lack the keyword "thread", which we can't add because original names, it's fine in the context of libultra but "z_priority_" could be easily confused with other priorities in "zelda" context imo

all in all I think "thread" should stay in the name, so if we still wanted to make `THREAD_PRI_` look more like `OS_PRIORITY_`, it would be maybe `Z_THREAD_PRIORITY_` (?) which huh why not but also doesn't seem needed?

src/boot/z_std_dma.c
Roman971: Oh I must have misread this and confused it with the msg variable, so yeah the cast makes a bit more sense.
I'm still on the fence about it because omitting it doesn't show up as a warning or anything and it's arguably useless, so it will likely be forgotten in the future (either in the repo itself, or in a modding scenario). But I guess it's fine either way

Roman971 (approved):
The couple things I mentioned are worth discussing further imo, but otherwise this looks good to me and we could probably get it in

src/boot/z_std_dma.c
Dragorn421: I'm thinking explicit casts to `OSMesg`/`OSMesg*` when that's not already the type of the argument being passed, would really help confusion between

- `osSendMesg` taking in a message value `OSMesg` (`void*` but also used to pass integers for example https://github.com/zeldaret/oot/blob/a9084ad65e272441e838723a9ebade9f4b01d9eb/src/code/audio_load.c#L1711)
- `osRecvMesg` taking in a pointer to where to store a message, aka `OSMesg*` (`void**`)

The pointer of pointer usage especially on a `void` makes it imo confusing if you're passing a value (or "pointer value" I guess since the type for the message value `OSMesg` is `void*`) or a pointer (a pointer to the "pointer value")😅
but maybe there should be no unneeded casts at all and we should just be smarter

Dragorn421 (approved):

PR #1179, opened 2022-03-16
Document En_Rd

tom-overton:
I already documented this in MM, so most of this involved porting my documentation over. Let me know if anything I did isn't agreeable with OoT's code styling.


src/overlays/actors/ovl_En_Rd/z_en_rd.c
```diff
@@ -515,273 +570,280 @@ void func_80AE3454(EnRd* this, GlobalContext* globalCtx) {
...
+                    this->fireTimer = 40;
                 } else {
                     Actor_SetColorFilter(&this->actor, 0x4000, 0xFF, 0, 8);
                 }
 
                 Actor_ApplyDamage(&this->actor);
                 if (this->actor.colChkInfo.health == 0) {
-                    func_80AE2630(globalCtx, &this->actor, 1);
-                    func_80AE3C20(this);
+                    EnRd_UpdateParentForOtherRedeads(globalCtx, &this->actor, 1);
```
Dragorn421, line 842: the last argument in `EnRd_UpdateParentForOtherRedeads` can be true/false

src/overlays/actors/ovl_En_Rd/z_en_rd.h
```diff
@@ -8,29 +8,75 @@ struct EnRd;
...
-    /* 0x0319 */ u8 unk_319;
-    /* 0x031A */ u8 unk_31A;
-    /* 0x031B */ u8 unk_31B;
-    /* 0x031C */ u8 unk_31C;
-    /* 0x031D */ u8 unk_31D;
+    /* 0x030C */ union {
+                    s16 animationJudderTimer;
+                    s16 deathTimer;
+                    s16 coffinRiseForwardAccelTimer;
+                };
```
Dragorn421, line 68: I think this should just be a generic `s16 timer`, as sad as it is
`union` doesn't seem to get much use in oot code, and also I think it's just confusing to use (accidentally use the same memory for two purposes)

Dragorn421 (commented):
Very nice job!

src/overlays/actors/ovl_En_Rd/z_en_rd.c
```diff
@@ -173,83 +205,97 @@ void EnRd_Destroy(Actor* thisx, GlobalContext* globalCtx) {
...
+/**
+ * This function does two things depending on whether setParent is true or false.
+ * - If setParent is true, this function sets thisx to be the parent for all Redeads
+ *   in the area that are capable of mourning. This is used right when thisx first
+ *   dies to make the other Redeads mourn it.
+ * - If setParent is false, this function nulls out the parent for all Redeads in the
+ *   are whose parents is thisx. This is used when thisx is fading away to make the
+ *   other Redeads stop mourning over it.
+ */
+void EnRd_UpdateParentForOtherRedeads(GlobalContext* globalCtx, Actor* thisx, s32 setParent) {
```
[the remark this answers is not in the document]
tom-overton: I went with `EnRd_UpdateMourningTarget`, since I feel that describes what it does precisely; it updates the target for the Redead's mourning, including the "no target" case when the third parameter is false. Let me know if you don't like the name, though.

src/overlays/actors/ovl_En_Rd/z_en_rd.h
tom-overton: In this specific case, I'd be relatively okay with just doing a single `s16 timer`. But I want to gently challenge what you said about `union`, because the devs *did* actually accidentally or purposefully use the same memory for multiple different purposes. Looking at `Obj_Tokeidai` in MM for example, there's a single struct variable that's used as a timer, as a translation amount, as acceleration, and as speed, all depending on the context. You can see another example in OoT with En_Tite, where actionVar1 is used as two different timers and as a state variable (the "actionVar + defines" thing that actor does is just a different way of implementing a union). Additionally, there are still lots of undocumented or semi-documented code in both codebases, which can make it hard to see *when* they're using the same memory for different purposes; it all looks the same when it's unk_XXX.

I agree that it can sort of be confusing to have multiple different variables that end up being backed by the same bit of memory, but there are clear cases where it's obvious that's what they did. This specific case *isn't* one of those clear cases, since yeah, they're all just timers. But IMO, the benefits of labeling the timers individually outweigh the drawbacks. Feel free to disagree, though.

Dragorn421: I would be way less on the fence with this if there weren't other timers outside the union like `sunsSongStunTimer`, which make it more of a pain to remember which timers are in the union

Even in the spots where a union makes sense because a timer is reused as a non-timer-like value, I'm still doubting a union was used originally, because it really saves only bytes of memory (nearly nothing), and if they had been worried about memory usage that much there was a lot of more important things to change that would matter way more

That said I do agree it's nicer for reading to know what timer is being used. So if others are ok with it I'd rather see something like:

```suggestion
    /* 0x030C */ union {
                    s16 animationJudder;
                    s16 death;
                    s16 coffinRiseForwardAccel;
                } timer;
```

which rules out confusion about which timers are in the union imo

tom-overton: > I'm still doubting a union was used originally, because it really saves only bytes of memory (nearly nothing), and if they had been worried about memory usage that much there was a lot of more important things to change that would matter way more

We can't really say for certain what they were thinking. Maybe one dev really cared about the memory usage of their actor. Even if the engine as a whole is inefficient with memory, that doesn't necessarily rule out that an individual piece of code might not be.

Yeah, I see what you mean about it being confusing to remember what timers are in the union. MM's version of Rd adds even *more* timers, including one that shares the same memory as the rotational velocity for one of the Redead's dances (this is what prompted me to start using `union` in the first place). I am totally fine with what you proposed here, so long as others are okay with it too.

src/overlays/actors/ovl_En_Rd/z_en_rd.c
```diff
@@ -515,273 +572,280 @@ void func_80AE3454(EnRd* this, GlobalContext* globalCtx) {
...
+
+    EnRd_SetupAction(this, EnRd_Stunned);
 }
 
-void func_80AE3ECC(EnRd* this, GlobalContext* globalCtx) {
-    if ((this->unk_318 != 0) && (this->unk_316 != 0)) {
-        this->unk_316--;
-        if (this->unk_316 >= 0xFF) {
+void EnRd_Stunned(EnRd* this, GlobalContext* globalCtx) {
+    if ((this->stunnedBySunsSong) && (this->sunsSongStunTimer != 0)) {
```
Dragorn421, line 759:
```suggestion
    if (this->stunnedBySunsSong && (this->sunsSongStunTimer != 0)) {
```

src/overlays/actors/ovl_En_Rd/z_en_rd.c
```diff
@@ -515,273 +572,280 @@ void func_80AE3454(EnRd* this, GlobalContext* globalCtx) {
...
-void func_80AE37BC(EnRd* this) {
-    Animation_Change(&this->skelAnime, &object_rd_Anim_004F94, 0.0f, 0.0f,
-                     Animation_GetLastFrame(&object_rd_Anim_004F94), ANIMMODE_ONCE, 0.0f);
-    this->unk_31B = 7;
-    EnRd_SetupAction(this, func_80AE3834);
+void EnRd_SetupAttemptPlayerFreeze(EnRd* this) {
+    Animation_Change(&this->skelAnime, &gGibdoRedeadLookBackAnim, 0.0f, 0.0f,
+                     Animation_GetLastFrame(&gGibdoRedeadLookBackAnim), ANIMMODE_ONCE, 0.0f);
+    this->action = EN_RD_ACTION_ATTEMPTING_PLAYER_STUN;
+    EnRd_SetupAction(this, EnRd_AttemptPlayerFreeze);
```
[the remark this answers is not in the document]
tom-overton: Yeah this came from when I was working on the actor in MM; `EnRd_AttemptingPlayerStun` was the old name for this function, and I just never got around to updating the enum.

src/overlays/actors/ovl_En_Rd/z_en_rd.h
```diff
@@ -8,29 +8,75 @@ struct EnRd;
...
+
+typedef enum {
+    /* -3 */ EN_RD_TYPE_GIBDO_RISING_OUT_OF_COFFIN = -3,
+    /* -2 */ EN_RD_TYPE_GIBDO = -2,
+    /* -1 */ EN_RD_TYPE_DOES_NOT_MOURN = -1,
+    /*  0 */ EN_RD_TYPE_DOES_NOT_MOURN_IF_WALKING = 0,
+    /*  1 */ EN_RD_TYPE_REGULAR,
+    /*  2 */ EN_RD_TYPE_CRYING,
+    /*  3 */ EN_RD_TYPE_INVISIBLE
+} EnRdType;
```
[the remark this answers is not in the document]
tom-overton: Had to cast it like this to get it to match:
```c
0, this->dyna.actor.shape.rot.y, 0, (u8)EN_RD_TYPE_GIBDO_RISING_OUT_OF_COFFIN);
```
Not sure if it's better to do that, or to make these defines instead. But I'll push this up just to show what it looks like.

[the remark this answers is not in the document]
Dragorn421: hm so the redead actor expects the lower byte as a s8, since it sign extends it to a s16 here:
```c
    if (this->actor.params & 0x80) {
        this->actor.params |= 0xFF00;
    } else {
        this->actor.params &= 0xFF;
    }
```
I think a (u8) cast is ok

there may be something more involved making more sense like `((s16)EN_RD_TYPE_GIBDO_RISING_OUT_OF_COFFIN) & 0xFF` idk, but I'm not convinced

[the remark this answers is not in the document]
tom-overton: I experimented with defining something like so:
```c
#define EN_RD_TYPE_COFFIN 0xFD
```
But when you try to use that in EnRd, that fails:
image

So I think what I have now is *probably* the best we'll get? I'll leave it up to you, though.

src/overlays/actors/ovl_En_Rd/z_en_rd.c
```diff
@@ -173,83 +205,97 @@ void EnRd_Destroy(Actor* thisx, GlobalContext* globalCtx) {
...
-    while (enemyIt != NULL) {
-        if ((enemyIt->id != ACTOR_EN_RD) || (enemyIt == thisx) || (enemyIt->params < 0)) {
-            enemyIt = enemyIt->next;
+/**
+ * This function does two things depending on whether shouldMourn is true or false.
+ * - If shouldMourn is true, this function sets thisx to be the parent for all Redeads
+ *   in the area that are capable of mourning. This is used right when thisx first
+ *   dies to make the other Redeads mourn it.
+ * - If shouldMourn is false, this function nulls out the parent for all Redeads in the
+ *   are whose parents is thisx. This is used when thisx is fading away to make the
```
Dragorn421, line 218: just a typo I didn't notice before
```suggestion
 *   area whose parent is thisx. This is used when thisx is fading away to make the
```

src/overlays/actors/ovl_En_Rd/z_en_rd.h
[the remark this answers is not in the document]
tom-overton: Sorry, didn't refresh to see your comment. `((s16)EN_RD_TYPE_GIBDO_RISING_OUT_OF_COFFIN) & 0xFF` does actually work, but for now I think I'll leave it as a `u8` cast. I think if other `Actor_Spawns` do weird stuff like this, then maybe it makes sense to make it semi-standard, but I'll keep it simple for now.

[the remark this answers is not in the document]
Dragorn421: I don't think there's a way to get rid of the assumptions made by the code in EnRd on how negative numbers are represented, so the simplest u8 cast also making those assumptions looks good to me

src/overlays/actors/ovl_En_Rd/z_en_rd.c
tom-overton: Good catch, though I think the grammatically correct thing to say is "all Redead*s* in the area whose parent*s* *are* thisx", since we're referring to multiple redeads. It's spread across two lines so it's difficult to see, and the "are" that's there right now is just totally wrong.

Dragorn421: Sure, I was indeed hesitating

src/overlays/actors/ovl_En_Rd/z_en_rd.c
```diff
@@ -173,83 +205,97 @@ void EnRd_Destroy(Actor* thisx, GlobalContext* globalCtx) {
...
-            enemyIt = enemyIt->next;
+/**
+ * This function does two things depending on whether shouldMourn is true or false.
+ * - If shouldMourn is true, this function sets thisx to be the parent for all Redeads
+ *   in the area that are capable of mourning. This is used right when thisx first
+ *   dies to make the other Redeads mourn it.
+ * - If shouldMourn is false, this function nulls out the parent for all Redeads in the
+ *   area whose parents are thisx. This is used when thisx is fading away to make the
+ *   other Redeads stop mourning over it.
+ */
```
fig02, line 220:
```
/**
 * The `parent` pointer if updated for all currently loaded redeads depending on the `shouldMourn` variable.
 *
 * If `shuoldMourn` is true, the parent of all other Redeads is set to this instance so they will mourn
 * this Redead.
 * 
 * If `shouldMourn` is false, the parent of all other Redeads is cleared so that they stop mourning.
 * This is done when the dead redead starts fading away
 */
 ```
Can add or tweak as needed, but imo I like to avoid references to actual variable names/code constructs in comments. At least when possible. So referring to "thisx" as "this instance" for example. The fact that its thisx is not important imo

File: src/overlays/actors/ovl_En_Rd/z_en_rd.c [hunk 9]
```diff
@@ -463,39 +516,43 @@ void func_80AE3260(EnRd* this, GlobalContext* globalCtx) {
...
-        case 2:
+            this->grabDamageTimer = 20;
+            // fallthrough
+
+        case EN_RD_GRAB_START:
+            Math_SmoothStepToS(&this->headYRotation, 0, 1, 0x5DC, 0);
+            Math_SmoothStepToS(&this->upperBodyYRotation, 0, 1, 0x5DC, 0);
+            // fallthrough
+
+        case EN_RD_GRAB_ATTACK:
```
fig02, line 550: we havent standardized switch styling at all, but imo fallthroughs look nice without a line break between the cases. it gets the point across of the fall through much more. Then cases that break properly can have line breaks to help emphasize that fact

src/overlays/actors/ovl_En_Rd/z_en_rd.h
```diff
@@ -8,29 +8,75 @@ struct EnRd;
...
-    /* 0x0319 */ u8 unk_319;
-    /* 0x031A */ u8 unk_31A;
-    /* 0x031B */ u8 unk_31B;
-    /* 0x031C */ u8 unk_31C;
-    /* 0x031D */ u8 unk_31D;
+    /* 0x030C */ union {
+                    s16 animationJudder;
+                    s16 death;
+                    s16 coffinRiseForwardAccel;
+                } timer;
```
fig02, line 68: we dont generally use unions soley for the purpose of giving a variable different names. At least its something we've actively avoided in the past, but iirc @Roman971 feels more strongly about that

fig02 (commented):
sorry for the very long delay on this review

src/overlays/actors/ovl_En_Rd/z_en_rd.h
```diff
@@ -8,29 +8,75 @@ struct EnRd;
 
 typedef void (*EnRdActionFunc)(struct EnRd*, GlobalContext*);
 
+#define EN_RD_GET_FLAGS(thisx) (((thisx)->params & 0xFF00) >> 8)
+
```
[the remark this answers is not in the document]
fig02: and all the enum prefixes as well

src/overlays/actors/ovl_En_Rd/z_en_rd.h
Roman971: Yeah I'm personally not a fan of using unions for naming purposes since it could be misleading very quickly and lead to strange bugs in modding for instance. It's not as bad if there is a name for the union itself (ie. the `timer` part here), but it can still be confused with a struct named `timer` for example, so it's not ideal.

Here I would just go for a generic `timer` variable, but a union is probably okay if we consider the specific names important enough

EllipticEllipsis: Other actors use defines for this sort of situation (EnTite and BossSst come to mind, there's probably a few more).

fig02: in this specific case I would probably just say a variable named `timer` is best, its what we do in many other actors. generic timers are expected to be action dependent id say

src/overlays/actors/ovl_En_Rd/z_en_rd.h
```diff
@@ -8,29 +8,75 @@ struct EnRd;
 
 typedef void (*EnRdActionFunc)(struct EnRd*, GlobalContext*);
 
+#define EN_RD_GET_FLAGS(thisx) (((thisx)->params & 0xFF00) >> 8)
```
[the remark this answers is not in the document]
fig02: imo we can start adding them, now that we are in documentation oriented phases.
Is there anything you want to discuss about them before we start implementing them in actors

src/overlays/actors/ovl_En_Rd/z_en_rd.h
tom-overton: In this specific case, I'll relent and use `timer`. However, it's very, very clear that this approach will not scale with every actor; take a look at Obj_Tokeidai in MM and try to figure out how you would document all that without a union or a define.

src/overlays/actors/ovl_En_Rd/z_en_rd.h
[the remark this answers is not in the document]
tom-overton: I'm cool with either approach, so just tell me what you want and I'll do it.

assets/xml/objects/object_rd.xml
```diff
@@ -1,106 +1,124 @@
...
+        <Animation Name="gGibdoRedeadGrabAttackAnim" Offset="0x4268"/> <!-- Original name is "re_dead_attack" -->
+        <Animation Name="gGibdoRedeadGrabEndAnim" Offset="0x46F8"/> <!-- Original name is "re_dead_attack_end" -->
+        <Animation Name="gGibdoRedeadGrabStartAnim" Offset="0x4ADC"/> <!-- Original name is "re_dead_attack_start" -->
+        <Animation Name="gGibdoRedeadLookBackAnim" Offset="0x4F94"/> <!-- Original name is "re_dead_back_turn" -->
+        <Animation Name="gGibdoRedeadWipingTearsAnim" Offset="0x57AC"/> <!-- Original name is "re_dead_cry" -->
+        <Animation Name="gGibdoRedeadSobbingAnim" Offset="0x5D98"/> <!-- Original name is "re_dead_cryB" -->
+        <Animation Name="gGibdoRedeadDeathAnim" Offset="0x6E88"/> <!-- Original name is "re_dead_down" -->
+        <Animation Name="gGibdoRedeadDamageAnim" Offset="0x74F0"/> <!-- Original name is "re_dead_hit" -->
+        <Animation Name="gGibdoRedeadStandUpAnim" Offset="0x8040"/> <!-- Original name is "re_dead_standing" -->
+        <Animation Name="gGibdoRedeadIdleAnim" Offset="0x87D0"/> <!-- Original name is "re_dead_wait" -->
```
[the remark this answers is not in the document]
tom-overton: I like the information they provide, esp. for objects that have really ambiguous or difficult-to-name animations. Dev names are oftentimes not super great, but in other times they can really help explain what they were thinking when they made them. Of course, this actor has super-plain names, so they don't help much. I'd be fine removing them, but I also don't see the harm of them at all.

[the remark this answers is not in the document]
tom-overton: I'm happy to sit down and do them for other objects if it's desired. Would certainly be A Project, but one that I would be willing to do.

src/overlays/actors/ovl_En_Rd/z_en_rd.c
```diff
@@ -534,25 +521,21 @@ void EnRd_Grab(EnRd* this, GlobalContext* globalCtx) {
...
+        case REDEAD_GRAB_INITIAL_DAMAGE:
             Animation_PlayLoop(&this->skelAnime, &gGibdoRedeadGrabAttackAnim);
             this->grabState++;
             globalCtx->damagePlayer(globalCtx, -8);
             func_800AA000(this->actor.xzDistToPlayer, 0xFF, 1, 0xC);
             this->grabDamageTimer = 20;
-            // fallthrough
-
-        case EN_RD_GRAB_START:
+        case REDEAD_GRAB_START:
```
Dragorn421, line 530: I think `// fallthrough` comments can/should stay, I think @fig02 only refered to the empty line after it being unneeded but I may be wrong

fig02: right, yeah sorry if that wasnt clear. Comments definitely help to make the fall through clear. I just like keeping the cases bunched together when they fall through into each other without a line break

assets/xml/objects/object_rd.xml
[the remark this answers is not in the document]
Dragorn421: imo assets xmls are so inconsistent right now (ordered by offsets or grouped by type, comments, explicit/implicit offsets, missing resources, not-so-good names, being undocumented...), that it doesn't matter to have quality info in one xml like this but not the others.

fig02 (approved):

Dragorn421 (approved):

Roman971 (approved):

PR #1180, opened 2022-03-16
Fixup `feetFloorFlags` doc and -> `feetFloorFlag`

Dragorn421:
From #1150 

See discussion in #1150 : https://github.com/zeldaret/oot/pull/1150#discussion_r810626523


Roman971 (approved):
I was originally against this change but I guess the idea is to name this field based on what it is "normally", so it fits when looking at the field from an actor or something, and I suppose it's fair to name it based on that

Dragorn421:
Yes it only represents a bitfield temporarily, for ~60 lines in `ActorShadow_DrawFeet`

PR #1181, opened 2022-03-16
Document more PathwayLists for scenes

Dragorn421:
Original PR: #1134 by mrneo240

Add `<Path>` nodes in all (I think) scene xmls


fig02 (approved):

Roman971 (approved):
On a side note, "Path" seems like the wrong name for this XML node since it describes a list of paths (and it even has an attribute called "NumPaths")

PR #1183, opened 2022-03-20
Fix misc 9

Dragorn421:


src/overlays/actors/ovl_Demo_Tre_Lgt/z_demo_tre_lgt.c
```diff
@@ -147,6 +147,11 @@ s32 DemoTreLgt_PostLimbDraw(GlobalContext* globalCtx, SkelAnimeCurve* skelCurve,
     }
 
     CLOSE_DISPS(globalCtx->state.gfxCtx, "../z_demo_tre_lgt.c", 448);
+
+    //! @bug missing return
+    // If the return value ends up being false (0), the limb won't draw (meaning no limb at all will draw).
+    // In MQ Debug, `Graph_CloseDisps` has the last instruction writing to v0 before this function ends.
+    // That instruction sets v0 to a non-NULL pointer, which is "true", so the limbs get drawn.
```
fig02, line 154: is skelcurve ovveride limb draw different?
normally, return false means that it *wont* override/null the dlist, and true will make it not draw

ill look at the code for it myself to check later, but curious if you know already

Dragorn421: I didn't notice but yeah apparently the return value is "reversed", nice catch

"normal" skelanime https://github.com/zeldaret/oot/blob/4729ac131a1f3d8ee28f59c41ed8792f51b09040/src/code/z_skelanime.c#L693
skelcurve https://github.com/zeldaret/oot/blob/4729ac131a1f3d8ee28f59c41ed8792f51b09040/src/code/z_fcurve_data_skelanime.c#L104

(so this comment is correct as far as "return true = draw" is concerned)

fig02: alright cool
we should prob avoid_ub this, apparently this was broken in gcc builds the whole time and i never noticed

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -2019,7 +2019,7 @@ s32 func_80834380(GlobalContext* globalCtx, Player* this, s32* itemPtr, s32* typ
         if (this->stateFlags1 & PLAYER_STATE1_23) {
             *typePtr = ARROW_NORMAL_HORSE;
         } else {
-            *typePtr = this->heldItemActionParam - 6;
+            *typePtr = this->heldItemActionParam - (PLAYER_AP_BOW - ARROW_NORMAL);
```
Roman971, line 2022: Pretty minor but I think it would be easier to understand if it was written like this: (base value + offset)
```suggestion
            *typePtr = ARROW_NORMAL + this->heldItemActionParam - PLAYER_AP_BOW;
```

include/z64.h
```diff
@@ -1403,7 +1403,7 @@ typedef enum {
     /* 16 */ F_B8
 } FloorID;
 
-// All arrays pointed in this struct are indexed by "map indexes"
+// All arrays pointed in this struct are indexed by "map indices"
```
Roman971, line 1406: I'm in the indexes team, but that's just personal preference and I understand why we would favour indices instead

src/overlays/actors/ovl_player_actor/z_player.c
Dragorn421: I didn't know this kind of thing didn't break matching 👍

include/z64.h
Dragorn421: as explained in the commit message I went with "indices" because it currently prevails, and seems vaguely more appropriate with respect to common usage, but we could also change "indices" to "indexes" instead (I would be personally defending "indices")

Roman971 (approved):

fig02 (approved):

PR #1184, opened 2022-03-20
Document EnInsect actor

Dragorn421:
Original PR: #1125


src/overlays/actors/ovl_En_Insect/z_en_insect.c
```diff
@@ -10,30 +10,39 @@
...
+void EnInsect_SetupCaught(EnInsect* this);
+void EnInsect_Caught(EnInsect* this, GlobalContext* globalCtx);
+void EnInsect_SetupDig(EnInsect* this);
+void EnInsect_Dig(EnInsect* this, GlobalContext* globalCtx);
+void EnInsect_SetupWalkOnWater(EnInsect* this);
+void EnInsect_WalkOnWater(EnInsect* this, GlobalContext* globalCtx);
+void EnInsect_SetupDrown(EnInsect* this);
+void EnInsect_Drown(EnInsect* this, GlobalContext* globalCtx);
+void EnInsect_SetupDropped(EnInsect* this);
+void EnInsect_Dropped(EnInsect* this, GlobalContext* globalCtx);
```
Dragorn421, line 35: @mzxrules didn't like "Dropped" but idk what to rename this to

src/overlays/actors/ovl_En_Insect/z_en_insect.h
```diff
@@ -17,16 +27,23 @@ typedef struct EnInsect {
...
     /* 0x0324 */ f32 unk_324;
     /* 0x0328 */ s16 unk_328;
     /* 0x032A */ u8 unk_32A;
 } EnInsect; // size = 0x032C
 
+typedef enum {
+    /* 0 */ INSECT_TYPE_PERMANENT,
+    /* 1 */ INSECT_TYPE_SPAWNED,
+    /* 2 */ INSECT_TYPE_FIRST_DROPPED,
+    /* 3 */ INSECT_TYPE_EXTRA_DROPPED // An additional bug spawned by dropping one from a bottle.
```
Dragorn421, line 46: @mzxrules didn't like these names but idk what to use instead

src/overlays/actors/ovl_En_Insect/z_en_insect.c
jpburnett: `EnInsect_Released`?

Dragorn421: @mzxrules please review if you still think this should be changed, I'll mark these as resolved

src/overlays/actors/ovl_En_Insect/z_en_insect.c
```diff
@@ -10,30 +10,39 @@
...
-static s16 D_80A7DEB4 = 0;
-static s16 D_80A7DEB8 = 0;
+/**
+ * The number of bugs caught this frame.
+ */
+static s16 sCaughtCount = 0;
+/**
+ * The number of active dropped bugs.
+ */
+static s16 sDroppedCount = 0;
```
fig02, line 45: these comments are prob not needed? the names seem self explanatory to me

src/overlays/actors/ovl_En_Insect/z_en_insect.c
```diff
@@ -74,7 +83,15 @@ static ColliderJntSphInit sColliderInit = {
...
-static u16 D_80A7DF10[] = { 0, 5, 7, 7 };
+/**
+ * The initial flags for a new bug, depending on its type.
+ */
+static u16 sInitInsectFlags[] = {
+    0,
+    INSECT_FLAG_0 | INSECT_FLAG_IS_SHORT_LIVED,
+    INSECT_FLAG_0 | INSECT_FLAG_1 | INSECT_FLAG_IS_SHORT_LIVED,
+    INSECT_FLAG_0 | INSECT_FLAG_1 | INSECT_FLAG_IS_SHORT_LIVED,
+};
```
fig02, line 94: can prob afford to just be a // comment, its taking up alot of space and looks like a function comment at a glance

src/overlays/actors/ovl_En_Insect/z_en_insect.c
```diff
@@ -148,105 +162,108 @@ s32 EnInsect_FoundNearbySoil(EnInsect* this, GlobalContext* globalCtx) {
...
+/**
+ * Update the crawl sound timer, and play the crawling sound when it reaches 0.
+ */
+void EnInsect_UpdateCrawlSfx(EnInsect* this) {
+    if (this->crawlSoundTimer > 0) {
+        this->crawlSoundTimer--;
         return;
     }
 
     Audio_PlayActorSound2(&this->actor, NA_SE_EN_MUSI_WALK);
```
fig02, line 174: we can maybe name it `crawlSoundDelay`? the current name makes it sound like the sound plays while the timer is non zero

src/overlays/actors/ovl_En_Insect/z_en_insect.c
```diff
@@ -10,30 +10,39 @@
 
 #define FLAGS 0
 
+#define IS_DROPPED(type) ((type) == INSECT_TYPE_FIRST_DROPPED || (type) == INSECT_TYPE_EXTRA_DROPPED)
+
```
fig02, line 14: I dont see this macro as being too beneficial personally, I think id rather the type checks be inline as is

src/overlays/actors/ovl_En_Insect/z_en_insect.h
```diff
@@ -9,6 +9,16 @@ struct EnInsect;
 
 typedef void (*EnInsectActionFunc)(struct EnInsect*, GlobalContext*);
 
+#define INSECT_FLAG_0 (1 << 0)
+#define INSECT_FLAG_1 (1 << 1)
+#define INSECT_FLAG_IS_SHORT_LIVED (1 << 2)
```
fig02, line 14: in general Id be fine with any mention of `insect` being `bug`, so these enums and the instance variables

fig02 (commented):
sorry for the late review on this.
I know most (or even all) of these changes are not yours, so I can do these on my own if you like

src/overlays/actors/ovl_En_Insect/z_en_insect.h
Dragorn421: tbh I'm not a fan of using "bug", solely because of the other meanings in computer science

fig02: i agree its a little weird, but I think context will be enough. no one calls these insect, its always bugs

fig02 (approved):
we can revisit the topic of insect -> bug when we start actually renaming actors

Roman971 (approved):

PR #1187, opened 2022-03-25
Mark bug in file choose

louist103:


src/overlays/gamestates/ovl_file_choose/z_file_nameset_PAL.c
```diff
@@ -46,6 +46,8 @@ void FileChoose_SetKeyboardVtx(GameState* thisx) {
         phi_t0 = -0x60;
 
         for (phi_t1 = 0; phi_t1 < 13; phi_t1++, phi_t3 += 4, phi_t2++) {
+            //! @bug D_80812544 is accessed out of bounds when drawing the empty space character phi_t2 == 64. Under normal
+            //! circumstances it reads a halfword from sNameLabelTextures.
```
fig02, line 50: I think we should try avoiding writing actual code in comments. So instead of `phi_t2 == 64` id just say "empty space character (value of 64)", or something.
you can also start the second sentance on the next line for a bit cleaner formatting, if you like

Roman971 (approved):

PR #1188, opened 2022-03-26
Update README.md

LOuroboros:
The package `python3-pip` is needed in order to install the pip package "colorama" used by this project, yet, the README makes no mention of this. I had to install it in order to build this project on my copy of Ubuntu v20.04 inside WSL2, at the very least, and I've seen other people having to go out of their way to install said package as well.
This PR intends to correct that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,13 +68,14 @@ The build process has the following package requirements:
 * build-essential
 * binutils-mips-linux-gnu
 * python3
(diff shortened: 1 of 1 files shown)


fig02:
we're going to change how our tooling works so that colorama wont be required for building. Appreciate the fix, but we are going to handle it in a slightly different way

PR #1191, opened 2022-03-29
Names for some view things

rogual:
I've used the term "apply" to mean "do actual gSP/gDP stuff based on the data in this view".

`View_Apply` and `View_ApplyTo` seem to use this system:
* The function takes a `mask` saying what to apply.
* If you give the `VIEW_XXX` flags it will use its `dirty` variable to only apply what has been set since the last time the thing was applied.
* If you give the `VIEW_FORCE_XXX` flags then it'll apply things unconditionally, ignoring the dirty flags.

Not sure that's worth documenting in a comment or the best place to do it, since it applies to more than one function.


include/z64.h
```diff
@@ -184,10 +184,22 @@ typedef struct {
...
-    /* 0x0120 */ s32    flags;
+    /* 0x0120 */ s32    dirty;
     /* 0x0124 */ s32    unk_124;
 } View; // size = 0x128
 
+#define VIEW_VIEWING 0x1
+#define VIEW_VIEWPORT 0x2
+#define VIEW_PROJECTION_PERSPECTIVE 0x4
+#define VIEW_PROJECTION_ORTHOGRAPHIC 0x8
+#define VIEW_ALL 0xF
```
Thar0, line 195:
```suggestion
#define VIEW_ALL (VIEW_VIEWING | VIEW_VIEWPORT | VIEW_PROJECTION_PERSPECTIVE | VIEW_PROJECTION_ORTHOGRAPHIC)
```
`VIEW_PROJECTION_ORTHOGRAPHIC` could possibly be worth shortening, since we use `ortho` everywhere else currently. Less sure but `VIEW_PROJECTION_PERSPECTIVE` could be shortened to `persp` possibly, although that's used less than `ortho` is.

include/z64.h
```diff
@@ -184,10 +184,22 @@ typedef struct {
...
+
+#define VIEW_FORCE_VIEWING 0x10
+#define VIEW_FORCE_VIEWPORT 0x20
+#define VIEW_FORCE_PROJECTION_PERSPECTIVE 0x40
+#define VIEW_FORCE_PROJECTION_ORTHOGRAPHIC 0x80
+#define VIEW_FORCE_ALL 0xF0
```
Thar0, line 201: This one should probably also be `|` of the above.

include/z64.h
```diff
@@ -184,10 +184,22 @@ typedef struct {
...
+#define VIEW_VIEWING 0x1
+#define VIEW_VIEWPORT 0x2
+#define VIEW_PROJECTION_PERSPECTIVE 0x4
+#define VIEW_PROJECTION_ORTHOGRAPHIC 0x8
+#define VIEW_ALL 0xF
+
+#define VIEW_FORCE_VIEWING 0x10
+#define VIEW_FORCE_VIEWPORT 0x20
+#define VIEW_FORCE_PROJECTION_PERSPECTIVE 0x40
+#define VIEW_FORCE_PROJECTION_ORTHOGRAPHIC 0x80
```
Thar0, line 200: Since the code shifts these right by 4 to get the un-forced flags, there's an intrinsic relationship between their values and changing one define in one set but not the other can cause unintended side-effects. I suggest defining these as the other flag define << 4 to make this totally clear.

include/z64.h
```diff
@@ -184,10 +184,22 @@ typedef struct {
     /* 0x0104 */ Vec3f  curDistortionOrientation;
     /* 0x0110 */ Vec3f  curDistortionScale;
     /* 0x011C */ u16    normal; // used to normalize the projection matrix
-    /* 0x0120 */ s32    flags;
+    /* 0x0120 */ s32    dirty;
```
Thar0, line 187: `dirtyFlags`? Might feel redundant but I think it's nice to still indicate it's a set of flags.

include/functions.h
```diff
@@ -1364,14 +1364,14 @@ void View_InitDistortion(View* view);
...
-s32 func_800AB9EC(View* view, s32 arg1, Gfx** p);
-s32 func_800ABE74(f32 eyeX, f32 eyeY, f32 eyeZ);
+void View_Apply(View* view, s32 mask);
+s32 View_ApplyPerspective(View* view);
+s32 View_ApplyOrtho(View* view);
+s32 View_ApplyToOverlayOrtho(View* view);
+s32 View_ApplyToOverlayPerspective(View* view);
+s32 View_UpdateViewingMatrix(View* view);
+s32 View_ApplyTo(View* view, s32 arg1, Gfx** p);
+s32 View_SanityCheckEyePosition(f32 eyeX, f32 eyeY, f32 eyeZ);
```
Thar0, line 1374: Purely a preference suggestion: `ErrorCheck`?

include/functions.h
```diff
@@ -1347,14 +1347,14 @@ void func_800AA178(u32);
 View* View_New(GraphicsContext* gfxCtx);
 void View_Free(View* view);
 void View_Init(View*, GraphicsContext*);
-void func_800AA358(View* view, Vec3f* eye, Vec3f* lookAt, Vec3f* up);
-void func_800AA3F0(View* view, Vec3f* eye, Vec3f* lookAt, Vec3f* up);
+void View_LookAt(View* view, Vec3f* eye, Vec3f* lookAt, Vec3f* up);
```
Thar0, line 1350: This updates more than the `lookAt` position so the name may be slightly odd at a glance, although I see the parallel between this and `gluLookAt` from OpenGL and `guLookAt(F)` from libultra which do the same thing. I don't necessarily suggest any changes but it may be worth pondering.

Thar0 (changes requested):
Nice job sorting out the flags and remaining function names.
Some suggestions/comments

include/z64.h
rogual: Nice 👍

src/code/z_view.c
```diff
@@ -51,80 +51,83 @@ void View_Init(View* view, GraphicsContext* gfxCtx) {
...
 
     view->eye = *eye;
     view->lookAt = *lookAt;
     view->up = *up;
-    view->flags |= 1;
+    view->dirty |= VIEW_VIEWING;
 }
 
-void func_800AA3F0(View* view, Vec3f* eye, Vec3f* lookAt, Vec3f* up) {
+// Unused. Maybe inlined into View_LookAt?
```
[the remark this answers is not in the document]
rogual: Interesting! Yeah, no idea really, just trying to think why there might be a separate LookAt that doesn't do the error correction or set the dirty flag.

src/code/z_view.c
```diff
@@ -255,17 +258,19 @@ s32 View_StepDistortion(View* view, Mtx* projectionMtx) {
     return true;
 }
 
-void func_800AAA50(View* view, s32 arg1) {
-    arg1 = (view->flags & arg1) | (arg1 >> 4);
+// Apply view to POLY_OPA_DISP, POLY_XLU_DISP, OVERLAY_DISP
```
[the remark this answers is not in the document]
rogual: 👍 c501306a6

include/z64.h
rogual: Shortened 273f422f2f915be7d78d40f3897aa555345c011f

include/z64.h
rogual: I did try that, but to be honest I kind of prefer this. Couple of reasons:
* I find this much easier to read, and it becomes a bit of a "mess of capital letters" the other way, so much so that to my eye it obscures the relationship it's trying to clarify.
* It's impossible to define any more flags without changing the whole shifting logic anyway.

include/z64.h
rogual: As for the `|`, not sure how strong the suggestion is intended to be, and if it's project policy I will of course just change it, but I personally find the `0xF` version a lot more readable, whereas the other way just starts to look like a fatiguing block of capital letters. It's already called `ALL` and there can only ever be four flags without rewriting the whole system so errors there seem unlikely.

include/functions.h
rogual: Yeah, I did deliberate a bit on this, but I think it's okay. The only other things it's doing are some quick input sanitization and setting the dirty flag, which I feel are acceptable things to do in a setter without explicitly calling them out in the name.

include/z64.h
fig02: seeing that its all of the flags being set together is much more important information to get at a glance, than just seeing some magic number being set. 

this is also why tharos suggestion is pretty important with the other flags being set as a shift of the others. The relationship between these flags is important and must be reflected in the define, over just being a bunch of hex numbers

include/functions.h
```diff
@@ -1347,14 +1347,14 @@ void func_800AA178(u32);
...
+void View_LookAtInternal(View* view, Vec3f* eye, Vec3f* lookAt, Vec3f* up);
```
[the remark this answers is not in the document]
rogual: I really don't want to give this function such an innocuous name. It's unused in the code, and probably no-one should use it, since the code really expects that dirty flag to get set.

include/z64.h
rogual: 35f4fae87

include/z64.h
rogual: Done 35f4fae87

include/functions.h
```diff
@@ -1364,14 +1364,14 @@ void View_InitDistortion(View* view);
...
-s32 func_800AB2C4(View* view);
-s32 func_800AB560(View* view);
-s32 func_800AB944(View* view);
-s32 func_800AB9EC(View* view, s32 arg1, Gfx** p);
-s32 func_800ABE74(f32 eyeX, f32 eyeY, f32 eyeZ);
+void View_Apply(View* view, s32 mask);
+s32 View_ApplyPerspective(View* view);
+s32 View_ApplyOrtho(View* view);
+s32 View_ApplyToOverlayOrtho(View* view);
+s32 View_ApplyToOverlayPerspective(View* view);
```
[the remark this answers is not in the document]
rogual: I think what I was going for was "Apply this view to the overlay (and this is the Ortho version of the function)" but I don't feel that strongly about it.

Changed! 45a25018a

include/functions.h
rogual: I do quite like the slight extra information you get from the `SanityCheck` name. It's specifically checking for enormous values — the eye position isn't invalid, exactly, it's just *insane*. I think this tracks with the general use of "sanity check" to mean checking for values that may not crash or anything but are unlikely to be what you wanted.

include/functions.h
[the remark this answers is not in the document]
rogual: Maybe `Unsafe` or `Unused` if you don't like `Internal`?

src/code/z_view.c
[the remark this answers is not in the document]
fig02: we should prob remove the mention of "inlined" for this reason. just saying its unused is probably enough, and you can elaborate on the fact that the error checking version already does the same thing

include/functions.h
Thar0: I see, I'm used to viewing numbers outside of [-32768, 32767] as more of an error than just a bogus value, since the geometry transformation code that runs on the RSP processor actually cannot handle values outside of this range properly, they would get truncated when loading into the 8x16b vector registers. This still doesn't crash though, so it's probably not any worse than what you already get with such large eye values, so it can go either way really.

fig02: that sounds like enough of a reason to call it an error to me either way though, if the rsp cant handle those numbers

include/functions.h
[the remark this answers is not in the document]
Thar0: I like `Internal` over `Unused`, calling something "unused" is to me the absolute last resort 😅 
`Unsafe` could definitely make sense though and would still communicate well that it's not really supposed to be used under any normal circumstances.

include/functions.h
rogual: Didn't know that about the RSP! Will change.

include/functions.h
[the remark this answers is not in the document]
EllipticEllipsis: Whatever it's called, it sounds like it shouldn't be in functions.h any more anyway.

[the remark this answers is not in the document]
fig02: unsafe means more to me than internal does, if that helps with deciding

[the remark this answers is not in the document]
EllipticEllipsis: Well, if it's not used externally, and we think it shouldn't be used externally, that suggests that not making it available externally is appropriate.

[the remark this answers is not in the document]
fig02: dunno, it seems to me like it was an alternate version of the call that existed to be used externally, but they ended up just using the safe version for everything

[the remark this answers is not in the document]
rogual: I wasn't sure if we ever took things out of functions.h so I left it in; would probably favour taking it out.

I really don't think anyone has a reason to call this function, as it probably won't do what they expect and there is a working version right next to it.

People seem to like `Unsafe`, so, renamed: e1eb2b03f

include/functions.h
rogual: 1719252db

include/functions.h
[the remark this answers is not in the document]
rogual: Yeah, that makes sense 👍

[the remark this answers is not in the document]
EllipticEllipsis: functions.h will be broken up eventually, though: it's horrible to open that file in VSCode with a small amount of ram because of its size and cross-references. And it's just bad organisationally to have such a huge header anyway.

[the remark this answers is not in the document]
fig02: whether or not if this particular function was meant to be internal or external is certainly debatable, cause of how little info we have. But I would personally lean on the side of intended to be external, cause  i dont see them making an internal helper function for this one thing. But its a gut feeling thing, dont really have any evidence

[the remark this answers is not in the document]
fig02: ofc yeah, it will be broken up when each system has its own header. but for now you can just view functions.h as the declarations for externally available functions

src/code/z_view.c
```diff
@@ -255,17 +263,23 @@ s32 View_StepDistortion(View* view, Mtx* projectionMtx) {
     return true;
 }
 
-void func_800AAA50(View* view, s32 arg1) {
-    arg1 = (view->flags & arg1) | (arg1 >> 4);
+/**
+ * Apply view to POLY_OPA_DISP, POLY_XLU_DISP, OVERLAY_DISP
+ */
+void View_Apply(View* view, s32 mask) {
```
Dragorn421, line 269: only `View_ApplyOrtho` appends to overlay_disp, and even then only sets the viewport. so I guess overlay_disp may be omitted here.

It doesn't matter for overlay_disp anyway since it's meant to be used to draw 2d rectangles (*) in screen coordinates, bypassing the viewport (and projection) entirely if I understand it right

(*) or so I thought, I looked at what uses `View_ApplyOrthoToOverlay` since it wouldn't make sense to set a viewport on overlay_disp then, and actually some interface/pause menu code does draw "normal" geometry with vertices instead of 2d rectangles to overlay_disp. So now idk

src/code/z_view.c
```diff
@@ -403,7 +424,10 @@ s32 func_800AB0A8(View* view) {
     return 1;
 }
 
-s32 func_800AB2C4(View* view) {
+/**
+ * Apply scissor, viewport and projection (ortho) to OVERLAY_DISP.
+ */
+s32 View_ApplyOrthoToOverlay(View* view) {
```
Dragorn421, line 430: I would go for an even longer name and name this `View_ApplyOrthoToOverlayDisp` to prevent confusion with a code overlay

src/code/z_view.c
```diff
@@ -438,7 +462,10 @@ s32 func_800AB2C4(View* view) {
     return 1;
 }
 
-s32 func_800AB560(View* view) {
+/**
+ * Apply scissor, viewport, view and projection (perspective) to OVERLAY_DISP.
+ */
+s32 View_ApplyPerspectiveToOverlay(View* view) {
```
Dragorn421, line 468: `View_ApplyOrthoToOverlay` says `Apply scissor, viewport and projection (ortho) to OVERLAY_DISP.`

I'm trying to figure out if the extra element in the `View_ApplyPerspectiveToOverlay` comment "`view`" is a typo or if there's actually a difference between the two ortho/persp variants?

include/z64.h
```diff
@@ -184,10 +184,22 @@ typedef struct {
...
+#define VIEW_VIEWPORT (1 << 1)
+#define VIEW_PROJECTION_PERSPECTIVE (1 << 2)
+#define VIEW_PROJECTION_ORTHO (1 << 3)
+#define VIEW_ALL (VIEW_VIEWING | VIEW_VIEWPORT | VIEW_PROJECTION_PERSPECTIVE | VIEW_PROJECTION_ORTHO)
+
+#define VIEW_FORCE_VIEWING (VIEW_VIEWING << 4)
+#define VIEW_FORCE_VIEWPORT (VIEW_VIEWPORT << 4)
+#define VIEW_FORCE_PROJECTION_PERSPECTIVE (VIEW_PROJECTION_PERSPECTIVE << 4)
+#define VIEW_FORCE_PROJECTION_ORTHO (VIEW_PROJECTION_ORTHO << 4)
+#define VIEW_FORCE_ALL (VIEW_FORCE_VIEWING | VIEW_FORCE_VIEWPORT | VIEW_FORCE_PROJECTION_PERSPECTIVE | VIEW_FORCE_PROJECTION_ORTHO)
```
Dragorn421, line 201: this define doesn't make sense, using both of `VIEW_FORCE_PROJECTION_PERSPECTIVE | VIEW_FORCE_PROJECTION_ORTHO` is contradictory

maybe `#define VIEW_FORCE_ALL_PERSPECTIVE (VIEW_FORCE_VIEWING | VIEW_FORCE_VIEWPORT | VIEW_FORCE_PROJECTION_PERSPECTIVE)` would be useful though. it could replace some `|`s. But I'm not a fan of how "all_perspective" looks

Dragorn421 (commented):
Very nice job

Dragorn421 (commented):
Also I'm not sure about the name for `dirtyFlags`

First, these flags are used externally to z_view which is weird for something that seems as sensitive as state-indicating flags.
If they were "dirty flags" I feel like it would have made more sense for them to be set in view_init and through functions rather than directly

Also, these flags are not reset in all functions, for example they aren't reset by calling `View_Apply` (unless I missed it). And I would expect it to, to reflect the undirty'ed state if that makes sense.

include/z64.h
```diff
@@ -184,10 +184,22 @@ typedef struct {
     /* 0x0104 */ Vec3f  curDistortionOrientation;
     /* 0x0110 */ Vec3f  curDistortionScale;
     /* 0x011C */ u16    normal; // used to normalize the projection matrix
-    /* 0x0120 */ s32    flags;
+    /* 0x0120 */ s32    dirtyFlags;
```
engineer124, line 187: Where does `dirty` come from? Is that a term used in computer graphics/game engine? It's not something I've heard of before

src/code/z_view.c
```diff
@@ -51,80 +51,88 @@ void View_Init(View* view, GraphicsContext* gfxCtx) {
...
-    view->flags |= 1;
+    view->dirtyFlags |= VIEW_VIEWING;
 }
 
-void func_800AA3F0(View* view, Vec3f* eye, Vec3f* lookAt, Vec3f* up) {
+/*
+ * Unused. View_LookAt is always used instead. This version is similar but
+ * is missing the input sanitization and the update to the dirtyFlags.
+ */
+void View_LookAtUnsafe(View* view, Vec3f* eye, Vec3f* lookAt, Vec3f* up) {
```
engineer124, line 80: What is this concept of sanitazation
```suggestion
/*
 * Unused. View_LookAt is always used instead. This version is similar but
 * is missing the input sanitization and the update to the dirtyFlags.
 */
void View_SetEyeAtUp(View* view, Vec3f* eye, Vec3f* lookAt, Vec3f* up) {
```
Could this not be called this? I'm not sure about this whole concept of input sanitization and security. It seems like a bit of a stretch to interpret it this way

src/code/z_view.c
```diff
@@ -51,80 +51,88 @@ void View_Init(View* view, GraphicsContext* gfxCtx) {
...
     view->eye = *eye;
     view->lookAt = *lookAt;
     view->up = *up;
```
engineer124, line 83:
```suggestion
    view->eye = *eye;
    view->at = *at;
    view->up = *up;
```
As an fyi, camera uses the term `at` and "viewing" should be made consistent. We've had the discussion before about `at` vs `lookAt` and concluded that we should be using `at`, but of course that conversation can always be opened again. But it should be consistent.

src/code/z_view.c
```diff
@@ -51,80 +51,88 @@ void View_Init(View* view, GraphicsContext* gfxCtx) {
...
+    view->dirtyFlags |= VIEW_PROJECTION_PERSPECTIVE;
     view->scale = scale;
 }
 
 void View_GetScale(View* view, f32* scale) {
     *scale = view->scale;
 }
 
-void func_800AA460(View* view, f32 fovy, f32 near, f32 far) {
+void View_SetPerspective(View* view, f32 fovy, f32 near, f32 far) {
```
engineer124, line 95:
```suggestion
void View_SetPerspective(View* view, f32 fovy, f32 zNear, f32 zFar) {
```
This function name makes me think this would return the projection matrix, not the parameters involved in projection matrix calculations. The names I thought of were either:
`_Get/SetPerspectiveFovZNearZFar` or alternatively `_Get/SetFovZNearZFar`

src/code/z_view.c
```diff
@@ -51,80 +51,88 @@ void View_Init(View* view, GraphicsContext* gfxCtx) {
...
 
     view->unk_124 = 0;
-    view->flags = 1 | 2 | 4;
+    view->dirtyFlags = VIEW_VIEWING | VIEW_VIEWPORT | VIEW_PROJECTION_PERSPECTIVE;
     View_InitDistortion(view);
 }
 
-void func_800AA358(View* view, Vec3f* eye, Vec3f* lookAt, Vec3f* up) {
+void View_LookAt(View* view, Vec3f* eye, Vec3f* lookAt, Vec3f* up) {
     if (eye->x == lookAt->x && eye->z == lookAt->z) {
```
engineer124, line 66:
```suggestion
void View_SetViewingEyeAtUp(View* view, Vec3f* eye, Vec3f* at, Vec3f* up) {
    if (eye->x == at->x && eye->z == at->z) {
```
What about something like this?

src/code/z_view.c
```diff
@@ -51,80 +51,88 @@ void View_Init(View* view, GraphicsContext* gfxCtx) {
...
-    view->flags |= 2;
+    view->dirtyFlags |= VIEW_VIEWPORT;
 }
 
 void View_GetViewport(View* view, Viewport* viewport) {
     *viewport = view->viewport;
 }
 
-void func_800AA550(View* view) {
+void View_ApplyScissor(View* view) {
```
engineer124, line 135:
```suggestion
void View_ApplyShrinkWindow(View* view) {
```
Scissor is the specific way a shrink-window is applied, but many sub-systems within the game can be scissored. There's an entire file called shrink_window.c that is used to determine how the window shrinks, and this is the function that applies it

src/code/z_view.c
```diff
@@ -276,16 +290,18 @@ s32 func_800AAA9C(View* view) {
...
     View_ViewportToVp(vp, &view->viewport);
     view->vp = *vp;
 
-    func_800AA550(view);
+    View_ApplyScissor(view);
 
     gSPViewport(POLY_OPA_DISP++, vp);
     gSPViewport(POLY_XLU_DISP++, vp);
 
+    // Apply perspective projection
```
engineer124, line 304: I don't know if "Apply" is appropriate here. First, this allocates space on the heap to store the projection matrix. Then it saves the pointer to it in the appropriate variables. Then, `guPerspective` is where the perspective matrix is actually calculated. "Applying" makes me think the projection is already obtained, and is being "applied" to a different matrix to transform it into the appropriate coordinate systems.

src/code/z_view.c
```diff
@@ -368,7 +386,10 @@ s32 func_800AAA9C(View* view) {
     return 1;
 }
 
-s32 func_800AB0A8(View* view) {
+/**
+ * Subroutine of View_Apply
+ */
+s32 View_ApplyOrtho(View* view) {
```
engineer124, line 392: See name suggestion above. I'm also not entirely sure about "Apply". I also don't think how it's called is important, so I don't think "Subroutine of View_Apply" is needed

src/code/z_view.c
```diff
@@ -479,13 +506,14 @@ s32 func_800AB560(View* view) {
     LogUtils_CheckNullPointer("viewing", viewing, "../z_view.c", 848);
     view->viewingPtr = viewing;
 
+    // This check avoids a divide-by-zero in guLookAt if eye == lookAt
```
engineer124, line 509:
```suggestion
    // This check avoids a divide-by-zero in guLookAt if eye == at
```

engineer124 (commented):
Nicely done! I've looked at this file extensively in the past before, and I also my own local view docs which may be a good point to compare to brainstorm ideas:
https://github.com/zeldaret/oot/compare/master...engineer124:view
(I never submitted it as a PR because there were things I wasn't set on and I never got around to finalizing it)

src/code/z_view.c
```diff
@@ -51,80 +51,88 @@ void View_Init(View* view, GraphicsContext* gfxCtx) {
...
     view->eye.z = -1.0f;
 
-    if (D_8012ABF0) {
-        if (D_8012ABF0 == 0) {}
+    if (sViewNotInitialized) {
+        if (sViewNotInitialized == 0) {}
         osSyncPrintf("\nview: initialize ---\n");
-        D_8012ABF0 = false;
+        sViewNotInitialized = false;
     }
```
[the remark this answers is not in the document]
rogual: You're right, I somehow totally blanked on the fact that this isn't per-view. I've renamed to `sLogOnNextViewInit`, since that's literally all it does.

include/z64.h
rogual: You're right, the "force all" mask doesn't make any sense. I've removed it. Would be nice to cut down on those `|`s too, though maybe not such a pressing issue.

a2e31a282cf4a47c09ede9eebd1adf74f0be015a

include/z64.h
rogual: Yeah, I'd say it's fairly commonly used. The pattern is usually that you want to "update" something from several places, but it would be a waste to actually do all those updates, so you just set the flag and then whatever code actually reads the value can look at the flag and do the actual update if necessary before using it.

It basically changes N expensive updates into N cheap flag sets and 1 expensive update.

https://stackoverflow.com/questions/13503243/what-does-dirty-flag-dirty-values-mean

src/code/z_view.c
rogual: Yeah, there's no security stuff here.

It's sanitization in that it quietly corrects its inputs if they would lead to a divide-by-zero (eye == at).

src/code/z_view.c
rogual: `view` refers to the view transform, as separate from the projection transform. The view transform is done by `guLookAt` and isn't done in the ortho version.

src/code/z_view.c
rogual: I agree, I also like `at` for the variable. Seems like the most common name for the eye/at/up triple, and is also the name used for the parameter to `guLookAt`.

rogual:
@[engineer124](https://github.com/engineer124) Ah, sorry, didn't see there was prior work on this. It's probably pretty annoying to have figured most of this stuff out already and then have some newcomer come and duplicate the work and do a PR.

Well, it's done now, so merge if useful. Happy to change most of the names if there's consensus. I do think one or two of the names, particularly `LookAt`, should stay, but I'm not at all attached to most of them (2D/3D, dirty, etc).

If we are going with this PR, I feel I should probably wait for consensus on the names to save renaming thing several times.

src/code/z_view.c
rogual: To be honest, I do rather prefer `LookAt` here. It's concise and is something of a standard name for setting a view transform from an eye position, target position and up direction. (c.f. `guLookAt`, `gluLookAt`.)

src/code/z_view.c
rogual: See also long discussion here on a name for this https://github.com/zeldaret/oot/pull/1191#discussion_r837722332

src/code/z_view.c
rogual: OK, I've removed the comments. I've taken the prototypes of these out of the header which is a better way of saying they're internal helper functions, which I think they probably are.

bc8ad7c6d

src/code/z_view.c
rogual: 9b79cf4ca

src/code/z_view.c
```diff
@@ -510,7 +541,7 @@ s32 func_800AB944(View* view) {
     return 1;
 }
 
-s32 func_800AB9EC(View* view, s32 arg1, Gfx** gfxp) {
+s32 View_ApplyTo(View* view, s32 mask, Gfx** gfxp) {
```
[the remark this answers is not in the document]
rogual: Anything starting with `View_Init` would look to me like it's initializing the view itself, rather than using it to initialize something else.

I see how `Apply` could be confusing, especially as it's a word also used for doing matrix multiplication, so am happy to use a different word.

I do quite like the `To` suffix on the function that takes an argument to operate on, though, with the non-`To` version operating on an implicit target. That feels neat to me.

src/code/z_view.c
Dragorn421: shrink_window comes from debug strings though so idk if we want to propagate it that much, there may be a better name.
But yeah scissoring is a low-level term so maybe there's a better higher level name to figure out for this (imo it can be left for later if we can't reach an agreement but I'm open to discussion)

What do you mean by "many sub-systems within the game can be scissored" though? which sub-systems?

rogual:
@Dragorn421 

> Also I'm not sure about the name for dirtyFlags
> 
> First, these flags are used externally to z_view which is weird for something that seems as sensitive as state-indicating flags.
> If they were "dirty flags" I feel like it would have made more sense for them to be set in view_init and through functions rather than directly
> 
> Also, these flags are not reset in all functions, for example they aren't reset by calling View_Apply (unless I missed it). And I would expect it to, to reflect the undirty'ed state if that makes sense.

You're right, it is odd and messy that they would be set from outside.

It is also rather odd that View_Apply doesn't clear the flag. It also doesn't check the flag properly except to determine ortho/perspective. That is, it doesn't check for VIEWING or VIEWPORT at all and always just applies those unconditionally. So it does feel like View_Apply doesn't use it as a dirty flag at all.

The flag is indeed only used that way in View_ApplyTo, the variant that takes a Gfx pointer.

And It's still kind of a wonky system because the view doesn't know if you're calling it with the same Gfx pointer you did before. I guess it just assumes you won't use the same view to build two display lists.

All in all, I'm happy to rename `dirtyFlag` back to `flags` if people would prefer.

src/code/z_view.c
rogual: If it's just restricting drawing to an axis-aligned rectangle, then `clip` or `clipRect` is a common term.

Dragorn421: "clip" means something else at the low level too unfortunately, and it's even related to scissoring so would be a very confusing choice (unfortunately x2)
( http://n64devkit.square7.ch/pro-man/pro12/12-03.htm#01 )

src/code/z_view.c
rogual: That's fair. `LookAt` feels like the right name to me but it's just preference. Consistency within this repo is probably more important.

src/code/z_view.c
fig02: drawing something to overlay doesnt necessarily mean its 2d. The A button on the interface for example is 3d, but is drawn to overlay. So im not sure this differentiation is sufficient

src/code/z_view.c
mzxrules: The big issue I have for a name like `View_SetViewingEyeAtUp` is that the operation being performed feels needlessly lost in a salad of words, that the name choice is driven too much by the arguments being passed into the function instead of being representative of the high level operation that the function performs. I guess I'm also not fully aware of why the concept of lookAt should be totally thrown away in this instance.

rogual: For some addditional context, every 3D graphics library I've seen that's high-level enough to include a function for setting a view transform from an eye/at/up triple calls it LookAt:

[guLookAt](https://ultra64.ca/files/documentation/online-manuals/man/n64man/gu/guLookAt.html)
[gluLookAt](https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/gluLookAt.xml)
[D3DMatrixLookAtLH/RH](https://docs.microsoft.com/en-us/windows/win32/direct3d9/d3dxmatrixlookatlh)
[glm::lookAt](https://glm.g-truc.net/0.9.9/api/a00668.html#gaa64aa951a0e99136bba9008d2b59c78e)
[bgfx: mtxLookAt](https://github.com/bkaradzic/bx/blob/master/src/math.cpp#L241)
[Unity: LookAt](https://docs.unity3d.com/ScriptReference/Matrix4x4.LookAt.html)
[Unreal: Look At](https://docs.unrealengine.com/4.26/en-US/BlueprintAPI/Utilities/Animation/LookAtFunction/)

Given that this is such a ubiquitous operation in CG, and one with such a standardized name, it would seem like a shame to just name it after its arguments.

Consistency within this repo is important too, though, so there is an argument to be made for using the name used in the decompiled camera code name here and perhaps considering changing everything to LookAt later.

src/code/z_view.c
rogual: Summary of possible names with objections raised so far:

`Scissor`: is a low-level term
`shrink_window`: is from debug strings. Also I personally find it confusing; when I first saw "shrink window" I thought of the screen-shrinking effect when the bell rings in MM. Feels like a bit of an engrish-y name for scissor/clipRect.
`clip`: is a low-level term

I don't think `scissor` or `clip` are particularly confusing here. Scissor is indeed a low-level term but it means basically the same thing at the low level and here.  "Clip" is a low-level term but it's also very commonly used to refer to a rectangular region that drawing is restricted to by application code, to speed things up or to implement masking. It doesn't generally cause much confusion, I think. It's very commonly seen in high-level graphics APIs and is rarely called anything other than "clipping",

src/code/z_view.c
rogual: OK, have removed the word "Apply" here and nearby. 46ad1a2cb

Dragorn421: @engineer124 `Graph_Alloc` is essentially the same as `POLY_OPA_DISP++` except it appends (or, prepends, I guess) to the tail instead of the head. The memory area is the same for both, `GfxPool.polyOpaBuffer` (in `gGfxPools`)

src/code/z_view.c
fig02: If it is the common term used everywhere, then we may need to discuss introducing that into the camera docs. But that will need some coordinating with engineer who has been doing all of the camera stuff. 
So, what are your thoughts @engineer124

src/code/z_view.c
fig02: to phrase this another way:
If you say `View_ApplyScissor` it implies just any old scissor value (would expect it to be a value given as an arg, for example)

where as `View_ApplyShrinkWindow` is saying that its applying the scissor with the value obtained from the shrink window system, which is what this does.

So i would probably lean toward naming it after the higher level system its interacting with

agreed though that shrink window is not the best name for that system, but it will change in the future when we actually start renaming og system names

rogual: Ah, thanks for the explanation about the letterboxing. `shrink_window` is fine for me. I don't have a strong opinion on the name for this, anyway, so am happy to just change the PR when there's consensus.

src/code/z_view.c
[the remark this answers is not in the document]
Dragorn421: I'm ok with `View_ApplyTo`. It's `View_Apply` but *to* a specific buffer 😃 

Another example of this add-to-implicit/add-to-explicit split can be `SkelAnime_DrawOpa` (implicit draw to opa, no argument for the buffer like `View_Apply`) / `SkelAnime_Draw` (argument for the buffer to append to, like `View_ApplyTo`)

More on that pattern:
`VisMono_Draw` takes a buffer to append to
`Gfx_DrawDListOpa`/`Gfx_DrawDListXlu` append to opa/xlu

Overall to go with these examples this should have `View_ApplyTo` -> `View_Apply` and `View_Apply` -> `View_ApplyOpaXlu`(`OverlayDisp`)
But I don't see `View_Apply` being `View_ApplyOpaXluOverlayDisp` especially since the `OverlayDisp` part depends on the flags

So I'm still ok with Apply and ApplyTo. it's not a big inconsistency deal, it also may only look inconsistent due to the examples I picked

src/code/z_view.c
Dragorn421: There is a difference actually, the cited `lookat`s functions are producing matrices, the `View_LookAt` function is only a setter for eye/at/up vectors that are then used in other functions to pass to `guLookAt`

if we still want to go with `LookAt` we may also want to change `View_SetPerspective`/`View_SetOrtho` to match `guPerspective`/`guOrtho` as well. For example in `Title_SetupView` it looks weird:
https://github.com/rogual/oot/blob/46ad1a2cb6bac4c40efa23432a7b03437e328f82/src/overlays/gamestates/ovl_title/z_title.c#L52-L54
```c
    View_SetPerspective(view, 30.0f, 10.0f, 12800.0f);
    View_LookAt(view, &eye, &lookAt, &up);
    View_Apply(view, VIEW_ALL);
```

but I feel like `View_LookAt` not producing a matrix is enough of an argument to not stick to `LookAt` and use `SetEyeAtUp` instead

rogual: A matrix is one possible representation of a coordinate transform. In my view, `View`'s interface is that it represents a coordinate transform, and the fact that it uses matrices internally (and uses the flag to decide when to update them) is an implementation detail. To "look at" something is to orient oneself. `LookAt` is something that makes sense for any representation of a coordinate transform, not just a matrix. For instance, I would feel that LookAt would be an equally good name even if Euler angles or quaternions were being used.

rogual: Not all of the cited `LookAt`s work on matrices. They tend to work on whatever representation of a Transform the library in question uses. For GL that's matrices, but for Unity and Unreal that's a `Transform` object which doesn't expose its internal representation.

include/z64.h
rogual: Changed back to `flags` bc841becfc21a7818d20832e6e50ff07589f081a

src/code/z_view.c
Dragorn421: I'm not dead set against LookAt, my main point is instead of (in `Title_SetupView`)
```c
    View_SetPerspective(view, 30.0f, 10.0f, 12800.0f);
    View_LookAt(view, &eye, &lookAt, &up);
    View_Apply(view, VIEW_ALL);
```
we should have either
```c
    View_Perspective(view, 30.0f, 10.0f, 12800.0f);
    View_LookAt(view, &eye, &lookAt, &up);
    View_Apply(view, VIEW_ALL);
```
or
```c
    View_SetPerspective(view, 30.0f, 10.0f, 12800.0f);
    View_SetEyeAtUp(view, &eye, &lookAt, &up);
    View_Apply(view, VIEW_ALL);
```

mzxrules: don't understand why `View_SetLookAt` isn't also a valid option; view's eye, at, up properties appear to be used exclusively for calculating the LookAt matrix.

src/code/z_view.c
rogual: Okay, I suppose there's an implied "as necessary" in this comment. These `xxx_DISP`s, as a set, are the target of the operation. I'm not trying to say that it always touches all of them.

I don't think it's necessary for this comment to detail exactly what happens when; you can read the code for that.

src/code/z_view.c
rogual: `SetLookAt` is fine for me too. Slight preference for just `LookAt` but I don't really care, I really think we're firmly in nitpick territory now.

src/code/z_view.c
Dragorn421: I do feel like it matters that overlay_disp is not touched at all in the perspective case
```suggestion
 * Apply view to POLY_OPA_DISP, POLY_XLU_DISP (and OVERLAY_DISP if ortho)
 */
void View_Apply(View* view, s32 mask) {
```

src/code/z_view.c
Dragorn421: well my only issue is again the `View_SetPerspective`/`View_LookAt` combo looks weird to me but if I'm the only one I'm ready to drop the matter

Dragorn421 (approved):

fig02 (approved):

Roman971 (approved):
Looks good to me 👍

PR #1193, opened 2022-04-03
Angle cleanup - macros names

Dragorn421:
- Get rid of the "F" in angle conversion macros names as discussed in previous angle cleanup PRs
- Move `BINANG_ROT180`, `BINANG_SUB` out of the block of macros defining angle conversions (I will petition to get rid of them eventually 😈)
- Move the two "approximate conversions" macros (which are used in camera-related code, almost exclusively) to `z64camera.h` and add a `CAM_` suffix
- Add proper degrees <-> binang conversion macros in z64math.h (unused by the game)

-----------

- Renamed macros:

Old name|New name
-|-
`DEGF_TO_BINANG`|`CAM_DEG_TO_BINANG` (moved to z64camera.h)
`RADF_TO_BINANG`|`RAD_TO_BINANG`
`RADF_TO_DEGF`|`RAD_TO_DEG`
`DEG_TO_RAD`|unchanged
`BINANG_TO_DEGF`|`CAM_BINANG_TO_DEG` (moved to z64camera.h)
`BINANG_TO_RAD`|unchanged
`BINANG_TO_RAD_ALT`|unchanged
`BINANG_TO_RAD_ALT2`|unchanged

- Moved without change: `BINANG_ROT180`, `BINANG_SUB`

- Added: `DEG_TO_BINANG`, `BINANG_TO_DEG`

- Removed: none

---------

Besides `include/z64math.h`, `z64camera.h` and formatting, this is a Replace-All (Ctrl+Shift+H) PR


fig02 (approved):

include/z64math.h
```diff
@@ -93,14 +96,14 @@ typedef struct {
...
-#define BINANG_SUB(a, b) ((s16)((a) - (b)))
+// Angle conversion macros
+#define DEG_TO_BINANG(degrees) (s16)((degrees) * (0x8000 / 180.0f))
+#define DEG_TO_BINANG_ALT(degrees) (s16)((degrees) * 182.04167f + .5f)
+#define RAD_TO_BINANG(radians) (s16)((radians) * (0x8000 / M_PI))
+#define RAD_TO_DEG(radians) ((radians) * (180.0f / M_PI))
 #define DEG_TO_RAD(degrees) ((degrees) * (M_PI / 180.0f))
-#define BINANG_TO_DEGF(binang) ((f32)(binang) * (360.0001525f / 65535.0f))
+#define BINANG_TO_DEG(binang) ((f32)(binang) * (180.0f / 0x8000))
+#define BINANG_TO_DEG_ALT(binang) ((f32)(binang) * (360.0001525f / 65535.0f))
```
Roman971, line 106: If the codebase only needs 1 macro then I don't think we should use an `_ALT` macro.
If the idea to provide/document a more accurate or more efficient version, I suggest leaving a comment mentioning it, and modders can simply replace the macro implementation with it if they want to.

Same comment for `DEG_TO_BINANG_ALT`

include/z64math.h
```diff
@@ -76,6 +76,9 @@ typedef struct {
     /* 0x06 */ s16 yaw;    // azimuthal angle
 } VecSph; // size = 0x08
 
+#define BINANG_ROT180(angle) ((s16)((angle) - 0x7FFF))
+#define BINANG_SUB(a, b) ((s16)((a) - (b)))
```
Roman971, line 80: Just a side note but I would agree with removing these macros. I think the codebase is way too inconsistent/hardcoded to try and abstract the fact that angles use s16 types and value ranges.

fig02: +1 from me, particularly for BINANG_SUB. 
I think the rotate 180 would not be immediately obvious to someone who isnt used to the pattern. So it provides a benefit in that sense. But I dont like the idea of macro-ifying the operation in general

include/z64math.h
fig02: This makes sense, as a modder id prob see theres an alt version and consider it "extra" without putting any thought into which is better to use. A comment providing a better implementation to use seems more helpful and obvious

Dragorn421: I would entirely agree if the vast majority of the code using these two "bad" macros wasn't camera code, which makes me think it was a macro "local to one dev person" and not a macro that was in a main header

`BINANG_TO_DEGF`:
- 8 in db_camera.c
- 9 in z_camera.c
- 2 in z_onepointdemo.c
- 1 in z_bg_spot02_objects.c (in cutscene code unsurprisingly)

`DEGF_TO_BINANG`:
- 4 in db_camera.c
- 61 in z_camera.c
- 4 in z_olib.c (olib being mostly a "math file" for z_camera)
- 1 in z_onepointdemo.c

if we had split headers maybe I would have tried putting it in a would-be "z_olib.h" as `_CAM` macros

but I understand not wanting "useless" macros. it's not like degrees<->binang is particularly useful anyway

Roman971: That makes a bit more sense yeah. I guess I wouldn't mind it as much if it had a more specific name/header instead.

Also I'm fine with unused macros if they belong to a library or something, but this is a general header whose sole purpose is to host defines/types used in the codebase so it feels a bit weird to have unused stuff in it.

Dragorn421: So do I do any or all of
- remove the two new macros
- rename the two "camera" _ALT macros to no suffix, _CAM or something
- move the two "camera" macros to some header, like z64camera idk

fig02: if the macros are camera specific and they arent the ideal form of the macro, id say move it to z64camera

fig02: they arent camera specific i guess, but they are clearly ogawa specific Kappa

Dragorn421: they aren't camera specific but there's a strong suspicion that only the camera dev used them (is "ogawa" their name?)
putting them in z64camera may pose dumb issues / force dumb includes when having proper includes down the line
if z_olib.h existed it would probably be a good place but it doesn't exist (yet)

fig02: camera dev is ogawa yes. He definitely did camera and the camera related stuff in olib, and probably one point demo as well.
z64camera seems fine for now and it can be moved to an appropriate place later, but others can weigh in

Roman971: Yeah camera is probably fine for now. Or we could leave the macros on their own in math.h for now and move it later

fig02: camera prob makes sense cause it seems for sure camera-dev-only. we'll know what to do with them when the time comes

Roman971: Regarding the "proper" unused macros, I don't really have an opinion on what to do about them. If you think it makes sense to have them with the other angle macros then it's probably fine. Otherwise feel free to remove them

Dragorn421: I
- removed the "proper" unused macros
- moved the two "camera" macros to z64camera.h
- removed the _alt suffix from the two "camera" macros
- added a comment near the two "camera" macros

fig02: fwiw i think the "better" version of the macros are worth keeping. I prob should have made that clear earlier, sorry.
In modding contexts ive personally made use of both degrees -> binang and vice versa. 
Maybe my uses were niche, but for angles in general i think in terms of degress, not radians or binang

fig02: I think in my ideal world, the bad camera ones are prefixed with CAM_, and the proper versions of the macros exist for modders to use. But I understand why including unused macros can be seen as not in scope

Dragorn421: So I'm not sure where to go on this because I don't have a particular opinion, and you two don't seem to have exactly the same idea

Currently (what roman wants?):
game's `DEG_TO_BINANG` and `BINANG_TO_DEG` in z64camera.h

Possibility (what fig wants?):
game's `CAM_DEG_TO_BINANG` and `CAM_BINANG_TO_DEG` in z64camera.h
more proper `DEG_TO_BINANG` and `BINANG_TO_DEG` in z64math.h

fig02: yes I would like to have the proper conversions available to modders out of the box, but its kinda a selfish request cause i already use those macros

I dont think roman has had the chance to respond, so I dont think you need to do anything on your end

Roman971: Yeah I also think there should be a CAM_ prefix if the macros are specific to the camera system, sorry if I didn't make it clear. 

And the unused macros are fine to have in z64math.h as well, tho I don't really have an opinion on that myself

fig02: so then yea i would add them even if they arent used in the original codebase personally

fig02 (approved):

Roman971 (approved):

PR #1197, opened 2022-04-11
Create GPL3.0 LICENSE

YusufKhan-gamedev:
This includes a verbatim license from mips_to_c


fig02:
As stated in the other PR. mips_to_c is a tool external to this project that people used on their own to aid with the process of decompilation. The code for mips_to_c is not included in this repo, so we do not need to share the same license.

PR #1206, opened 2022-04-25
Move asset extraction to makefile

Dragorn421:
- Remove `extract_assets.py`
- Make `msgdis.py` executable and able to be passed command line arguments
- Get the makefile to handle asset extraction (`make extract_assets`)

The file `.extracted-assets.json` which used to store the timestamps at which assets had been extracted, has been removed and functionally replaced by empty files in each assets folder named `.stamp_extracted`. These files are set in the makefile to depend on the xml, and rebuilding them triggers the asset's extraction with ZAPD and updating the stamp file's timestamp to "now" (with `touch`). Their timestamp effectively represents the time of extraction.
Removing a `.stamp_extracted` file would result in the corresponding asset being reextracted on the next extract_assets run.

The `.stamp_extracted` could also be put under `build/assets/...`, but then they would be deleted by `make clean`, possibly causing unsollicited asset reextraction.

# Extracting assets now

`./extract_assets.py` becomes `make extract_assets`

Passing flags to zapd (previously `./extract_assets.py`'s `-u`, `-z`) can be achieved with
`make extract_assets ZAPD_FLAGS="--warn-unaccounted -Werror"`

Extracting assets over even if the xmls haven't been modified (previously `./extract_assets.py`'s `-f`) can be achieved with
`make extract_assets --always-make` (it's a `make` option)

Extracting with a single xml (previously `./extract_assets.py`'s `-s`) can be achieved with
`make assets/objects/object_daiku/.stamp_extracted`

Of course all of the above examples can be combined.

# Notes

The makefile now has rules for `assets/text/message_data.h` and `assets/text/message_data_staff.h`, so these can be extracted if missing when the makefile needs them (there are rules depending on them). It's not an issue, just pointing it out.

`./tools/msgdis.py` is called separately for the text and the staff-text, whereas in old `extract_assets.py` it was called once for both files. The only issue would be performance but I don't think it's major. This could be addressed by changing the makefile if it bothers someone, but it's not an issue otherwise.

One can prevent an asset file from being extracted by removing its xml from the makefile's list of assets xml files. For example:
`ASSET_XML := $(filter-out assets/xml/objects/object_daiku.xml,$(ASSET_XML))`
right after the first `ASSET_XML` definition (`ASSET_XML := $(shell find assets/xml/ -name "*.xml")`), will result in not calling zapd for object_daiku.
This may be an acceptable temporary monkeypatch for custom assets in modding

I didn't update the tutorial which still references `./extract_assets.py`. idk at this point if the tutorial is to be preserved for history or kept updated?


louist103:
Is there any benefit to doing it like this? IE performance or memory? It seems like it makes the `Makefile` more complicated with little benefit.

Dragorn421:
This replaces `extract_assets.py` (164 lines, with basically a make -j implementation) with a dozen lines in `msgdis.py` and 15 lines in the `Makefile`, so I think it's simpler with these changes
Also `extract_assets.py` takes a few seconds to start up, so removing it fixes that delay 😎

This can be moved to another makefile though if it really bloats the main makefile too much (I don't think it does).

louist103:
> Also `extract_assets.py` takes a few seconds to start up, so removing it fixes that delay 😎
Should have opened with that lol.

PR #1211, opened 2022-05-01
Document `eventChkInf` usage of frogs played songs flags

Dragorn421:
Mostly simply extending the existing doc on frogs to the relevant `eventChkInf` bits


src/overlays/actors/ovl_En_Fr/z_en_fr.c
```diff
@@ -105,13 +105,26 @@ static EnFrPointers sEnFrPointers = {
...
-static u16 sSongIndex[] = {
-    EVENTCHKINF_D1_MASK, EVENTCHKINF_D2_MASK, EVENTCHKINF_D4_MASK, EVENTCHKINF_D3_MASK,
-    EVENTCHKINF_D5_MASK, EVENTCHKINF_D6_MASK, EVENTCHKINF_D0_MASK, 0,
+#define GET_EVENTCHKINF_SONGS_FOR_FROGS(frogSongIndex)             \
+    (gSaveContext.eventChkInf[EVENTCHKINF_SONGS_FOR_FROGS_INDEX] & \
+     sFrogSongIndexToEventChkInfSongsForFrogsMask[frogSongIndex])
+
+#define SET_EVENTCHKINF_SONGS_FOR_FROGS(frogSongIndex)             \
+    gSaveContext.eventChkInf[EVENTCHKINF_SONGS_FOR_FROGS_INDEX] |= \
+        sFrogSongIndexToEventChkInfSongsForFrogsMask[frogSongIndex];
```
fig02, line 114: should we potentially use a FROG_ prefix to make this clear that the macro is for the frogs? as opposed to a more general eventchkinf macro
this would replace `FOR_FROGS` at the end

Dragorn421: the idea is searching for `EVENTCHKINF_SONGS_FOR_FROGS` would bring up everything related to those flags

`FROG_SET_EVENTCHKINF_SONGS_FOR_FROGS`
or `EVENTCHKINF_SONGS_FOR_FROGS_SET`
would be other options I think

but `FROG_SET_EVENTCHKINF_SONGS` would break discovering those uses, except if all other names of frog flags stuff follow that

Dragorn421: Also the name I went with mimics the generic `GET/SET_EVENTCHKINF()` macro

fig02: that similarity is what my issue was to begin with-- it makes it seem like it belongs to those set of global macros when its not. Its local to the actor and uses data that belongs to that overlay

but yea I dont feel too strongly about it to need it changed, its probably fine

Dragorn421: So like `GET_EVENTCHKINF_SONGS_FOR_FROGS(frogSongIndex)` is the contraction of `GET_EVENTCHKINF` and `EVENTCHKINF_SONGS_FOR_FROGS` as in "get the relevant `EVENTCHKINF_SONGS_FOR_FROGS` bit in eventChkInf"

If the macro should be named at a higher level, it may be something like `FROG_HAS_SONG_BEEN_PLAYED(frogSongIndex)`/`FROG_SET_SONG_PLAYED(frogSongIndex)` ?

fig02: yeah those names are alot more helpful imo and help with the issue i brought up. If you are ok with using those names I think I would prefer it

fig02 (approved):

Roman971 (approved):

PR #1213, opened 2022-05-02
Introduce DEFINE_ENTRANCE for entrance table and create entrance enum

fig02:
This implements a `/table/include` table for the entrance table.
An enum is introduced for entrance indices and used in code

Note: values 0x7FF9 - 0x7FFF have a special meaning in player and should have defines introduced. But its 5am and I cannot understand how this thing even works. Suggestions are welcome if anyone knows
Relevant arrays are `D_808544F8` and `D_80854514`


Dragorn421 (changes requested):
Nice 👏
I like the `ENTR_` prefix 

Would the idea be to name the spawns eventually too?

I think for the first setup of each group of entrances, the enum name should omit the `_0` suffix indicating the setup, as it's irrelevant when used in code (it's more used as the base entrance index which the actual setup is added to, rather than setup 0 itself)

----------------

I found a single spot missing entrance enum usage: `Environment_WarpSongLeave`

About the special `0x7FF9`-`0x7FFF` values:

<details>

Searching for regex `sceneExitList([^;]|\n)*0x7` under `assets/scenes/` shows that the following scenes use them:

Name|Friendly name|Exit entrance|Exit id
-|-|-|-
`daiyousei_izumi`|Great fairy fountain (magic, double magic, double defense)|`0x7FFE`|0
`syatekijyou`|Shooting gallery|`0x7FFA`|0
`yousei_izumi_tate`|Fairy fountain|`0x7FFF`|0
`yousei_izumi_yoko`|Great fairy fountain (spells)|`0x7FF9`|0
`kakusiana`|Grottos|`0x7FFF`|0
`shop1`|Bazaar|`0x7FFC`|0

At first glance it looks like these are exactly the "reused" scenes that map to several entrances, so a reasonable assumption would be that this wacky thing implements just that, reusable scenes where "going back" works as expected with a dynamic exit instead of always the same.

Afaict only `func_80839034` from z_player plays a role code-wise, and that function is about exits-triggered scene transitions

The two arrays holding the data are used like

```c
globalCtx->nextEntranceIndex = D_808544F8[D_80854514[globalCtx->nextEntranceIndex - ENTR_0x7FF9] + globalCtx->curSpawn];
```

I guess each reused scene has a different set of entrances in the entrance table, and each set has a different "spawn", which put Link in the same spot but still serves as indicating the entrance group that was used, hence what scene the player comes from

Seeing `this->fountainType = globalCtx->curSpawn;` in `BgDyYoseizo_Init` (great fairy actor) seems to confirm that

I would see the arrays like this:

```c
s16 D_808544F8[] = {
    /*  0 */ ENTR_SPOT16_4_0, // DMT from Magic Fairy Fountain
    /*  1 */ ENTR_SPOT17_3_0, // DMC from Double Defense Fairy Fountain
    /*  2 */ ENTR_SPOT15_2_0, // Hyrule Castle from Dins Fire Fairy Fountain

    /*  3 */ ENTR_SPOT01_9_0,     // Kakariko from Potion Shop
    /*  4 */ ENTR_MARKET_DAY_5_0, // Market (child day) from Potion Shop

    /*  5 */ ENTR_SPOT01_3_0,     // Kakariko from Bazaar
    /*  6 */ ENTR_MARKET_DAY_6_0, // Market (child day) from Bazaar

    /*  7 */ ENTR_SPOT01_11_0,      // Kakariko from House of Skulltulas
    /*  8 */ ENTR_MARKET_ALLEY_2_0, // Back Alley (day) from Bombchu Shop

    /*  9 */ ENTR_SPOT01_10_0,    // Kakariko from Shooting Gallery
    /* 10 */ ENTR_MARKET_DAY_8_0, // Market (child day) from Shooting Gallery

    /* 11 */ ENTR_SPOT08_5_0, // Zoras Fountain from Farores Wind Fairy Fountain
    /* 12 */ ENTR_SPOT15_2_0, // Hyrule Castle from Dins Fire Fairy Fountain
    /* 13 */ ENTR_SPOT11_7_0, // Desert Colossus from Nayrus Love Fairy Fountain
};

u8 D_80854514[] = {
    /* 0x7FF9 */ 11,
    /* 0x7FFA */ 9,
    /* 0x7FFB */ 3,
    /* 0x7FFC */ 5,
    /* 0x7FFD */ 7,
    /* 0x7FFE */ 0,
};
```

Where `D_808544F8` is split into groups indexed by `curSpawn`, and the start of each group is in `D_80854514`

Interestingly `0x7FFB` and `0x7FFD` are not used
For `0x7FFB` I assume they added the passage to the back of the potion shop in the kakariko version later, so couldn't use the same scene then
For `0x7FFD` I guess at some point either bombchu shop or house of skulltulas wasn't a thing, and they replaced either with the other

Anyway, I think 0x7FF9 to 0x7FFE should be named according to the scenes that use them

For 0x7FFF aka grottos/fairy fountains I'm less sure what's going on but it's more obvious/known what's going on

</details>

src/overlays/actors/ovl_En_Heishi3/z_en_heishi3.c
```diff
@@ -203,7 +203,7 @@ void func_80A55D00(EnHeishi3* this, GlobalContext* globalCtx) {
     if ((Message_GetState(&globalCtx->msgCtx) == TEXT_STATE_EVENT) && Message_ShouldAdvance(globalCtx) &&
         (this->respawnFlag == 0)) {
         SET_EVENTCHKINF(EVENTCHKINF_4E);
-        globalCtx->nextEntranceIndex = 0x47E; // Hyrule Castle from Guard Capture (outside)
+        globalCtx->nextEntranceIndex = ENTR_SPOT15_4_0;
```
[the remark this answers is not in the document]
fig02: I intended to remove all of them, mainly because its not a pattern that we do consistently at all

It does hurt readability short term because the scene names are bad, but that will be fixed eventually. If we had the entrance comments everywhere it might make more sense, but it seems weird to only have them in a few places

fig02:
I can see the argument for leaving the 0 out for scene setup 0, but Im not too much a fan of the inconsistency that would introduce. It is true that only setup 0 is relevant for the entrances in code, but it feels kinda weird to leave that information out for that case

If its what people want to do though then we can change over to that

include/tables/entrance_table.h
```diff
@@ -0,0 +1,1932 @@
+
```
Dragorn421, line 1:
```suggestion
```

include/tables/entrance_table.h
```diff
@@ -0,0 +1,1932 @@
...
+/* 0x2E3 */ DEFINE_ENTRANCE(ENTR_SPOT20_2_1, SCENE_SPOT20, 2, false, true, TRANS_TYPE_FADE_BLACK, TRANS_TYPE_FADE_BLACK)
+/* 0x2E4 */ DEFINE_ENTRANCE(ENTR_SPOT20_2_2, SCENE_SPOT20, 2, false, true, TRANS_TYPE_FADE_WHITE, TRANS_TYPE_FADE_WHITE)
+/* 0x2E5 */ DEFINE_ENTRANCE(ENTR_SPOT20_2_3, SCENE_SPOT20, 2, false, true, TRANS_TYPE_FADE_BLACK, TRANS_TYPE_FADE_BLACK)
+
+/* 0x2E6 */ DEFINE_ENTRANCE(ENTR_SPOT20_3, SCENE_SPOT20, 3, false, true, TRANS_TYPE_FADE_WHITE, TRANS_TYPE_FADE_WHITE)
+/* 0x2E7 */ DEFINE_ENTRANCE(ENTR_SPOT20_3_1, SCENE_SPOT20, 3, false, true, TRANS_TYPE_FADE_BLACK, TRANS_TYPE_FADE_BLACK)
+/* 0x2E8 */ DEFINE_ENTRANCE(ENTR_SPOT20_3_2, SCENE_SPOT20, 3, false, true, TRANS_TYPE_FADE_WHITE, TRANS_TYPE_FADE_WHITE)
+/* 0x2E9 */ DEFINE_ENTRANCE(ENTR_SPOT20_3_3, SCENE_SPOT20, 3, false, true, TRANS_TYPE_FADE_BLACK, TRANS_TYPE_FADE_BLACK)
+
+// Labeled as "Test Area" in Map Select
```
Dragorn421, line 927: Is this comment all that relevant? This table is so big probably no one is going to read it, unless we really plan to have comments on every entrance group?

fig02: this block is really special. Everything in this group is entirely unrelated, other than the fact that you can access them via map select by "test area" and changing the cutscene index

People will see it if they see test area in map select as pointing to the archery house for some reason, and care to go and check out the group of entrances that its near

Also note that I was forced to add `TEST_` to the name, cause otherwise it conflicts with the actual archery shop spawn 0 enum. So its a weird special case

Dragorn421 (approved):

Roman971:
Some conflict to fix then we can probably merge this

PR #1214, opened 2022-05-03
Fixup weird horse flags usage

Dragorn421:
I don't know what's going on with these, at least not enough to actually document them. At the very least they all seem to be horse-related

I'm not sure where to put the enum and macros either


include/z64save.h
```diff
@@ -644,31 +644,50 @@ typedef enum {
...
+    /* 6 */ EVENTINF_HORSES_STATE_6,
+    /* 7 */ EVENTINF_HORSES_STATE_7
+} EventInfHorsesState;
+// "InRaceSeq"
+#define GET_EVENTINF_HORSES_STATE \
+    ((gSaveContext.eventInf[EVENTINF_HORSES_INDEX] & EVENTINF_HORSES_STATE_MASK) >> EVENTINF_HORSES_STATE_SHIFT)
+#define SET_EVENTINF_HORSES_STATE(v)                                                   \
+    gSaveContext.eventInf[EVENTINF_HORSES_INDEX] =                                     \
+        (gSaveContext.eventInf[EVENTINF_HORSES_INDEX] & ~EVENTINF_HORSES_STATE_MASK) | \
+        ((v) << EVENTINF_HORSES_STATE_SHIFT);
```
fig02, line 681: some line breaks here would prob make it look nicer

fig02 (commented):
as for placing all of these extra flag defines, do we think there will be alot of them going forward? should flags get their own header instead of crowding up the save header?

fig02 (approved):

Roman971 (approved):

include/z64save.h
```diff
@@ -644,31 +644,55 @@ typedef enum {
...
+    /* 2 */ EVENTINF_HORSES_STATE_2,
+    /* 3 */ EVENTINF_HORSES_STATE_3,
+    /* 4 */ EVENTINF_HORSES_STATE_4,
+    /* 5 */ EVENTINF_HORSES_STATE_5,
+    /* 6 */ EVENTINF_HORSES_STATE_6,
+    /* 7 */ EVENTINF_HORSES_STATE_7
+} EventInfHorsesState;
+
+// "InRaceSeq"
+#define GET_EVENTINF_HORSES_STATE() \
```
louist103, line 678: I don't really like how these look with `()`.  I feel like it will make some people think its a function.

fig02: it essentially behaves as one, its a natural way for getters and setters to look

fig02: it was specifically requested by roman here https://github.com/zeldaret/oot/pull/1214#discussion_r867361339
and I agree with the choice, it looks better this way imo

fig02: I am going to merge this now as it has enough approvals to do so. But if anyone else shares this same sentiment and it needs further discussion, feel free to do so

PR #1216, opened 2022-05-07
Add offsets to map 48x85

louist103:
I got tired of looking at the warnings.


Dragorn421 (approved):
Unfortunate but +1

Another option would be passing `-Wno-missing-offsets` to zapd (I think), but I think we're better off having explicit offsets anyway

fig02 (approved):

Roman971 (approved):

PR #1217, opened 2022-05-10
Document entrances 0x7FF9-0x7FFF as "Return Entrances"

Dragorn421:


fig02:
naming them "Exit Entrances" is a problem for two reasons imo:
exit has a meaning in scenes with the exit index/list
"entrance exit" are opposites of each other, so it sounds kinda strange

imo naming it like "return entrance" or something would possibly be better, cause its an entrance that returns you to the right place that you came from in shared scenes

fig02 (approved):

Roman971 (approved):
I wonder if these values should be part of the main enum, but I guess that would only matter when using the enum as a type, so it's probably fine either way

fig02:
it probably makes more sense separate, cause they dont mean anything to the entrance system/table. They only have meaning within player

Roman971:
It's true that they are translated by the player system but they are still intended to be used everywhere else, alon with other entrance indexes (eg. scene exit lists), so I don't know if it's right to treat them as separate

PR #1218, opened 2022-05-10
Various minor cleanups

Roman971:
Another batch of small cleanups. Here's the list by commit:
- Fix an assert in `z_en_horse.c`
- Fix some UB with a sceneFlags access in `z_en_elf.c`
- Add `Object_Spawn` to public object functions (used externally in MM and works as a standalone)
- Fix float literals to always use lowercase e for the exponent (as discussed in discord)
- Remove "mininap" in one of the map data variables (actually used for the pause menu map as well)


include/z64.h
```diff
@@ -1433,7 +1433,7 @@ typedef struct {
     /* 0x2C */ s16*  owMinimapPosX;
     /* 0x30 */ s16*  owMinimapPosY;
     /* 0x34 */ s16 (*owCompassInfo)[4]; // [X scale, Y scale, X offset, Y offset]
-    /* 0x38 */ s16*  dgnMinimapTexIndexBase; // dungeon minimap texture index base
+    /* 0x38 */ s16*  dgnTexIndexBase; // dungeon texture index base
```
fig02, line 1436: technically not a part of your changes, but the dgn abbreviation is probably not needed in general? the name would be on the longer side for sure, but now that its shortened with your new changes it might be worth

I just dont see anyone use dgn ever, even tho its pretty easy to figure out what it means

Roman971: The abbreviation is used for many other names in there, including some pretty longer ones, so I kept it for consistency. Also this whole struct is already pretty verbose so I don't really think we should get rid of `dgn` (or `ow` for that matter)

fig02: I dont think I agree, but it can be talked about later in context with those other variables separate from this set of changes

fig02 (approved):

include/z64.h
Roman971: Something I forgot to mention and might not be obvious here but all of these fields are arrays (with many 2d or even 3d arrays), so short names are more important than usual

src/overlays/actors/ovl_En_Ru1/z_en_ru1_cutscene_data.c
```diff
@@ -5,21 +5,21 @@
...
-        CS_PLAYER_ACTION(0x0002, 109, 135, 0x0000, 0x6C16, 0x0000, 12, -340, -2810, 100, -340, -2991, 1.1393037E-29f, 0.0f, 1.4E-45f),
-        CS_PLAYER_ACTION(0x0005, 135, 199, 0x0000, 0x6C16, 0x0000, 100, -340, -2991, 100, -340, -2991, 1.1393037E-29f, 0.0f, 1.4E-45f),
-        CS_PLAYER_ACTION(0x0003, 199, 219, 0x0000, 0x6C16, 0x0000, 100, -340, -2991, 100, -340, -2991, 1.1393037E-29f, 0.0f, 1.4E-45f),
-        CS_PLAYER_ACTION(0x0005, 219, 259, 0x0000, 0x6C16, 0x0000, 100, -340, -2991, 100, -340, -2991, 1.1393037E-29f, 0.0f, 1.4E-45f),
-        CS_PLAYER_ACTION(0x0003, 259, 276, 0x0000, 0x6C16, 0x0000, 100, -340, -2991, 100, -340, -2991, 1.1393037E-29f, 0.0f, 1.4E-45f),
-        CS_PLAYER_ACTION(0x0002, 276, 297, 0x0000, 0x6C16, 0x0000, 100, -340, -2991, 182, -340, -3132, 1.1393037E-29f, 0.0f, 1.4E-45f),
-        CS_PLAYER_ACTION(0x0005, 297, 843, 0x0000, 0x6C16, 0x0000, 182, -340, -3132, 182, -340, -3132, 1.1393037E-29f, 0.0f, 1.4E-45f),
+        CS_PLAYER_ACTION(0x0027, 0, 50, 0x0000, 0x6C16, 0x0000, 12, -340, -2810, 12, -340, -2810, 1.1393037e-29f, 0.0f, 1.4e-45f),
```
Thar0, line 8: Out of scope but we probably should change these from floats eventually. It's not really clear which is correct since there are cases where both float and int look more correct, and there's even some cases where things that look reasonable as floats and things that are clearly not floats are side-by-side:
```c
CS_NPC_ACTION(0x0001, 0, 234, 0x0000, 0x4000, 0x0000, -1665, 52, -44, -1665, 52, -44, 1.1393037197548209e-29f, 0.0f, 1.401298464324817e-45f),
CS_NPC_ACTION(0x0002, 234, 241, 0x41F8, 0x0000, 0x0000, -1665, 52, -44, -1603, 130, -47, 8.857142448425293f, 11.142857551574707f, -8.857142448425293f),
CS_NPC_ACTION(0x0002, 241, 280, 0x4031, 0x0000, 0x0000, -1603, 130, -47, -549, 130, -52, 27.0256404876709f, 0.0f, -27.0256404876709f),
```
These last three fields are never used though afaict so it's just more garbage cs data to add to the already large amount 😔

Thar0 (approved):

src/overlays/actors/ovl_En_Elf/z_en_elf.c
```diff
@@ -1406,13 +1406,10 @@ void func_80A053F0(Actor* thisx, GlobalContext* globalCtx) {
...
-        nREG(80) = gSaveContext.sceneFlags[127].chest;
+        nREG(80) = HIGH_SCORE(HS_HBA);
 
-        if (nREG(81) != 0) {
-            if (gSaveContext.sceneFlags[127].chest) {
-                LOG_NUM("z_common_data.memory.information.room_inf[127][ 0 ]", gSaveContext.sceneFlags[127].chest,
-                        "../z_en_elf.c", 2595);
-            }
+        if ((nREG(81) != 0) && (HIGH_SCORE(HS_HBA) != 0)) {
+            LOG_NUM("z_common_data.memory.information.room_inf[127][ 0 ]", HIGH_SCORE(HS_HBA), "../z_en_elf.c", 2595);
```
EllipticEllipsis, line 1412: Since it was decided to go this way, there should at least be a comment here explaining why the string and the variable looks nothing alike (since presumably this was formed with stringification originally).

Dragorn421: I'll try to do something in #1223 , see you there :)

PR #1220, opened 2022-05-10
Remove many redundant parentheses (eg. in boolean conditions)

Roman971:
Tried to find and fix most cases of redundant parens for boolean conditions like `if ((foo))` or `if (foo && (!(bar & 1)))`.
Also found a few function calls and switches with double (or even triple) parens so I fixed those as well

Note: To be clear, I didn't get rid of any parens that exist to not rely on operator precendence (eg. having `if ((foo == 1) && (foo == 2))` instead of `if (foo == 1 && foo == 2)`). I'm actually in favor of that personally

### DIFF
--- a/src/code/z_debug.c
+++ b/src/code/z_debug.c
@@ -67,7 +67,7 @@ void func_8006376C(u8 x, u8 y, u8 colorId, const char* text) {
 
         i = 0;
         bufText = buf->text;
-        while ((*bufText++ = *text++)) {
+        while ((*bufText++ = *text++) != 0) {
             if (i++ > 0x14) {
                 break;
             }
--- a/src/overlays/actors/ovl_Bg_Jya_Lift/z_bg_jya_lift.c
+++ b/src/overlays/actors/ovl_Bg_Jya_Lift/z_bg_jya_lift.c
@@ -137,7 +137,7 @@ void BgJyaLift_Update(Actor* thisx, GlobalContext* globalCtx2) {
     }
     if ((this->dyna.unk_160 & 4) && ((this->unk_16B & 4) == 0)) {
         Camera_ChangeSetting(globalCtx->cameraPtrs[MAIN_CAM], CAM_SET_DIRECTED_YAW);
-    } else if (((this->dyna.unk_160) & 4) == 0 && ((this->unk_16B & 4)) &&
+    } else if (((this->dyna.unk_160) & 4) == 0 && (this->unk_16B & 4) &&
                (globalCtx->cameraPtrs[MAIN_CAM]->setting == CAM_SET_DIRECTED_YAW)) {
         Camera_ChangeSetting(globalCtx->cameraPtrs[MAIN_CAM], CAM_SET_DUNGEON0);
     }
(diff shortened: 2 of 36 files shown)


src/overlays/actors/ovl_En_Mk/z_en_mk.c
```diff
@@ -291,7 +291,7 @@ void EnMk_Update(Actor* thisx, GlobalContext* globalCtx) {
     Actor_MoveForward(&this->actor);
     Actor_UpdateBgCheckInfo(globalCtx, &this->actor, 0.0f, 0.0f, 0.0f, UPDBGCHECKINFO_FLAG_2);
 
-    if ((!(this->flags & 2)) && (SkelAnime_Update(&this->skelAnime))) {
+    if (!(this->flags & 2) && (SkelAnime_Update(&this->skelAnime))) {
```
Dragorn421, line 294:
```suggestion
    if (!(this->flags & 2) && SkelAnime_Update(&this->skelAnime)) {
```

src/code/z_debug.c
```diff
@@ -67,7 +67,7 @@ void func_8006376C(u8 x, u8 y, u8 colorId, const char* text) {
 
         i = 0;
         bufText = buf->text;
-        while ((*bufText++ = *text++)) {
+        while ((*bufText++ = *text++) != 0) {
```
Dragorn421, line 70:
```suggestion
        while ((*bufText++ = *text++) != '\0') {
```

Dragorn421 (approved):

fig02 (approved):

src/overlays/actors/ovl_Bg_Jya_Lift/z_bg_jya_lift.c
```diff
@@ -137,7 +137,7 @@ void BgJyaLift_Update(Actor* thisx, GlobalContext* globalCtx2) {
     }
     if ((this->dyna.unk_160 & 4) && ((this->unk_16B & 4) == 0)) {
         Camera_ChangeSetting(globalCtx->cameraPtrs[MAIN_CAM], CAM_SET_DIRECTED_YAW);
-    } else if (((this->dyna.unk_160) & 4) == 0 && ((this->unk_16B & 4)) &&
+    } else if (((this->dyna.unk_160) & 4) == 0 && (this->unk_16B & 4) &&
```
EllipticEllipsis, line 140:
```suggestion
    } else if (((this->dyna.unk_160 & 4) == 0) && (this->unk_16B & 4) &&
```

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.c
```diff
@@ -384,7 +384,7 @@ void ObjSwitch_FloorUp(ObjSwitch* this, GlobalContext* globalCtx) {
             CollisionCheck_SetAC(globalCtx, &globalCtx->colChkCtx, &this->tris.col.base);
         }
     } else {
-        switch ((this->dyna.actor.params >> 4 & 7)) {
+        switch (this->dyna.actor.params >> 4 & 7) {
```
EllipticEllipsis, line 387: These could do with () for disambiguation too, but they'll be replaced by macros soon enough anyway so I suppose it doesn't matter for now.

Roman971: Good point, but yeah there are a lot of expressions like these in the file so they should probably be cleaned up separately

PR #1222, opened 2022-05-11
Fix misc 12

Dragorn421:


src/code/z_kankyo.c
```diff
@@ -1638,14 +1638,14 @@ void Environment_DrawRain(GlobalContext* globalCtx, View* view, GraphicsContext*
...
+            u8 materialFlag = 0;
 
             for (i = 0; i < globalCtx->envCtx.unk_EE[1]; i++) {
-                if (!firstDone) {
+                if (materialFlag == 0) {
                     func_80093D84(gfxCtx);
                     gDPSetEnvColor(POLY_XLU_DISP++, 155, 155, 155, 0);
                     gDPSetPrimColor(POLY_XLU_DISP++, 0, 0, 255, 255, 255, 120);
-                    firstDone++;
+                    materialFlag++;
```
fig02, line 1648: this is probably something we need to discuss on a larger scale. But imo just because it becomes 1 via an increment doesnt mean its not a bool anymore. Its purpose is clearly to say that its true that the material has been applied

fig02: (I do dunderstand that if a bool type actually existed this wouldnt be allowed. But it doesnt, and so they could get away with setting true by incrementing like this. It still makes it more readable to keep it as its bool form imo)

fig02 (approved):

Roman971 (approved):

PR #1223, opened 2022-05-11
Doc misc 5

Dragorn421:


src/code/code_80097A00.c
```diff
@@ -11,26 +11,67 @@ u32 gBitFlags[] = {
...
+    0xF << (EQUIP_TYPE_TUNIC * 4),  // EQUIP_TYPE_TUNIC
+    0xF << (EQUIP_TYPE_BOOTS * 4),  // EQUIP_TYPE_BOOTS
 };
-u32 gUpgradeNegMasks[] = {
-    0xFFFFFFF8, 0xFFFFFFC7, 0xFFFFFE3F, 0xFFFFF1FF, 0xFFFFCFFF, 0xFFFE3FFF, 0xFFF1FFFF, 0xFF8FFFFF,
+u16 gEquipNegMasks[EQUIP_TYPE_MAX] = {
+    (u16) ~(0xF << (EQUIP_TYPE_SWORD * 4)),  // EQUIP_TYPE_SWORD
+    (u16) ~(0xF << (EQUIP_TYPE_SHIELD * 4)), // EQUIP_TYPE_SHIELD
+    (u16) ~(0xF << (EQUIP_TYPE_TUNIC * 4)),  // EQUIP_TYPE_TUNIC
+    (u16) ~(0xF << (EQUIP_TYPE_BOOTS * 4)),  // EQUIP_TYPE_BOOTS
```
fig02, line 24: i get it was probably done for consistency with the other arrays in the file. but the comments dont feel needed since the enum is right there in the definition

Dragorn421: I agree partially, imo deriving the corresponding type from the value would just be a (fair) assumption but I feel like alleviating any doubt is better

fig02 (approved):

src/overlays/actors/ovl_En_Ge1/z_en_ge1.c
```diff
@@ -629,6 +629,7 @@ void EnGe1_TalkNoPrize_Archery(EnGe1* this, GlobalContext* globalCtx) {
 void EnGe1_TalkAfterGame_Archery(EnGe1* this, GlobalContext* globalCtx) {
     CLEAR_EVENTINF(EVENTINF_HORSES_08);
     LOG_NUM("z_common_data.yabusame_total", gSaveContext.minigameScore, "../z_en_ge1.c", 1110);
+    // The debug string suggests an out-of-bounds read, see the other occurrence of this.
     LOG_NUM("z_common_data.memory.information.room_inf[127][ 0 ]", HIGH_SCORE(HS_HBA), "../z_en_ge1.c", 1111);
```
Roman971, line 633: I think the current theory is that their `sceneFlags` had a size of at least 128, so it might be better to mention that instead of referring to OOB which is supposedly less likely

Dragorn421: See the other comment on the similar issue in z_en_elf, should that one be amended too?

I'm not sure in what way to change it, I wrote it as "suggests we should have an out-of-bounds read" (given our structs), not "suggests the original code had an out of bounds read"

Maybe: "With the current `SaveContext` struct definition, the expression in the debug string is an out-of-bounds read, see the other occurrence of this."
? and also write more in the other spot?

Roman971: To me this currently reads as "suggests there was an out-of-bounds read", so yeah I would word it differently if that's not the intent. That being said, I don't think it's very useful to comment on what the debug string suggests we should do (and we generally don't do it in other places). And if we're going to comment on the debug string, I think it would be a lot more relevant to explain why our struct layout differs from theirs and is shown by these strings, if that makes sense

fig02 (approved):

Roman971 (approved):

PR #1227, opened 2022-05-16
Doc/Cleanup pass on object handling

Dragorn421:
- `ObjectStatus status[OBJECT_EXCHANGE_BANK_MAX];` -> `ObjectEntry slots[19];`
- `Object_InitBank` -> `Object_Init`
- `Object_Spawn` -> `Object_SpawnPersistent`
- `Object_GetIndex` -> `Object_GetSlot`
- "obj bank index" -> "object slot" (`objectCtx.slots[actor.objectSlot]`)
- Cleanup around usage of object slots, and the `WaitForObject` scheme
- other less important renames and cleanup

Sorry for the big PR, starting to cleanup/rename "obj bank index" stuff was a deeper rabbit hole than I expected it to be. It was far from a replace-all change too, the names in master aren't super consistent


include/z64.h
```diff
@@ -777,16 +777,19 @@ typedef struct {
     /* 0x08 */ DmaRequest  dmaRequest;
     /* 0x28 */ OSMesgQueue loadQueue;
     /* 0x40 */ OSMesg   loadMsg;
-} ObjectStatus; // size = 0x44
+} ObjectLoadEntry; // size = 0x44
+
+// 1000 kilobytes
+#define OBJECT_SPACE_SIZE_BASE (1000 * 1024)
```
Roman971, line 783: I suggest having this define in `z_scene.c`, maybe next to the function where it's used, since it's not intended for use externally I'd say

include/z64actor.h
```diff
@@ -147,7 +147,7 @@ typedef struct Actor {
     /* 0x004 */ u32 flags; // Flags used for various purposes
     /* 0x008 */ PosRot home; // Initial position/rotation when spawned. Can be used for other purposes
     /* 0x01C */ s16 params; // Configurable variable set by the actor's spawn data; original name: "args_data"
-    /* 0x01E */ s8 objBankIndex; // Object bank index of the actor's object dependency; original name: "bank"
+    /* 0x01E */ s8 objectLoadEntryIndex; // Object load entry index corresponding to the actor's object; original name: "bank"
```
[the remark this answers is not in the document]
Dragorn421: Get back to me on object -> obj after seeing what I changed this to :sweat_smile:

include/z64.h
```diff
@@ -777,16 +777,19 @@ typedef struct {
...
-    /* 0x0008 */ u8     num; // number of objects in bank
-    /* 0x0009 */ u8     unk_09;
-    /* 0x000A */ u8     mainKeepIndex; // "gameplay_keep" index in bank
-    /* 0x000B */ u8     subKeepIndex; // "gameplay_field_keep" or "gameplay_dangeon_keep" index in bank
-    /* 0x000C */ ObjectStatus status[OBJECT_EXCHANGE_BANK_MAX];
+    /* 0x0008 */ u8     loadEntriesCount; // total amount of used load entries
+    /* 0x0009 */ u8     persistentLoadEntriesCount; // amount of load entries that won't be reused when loading a new object list (when loading a new room)
+    /* 0x000A */ u8     mainKeepIndex; // "gameplay_keep" index in loadEntries
+    /* 0x000B */ u8     subKeepIndex; // "gameplay_field_keep" or "gameplay_dangeon_keep" index in loadEntries
+    /* 0x000C */ ObjectLoadEntry loadEntries[19];
```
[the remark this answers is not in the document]
Dragorn421: I went the even crazier route of using "entry"/"object entry" everywhere, it actually works out quite well. no more bank/load discussion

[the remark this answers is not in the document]
Roman971: I'm not a fan of using that everywhere personally. I think it's good to use the term entry for the struct in the context, but we lose a lot of information by not having "index" in the other names. This is especially apparent for things like `objectEntry = Object_GetEntry(&play->objectCtx, actorInit->objectId);`, which are everywhere and are specifically intended to get the index which can then be used to get info on the object. I would agree with the name if the function returned a pointer to the entry ofc, but that's not the case so it's confusing imo.

That being said, I don't think "index" or "entry index" alone is enough, because it's not just the index in the list of object entries, it's a fixed value that won't change in the current room, so actors often store it once and continue using it over their lifetime. It's almost an ID in that sense, but we've been avoiding that in order to not get it confused with the actual object ID.

All this is why I think the term "bank index" fits the best among everything we've mentioned so far. It's recognizable enough that you can quickly associate it with this pattern and its usage as an index, while not being too easy to mistake with the actual "object id".

It could be nice to find an even better alternative that fits all these boxes tho, but I don't have any ideas myself atm

include/functions.h
```diff
@@ -1181,11 +1181,11 @@ void Sample_Init(GameState* thisx);
...
-s32 Object_Spawn(ObjectContext* objectCtx, s16 objectId);
-void Object_InitBank(PlayState* play, ObjectContext* objectCtx);
-void Object_UpdateBank(ObjectContext* objectCtx);
-s32 Object_GetIndex(ObjectContext* objectCtx, s16 objectId);
-s32 Object_IsLoaded(ObjectContext* objectCtx, s32 bankIndex);
+s32 Object_SpawnPersistent(ObjectContext* objectCtx, s16 objectId);
+void Object_InitContext(PlayState* play, ObjectContext* objectCtx);
+void Object_UpdateLoadEntries(ObjectContext* objectCtx);
+s32 Object_GetLoadEntryIndex(ObjectContext* objectCtx, s16 objectId);
+s32 Object_IsLoadEntryLoaded(ObjectContext* objectCtx, s32 loadEntryIndex);
```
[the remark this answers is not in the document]
Roman971: I agree we can't really abstract the loading part away, but I don't think we need to have "load" in every name either, especially since anyone reading a bit of the code will quickly see all the `Object_IsLoaded`-type calls stuff and figure out it's important

include/z64.h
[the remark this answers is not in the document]
Dragorn421: I think the only reason "bank" would be better than "entry" is because it's what has been used so far / what comes from the debug string, there doesn't seem to be anything else to it

Furthermore the only way I see "bank" making sense is the "(computing) A contiguous block of memory that is of fixed, hardware-dependent size, but often larger than a page and partitioning the memory such that two distinct banks do not overlap." sense ( https://en.wiktionary.org/wiki/bank#Noun_3 ("bank" is also a super overloaded word)) which is somewhat accurate (besides the load stuff and the entries being only a pointer to the "bank", not the "bank" itself), but doesn't really have anything to do with the engine's objects in particular

"entry" is more generic which makes it counter intuitively more appropriate than "bank" which as a word is overloaded

I think it's ok to use `entry` as an index without saying `entryIndex`. If you want I can change the few `entry` pointers to `entryPtr`, that would be less verbose than changing everything else to `entryIndex`

src/code/z_scene.c
```diff
@@ -3,116 +3,126 @@
 
 RomFile sNaviMsgFiles[];
 
-s32 Object_Spawn(ObjectContext* objectCtx, s16 objectId) {
+// called by
+// - initcontext: for gkeep
+// - spawn list scene command: for link's object
+// - special files scene command: for subkeep object if any
+s32 Object_SpawnPersistent(ObjectContext* objectCtx, s16 objectId) {
```
fig02, line 10: I dont think its important to note what calls it
However I think the danger of using this function should be pointed out. I didnt realize until now the way that unk_09 (`numPersistentEntries`) works.
The wording can be worked on if needed, but these are the important points i think
```
Spawn an object file of the specified ID that will persist through room changes.

NOTE: This function is NOT meant to be called externally to spawn object files on the fly.
When an object is spawned with this function, any objects that come before it in the entry list will be treated as persistent. 
This function is only meant to be called internally by the object system BEFORE the object list from the room is processed.
```

fig02, line 10: I dont think its important to note what calls it
However I think the danger of using this function should be pointed out. I didnt realize until now the way that unk_09 (`numPersistentEntries`) works.
The wording can be worked on if needed, but these are the important points i think
```
Spawn an object file of the specified ID that will persist through room changes.

NOTE: This function is NOT meant to be called externally to spawn object files on the fly.
When an object is spawned with this function, any objects that come before it in the entry list will be treated as persistent. 
This function is only meant to be called internally by the object system BEFORE the object list from the room is processed.
```

fig02: also, we should decide on persistent vs permanent. Actors use permanent to describe the same thing for their overlay allocation type. itd be good if they both used the same

include/z64.h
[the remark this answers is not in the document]
fig02: I am usually a hater of the word "index" being added to things, but I think it does matter here. just `entry` makes it seem like the entry itself, it just doesnt feel right

fig02:
@MNGoldenEagle suggested using the terminology "slot" to help with the "entry" discussion, and I think this could work really well. Ive already called them slots on my own terms when describing things about the system outside of the codebase

`ObjectEntry entries[19];` -> `ObjectEntry slots[19];` (imo the struct can stay as entry, there are slots for different entries)
`Object_GetEntry` -> `Object_GetSlot`
`Object_IsEntryLoaded` -> `Object_IsLoaded` (see my pr comment above on why i think this change makes sense)
`actor.objectEntry` -> `actor.objectSlot`

unlike with entry, I dont think the word "index" is needed to clarify that its a number for the entry/slot. It works out well imo

include/functions.h
```diff
@@ -456,7 +456,7 @@ void func_800304DC(PlayState* play, ActorContext* actorCtx, ActorEntry* actorEnt
 void Actor_UpdateAll(PlayState* play, ActorContext* actorCtx);
 s32 func_800314D4(PlayState* play, Actor* actor, Vec3f* arg2, f32 arg3);
 void func_800315AC(PlayState* play, ActorContext* actorCtx);
-void func_80031A28(PlayState* play, ActorContext* actorCtx);
+void Actor_KillActorsWithUnloadedObject(PlayState* play, ActorContext* actorCtx);
```
fig02, line 459: feel like theres a more elegant way to name this but I dont have any immiediate idea. Mainly want to remove the redundant use of the word "actor"
`Actor_CleanupBasedOnObject` `Actor_KillAllWithMissingObject`, idk

Dragorn421: Note it's used exactly once

I don't have much better suggestion than what it is, yours or `Actor_KillActorsWhichObjectIsNotLoaded` idk

I think it should have `Kill` though since it `Actor_Kill`s and there are the various renowned ways to "remove" actors

src/code/z_scene.c
```diff
@@ -3,116 +3,135 @@
...
     }
 
-    objectCtx->num = objectCtx->unk_09 = 0;
-    objectCtx->mainKeepIndex = objectCtx->subKeepIndex = 0;
+    objectCtx->numEntries = objectCtx->numPersistentEntries = 0;
+    objectCtx->mainKeepSlot = objectCtx->subKeepSlot = 0;
 
-    for (i = 0; i < OBJECT_EXCHANGE_BANK_MAX; i++) {
-        objectCtx->status[i].id = OBJECT_INVALID;
+    for (i = 0; i < ARRAY_COUNT(objectCtx->slots); i++) {
```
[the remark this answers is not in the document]
Dragorn421: I don't really see a reason it should be a define tbh, besides the debug string saying it originally was

I also think the code reads better with array_count since you already gotta watch out for iterations from / up to `numEntries`/`numPersistentEntries`, no need to add another name imo?

include/functions.h
fig02: `Actor_KillAllWithMissingObject` is probably decent then. it removes the redundant "actor" I didnt like, and I think abstracting it to "missing" instead of unloaded makes it sound a bit less verbose, gets the point across that the idea is to clean up the actor list for those whos object is now not there after changing rooms

Dragorn421: alright sounds good, I'll change it to that

src/code/z_scene.c
```diff
@@ -3,116 +3,135 @@
...
+ * This only starts loading the object, the data will not be available immediately when the function returns.
+ *
+ * Used for gameplay_keep, Link's object, and the subkeep if any.
+ *
+ * @return The new object slot corresponding to the requested object ID.
+ *
+ * @note This function is not meant to be called externally to spawn object files on the fly.
+ * When an object is spawned with this function, all objects that come before it in the entry list will be treated as
+ * persistent, which will likely cause the memory available for object space to run out.
+ * This function is only meant to be called internally on scene load, before the object list from any room is processed.
```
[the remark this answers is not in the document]
Dragorn421: Good point about the slots, I'll amend the comment

I don't think capital letters are good practice, if it's so important we should more properly use `@warning`
(doxygen has `@remark`, `@note`, `@attention`, `@warning`)

<details>

```c
/**
 * Spawn an object file of a specified ID that will persist through room changes.
 *
 * This only starts loading the object, the data will not be available immediately when the function returns.
 *
 * Used for gameplay_keep, Link's object, and the subkeep if any.
 *
 * @return The new object slot corresponding to the requested object ID.
 *
 * @remark This function is not meant to be called externally to spawn object files on the fly.
 * @note When an object is spawned with this function, all objects that come before it in the entry list will be treated as
 * @attention persistent, which will likely cause the memory available for object space to run out.
 * @warning This function is only meant to be called internally on scene load, before the object list from any room is processed.
 */
```

image

</details>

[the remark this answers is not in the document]
fig02: I think theyre fine to use when emphasis is needed. I dont want to start using yet another doxygen tag for this, it just makes it read nicer with the emphasis that is intended in the message lol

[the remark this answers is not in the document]
Dragorn421: I really don't like documentation screaming at me or showing emotion

This function has been wrongly used for a while, but just naming it with `Persistent` and the `@note` is enough imo to document the behavior.
And when I'll document the rest of object handling I'll add an example (or just a `@see`) of how to actually request objects be loaded I guess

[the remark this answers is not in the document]
fig02: Guess others will have to weigh in, cause I just dont agree. When we dont have access to things like italics and bold letters, caps are fine to use for important words like NOT, imo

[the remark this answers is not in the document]
Dragorn421: Oh we can do italics and bold
```c
@note *This function* is **not** meant to
```
image

Not that I think it's appropriate here either but I guess it's still more professional looking than caps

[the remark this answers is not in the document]
fig02: I mean within the code itself, we cant italicize or bold. I personally dont care too much for how doxygen output looks, the appearance in the code is a priority for me. And the caps gives the exact emphasis that I want on the comments

Some of our existing docs already do this, and even gbi from nintendo does. I dont see it as "unprofessional", just making use of what you have to get the correct tone across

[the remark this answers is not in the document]
Dragorn421: And for me the screaming tone is incorrect, as documentation shouldn't carry emotion

More opinions needed :smile:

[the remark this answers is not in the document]
EllipticEllipsis: As I said in #1265 , I think it's worth having something stronger than `@note` for this. I also don't see why we shouldn't use Markdown, at least, since it's used on Discord, GitHub, in the README.md, ...

src/code/z_scene.c
```diff
@@ -3,116 +3,133 @@
 
 RomFile sNaviMsgFiles[];
 
-s32 Object_Spawn(ObjectContext* objectCtx, s16 objectId) {
+/**
+ * Spawn an object file of a specified ID that will persist through room changes.
+ *
+ * This only starts loading the object, the data will not be available immediately when the function returns.
```
Roman971, line 9: This function should fully load the object afaict, since it uses `DmaMgr_SendRequest1` which blocks the thread until the dma is done. And I don't think there is anything else that would prevent the object from being available immediately (?)

src/code/z_scene.c
```diff
@@ -3,116 +3,133 @@
...
-    return objectCtx->num - 1;
+    return objectCtx->numEntries - 1;
 }
 
-void Object_InitBank(PlayState* play, ObjectContext* objectCtx) {
-    PlayState* play2 = play; // Needs to be a new variable to match (possibly a sub struct?)
+// 1000 kilobytes
+#define OBJECT_SPACE_SIZE_BASE (1000 * 1024)
+
+void Object_InitContext(PlayState* play, ObjectContext* objectCtx) {
```
Roman971, line 52: `Object_InitContext` doesn't seem great since it makes it sound like the function only initializes the context struct, and it doesn't make it clear that it also allocates the space and spawns the initial object(s), which is just as important if not more imo.
However I don't really have a suggestion either, aside from referring to the space and system as a whole in some way (ie. the "bank" in the previous naming)

Roman971 (commented):
To be honest I think the switch to the term "object slot" over "object bank index" is a downgrade, but I agree it seems better than the "entry" names, even tho it shares some of its issues imo (particularly with names like `Object_GetSlot` whose purpose and return value isn't clear at first glance).

That being said, it's probably the best we'll get if we want to drop both "bank" and "index" entirely, so I'm fine with it if other people are.

I do have a few mostly unrelated comments however

src/code/z_scene.c
Dragorn421: My mistake :fearful:

src/overlays/actors/ovl_Bg_Haka_Sgami/z_bg_haka_sgami.c
```diff
@@ -169,16 +169,16 @@ void BgHakaSgami_Init(Actor* thisx, PlayState* play) {
     Effect_Add(play, &this->blureEffectIndex[1], EFFECT_BLURE1, 0, 0, &blureInit);
 
     if (thisx->params == SCYTHE_TRAP_SHADOW_TEMPLE) {
-        this->requiredObjBankIndex = Object_GetIndex(&play->objectCtx, OBJECT_HAKA_OBJECTS);
+        this->waitObjectSlot = Object_GetSlot(&play->objectCtx, OBJECT_HAKA_OBJECTS);
```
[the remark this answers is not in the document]
fig02: fwiw I agree here. the wait makes sense for the function that ends up waiting, but not the slot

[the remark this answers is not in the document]
Dragorn421: Changed

I don't remember if I did clean-up on the `_WaitForObject` function names but I guess I'll leave what's left to another time

src/code/z_scene.c
Dragorn421: -> `Object_Init` I guess

fig02 (approved):

src/code/z_scene.c
Roman971: `Object_Init` sounds very confusing since it intializes the whole space, and not a specific object, which kinda goes against how we use `<Thing>_Init` in the rest of the codebase (eg. <Actor>_Init)

Dragorn421: `Interface_Init`, `Message_Init`, `Map_Init`, `MapMark_Init`, `Skybox_Setup`, `Skybox_Init` use `GameState_Alloc` and load stuff then, maybe `Object_Setup` ? `Object_SetupContext` ?

Or there is `func_80097dd8`, there are more Object_ system functions to name anyway, and the other function `func_80096FE8` needs a name but also has `GameState_Alloc`/load data combo and I don't think it'll work as a `_Init` (the call to it is also weirdly hidden in a printf...)

Dragorn421: we can also change the system name to "objects" since it handles several objects anyway. `Objects_Init`

Roman971: `Object_Setup` has the same issue as `Object_Init` imo, it sounds like it does the work needed for an actual "object". And `Object_SetupContext` isn't really better than `Object_InitContext` in that the function does more than just context initialization or setup. Tho I do think `Object_InitContext` is better than `Object_Init`/`Object_Setup`, fwiw

Among the init functions you mentioned, almost all are init functions for a system where there is no ambiguity between initializing the system or one of the instances it manages (ie. there is only 1 "interface" or 1 "skybox" at a time, and that's what the systems manage), so the names work well. The only exception would be `Message_Init`, since you might mistake it as initializing a specific text aka. a message, and I would argue it's also misleading to some extent (tho we primarily use "text" in our names so it might be ok).
But for a system like objects, I think we should use the same distinction made in the actor or effect systems for instance, where we use names like `Actor_UpdateAll` to differentiate from `Actor_Update`

Dragorn421: there is `Object_UpdateEntries`, `Object_InitEntries` ?

Roman971: It doesn't initialize the entries either tho, it just intializes the space (aka. the bank).

And to be clear I don't think `Object_InitContext` is unusable either, it's probably the best among everything that's been suggested so far. I just think it's a downgrade over a clear name like `Object_InitBank` (tho you can change "bank" by whichever equivalent term you want to replace it with). And I feel like we're going to have similar issues when naming the rest of the system in the future, but that might just be me

Dragorn421: the main logic is in `Scene_CommandObjectList` (popping the slots from the first mismatch between the two object lists), the few functions left are unrelated to the memory management aspect afaict (well, one is 100% memory management but just loading back objects that were overwritten by kaleidoscope, to exactly where they were before)

A critic of the suggested change could also be that `ObjectEntry slots[19];` looks weird, maybe `ObjectEntry`'s name should change but I can't find a much better name, and the stuff using "entry" is internal anyway.
The main point with changing objBankIndex to objectSlot is to get a unique and less confusing (objBankIndex // objIndex) way to refer to these entries and externally only `ObjectContext.slots` is used
(that confusion was also a straight path from the devs' names apparently:
https://github.com/zeldaret/oot/blob/2e18eb334bad77db0625549c62849333705214b8/src/code/z_actor.c#L2747-L2749
this debug strings calls an objectId "bank", this do be great naming)

fig02: I think regardless of how the original game is using the term bank, we can just use it as a term to describe the thing that holds all of these entries. Or we can use the term "pool", we just need something. Point is its the name of the thing that has slots that you have these load entries in

src/code/z_scene.c
```diff
@@ -3,116 +3,133 @@
...
+    objectCtx->numPersistentEntries = objectCtx->numEntries;
 
-    return objectCtx->num - 1;
+    return objectCtx->numEntries - 1;
 }
 
-void Object_InitBank(PlayState* play, ObjectContext* objectCtx) {
-    PlayState* play2 = play; // Needs to be a new variable to match (possibly a sub struct?)
+// 1000 kilobytes
+#define OBJECT_SPACE_SIZE_BASE (1000 * 1024)
```
mzxrules, line 50: I think this define should be removed. There's no consistent relationship between the default spaceSize and the spaceSize set in the other use cases. To illustrate, imagine a mod that increases the amount of memory available to the game without modifying any of the assets. If we increase `OBJECT_SPACE_SIZE_BASE` to `(1050 * 1024)`, then the other cases will needlessly allocate an extra 50 kilobytes of ram automatically.

Dragorn421: I'm not sure how you want this to be addressed...
If you change something then yeah you should be aware of the effects

Maybe you want something like
`spaceSize = OBJECT_SPACE_SIZE_BASE + (50 * 1024);`
->
`spaceSize = MAX(OBJECT_SPACE_SIZE_BASE, 1050 * 1024);`
?

src/code/z_scene.c
Dragorn421: *Deletes comment to avoid getting in the "bank" argument*

Did we rule out `Object_InitSpace` ?

src/code/z_scene.c
mzxrules: Sort of yes, but is `OBJECT_SPACE_SIZE_BASE` actually representative of the minimum spaceSize requirement though?

Dragorn421: I know that it's not the minimum for pause menu to work though it's very tight

BASE was more like "baseline" since it's the "default" for most scenes

src/code/z_scene.c
Dragorn421: `Object_Allocate` ? c.f. `BgCheck_Allocate`

fig02: `Object_Allocate` on the surface sounds like its allocating a single object. even tho of course thats not how the system works at all, I think it gives the wrong impression. and the function is doing more than just allocating some space anyway

`Object_InitSpace` is a step in the right direction I suppose, but I think i do agree with roman here that it needs some sort of concrete name. "space" isnt giving a name to the thing thats holding the slots, its just referencing to this chunk of memory that this thing will go. So it doesnt feel 100% right to me

I honestly dont see what the huge issue is with bank. The bank can exist as its own entity that has slots like we are currently calling them

Dragorn421: We currently have the object context, which manages the object space by dividing it between slots

I think that's enough words and terminology to describe the system, I don't see what a single use of "bank" would refer to
That's why I thought InitContext or InitSpace would work, they're established and well-used terms within the system, countrarily to Bank

To summarize, if I read "init bank" in the object system I'd be like "what is bank"

hensldm: Been reading/thinking through this for a while now.
Currently it seems like `space` is already kind of used for this terminology referring to the full object space (based on the struct members in ObjectContext `spaceStart` and `spaceEnd`, and the stack variable `spaceSize` (unless these are for something else)). Based on that I think `InitSpace` would be fine. Though I think `bank` or `pool` also makes sense for this terminology referring to the full object space. Perhaps we could just use `slots` to refer to this space as well?

src/code/z_scene.c
hensldm: I'll say I'm not sure how much the macro adds, and don't really see a correlation between `OBJECT_SPACE_SIZE_BASE` and the size allocated in each scenario.
`spaceSize = MAX(OBJECT_SPACE_SIZE_BASE, 1050 * 1024);` seems better imo.

src/code/z_scene.c
fig02: > Perhaps we could just use slots to refer to this space as well?

I think its important to not conflate the actual space/buffer where the objects go and the slots. The slots are definitely separate, and the amount of slots and their existence doesnt really have any bearing on the buffer and how big it is. the 19 slots oot has is pretty arbitrary, and isnt really whats getting initialized here

fig02: as for InitSpace, it def makes sense based on how we are calling this space with variables like `spaceStart` and `spaceEnd`, but something just sounds weird about init space lol
Something like bank feels a little more concrete, but has been a contentious term to use so far. 
Maybe just buffer since thats all it really is. InitBuffer/AllocateBuffer, idk

fig02: `InitBuffer` is probably more appropriate since on top of allocating the space it also spawns gameplay keep

Dragorn421: I don't have much to add to https://github.com/zeldaret/oot/pull/1227#discussion_r963341163 (I have the same objections to Buffer as Bank, in the sense that if I was to agree to Buffer I would also agree to Bank)

How I see it is `InitSpace` would initialize the space (allocate the memory) and its partition (resets the slots), and start filling it by spawning gkeep

fig02: oh, yeah I did forget that there is a loop that set the slot ids to OBJECT_INVALID.

I dont think buffer has the same issues/implications as bank though. buffer tells it like it is, its a raw buffer with no sort of managment system (ie its not a heap/linked list/stack or anything). Whereas bank is a term that doesnt really give you any information, and as you say in that old comment, you ask youself "what is the bank". buffer is pretty clear, its the buffer where the files themselves go.

hensldm (approved):
Seems good to me. Hopefully we can agree on a name for `func_80097DD8` eventually.

src/code/z_scene.c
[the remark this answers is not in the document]
AngheloAlf: Personally I prefer the ARRAY_COUNT solution instead of making extra defines for this kind of stuff.
It may be worth noting somewhere this was a macro and the original name (as a way of preserving history or something idk), probably in the struct itself

src/code/z_scene.c
AngheloAlf: Personally I think `Object_InitContext` is a fine name, it isn't like we can add all the stuff a function does in the name itself. Something similar happens on `Actor_InitContext` where that function also resets the loaded entries of the overlay table, initialize the target context and the farore's wind stuff
I think it could be noted as a comment what this function does, like allocating space and initializing objects.

AngheloAlf (approved):
LGTM

src/code/z_scene.c
```diff
@@ -3,116 +3,130 @@
...
+    objectCtx->numPersistentEntries = objectCtx->numEntries;
 
-    return objectCtx->num - 1;
+    return objectCtx->numEntries - 1;
 }
 
-void Object_InitBank(PlayState* play, ObjectContext* objectCtx) {
-    PlayState* play2 = play; // Needs to be a new variable to match (possibly a sub struct?)
+void func_80097DD8(PlayState* play, ObjectContext* objectCtx) {
+    PlayState* play2 = play;
```
fig02, line 50: Starting this as a separate comment to start fresh since it has been so long, though technically it is a continuation of https://github.com/zeldaret/oot/pull/1227#discussion_r898398992

I believe that `Object_InitContext` is a fine name for this. While I do understand concerns that it initializes more than just the literal `ObjectContext` struct, it gets the point of the function across. We use this method of naming in similar places like `Actor_InitContext` which does more than just init the literal context structure. It initializes the overlay table, copies the identity matrix to the matrices in `play`, and inits fw data which mostly concerns save context. It is still actor related things. 

In that same vein, I think `Object_InitContext` is more than suitable for this function name in the same way. Its for the Object system generally. 

Note, I named this function with InitContext in the mm version of this PR, which is still under review: https://github.com/zeldaret/mm/pull/1361

mzxrules: My current opinion is that `Object_InitContext` feels like a bad name moreso because the system name of `Object_` isn't very good and ideally should be renamed. But if that's not on the cards atm, I agree with fig in calling it `Object_InitContext` just to get it merged for now. 

But if we're up to giving the system a better name, I find it hard not to borrow from og. Perhaps it's because I've been doing a lot of Atari2600 programming lately, but I think I'd prefer to keep the bank swapping terminology given how similar it is to how you'd handle bank switching in older hardware, just in a virtualized form.

For renames I'd personally would like to go with something like this:

`Object_` -> `ObjectExchange_`
`ObjectContext` -> `ObjectExchange`
`ObjectEntry` -> `ObjectBank` or `ObjectExchangeBank`
`ObjectContext.slot` -> `ObjectExchange.bankSlot`

`Object_SpawnPersistent` -> `ObjectExchange_AllocBank`?
`func_80097DD8` -> `ObjectExchange_Init`
`Object_UpdateEntries` -> `ObjectExchange_FetchBanksAsync`
`Object_GetSlot` -> `ObjectExchange_GetBankSlot`
`Object_IsLoaded` -> `ObjectExchange_IsBankLoaded`
`func_800981B8` -> `ObjectExchange_FetchBanksSync`
`func_800982FC` -> `ObjectExchange_BankSwitch`

Dragorn421: Define "bank"

mzxrules: My personal opinion as a programmer is that a bank / memory bank is just a segment of contiguous random access memory. But the definitions I've seen all associate memory banks to physical hardware. For example in bank swapping you aren't spending time transferring data into main memory; a bank swap switches out what physical memory the main processor memory is able to access directly. 

In my atari 2600 game for example, I use an 3E+ bankswitching scheme which splits processor memory into 4 1k segments that can be arbitrarily swapped with 64 1k read-only memory segments and 64 0.5k read 0.5k write segments. Bankswitching is done by passing in an 8 bit value storing the main memory segment id and the 0-63 memory bank id to either the read-only memory or read-write register.

More back to Z64, if I were to describe the object swapping system in terms of bank swapping, `ObjectEntry` would represent the processor accessible memory banks, while the object files themselves would represent external memory banks you'd swap in.

Dragorn421: So with the definition of "bank = an area of memory that we can effectively easily swap the data it maps to"
(note: "bank" is not even a good term to use at all from being so poorly defined)

Then the object system doesn't have banks:
You can't swap data in slots besides the highest-indexed/top one, so that fails that definition of "bank"

hensldm: Fwiw, `Object_InitContext` seems fine/best to me.

src/code/z_scene.c
```diff
@@ -250,50 +264,52 @@ void Scene_CommandObjectList(PlayState* play, SceneCmd* cmd) {
...
         i++;
         k++;
-        objectEntry++;
-        status++;
+        objectListEntry++;
+        entry++;
     }
 
-    ASSERT(cmd->objectList.length <= OBJECT_EXCHANGE_BANK_MAX,
+    ASSERT(cmd->objectList.length <= ARRAY_COUNT(play->objectCtx.slots),
```
mzxrules, line 299: Noticed that the ASSERT here is flawed; It should be checking for `cmd->objectList.length + play->objectCtx.numPersistentEntries <= ARRAY_COUNT(play->objectCtx.slots)` instead

fig02: good catch, though yeah we cant do anything about it cause its their mistake. can add a comment though

src/code/z_scene.c
mzxrules: > Can we use a naming convention that better adjusts to this game (or at least to the N64 hardware) instead of terminology from the Atari 2600? 

Bank switching was still a relevant technique in console game development in the 90s and early 2000s. I only mention the 2600 because I'm more familiar with the system than the Gameboy/Gameboy Color or SNES.

@Dragorn421 Dunno how to best respond, so i'll start off by admitting that I missed that the variable `objectCtx->slots[i+1].segment` where `i` is the updated slot happens to be updated when `Object_SpawnPersistent` and `func_800982FC` are called (though the later is a bit particular since the update occurs in `Scene_CommandObjectList`, meaning the system doesn't fully manage it's memory internally). 

>(note: "bank" is not even a good term to use at all from being so poorly defined)

I don't think so.

> So with the definition of "bank = an area of memory that we can effectively easily swap the data it maps to"
Then the object system doesn't have banks: You can't swap data in slots besides the highest-indexed/top one, so that fails that definition of "bank"

Weird thing about that. It's likely intended that slots are reserved sequentially, but `func_800982FC`'s implementation allows any arbitrary slot to be swapped over with a new object file. If you had same sized segments like the warp pad texture objects, you could safely hotswap them with `func_800982FC` calls, making `func_800982FC` behave very similarly to swapping banks, with about the same level of safety.

I've thought more on it, and would like to suggest this set of function names instead. I don't think these names are 100% perfect, but at the same time the original system could arguably use a few rewrites to make things cleaner/safer.

`Object_SpawnPersistent` -> `ObjectExchange_AllocPersistentSegment
`func_80097DD8` -> `ObjectExchange_Init`
`Object_UpdateEntries` -> `ObjectExchange_FetchSegmentsAsync`
`Object_GetSlot` -> `ObjectExchange_FindSlot` <- could stay Get, but Find is a little more accurate imo. 
`Object_IsLoaded` -> `ObjectExchange_IsSegmentLoaded`
`func_800981B8` -> `ObjectExchange_FetchSegmentsSync`
`func_800982FC` -> `ObjectExchange_AllocSegment`

I'm not really a fan of the concept of object files being "spawned".

src/code/z_scene.c
fig02: @Dragorn421 can add this in this pr if you like, would even consider it an `@bug `

Dragorn421:
I would like the names for object_initcontext, object_spawnpersistent, and others, as well as commenting on the flawed assert, to be done in another PR (get this one over with)

fig02:
`Obejct_SpawnPersistent` is already named? what else needs to be done there?
I think it makes most sense to name Object_InitContext with this set of changes cause of how important it is. But any other cleanups, I agree

Dragorn421:
I was referring to mzx' suggestions which seemed to put an emphasis on the spawnpersistent name
I named Object_InitContext

fig02 (approved):

Roman971 (approved):

PR #1229, opened 2022-05-17
PlayState rename preparations

fig02:
- `Gameplay_` -> `Play_`
- play refers to itself as `this` internally

The actual rename will take place in a separate pr


fig02:
idea would be to call it `play` externally, including in declarations in functions.h.

Theres a discussion to be had about the inconsistencies that introduces. The options are to call it `play` in the external function declarations (current), or make the "public" functions in z_play.c use `play` instead of `this`

I think what I have currently makes more sense, as that inconsistency would be pretty confusing in z_play itself. The function declarations technically dont even need a name for the variable, but it helps that they do.
And fwiw intellisense (and probably other ides) will show the name from the header declaration, not from the function itself. So when trying to use these functions you would see `play`

Roman971 (approved):

PR #1230, opened 2022-05-17
Documentation pass centered on z_player_lib

Dragorn421:
- Name a bunch of data lying in z_player_lib
- For usage of arrays listed in `sPlayerDListGroups` (e.g. `gPlayerLeftHandOpenDLs`), consistently use pointer arithmetics instead of array indexing as the operations happen in an inconsistent order across code (`+4`,`+age`,`+lodoffset`,`+shield*4`,...), and arrays are 1D anyway
- Name, comment, fixup various things in z_player_lib


Roman971 (approved):
I'm not really not a fan of the array access changes since so much pointer arithmetic is a downgrade imo, even for the sake of consistency, but it's not a big deal either way

PR #1233, opened 2022-05-19
Doc pass on scene draw configs: names

Dragorn421:
Name every draw config
- for draw configs used by a single scene, name the config from the scene (debug) name (eg `SCENE_SPOTxx`)
- for draw configs used by several similarly named scenes, name the config by the common name stem (e.g. `BDAN`, `HAIRAL_NIWA`)
- for draw configs used by several scenes fitting a description, name the config by that description (GanonCastleExterior, FairyFountain, GreatFairyFountain)
- for other cases, name the draw config based on what it is used for (`GRAVE_EXIT_LIGHT_SHINING`, `CALM_WATER`)


include/tables/scene_table.h
```diff
@@ -9,114 +9,114 @@
...
-/* 0x5E */ DEFINE_SCENE(spot13_scene, g_pn_37, SCENE_SPOT13, 12, 0, 0)
-/* 0x5F */ DEFINE_SCENE(spot15_scene, g_pn_38, SCENE_SPOT15, 13, 0, 0)
-/* 0x60 */ DEFINE_SCENE(spot16_scene, g_pn_39, SCENE_SPOT16, 14, 0, 0)
-/* 0x61 */ DEFINE_SCENE(spot17_scene, g_pn_40, SCENE_SPOT17, 15, 0, 0)
-/* 0x62 */ DEFINE_SCENE(spot18_scene, g_pn_41, SCENE_SPOT18, 16, 0, 0)
-/* 0x63 */ DEFINE_SCENE(spot20_scene, g_pn_42, SCENE_SPOT20, 17, 0, 0)
-/* 0x64 */ DEFINE_SCENE(ganon_tou_scene, g_pn_43, SCENE_GANON_TOU, 36, 0, 0)
+/* 0x00 */ DEFINE_SCENE(ydan_scene, g_pn_06, SCENE_YDAN, SCENE_DRAW_CONFIG_YDAN, 1, 2)
+/* 0x01 */ DEFINE_SCENE(ddan_scene, g_pn_08, SCENE_DDAN, SCENE_DRAW_CONFIG_DDAN, 1, 3)
+/* 0x02 */ DEFINE_SCENE(bdan_scene, g_pn_07, SCENE_BDAN, SCENE_DRAW_CONFIG_BDAN, 1, 4)
```
fig02, line 14: I know the intention is for the prefix to be `SCENE_DRAW_CONFIG_`, but that first underscore makes them seem like `SCENE_`, especially when placed next to the actual scene enum in the table like this

I would personally be ok with just `DRAW_CONFIG_`, but not sure if thats too general for everyones tastes. alternatively we can do like SCENECFG, idk

include/tables/scene_table.h
```diff
@@ -9,114 +9,114 @@
...
+/* 0x13 */ DEFINE_SCENE(bdan_boss_scene, none, SCENE_BDAN_BOSS, SCENE_DRAW_CONFIG_BDAN, 0, 0)
+/* 0x14 */ DEFINE_SCENE(moribossroom_scene, none, SCENE_MORIBOSSROOM, SCENE_DRAW_CONFIG_NONE, 1, 0)
+/* 0x15 */ DEFINE_SCENE(FIRE_bs_scene, none, SCENE_FIRE_BS, SCENE_DRAW_CONFIG_HIDAN, 0, 0)
+/* 0x16 */ DEFINE_SCENE(MIZUsin_bs_scene, none, SCENE_MIZUSIN_BS, SCENE_DRAW_CONFIG_MIZUSIN_BS, 0, 0)
+/* 0x17 */ DEFINE_SCENE(jyasinboss_scene, none, SCENE_JYASINBOSS, SCENE_DRAW_CONFIG_NONE, 0, 0)
+/* 0x18 */ DEFINE_SCENE(HAKAdan_bs_scene, none, SCENE_HAKADAN_BS, SCENE_DRAW_CONFIG_HAKADAN, 0, 0)
+/* 0x19 */ DEFINE_SCENE(ganon_boss_scene, none, SCENE_GANON_BOSS, SCENE_DRAW_CONFIG_NONE, 0, 0)
+/* 0x1A */ DEFINE_SCENE(ganon_final_scene, none, SCENE_GANON_FINAL, SCENE_DRAW_CONFIG_GANON_FINAL, 0, 0)
+/* 0x1B */ DEFINE_SCENE(entra_scene, none, SCENE_ENTRA, SCENE_DRAW_CONFIG_NONE, 0, 0)
+/* 0x1C */ DEFINE_SCENE(entra_n_scene, none, SCENE_ENTRA_N, SCENE_DRAW_CONFIG_NONE, 0, 0)
```
fig02, line 40: I view NONE as more of a DEFAULT, cause it still has a little bit of configuration going on. it just sets all of the segments to an empty dlist and such. It also parallels nicely with `sDefaultDisplayList`

Dragorn421: Yeah I was hesitant too, but `_default` also suggests that what it does would be done in all other configs, which isn't the case,
And afaik no scene relies on the defaults `_none` sets. It's also not universal defaults either, why would an empty dlist be the sensible default instead of null bytes (e.g. a black rgba16 image).
The intent with `_none` seems more "I don't care about this" to me, which to me feels more `_none`y than `_default`y

I've also considered renaming `sDefaultDisplayList` but wasn't sure what to. Maybe `sDrawConfigNoneDList`.
I think the main reason it exists as an array at all is to not copypaste the same thing for opa and xlu

include/tables/scene_table.h
Dragorn421: ah that's unfortunate, didn't think of that as an issue.

If we change "scene draw config" to "draw scene config" (notice the slight shift in what it means 👀 I like it better), then the enum can be `DRAW_SCENE_CONFIG_` which I think would be good? thoughts?

(`DRAW_CONFIG_SCENE_` also being bad for names like `DRAW_CONFIG_SCENE_BDAN`, it straight up contains `SCENE_BDAN`)

include/tables/scene_table.h
fig02: Default just means its the one to use when it doesnt need anything special. Its a perfect description of what it does imo.
I dont think its accurate to say that it does nothing. And default doesnt have to mean that every other case needs to do it as well. Its just the case that you default to when you dont need any of the specific ones

include/tables/scene_table.h
Roman971: I would personally be fine with `DRAW_CONFIG_` since the context makes it clear what it's meant for, and I doubt it would ever conflict with anything else

`DRAW_SCENE_CONFIG_` sounds a bit odd and is longer than I would like, but I don't have any other argument against it

Dragorn421: I feel like `draw_config` is too generic, for example there are actors using `this->drawConfig`

what about a `SDC_` prefix (short for Scene Draw Config), that should actually not clash with anything (?) and would be less verbose in the scene table (not that I think acronyms should be encouraged but I'm out of ideas)

EllipticEllipsis: `SCENEDRAW_CONFIG_`? Make scene drawing its own system?

Roman971: I think a prefix like `SDC_` could be fine. I'm not a fan of acronyms in general either, but this enum is only meant to be used in the scene table where the context and argument position already gives out its purpose, so we really just need a distinct prefix and it achieves that

fig02 (approved):

Roman971 (approved):

PR #1235, opened 2022-05-20
Overlay Functions Cleanup

hensldm:
This involves cleanup and additional documentation around the three `Overlay` functions inspired by my load PR in MM. (https://github.com/zeldaret/mm/pull/792).
* Overlay_Load
* Overlay_AllocateAndLoad
* Overlay_Relocate

Note changing the second argument type of `bzero` allows for the removal of the `bssSize` shenanigans in `Overlay_Load` and seems consistent with the declaration in ultralib.


fig02 (approved):

include/z64.h
```diff
@@ -1643,6 +1643,17 @@ typedef struct ArenaNode {
     /* 0x28 */ u8 unk_28[0x30-0x28]; // probably padding
 } ArenaNode; // size = 0x30
 
+#define RELOC_SECTION(reloc) ((reloc) >> 0x1E)
```
Dragorn421, line 1646: I've always thought it was clearer to use decimal for shifting, as the binary arrangement of bits doesn't matter here
Unless this define is standard or I'm alone on this, in which case then *sigh* so be it

```suggestion
#define RELOC_SECTION(reloc) ((reloc) >> 30)
```
reads easier to me as "get the 32-30=2 higher bits" (instead of 32-0x1E)

src/code/relocation.c
```diff
@@ -37,55 +37,60 @@ void Overlay_Relocate(void* allocatedVRamAddress, OverlayRelocationSection* over
...
-                 * jump tables.
-                 */
+
+        switch (RELOC_TYPE(reloc)) {
+            case R_MIPS_32 << 24:
+                // Handles 32-bit address relocation, used for things such as jump tables and pointers in data.
+                // Just relocate the full address.
+
+                // Check address is valid for relocation
                 if ((*relocDataP & 0xF000000) == 0) {
```
Dragorn421, line 49: not your change but
```suggestion
                if ((*relocDataP & 0x0F000000) == 0) {
```
would be easier to read imo / less confusion-prone
(same with `0x3FFFFFF` and `0xFFFFFFF` later I guess)

I'm not sure what this check is meant for to be honest, could have something to do with segmented addresses but those symbols aren't in relocation data obviously

src/code/relocation.c
```diff
@@ -37,55 +37,60 @@ void Overlay_Relocate(void* allocatedVRamAddress, OverlayRelocationSection* over
...
-                luiInstRef = luiRefs[((*relocDataP >> 0x15) & 0x1F)];
-                if ((vaddr & 0xF000000) == 0) {
-                    relocOffset = vaddr - (u32)vRamAddress;
-                    vaddr = (s16)relocData;
+
+            case R_MIPS_LO16 << 24:
+                // Handles relocation for a hi/lo pair, part 2.
+                // Grab the stored LUI (hi) from the `R_MIPS_HI16` section using the `rs` register of the instruction.
+                // The full address is calculated, relocated, and then used to update both the LUI and lo instructions.
+                // If the lo part is negative, add 1 to the LUI.
```
Dragorn421, line 83:
```suggestion
                // If the lo part is negative, add 1 to the hi value.
```
or something, this could be read as "changing the opcode"

Dragorn421 (commented):
nice

About `bzero`: yes apparently it was `s32` originally. (`extern void     bzero(void *, int);` http://n64devkit.square7.ch/header/os_libc.htm )

+1 on moving `Overlay_Load` to a new file. The name doesn't really matter as it will probably change later, `load.c` is good for now imo

src/code/relocation.c
hensldm: It is a sanity check to verify that the address is actual valid for relocation (see comment above the if statement). See https://discord.com/channels/688807550715560050/688850564595187788/965708627165970462, Tharo's message, which explains why this is the case.

Dragorn421: idk I don't see why the vram address couldn't go past 0x80FFFFFF in theory, overlays currently start at vram `0x80800000` but why not `0x81000000`, if it wasn't for this piece of code

Dragorn421 (approved):

include/z64.h
```diff
@@ -1643,6 +1643,17 @@ typedef struct ArenaNode {
...
+#define RELOC_SECTION(reloc) ((reloc) >> 30)
+#define RELOC_OFFSET(reloc) ((reloc) & 0xFFFFFF)
+#define RELOC_TYPE(reloc) ((reloc) & 0x3F000000)
+
+typedef enum OverlayRelocationType {
+    R_MIPS_32 = 2,
+    R_MIPS_26 = 4,
+    R_MIPS_HI16 = 5,
+    R_MIPS_LO16 = 6
+} OverlayRelocationType;
```
Thar0, line 1655: These are more generally MIPS relocation types, you can find them elsewhere such as ELF files for any MIPS machine. It's somewhat confusing to call the "Overlay" relocation types since it implies that they're something specific to the overlay implementation

Thar0 (commented):
One thing

include/z64.h
Dragorn421: True, for example `mips-linux-gnu-objdump --disassemble --reloc build/src/overlays/actors/ovl_En_Daiku/z_en_daiku.o` (excerpt)
```
    1488:       3c040000        lui     a0,0x0
                        1488: R_MIPS_HI16       .data
    148c:       15e10012        bne     t7,at,14d8 <EnDaiku_PostLimbDraw+0x8c>
    1490:       8fa50048        lw      a1,72(sp)
    1494:       24840124        addiu   a0,a0,292
                        1494: R_MIPS_LO16       .data
    1498:       0c000000        jal     0 <EnDaiku_ChangeAnim>
                        1498: R_MIPS_26 Matrix_MultVec3f
    149c:       24a50038        addiu   a1,a1,56
```

hensldm: Makes sense. Changed it to `MIPSRelocationType`.

Thar0 (approved):

include/z64.h
```diff
@@ -1643,6 +1643,17 @@ typedef struct ArenaNode {
     /* 0x28 */ u8 unk_28[0x30-0x28]; // probably padding
 } ArenaNode; // size = 0x30
 
+#define RELOC_SECTION(reloc) ((reloc) >> 30)
+#define RELOC_OFFSET(reloc) ((reloc) & 0xFFFFFF)
+#define RELOC_TYPE(reloc) ((reloc) & 0x3F000000)
```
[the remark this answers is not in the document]
hensldm: Yeah, that makes sense, I hadn't thought of that. Perhaps `RELOC_TYPE_SHIFTED24` would be better (maybe too verbose?) to show that it is left shifted up 24 bits?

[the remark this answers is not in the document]
Dragorn421: `RELOC_TYPE_MASK(reloc) ((reloc) & 0x3F000000)` ?

or alternative route:
```c
#define RELOC_TYPE_MASK 0x3F
#define RELOC_TYPE_SHIFT 24

switch (reloc & (RELOC_TYPE_MASK << RELOC_TYPE_SHIFT) {
    case R_MIPS_32 << RELOC_TYPE_SHIFT:
}
```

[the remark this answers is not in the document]
hensldm: Hmm, I think I like a mix of the two

```c
RELOC_TYPE_MASK(reloc) ((reloc) & 0x3F000000)
#define RELOC_TYPE_SHIFT 24

switch (RELOC_TYPE_MASK(reloc)) {
    case R_MIPS_32 << RELOC_TYPE_SHIFT:
}
```

fig02 (approved):

Dragorn421 (approved):

Roman971 (approved):

PR #1239, opened 2022-05-28
Fix a number of warnings when compiling with GCC

Thar0:
- Mostly deals with `-Wimplicit-fallthrough` warnings by adding a `FALLTHROUGH` macro that expands to `__attribute__((fallthrough))` when `__GNUC__` is defined.
 - Adds a couple of missing breaks I noticed while adding the above.
 - Also introduces an `UNUSED` macro, removes code that was only present to suppress set-but-unused warnings.
 - Fixes `-Wabsolute-value` warning that occurred when `fabsf` was passed a non-float, casts them to float first.
 - Introduces `BAD_RETURN` macro for functions that have a return type but return nothing, used in skelanime and player, and fixes some return types that still match.
 - Adds some more AVOID_UB returns, and fixes a return in z_std_dma that still matches.

There are still a lot of warnings left, I doubt it's realistic to fix them all however as most are `-Wmaybe-uninitialized`


src/code/z_parameter.c
```diff
@@ -2472,6 +2472,7 @@ void Magic_Update(PlayState* play) {
                 sMagicBorderR = sMagicBorderG = sMagicBorderB = 255;
             }
             // fallthrough (flash border while magic is being consumed)
+            FALLTHROUGH;
```
Dragorn421, line 2475:
```suggestion
            FALLTHROUGH; // flash border while magic is being consumed
```

src/code/relocation.c
```diff
@@ -100,14 +100,17 @@ void Overlay_Relocate(void* allocatedVRamAddress, OverlayRelocationSection* over
...
             case 0x4000000:
                 dbg += 0xA;
+                FALLTHROUGH;
             case 0x6000000:
                 if (gOverlayLogSeverity >= 3) {
                     osSyncPrintf("%02d %08x %08x %08x ", dbg, relocDataP, relocatedValue, relocatedAddress);
                     osSyncPrintf(" %08x %08x %08x %08x\n", ((u32)relocDataP + (u32)vRamAddress) - allocu32, relocData,
                                  unrelocatedAddress, relocOffset);
                 }
+                // No break required for matching
```
Dragorn421, line 113: this is ambiguous, can go either as "adding `break` still matches" or "adding `break` prevents matching"

src/overlays/actors/ovl_Bg_Hidan_Hamstep/z_bg_hidan_hamstep.c
```diff
@@ -102,8 +102,8 @@ s32 BgHidanHamstep_SpawnChildren(BgHidanHamstep* this, PlayState* play2) {
     s16 params;
     PlayState* play = play2;
 
-    pos = pos; // Required to match
     pos.y = this->dyna.actor.home.pos.y - 100.0f;
+    pos = pos; // Required to match
```
Dragorn421, line 106: why not :smile:

src/overlays/actors/ovl_En_Gm/z_en_gm.c
```diff
@@ -203,11 +203,13 @@ void func_80A3DC44(EnGm* this, PlayState* play) {
...
                 SET_INFTABLE(INFTABLE_B0);
+                FALLTHROUGH;
             case 3:
                 this->actionFunc = func_80A3DD7C;
                 return;
             case 1:
                 SET_INFTABLE(INFTABLE_B1);
+                FALLTHROUGH;
             case 2:
                 this->actionFunc = EnGm_ProcessChoiceIndex;
```
Dragorn421, line 214:
```suggestion
                this->actionFunc = EnGm_ProcessChoiceIndex;
                FALLTHROUGH;
```

src/overlays/actors/ovl_En_Dy_Extra/z_en_dy_extra.c
```diff
@@ -91,7 +91,8 @@ void EnDyExtra_Draw(Actor* thisx, PlayState* play) {
     static Color_RGBA8 primColors[] = { { 255, 255, 170, 255 }, { 255, 255, 170, 255 } };
     static Color_RGBA8 envColors[] = { { 255, 100, 255, 255 }, { 100, 255, 255, 255 } };
     static u8 D_809FFC50[] = { 0x02, 0x01, 0x01, 0x02, 0x00, 0x00, 0x02, 0x01, 0x00, 0x02, 0x01, 0x00, 0x02,
-                               0x01, 0x00, 0x02, 0x01, 0x00, 0x02, 0x01, 0x00, 0x02, 0x01, 0x00, 0x01, 0x02 };
+                               0x01, 0x00, 0x02, 0x01, 0x00, 0x02, 0x01, 0x00, 0x02, 0x01, 0x00, 0x01, 0x02,
+                               0x00 };
```
Dragorn421, line 95: this file is oddly formatted
also `BAD_RETURN(s32)` is put on its own line by clang-format 11.0.1

src/overlays/actors/ovl_En_Go2/z_en_go2.c
```diff
@@ -766,6 +776,9 @@ u16 EnGo2_GetTextId(PlayState* play, Actor* thisx) {
                 return EnGo2_GetTextIdGoronMarketBazaar(play, this);
         }
     }
+#ifdef AVOID_UB
+    return faceReaction; // always 0 at this point
```
Dragorn421, line 780: imo this could use a comment like "at this point in vanilla mq debug v0 is the return from Text_GetFaceReaction"

Dragorn421 (commented):
more `FALLTHROUGH` potential in two `// fall-through` occurrences

src/overlays/actors/ovl_Bg_Hidan_Hamstep/z_bg_hidan_hamstep.c
Thar0: This actually suppresses another warning but I forgot to mention it in the PR description

Dragorn421: why is it happy with pos=pos after just pos.y gets initialized... :shrug:

src/overlays/actors/ovl_En_Go2/z_en_go2.c
```diff
@@ -766,6 +776,9 @@ u16 EnGo2_GetTextId(PlayState* play, Actor* thisx) {
                 return EnGo2_GetTextIdGoronMarketBazaar(play, this);
         }
     }
+#ifdef AVOID_UB
+    return faceReaction; // the return vlue of faceReaction is always in v0 at this point
```
Dragorn421, line 780:
```suggestion
    return faceReaction; // the return value of Text_GetFaceReaction, faceReaction, is always in v0 at this point
```

fig02 (approved):

Dragorn421 (approved):

Roman971 (approved):

PR #1241, opened 2022-05-30
Remove `BINANG_ROT180`

Dragorn421:
Remove `BINANG_ROT180`, using as few explicit casts as possible

`BINANG_SUB` is next

### DIFF
--- a/include/z64math.h
+++ b/include/z64math.h
@@ -76,7 +76,6 @@ typedef struct {
     /* 0x06 */ s16 yaw;    // azimuthal angle
 } VecSph; // size = 0x08
 
-#define BINANG_ROT180(angle) ((s16)((angle) - 0x7FFF))
 #define BINANG_SUB(a, b) ((s16)((a) - (b)))
 
 #define LERP(x, y, scale) (((y) - (x)) * (scale) + (x))
--- a/src/code/db_camera.c
+++ b/src/code/db_camera.c
@@ -710,7 +710,7 @@ void DbCamera_Update(DbCamera* dbCamera, Camera* cam) {
                 DbCamera_AddVecSph(sp7C, sp7C, &spFC);
             } else {
                 spFC.pitch = -spFC.pitch;
-                spFC.yaw = BINANG_ROT180(sp104.yaw);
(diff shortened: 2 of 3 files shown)


engineer124 (approved):

EllipticEllipsis:
Is this actually an improvement? Replacing weird macro by magic number?

EllipticEllipsis:
I don't think the physical size of the expression is of particular importance.

> I wouldn't call 0x7FFF a magic number, it's close enough to 180° as a binary angle
That's exactly why it is a magic number: 0x8000 is obviously half a turn, 0x7FFF is off by one for no apparent reason, hence magic. It's the same as the problems with the clock time macros being off by 1 or 2 frequently.

And yes, it is just one dev's weird habit, but the irony is by removing the macro you make it more visible, hence people are more likely to use it.

fig02:
I would say the macro existing makes it more visible than not, youre more likely to see that macro near other conversion macros and be likely to use it yourself. as opposed to it just being done in-place where its needed

fig02 (approved):

EllipticEllipsis:
Ah, that's why you put it in a file-specific header or the file itself.

Roman971 (approved):

fig02:
its really no different than any of the other hex angles that are in the codebase currently. I dont think this one deserves special treatment

Dragorn421:
Anyone still attached to
```c
#define BINANG_ROT180(angle) ((s16)((angle) - 0x7FFF))
```
?

Dragorn421:
Bumping this here from discord,

I'm open to replacing `BINANG_ROT180(x)` with `x - DEG_TO_BINANG(179.999f)` if that would be better than `x - 0x7FFF` to opponents of the macro removal

We can then argue if it should be `179.994` or `179.999` or `179.996178323729`,
and if this approach should be generalized to all binary angles

fig02:
i am going to merge this as we havent been able to get any more opinions. Of the people who have contributed to the conversation, there are more in favor of removing it.

we can talk the possible wider usage of degrees at a later time

PR #1245, opened 2022-05-30
Use `DMG_` defines in code

Dragorn421:
For example
- `dmgFlags & 0x0800` -> `dmgFlags & DMG_ARROW_FIRE`
- `dmgFlags & 0x10060000` -> `dmgFlags & (DMG_UNKNOWN_1 | DMG_MAGIC_ICE | DMG_MAGIC_FIRE)`

`dmgFlags` in collider init data is left, I'd like to do it in another PR as that's a beast of its own

Also there may be another pass to do to revise DMG_-grouping defines


fig02:
feel like we should just do it for everything and then make group macros for the ones it feels appropriate for after.
unless people feel that leaving it as hex will make it easier to search and make groups for them

fig02 (approved):

Roman971 (approved):

PR #1247, opened 2022-05-31
Create `save_flags.h` header

Dragorn421:
- Move save flags defines (e.g. `EVENTCHKINF_...`) from z64save.h to new save_flags.h
- Move save flags macros from macros.h to save_flags.h
- Include save_flags.h where needed (not in global headers)
- Cleanup includes, ordering mostly like global/actors/specific/assets

Note: `save_flags.h` is included in 144 files out of 804 (`grep 'save_flags\.h' -r src -l | wc`, `find src -name *.c | wc`)


fig02:
I think it would be fair to include the new file into z64save.h (eventually going to be named after whatever save context is called, I imagine) then we wouldnt need to do this in all the files its needed

at least its related enough to me to make sense. dunno what our long term plans are for includes like this

Dragorn421:
I think our goal should be to split headers as much as we can, and I think we can start now with things that we know won't change much in the future
(for example here what the save flags are is well-defined, and we know they're (nearly?) the same in every version)

The end goal for header splitting, far down the road, being to handle dependencies on header in the build system

fig02:
since you need access to save context to use these, it seems logical to me to have it included into the save header. then the macros are just available when the save header is included

Dragorn421:
Header splitting needs to be discussed anyway so there's no point really going through this now I guess

PR #1250, opened 2022-06-01
Fix VTX data for sasatest

louist103:
ZAPD was leaving gaps in the VTX data for this room.
Fixed by manually creating one large VTX array.

### DIFF
--- a/assets/xml/scenes/test_levels/sasatest.xml
+++ b/assets/xml/scenes/test_levels/sasatest.xml
@@ -3,6 +3,9 @@
         <Scene Name="sasatest_scene" Offset="0x0"/>
     </File>
     <File Name="sasatest_room_0" Segment="3">
+        <Array Name="gSasatestRoom0Vtx_00370" Count="289" Offset="0x370">
+            <Vtx/>
+        </Array>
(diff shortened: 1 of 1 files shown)


louist103:
> This looks more like a ZAPD bug that should be addressed in ZAPD, isn't it?
> 
> For example `gsSPVertex(&sasatest_room_0Vtx_000370[28], 8, 1)` clearly uses vertices 28 to 35 (included) from `sasatest_room_0Vtx_000370` but the vertex array as currently extracted is only of length `33` (for example vertex 33 is extracted as "unaccounted", `sasatest_room_0_unaccounted_000580`)

It might be a ZAPD bug but its definitely an edge case since this doesn't happen anywhere else.
It also makes sense for this room to extract the VTX as one large array since looking at the DList it looks like the whole floor is one VTX array and there are no textures.

Dragorn421:
The array absolutely should be in one piece, but it would be better / more worthwhile to fix the issue in ZAPD directly rather than hack a solution

It's not like we typically extract Vtx arrays explicitly, we mostly don't, so it would be way better for ZAPD to do better on its own on that front

fig02:
agreed, this should be handled zap side

PR #1252, opened 2022-06-03
Extract skyboxes and backgrounds

louist103:
At long last its here. Backgrounds and skyboxes extracted in beautiful RGBA32.

### DIFF
--- a/assets/xml/textures/backgrounds.xml
+++ b/assets/xml/textures/backgrounds.xml
@@ -1,0 +1,246 @@
+<Root>
+    <File Name="vr_ALVR_static">
+        <Texture Name="gMarketPotionShopBgTex" OutName="market_potion_shop" Format="ci8" Width="256" Height="256" Offset="0x0" ExternalPalette="vr_ALVR_pal_static" ExternalPaletteOffset="0x0"/>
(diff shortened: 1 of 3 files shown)


Dragorn421 (commented):
Nice!

Should we have any `SplitTlut="false"`? Since it's the default it can be omitted afaik, but should it?

Is there any point to `number 0` (in the spec) besides a "cleaner" z64.map file?

louist103:
> Nice!
> 
> Should we have any `SplitTlut="false"`? Since it's the default it can be omitted afaik, but should it?
> 
> Is there any point to `number 0` (in the spec) besides a "cleaner" z64.map file?

I added it to be consistent with the others but I have no attachment to it. 
I added number zero cause AFAIK its technically correct but it seems like I didn't add it for title cards, again I can take it out if needed.

Dragorn421:
Did we agree on removing `number 0` from the spec in the end?

louist103:
> Did we agree on removing `number 0` from the spec in the end?

I still personally think keeping `number 0` doesn't make the spec look cluttered enough to justify the map having nonsense values.
However I know the leads want it gone so if they want it gone I can remove it.

fig02:
I think its probably best if `number 0` is removed
I dont have much of an opinion on the split TLUT thing in the xmls

also agree with romans comment above that needs to be addressed

Dragorn421:
On `SplitTlut="false"` it will probably take a skybox documentation pass to tell if they should be kept (from what I'm seeing they should)
I'd keep them, we can remove them later if needed, it's easier than adding them back

fig02 (approved):

Roman971 (approved):

PR #1253, opened 2022-06-03
Include ``gameplay_keep`` in ``object_link_boy`` and ``ovl_Boss_Ganon``

Yanis42:
I noticed we can include ``gameplay_keep`` in these files to replace raw hex inside some DLs (found this with the hammer DL)


Dragorn421 (approved):
hackermans

Roman971 (approved):
👍

fig02 (approved):

PR #1254, opened 2022-06-03
Rename and move handwritten asm files

Thar0:
And adds a makefile rule to build asm files in `src/`.
Also removed some unnecessary `.set noat` directives from files that don't manually (or otherwise) use `$at`.


Dragorn421 (approved):

fig02 (approved):

Makefile
```diff
@@ -300,6 +300,9 @@ build/baserom/%.o: baserom/%
 build/asm/%.o: asm/%.s
```
Roman971, line 300: Couldn't we get rid of `asm/` related stuff entirely with this PR? For example this recipe is no longer needed, and `ASM_DIRS` could be reduced to `data/` and maybe renamed

Roman971 (commented):
This messes up the progress script because it assumes handwritten asm files aren't in `src` 😅
```
./progress.py
3856536 total bytes of decompilable code

3856536 bytes decompiled in src 100.0%

36384/31408 bytes decompiled in boot 115.84309730005094%

1003960/999984 bytes decompiled in code 100.39760636170179%

2812000/2812000 bytes decompiled in overlays 100.0%
```
I don't think we can identify C vs asm from the map file, so I guess the script would have to subtract the decompiled/matched size of each section with the handwritten size to account for the asm. Unless someone has a better idea?

Makefile
Thar0: I left it in so it wouldn't need to be added back when `asm/` is put to use again for other versions, but I can remove it if you think it's better for now

Roman971: I considered that but I doubt `asm/` itself would be used for other versions. We would probably only ever need `asm/non_matchings` which isn't handled by the makefile. And either way it's easy enough to get it back from the git history if we do end up needing it

Dragorn421 (approved):

fig02 (approved):

Roman971 (approved):

PR #1257, opened 2022-06-04
Update clang-tidy configuration in format.sh and apply fixes for a new check

Roman971:
I updated the compiler options in `format.sh` to fix some errors when running clang-tidy. And I added a check for actually having clang-tidy because I noticed the script would continue without erroring if you happened to not have clang-tidy installed.

I also enabled and applied fixes from the check for ['readability-inconsistent-declaration-parameter-name'](https://clang.llvm.org/extra/clang-tidy/checks/readability-inconsistent-declaration-parameter-name.html) which I was talking about in discord the other day. Basically this check finds inconsistencies in argument names between function definitions 
 and declarations (eg. in `functions.h`), and tries to fix them using names in the definition as the "correct" ones.
This last part may be controversial but I think the consistency is overall better, and I thought I would at least show how it works so we can decide whether we go for it.


fig02 (approved):
still not a fan of the external declarations for `play` and player lib functions using `this` as their arg name, but respect the consistency it will add

format.sh
```diff
@@ -3,7 +3,7 @@
 FORMAT_VER="11"
 FORMAT_OPTS="-i -style=file"
 TIDY_OPTS="-p . --fix --fix-errors"
-COMPILER_OPTS="-fno-builtin -std=gnu90 -Iinclude -Isrc -D_LANGUAGE_C -DNON_MATCHING"
+COMPILER_OPTS="-fno-builtin -fsyntax-only -std=gnu90 -Iinclude -Isrc -Iassets -Ibuild -I. -D_LANGUAGE_C -DNON_MATCHING"
```
Dragorn421, line 6:
```suggestion
COMPILER_OPTS="-fno-builtin -fsyntax-only -std=gnu90 -Iinclude -Isrc -Ibuild -I. -D_LANGUAGE_C -DNON_MATCHING"
```
smh sabotaging your own work

more widely this could be copypasted from `CC_CHECK` in the makefile
https://github.com/zeldaret/oot/blob/6f0b8f1e7483524cc967536c50cb02e431d534c9/Makefile#L133

format.sh
```diff
@@ -34,6 +34,12 @@ else
     fi
 fi
 
+if [ ! $(command -v clang-tidy) ]
```
Dragorn421, line 37: I'm illiterate in bash but it should probably prioritize `clang-tidy-$FORMAT_VER` like for clang-format ?

Or add `TIDY_VER` ?

Roman971: There is no difference or reason to prioritize a specific version for clang-tidy afaik, so it seems unnecessary to me. It's a different story for clang-format because we want to prioritize using clang-format-11 instead of the default version for whatever package manager you're using.

Dragorn421: Ok!

format.sh
Roman971: Good catch on the `-Iassets` 👍

If by copypasting from `CC_CHECK` you meant adding the other stuff, I don't think it makes any difference in the context of clang-tidy, since it mostly just needs to be able to preprocess and compile the file afaik

Dragorn421: oh I thought those were also passed to clang-format, nvm then :sweat_smile:

Dragorn421 (approved):

PR #1264, opened 2022-06-08
Enable more IDO warnings and apply fixes

Roman971:
Following the example of https://github.com/zeldaret/mm/pull/781, I enabled more IDO warnings with `-fullwarn` and added `-verbose` to get more details in warnings. Then I fixed every new warning generated as well as a few old warnings we still had in the repo. Just like the MM PR, I had to disable one new type of warning (number 516) because it's unusable in our case as it warns for any 64 bit integer (eg. in every texture). 

I also updated asm-processor to use a `#pragma` instead of the former `EARLY` includes for cs data, which avoids another type of warning and is an overall improvement.

---

Note:
For normal linux builds, the only warnings left after this are the GCC warnings from `-Wtype-limits`, which are basically unfixable without changing the code logic. So at this point our options are either to keep them forever, to disable them globally with a simple `-Wno-type-limits` in the makefile, or to ignore them locally with pragmas surrounding the issue.
For future reference, this is how it would look if we disabled them locally:
```c
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wtype-limits"
    // rot.x < 0x8000 is always true since it's a s16
    if ((this->actor.shape.rot.x > 0) && (this->actor.shape.rot.x < 0x8000)) {
#pragma GCC diagnostic pop
```

### DIFF
--- a/include/z64cutscene.h
+++ b/include/z64cutscene.h
@@ -144,7 +144,7 @@ typedef enum {
  * on its own line.
  *
  * Files that contain this type that are included in other C files
- * must be preceded by a '#pragma asmproc recurse qualifier' to
+ * must be preceded by a '#pragma asmproc recurse' qualifier to
  * inform asm-processor that it must recursively process that include.
  *
  * Example:
--- a/include/z64cutscene.h
+++ b/include/z64cutscene.h
@@ -144,10 +144,12 @@ typedef enum {
  * on its own line.
  *
  * Files that contain this type that are included in other C files
- * must include an 'EARLY' qualifier to inform asm-processor that it
- * must recursively process that include.
+ * must be preceded by a '#pragma asmproc recurse qualifier' to
+ * inform asm-processor that it must recursively process that include.
  *
- * Example: #include "file.c" EARLY
+ * Example:
+ * #pragma asmproc recurse
+ * #include "file.c"
  */
 typedef union CutsceneData {
     s32 i;
--- a/src/code/graph.c
+++ b/src/code/graph.c
@@ -31,6 +31,9 @@ void Graph_FaultClient(void) {
     osViSwapBuffer(nextFb);
 }
 
+// TODO: merge Gfx and GfxMod to make this function's arguments consistent
+void UCodeDisas_Disassemble(UCodeDisas*, Gfx*);
+
 void Graph_DisassembleUCode(Gfx* workBuf) {
     UCodeDisas disassembler;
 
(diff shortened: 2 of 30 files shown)


include/z64cutscene.h
```diff
@@ -144,10 +144,12 @@ typedef enum {
  * on its own line.
  *
  * Files that contain this type that are included in other C files
- * must include an 'EARLY' qualifier to inform asm-processor that it
- * must recursively process that include.
+ * must be preceded by a '#pragma asmproc recurse qualifier' to
```
Thar0, line 147:
```suggestion
 * must be preceded by a '#pragma asmproc recurse' qualifier to
```

Thar0 (commented):
Tiny thing

src/overlays/actors/ovl_En_Tp/z_en_tp.c
```diff
@@ -160,7 +160,8 @@ void EnTp_Init(Actor* thisx, PlayState* play2) {
             next = (EnTp*)Actor_Spawn(&play->actorCtx, play, ACTOR_EN_TP, this->actor.world.pos.x,
                                       this->actor.world.pos.y, this->actor.world.pos.z, 0, 0, 0, 0 * i);
 
-            if (0 * i) {} // Very fake, but needed to get the s registers right
+            if ((0 * i) != 0) {} // Very fake, but needed to get the s registers right
```
EllipticEllipsis, line 163: Synray actually had a better idea for this, might be worth using instead: https://discord.com/channels/688807550715560050/688851337085190255/887885019438272523

tools/asm_processor/build.py
```diff
@@ -1,64 +1,114 @@
 #!/usr/bin/env python3
```
EllipticEllipsis, line 1: Now that we have two files from asm_processor, is it worth a subrepo?

EllipticEllipsis (approved):
I love the smell of a warning-free repo. Couple of thoughts.

I'd be in favour of disabling the range warning globally and commenting wherever it's a problem rather than employing more GCC extension stuff.

tools/asm_processor/build.py
Dragorn421: On that note it would be nice to include the LICENSE file from asm-processor.

src/code/graph.c
```diff
@@ -31,6 +31,9 @@ void Graph_FaultClient(void) {
     osViSwapBuffer(nextFb);
 }
 
+// TODO: merge Gfx and GfxMod to make this function's arguments consistent
+void UCodeDisas_Disassemble(UCodeDisas*, Gfx*);
```
Dragorn421, line 35: Why put this here and not in `functions.h` ? (note `UCodeDisas_Disassemble` is commented out in `functions.h`)

Roman971: The function takes a `GfxMod*` in `ucode_disas.c` so it can't be defined as taking `Gfx*` in `functions.h` yet. This is why I added a TODO so we don't forget to fix it.

fig02 (approved):

Roman971:
Updated asm-processor with the latest change. Hopefully it should be back to being compatible with python 3.6.

Also, regarding setting it up as a subrepo, I think we could do that in a dedicated PR soon (using the alternative Dragorn mentioned above for the encoding flags)

fig02 (approved):

Dragorn421 (approved):

PR #1265, opened 2022-06-09
Update docs/Documenting.md

Dragorn421:
Let's actually formalize an agreement on how to handle documentation


Dragorn421:
Also we need to discuss the documentation style on structs because `// comments` on struct fields are completely ignored by doxygen (it expects `///`, `//!`, `/**` or `/*!` https://doxygen.nl/manual/docblocks.html )
And it would be great to be able to expand explanations on more than a single line

docs/Documenting.md
```diff
@@ -2,44 +2,57 @@
...
+Note:
+
+- Documentation for self-explanatory arguments (`@param`) and return values (`@return`) may be omitted.
+- `@param` commands should not have empty lines in between.
+- Different commands (main description, `@param`, `@return`, ...) should be separated by empty lines.
+
+Other directives that may be used for documenting functions:
+
+- `@see` to reference something else ([see below](#linking-related-information)).
+- `@note` to bring attention to some of the function behavior.
```
EllipticEllipsis, line 48: I think it is appropriate to have something stronger than `@note` for some things where stuff will break badly if instructions are not followed.

docs/Documenting.md
```diff
@@ -48,53 +61,59 @@ s32 foo;
...
- * \f[ \textrm{Your LaTeX Here} \f]
- */
-```
+`@see` may also be used for documenting files or variables.
+
+### HTML and LaTeX
+
+It is possible to include HTML and LaTeX in documentation comments.
+
+However it is prefered not to do so, so that the raw text stays readable when looked at as plain text, for example when browsing the source code.
```
EllipticEllipsis, line 119: I do think there are systems where LaTeX is useful enough to make it worth it (the matrix stuff, probably sys_math3d), but I know I'm in the minority.

docs/Documenting.md
```diff
@@ -48,53 +61,59 @@ s32 foo;
...
+
 ```c
 /**
- * Save File Data
- * @see SaveContext
+ * Sets the next framebuffer to the framebuffer associated to `task`.
+ * If there is no current buffer or it is time to swap, this buffer will be swapped to
+ * immediately, otherwise it will be swapped to later in Sched_HandleRetrace.
+ *
+ * @see Sched_HandleRetrace
```
EllipticEllipsis, line 106: Does Doxygen generate the correct links if function names are not followed by `()`?

docs/Documenting.md
Dragorn421: Yeah and i think the default should be `@remark` since `@note` is not for trivia stuff but it would end up being / already is used this way

The problem would be that graduation is completely arbitrary though, or can we come up with actual clear "rules"?

I made this to showcase `@remark`, `@note`, `@attention`, `@warning`

<details>

```c
/**
 * Spawn an object file of a specified ID that will persist through room changes.
 *
 * This only starts loading the object, the data will not be available immediately when the function returns.
 *
 * Used for gameplay_keep, Link's object, and the subkeep if any.
 *
 * @return The new object slot corresponding to the requested object ID.
 *
 * @remark This function is not meant to be called externally to spawn object files on the fly.
 * @note When an object is spawned with this function, all objects that come before it in the entry list will be treated as
 * @attention persistent, which will likely cause the memory available for object space to run out.
 * @warning This function is only meant to be called internally on scene load, before the object list from any room is processed.
 */
```

image

</details>

EllipticEllipsis: I usually use `@remark` like I'd use "Remark:" in a paper, as a tangential thing that's not usage-specific, like what the original name of the function is if known.

EllipticEllipsis: I usually use `@remark` like I'd use "Remark:" in a paper, as a tangential thing that's not usage-specific, like what the original name of the function is if known.

docs/Documenting.md
Dragorn421: Maybe we can find a compromise where LaTeX is its own part and doesn't add information besides math, and there's a main separate explanation part that is readable as plain text?

Do you have an example in MM maybe, there's not much LaTeX in OoT and idk how to find it

docs/Documenting.md
EllipticEllipsis: `@warning` is tricky because it can be mixed up with compiler warnings, but I don't love `@attention`.

docs/Documenting.md
```diff
@@ -2,44 +2,57 @@
...
 
+The documentation can then be browsed by opening `docs/doxygen/html/index.html` in a web browser.
+
 ## Documenting Functions
-For functions, a description of the function's purpose should go above the function:
+
+Any comments inside functions, except bugs ([see below](#documenting-bugs)), should use `//`-style comments, even if spanning over multiple lines.
+
+A simple example of documenting a function withjust a description (note the leading `/**`):
+
```
fig02, line 17: need a space between "with" and "just"

fig02 (commented):
I read over this document again after not having read it in a long time, and everything in it mostly still applies to how we approach documentation today.

We still dont really have clear preferences on different doxygen tags like "@note". I tend to not use them at all and just the word "note:" or "see:"
We can either take that section out or make it clear that we arent strict about it at the moment.

I think overall its clear that most people prefer reading docs straight in the source code and dont even bother with generating docs with doxygen.

docs/Documenting.md
Dragorn421: yes

fig02:
To this day I maintain that zero people on the planet actually generate the doxygen docs and use them. It feels almost pointless to be using any of its decorators and other syntax. But if we want to keep using the tags we do use then I wont argue!

Dragorn421:
Yeah I don't doxygen is that useful but it does give a bit of a standard way to organize comments at least
The fact it's pretty much unused is the motivation for why I rewrote the document with the idea of restricting the doxygen features allowed to be used

I think this is good to merge

fig02 (approved):

PR #1266, opened 2022-06-09
Some doc on actor overlays alloc types

Dragorn421:
- `ALLOCTYPE_` -> `ACTOROVL_ALLOC_` (`ACTOR_OVERLAY_ALLOC_` would be fully accurate though...)
- `ACTOROVL_ALLOC_` are now defines, since the engine uses bitwise operations for some reason...
- `AM_FIELD_SIZE` -> `ACTOROVL_ABSOLUTE_SPACE_SIZE`
- fancy documentation
- `PLAYER_STATE3_6` -> `PLAYER_STATE3_RESTORE_NAYRUS_LOVE`


include/tables/actor_table.h
```diff
@@ -11,474 +11,474 @@
  *
  * DEFINE_ACTOR_UNSET is needed to define empty entries from the original game
  */
-/* 0x0000 */ DEFINE_ACTOR_INTERNAL(Player, ACTOR_PLAYER, ALLOCTYPE_NORMAL, "Player")
+/* 0x0000 */ DEFINE_ACTOR_INTERNAL(Player, ACTOR_PLAYER, ACTOR_ALLOC_NORMAL, "Player")
```
EllipticEllipsis, line 14: Why not drop the ACTOR_ALLOC_ part in this table? It can be inserted with macros, and it's not like this cooperates properly with IDEs anyway.

include/z64actor.h
```diff
@@ -8,7 +8,6 @@
 
 #define ACTOR_NUMBER_MAX 200
 #define INVISIBLE_ACTOR_MAX 20
-#define AM_FIELD_SIZE 0x27A0
 #define MASS_IMMOVABLE 0xFF // Cannot be pushed by OC colliders
```
EllipticEllipsis, line 11: These aren't really the same sort of thing, so there should probably at least be a blank line here.

include/z64actor.h
```diff
@@ -38,11 +37,45 @@ typedef struct {
...
 
-typedef enum {
-    /* 0 */ ALLOCTYPE_NORMAL,
-    /* 1 */ ALLOCTYPE_ABSOLUTE,
-    /* 2 */ ALLOCTYPE_PERMANENT
-} AllocType;
+/**
+ * @see ACTOR_ALLOC_ABSOLUTE
+ */
+#define ACTOR_OVERLAY_ABSOLUTE_SPACE_SIZE 0x27A0
```
EllipticEllipsis, line 43: I mourn the Eva reference.

include/z64actor.h
```diff
@@ -38,11 +37,45 @@ typedef struct {
...
+ * `ACTOR_ALLOC_` defines indicate how an actor overlay should be loaded.
+ *
+ * @note Bitwise or-ing `ACTOR_ALLOC_` types is not meaningful.
+ * The `ACTOR_ALLOC_` types are 0, 1, 2 but checked against with a bitwise and.
+ *
+ * @see ACTOR_ALLOC_ABSOLUTE
+ * @see ACTOR_ALLOC_PERSISTENT
+ * @see actor_table.h
+ */
+#define ACTOR_ALLOC_NORMAL 0
```
EllipticEllipsis, line 58: Do these all render properly when generating the Doxygen? I've had them not stick to macros properly before.

include/z64actor.h
```diff
@@ -38,11 +37,45 @@ typedef struct {
...
+#define ACTOR_OVERLAY_ABSOLUTE_SPACE_SIZE 0x27A0
+
+/**
+ * The actor overlay should be allocated memory for when loading,
+ * and the memory deallocated when there is no more actor using the overlay.
+ *
+ * `ACTOR_ALLOC_` defines indicate how an actor overlay should be loaded.
+ *
+ * @note Bitwise or-ing `ACTOR_ALLOC_` types is not meaningful.
+ * The `ACTOR_ALLOC_` types are 0, 1, 2 but checked against with a bitwise and.
```
EllipticEllipsis, line 52: This sounds more like an enum at this point.

include/z64actor.h
```diff
@@ -38,11 +37,45 @@ typedef struct {
...
+ * @see actor_table.h
+ */
+#define ACTOR_ALLOC_NORMAL 0
+/**
+ * The actor overlay should be loaded to "absolute space".
+ *
+ * Absolute space is a fixed amount of memory allocated once.
+ * The overlay will still need to be loaded again if at some point there is no more actor using the overlay.
+ *
+ * @note Only one such overlay may be loaded at a time. This is not cleanly asserted.
```
EllipticEllipsis, line 65: Not sure what you're going for with the last sentence, could you clarify?

include/z64actor.h
Dragorn421: Well with an enum I would expect slapping another value (3) after it would work, but currently it would need to be `1 << 2`  (4), and not 3 which has the 1 and 2 bits set...

include/tables/actor_table.h
Dragorn421: because it breaks grepability (can't grep / ctrl shift F it) :/

EllipticEllipsis: You could say the same thing for leaving `ovl` off the names.

include/z64actor.h
EllipticEllipsis: Mm, rather frustrating.

include/tables/actor_table.h
Dragorn421: yeah I do think it would be nice to properly refer the segments

but that's not happening as we would then need another argument for the actorinit symbol name (or rename the actorinit symbols to use the overlay segment name), and I feel like that wouldn't have much support?

include/z64actor.h
mzxrules: for a more serious comment, there should be some sort of correlation to this number and the reservation size. Ideally this shouldn't be a constant value

Dragorn421: Codegen won't let this not be a constant

EllipticEllipsis: We have some ghastly construct in MM, let me see if I can find what we did.

EllipticEllipsis: Ah, apparently we just ended up with
```c
// This value is hardcoded to be the size of ovl_Arrow_Fire which currently is the biggest actor that uses the AM_FIELD.
#define AM_FIELD_SIZE SEGMENT_SIZE(ovl_Arrow_Fire)
```

but we certainly considered using a `MAX`.

Dragorn421: well maybe I missed something but in oot no "absolute alloc" actor overlay is `0x27A0` bytes

also how does this even match, if this relies on symbols the codegen should be ruined?

EllipticEllipsis:
```c
#define SEGMENT_START(segment) (_ ## segment ## SegmentStart)
#define SEGMENT_END(segment) (_ ## segment ## SegmentEnd)
#define SEGMENT_SIZE(segment) ((uintptr_t)SEGMENT_END(segment) - (uintptr_t)SEGMENT_START(segment))
```
They're in terms of segment symbols, it might get away with it because it's just a function argument?

AngheloAlf: > also how does this even match, if this relies on symbols the codegen should be ruined?

Yes, MM actually relies on the vram symbols and performs a subtraction: 
```mips
/* 015274 800BAD34 3C098092 */  lui         $t1, %hi(ArrowIce_SetupAction)
/* 015278 800BAD38 25292430 */  addiu       $t1, $t1, %lo(ArrowIce_SetupAction)
/* 01527C 800BAD3C 1480000A */  bnez        $a0, .L800BAD68
/* 015280 800BAD40 3C0A8092 */   lui        $t2, %hi(ArrowFire_SetupAction)
/* 015284 800BAD44 254A0340 */  addiu       $t2, $t2, %lo(ArrowFire_SetupAction)
/* 015288 800BAD48 012A2023 */  subu        $a0, $t1, $t2
/* 01528C 800BAD4C AFA30024 */  sw          $v1, 0x24($sp)
/* 015290 800BAD50 0C040B22 */  jal         ZeldaArena_MallocR
/* 015294 800BAD54 AFA70038 */   sw         $a3, 0x38($sp)
```

There's no way this could match on OoT, since it has a hardcoded value instead
```mips
/* 015264 800320C4 3C058013 */  lui         $a1, %hi(STR_801366E8)
/* 015268 800320C8 24A566E8 */  addiu       $a1, $a1, %lo(STR_801366E8)
/* 01526C 800320CC 8D8D013C */  lw          $t5, 0x13C($t4)
/* 015270 800320D0 240427A0 */  addiu       $a0, $zero, 0x27A0
/* 015274 800320D4 55A0000F */  bnel        $t5, $zero, .L80032114
/* 015278 800320D8 8FB90060 */   lw         $t9, 0x60($sp)
/* 01527C 800320DC 0C01EBE3 */  jal         ZeldaArena_MallocRDebug
/* 015280 800320E0 00003025 */   move       $a2, $zero
```

To add more weirdness to this, I checked what OoT 1.0 does, and it has a hardcoded value too, but a different one...
```mips
/* 0140FC 8002519C 14600007 */  bnez        $v1, .L800251BC
/* 014100 800251A0 00000000 */   nop
/* 014104 800251A4 0C019B0E */  jal         ZeldaArena_MallocR
/* 014108 800251A8 240424E0 */   addiu      $a0, $zero, 0x24E0
/* 01410C 800251AC 8FAC0058 */  lw          $t4, 0x58($sp)
```

Dunno why they liked to change this value so much

include/z64actor.h
fig02: hes saying that only one overlay can be loaded to the absolute space at a time, and that its not clearly checked/asserted for this to be true. magic items can conflict and overwrite each other

include/tables/actor_table.h
fig02: I do think that ACTOR_ being the prefix poses the same problem as the scene draw configs from a little while ago-- it looks like an actor id enum
would prefer ACTORALLOC_ or something, idk

include/z64actor.h
EllipticEllipsis: Is "This is not cleanly asserted" some sort of technical language I'm not familiar with? I'm not sure what it's supposed to mean if it isn't

fig02: asserted in this context meaning checked/enforced. can be with an assert macro for example, or just checked in some capacity
its not clean in this case, cause it is indirectly enforced by most magic items restricting each other from executing while another is going. but its not as effective as it could be (as seen by some scenarios crashing due to an overlap in the absolute space)

EllipticEllipsis: I think it should just say it in plain English rather than jargon, then. I had a look on Google and neither of those terms seems to be standardised (and "asserted" is ambiguous anyway, so seems even less appropriate). No reason to not be specific in documentation, it's not a variable name we want to keep short.

include/tables/actor_table.h
Dragorn421: `ACTOROVL_ALLOC_` ? (reminder this deals with overlays not the actor instances)

include/z64actor.h
Dragorn421: if we do it's not due to these enums being actually bitmasks I hope, I'll change 'em :sunglasses:

include/z64actor.h
Dragorn421: Yes?:

<details>

image

</details>

include/tables/actor_table.h
fig02: yea seems good

include/z64actor.h
```diff
@@ -38,11 +39,48 @@ typedef struct {
...
+ *
+ * @note Bitwise or-ing `ACTOROVL_ALLOC_` types is not meaningful.
+ * The `ACTOROVL_ALLOC_` types are 0, 1, 2 but checked against with a bitwise and.
+ *
+ * @see ACTOROVL_ALLOC_ABSOLUTE
+ * @see ACTOROVL_ALLOC_PERSISTENT
+ * @see actor_table.h
+ */
+#define ACTOROVL_ALLOC_NORMAL 0
+/**
```
fig02, line 61: id put a break between all of these to seperate them. other than that looks good to me

fig02 (approved):

EllipticEllipsis (approved):

include/z64actor.h
```diff
@@ -38,11 +39,50 @@ typedef struct {
...
 
-typedef enum {
-    /* 0 */ ALLOCTYPE_NORMAL,
-    /* 1 */ ALLOCTYPE_ABSOLUTE,
-    /* 2 */ ALLOCTYPE_PERMANENT
-} AllocType;
+/**
+ * @see ACTOROVL_ALLOC_ABSOLUTE
+ */
+#define ACTOR_OVERLAY_ABSOLUTE_SPACE_SIZE 0x27A0
```
engineer124, line 45:
```suggestion
#define ACTOROVL_ABSOLUTE_SPACE_SIZE 0x27A0
```
Should it be this for consistency?

Dragorn421: I thought about it and I don't know either if it would be better. The other names use actorovl_ more to keep it short than to be different, but they are also different so I don't think consistency matters.

All in all, I don't know

Anyone wanna take a decision? :pray:

fig02: kinda torn as well. I think it makes sense for it to share the same prefix as the type, they at least concern the same specific parts of the zelda heap in relation to actor overlays

Dragorn421:
Now with two extra `@bug`s, i knew this system was janky :angry:
Thanks to #glitch-hunting from the speedrunning discord

fig02 (approved):

src/overlays/actors/ovl_Magic_Dark/z_magic_dark.c
```diff
@@ -79,6 +79,11 @@ void MagicDark_DiamondUpdate(Actor* thisx, PlayState* play) {
     if (1) {}
 
     // See `ACTOROVL_ALLOC_ABSOLUTE`
+    //! @bug This condition is too broad, the actor will also be killed by warp songs. But warp songs do not use an
+    //! actor which overlay uses `ACTOROVL_ALLOC_ABSOLUTE`. There is no reason to kill the actor in this case.
```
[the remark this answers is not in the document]
Dragorn421: I want to prevent confusion between the allocation of the actor instance and of the actor overlay

Is there a way in proper English to phrase this not weirdly?
> But warp songs do not use an actor which overlay uses `ACTOROVL_ALLOC_ABSOLUTE`.
> But warp songs do not use an actor which uses an overlay which uses `ACTOROVL_ALLOC_ABSOLUTE`.
> But warp songs do not use an actor that uses an overlay which uses `ACTOROVL_ALLOC_ABSOLUTE`.
> But warp songs do not use an actor that uses an overlay with `ACTOROVL_ALLOC_ABSOLUTE`.
> But warp songs do not use an actor that relies on an overlay with `ACTOROVL_ALLOC_ABSOLUTE`.

? (NotLikeThis)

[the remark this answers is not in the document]
fig02: I feel like in this context, you dont need to be hyper specific about the mechanics of the overlay allocation.
If someone sees this and wants to learn more, they are going to read into how its actually working. And not to mention the define being referenced in the same setance specifies `ACTOROVL_`

I think "actor which uses" is fine, with it being implied that we are talking about the overlay of that actor

fig02 (approved):

Roman971 (approved):
The types were much better as an enum imo, and it just feels weird to make them into sort-of flags (not even real flags since the values can't live together and 0 is its own value). But I guess it's a matter of preference either way.

Also the comments in the header feel too verbose and redundant to me, but I don't have a clear suggestion atm and I don't want to hold this for too long, so I'll approve now and probably make a PR with my suggestion in the future

Dragorn421:
> The types were much better as an enum imo, and it just feels weird to make them into sort-of flags (not even real flags since the values can't live together and 0 is its own value).

Well the game uses `&`, we could switch back to an enum but it would need to use ` = (1 << n)` imo to document the bit-check usage

> Also the comments in the header feel too verbose and redundant to me

Besides the blocky `/** * */` syntax I don't think they are, but I agree it's above the typical verbosity in the repo (which is imo too low, too close to none)

fig02:
fwiw, after reading it a second time, I also think the wording can be adjusted a bit on these comments.
if roman is going to offer suggestions at a later time I will save my comments until then and we can just merge now

PR #1268, opened 2022-06-10
Small doc for Item00 actor

krm01:
- `EnItem00.unk_15A` -> `EnItem00.despawnTimer`
- `func_8001E5C8` -> `Item00_BounceAfterCollected`

### DIFF
--- a/include/z64actor.h
+++ b/include/z64actor.h
@@ -282,7 +282,7 @@ typedef struct EnItem00 {
     /* 0x154 */ s16 unk_154;
     /* 0x156 */ s16 unk_156;
     /* 0x158 */ s16 unk_158;
-    /* 0x15A */ s16 unk_15A;
+    /* 0x15A */ s16 despawnTimer;
     /* 0x15C */ f32 scale;
     /* 0x160 */ ColliderCylinder collider;
 } EnItem00; // size = 0x1AC
--- a/src/code/z_en_item00.c
+++ b/src/code/z_en_item00.c
@@ -13,7 +13,7 @@ void EnItem00_Draw(Actor* thisx, PlayState* play);
 void func_8001DFC8(EnItem00* this, PlayState* play);
 void func_8001E1C8(EnItem00* this, PlayState* play);
 void func_8001E304(EnItem00* this, PlayState* play);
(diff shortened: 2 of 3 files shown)


Yanis42:
From my notes:
- ``unk_158`` is a bool to prevent important items like heart pieces from despawning
- ``unk_156`` is the timer (used with unk_158) to make the drop blink before despawning
- ``unk_15A`` is the global despawn timer like you named it
- ``func_8001E5C8`` seems to be the function used to move the item over Link's head when you're close enough (it handles bounce effect for rupees, also that actor doesn't check the colliders but the distance between Link and the actor to trigger the collection and its visual effects)

I'm not sure what I said is 100% accurate though

krm01:
@Yanis42 I came to the same conclusion about the other two fields you mentioned (`unk_156` and `unk_158`) but didn't have sensible names for them :) I didn't review the other functions very closely but that sounds right.

Yanis42:
> @Yanis42 I came to the same conclusion about the other two fields you mentioned (`unk_156` and `unk_158`) but didn't have sensible names for them :) I didn't review the other functions very closely but that sounds right.

``unk_158`` could be named like ``allowDespawn`` or ``enableDespawn`` (or ``disableDespawn`` depending on how it's used in the code), ``unk_156`` could be named ``blinkTimer`` or ``itemBlinkTimer`` (or drop instead of item, or both idk)

Yanis42:
@krm01 also you should join the decomp discord, it would be easier to talk about naming those (there's the link on the readme)

fig02:
i would say that `EnItem00_BounceAfterCollected` can just be `_Collected`, the little bouncing it does while floating above players head is pretty unimportant imo. The main point is that it goes above players head and waits to despawn while being collected. 
id describe that action in general as "being collected"

krm01:
@fig02 I agree the main intent is it appearing over the player's head, I figured "bounce" captured that rather than "FloatAbovePlayerHead" :P but I think you are right that just Collected captures it well enough, I do think that part is important (vs the suggestion by @Yanis42 `MoveItemPosToPlayer`, I think is missing that "being collected" information)

fig02 (approved):

Roman971 (approved):

PR #1269, opened 2022-06-10
Minor cleanups in z_std_dma.c and yaz0.c

Roman971:
* Improved ram arguments to use pointers and reduce the amount of explicit casts
* Changed a few conditions to not use boolean checks for non-booleans
* Slightly renamed a few yaz0 variables to better fit their purpose


src/boot/z_std_dma.c
```diff
@@ -336,17 +336,17 @@ void DmaMgr_ThreadEntry(void* arg) {
...
 }
 
-s32 DmaMgr_SendRequestImpl(DmaRequest* req, u32 ram, u32 vrom, u32 size, u32 unk, OSMesgQueue* queue, OSMesg msg) {
+s32 DmaMgr_SendRequestImpl(DmaRequest* req, void* ram, u32 vrom, u32 size, u32 unk, OSMesgQueue* queue, OSMesg msg) {
     static s32 sDmaMgrQueueFullLogged = 0;
 
-    if ((1 && (ram == 0)) || (osMemSize < ram + size + 0x80000000) || (vrom & 1) || (vrom > 0x4000000) || (size == 0) ||
-        (size & 1)) {
+    if ((1 && (ram == NULL)) || (osMemSize < OS_K0_TO_PHYSICAL(ram) + size) || (vrom & 1) || (vrom > 0x4000000) ||
+        (size == 0) || (size & 1)) {
```
Thar0, line 343: Is it worth `& 1` -> `% 2 != 0` to make it more explicit that it's checking alignment?

Thar0 (approved):

fig02 (approved):

src/boot/z_std_dma.c
Roman971: Could be worth yeah, tho I would prefer keeping that for a separate set of changes, maybe one dedicated to cleaning up modulos

Dragorn421 (approved):

PR #1270, opened 2022-06-11
Macros for `Entranceinfo.field`

Dragorn421:
Suggestions for shorter names welcome!


include/tables/entrance_table.h
```diff
@@ -7,8 +7,8 @@
  *    - Argument 3: Spawn number for this entrance
  *    - Argument 4: Toggle if bgm should continue during the transition using this entrance (true or false)
  *    - Argument 5: Toggle if a title card should display when using this entrance (true or false)
- *    - Argument 6: Transition type when entering using this entrance
- *    - Argument 7: Transition type when exiting using this entrance
+ *    - Argument 6: Default transition type when entering using this entrance (second half of a scene transition)
+ *    - Argument 7: Default transition type when exiting using this entrance (first half of a scene transition)
```
fig02, line 11: does the word default add anything here? I view it as *the* transition that is defined for this entrance, and the fact that you can override after the fact if you wanted to is a pretty separate thing that doesnt need to be reflected in the name
(I would also take DEFAULT out of the macro names personally)

include/z64.h
```diff
@@ -1344,6 +1344,19 @@ typedef struct {
     /* 0x24 */ s16 unk_24;
 } struct_80034A14_arg1; // size = 0x28
 
+#define ENTRANCE_INFO_FIELD_CONTINUE_BGM_FLAG (1 << 15)
+#define ENTRANCE_INFO_FIELD_DISPLAY_TITLE_CARD_FLAG (1 << 14)
+#define ENTRANCE_INFO_FIELD_DEFAULT_END_TRANSITION_TYPE_MASK 0x3F80
```
fig02, line 1349: I think we can probably leave out `FIELD`
I know that the member accessed to get this data is called "field", but its just called that cause it needs to be one generic variable. entrance info is enough to tell you what it is, imo

include/tables/entrance_table.h
Dragorn421: Well it corresponds to `TRANS_NEXT_TYPE_DEFAULT`, but maybe

> Argument 6: Transition type when entering using this entrance (second half of a scene transition) (can be overridden in specific scenarios)

?
Or where would we document this

fig02: I guess i dont think it needs to be mentioned from within the entrance table at all. If youre overriding the transition type, youre manually setting up scene loads which is already sidestepping the usual process for how the entrance table is accessed and stuff. It feels separate to me

Dragorn421: Well there are cases that are not going out of your way to encounter, like `nextTransitionType = TRANS_TYPE_FADE_WHITE` when leaving grottos, dungeon warps, warp songs, but maybe it's fine to gloss over?

fig02: I think its at least not important as far as the entrance table is concerned yeah
we can add more description to the variable itself in the `play` struct if we want, to indicate that it lets you override what type is defined in the entrance table

theres a comment on `TRANS_NEXT_TYPE_DEFAULT` to indicate this, but putting it on the variable itself also seems reasonable

fig02 (approved):

Roman971 (approved):

PR #1272, opened 2022-06-11
Document flag for swinging bottle

EllipticEllipsis:
Happy to add a comment, but for once I think this is pretty self-explanatory, especially since it's hardly used.

Also fix VSCode complaining about the fallthrough macro in `audio_effects.c`

### DIFF
--- a/include/z64cutscene_commands.h
+++ b/include/z64cutscene_commands.h
@@ -98,6 +98,7 @@
  */
 #define CS_LIGHTING_LIST(entries) CS_CMD_SET_LIGHTING, CMD_W(entries)
 
+
 /**
  * ARGS
  *   s16 setting (m), s16 startFrame (s), s16 endFrame (e)
--- a/src/code/audio_effects.c
+++ b/src/code/audio_effects.c
@@ -242,8 +242,8 @@ f32 Audio_AdsrUpdate(AdsrState* adsr) {
         case ADSR_STATE_START_LOOP:
             adsr->envIndex = 0;
             adsr->action.s.state = ADSR_STATE_LOOP;
-            FALLTHROUGH;
         retry:;
+            FALLTHROUGH;
         case ADSR_STATE_LOOP:
             adsr->delay = adsr->envelope[adsr->envIndex].delay;
             switch (adsr->delay) {
--- a/src/overlays/actors/ovl_player_actor/z_player.c
+++ b/src/overlays/actors/ovl_player_actor/z_player.c
@@ -12378,7 +12378,7 @@ void func_8084EAC0(Player* this, PlayState* play) {
     }
 }
 
-static BottleCatchInfo sBottleCatchInfo[] = {
+static BottleCatchInfo sBottleCatchInfos[] = {
     { ACTOR_EN_ELF, ITEM_FAIRY, PLAYER_AP_BOTTLE_FAIRY, 0x46 },
     { ACTOR_EN_FISH, ITEM_FISH, PLAYER_AP_BOTTLE_FISH, 0x47 },
     { ACTOR_EN_ICE_HONO, ITEM_BLUE_FIRE, PLAYER_AP_BOTTLE_FIRE, 0x5D },
@@ -12397,7 +12397,7 @@ void func_8084ECA4(Player* this, PlayState* play) {
     if (LinkAnimation_Update(play, &this->skelAnime)) {
         if (this->unk_84F != 0) {
             if (this->unk_850 == 0) {
-                Message_StartTextbox(play, sBottleCatchInfo[this->unk_84F - 1].textId, &this->actor);
+                Message_StartTextbox(play, sBottleCatchInfos[this->unk_84F - 1].textId, &this->actor);
                 Audio_PlayFanfare(NA_BGM_ITEM_GET | 0x900);
                 this->unk_850 = 1;
             } else if (Message_GetState(&play->msgCtx) == TEXT_STATE_CLOSING) {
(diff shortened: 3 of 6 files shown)


src/code/audio_effects.c
```diff
@@ -242,8 +242,8 @@ f32 Audio_AdsrUpdate(AdsrState* adsr) {
         case ADSR_STATE_START_LOOP:
             adsr->envIndex = 0;
             adsr->action.s.state = ADSR_STATE_LOOP;
-            FALLTHROUGH;
         retry:;
+            FALLTHROUGH;
```
Dragorn421, line 246: :monocle_face:

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -12443,7 +12443,7 @@ void func_8084ECA4(Player* this, PlayState* play) {
     }
 
     if (this->skelAnime.curFrame <= 7.0f) {
-        this->stateFlags1 |= PLAYER_STATE1_1;
+        this->stateFlags1 |= PLAYER_STATE1_SWINGING_BOTTLE;
     }
 }
```
fig02, line 12448: somewhere in this function would be a good place to mark this bug https://www.youtube.com/watch?v=xTWSt7GJoSo
understand if you dont want to be the one to do that tho

Dragorn421 (approved):
noo gonna steal the first documented player flag award :crying_cat_face:

src/overlays/actors/ovl_player_actor/z_player.c
EllipticEllipsis: So the point here is that it `func_8083C0E8` will change the animation, so this can pass precisely once the final time this function runs for a swing?

fig02: yea, there should have either been an early return after setting up the next action, or this block of setting the flag should have been higher up. i think both are sufficient fixes

EllipticEllipsis: Yeah, I just wanted to make sure I understood the exact nature the glitch, I saw what you said about fixes in the video.

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -12397,14 +12397,15 @@ void func_8084ECA4(Player* this, PlayState* play) {
...
                 Audio_PlayFanfare(NA_BGM_ITEM_GET | 0x900);
                 this->unk_850 = 1;
             } else if (Message_GetState(&play->msgCtx) == TEXT_STATE_CLOSING) {
                 this->unk_84F = 0;
                 func_8005B1A4(Play_GetCamera(play, CAM_ID_MAIN));
             }
         } else {
+            // Changes the animation, which causes the bug at the bottom of the function.
             func_8083C0E8(this, play);
         }
```
fig02, line 12410: Id say this comment is probably not needed, only cause it will be clear that it changes the animation when its named in the future and the bug comment at the bottom refers directly to it

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -12442,6 +12443,8 @@ void func_8084ECA4(Player* this, PlayState* play) {
...
     }
 
+    //! @bug because func_8083C0E8 will change the animation and execution passes through here afterwards, this check
+    //! can pass for a single frame of the new animation the final time this function runs in a swing.
     if (this->skelAnime.curFrame <= 7.0f) {
```
fig02, line 12448: can add a sentence like "the end result is that the bottle will be active for a single frame long after the bottle was intended to be finished being active"

just to put it into perspective of what this really means

src/overlays/actors/ovl_player_actor/z_player.c
EllipticEllipsis: I thought it was worth it at least for now, but I can remove it.

include/z64cutscene_commands.h
```diff
@@ -98,6 +98,7 @@
  */
 #define CS_LIGHTING_LIST(entries) CS_CMD_SET_LIGHTING, CMD_W(entries)
 
+
```
Dragorn421, line 101: :monocle_face: x2

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -12442,8 +12442,11 @@ void func_8084ECA4(Player* this, PlayState* play) {
         }
     }
 
+    //! @bug because func_8083C0E8 will change the animation and execution passes through here afterwards, this check
+    //! can pass for a single frame of the new animation the final time this function runs in a swing. This means that
+    //! the bottle can be detected as swinging for one frame well the active-looking part of the swinging animation.
```
Dragorn421, line 12447: the last sentence is missing a word after "one frame well"

If I can suggest entirely different wording,
```c
    //! @bug The animation is changed above by func_8083C0E8, which changes curFrame to 0.
    //! This causes the flag to be set for one frame at a time when it does not look like Link is swinging the bottle.
```

fig02: agreed on the new wording

also, I believe `func_808322D0` in the "caught something in the bottle" case will cause the same glitch. less impactful though because you will be in a cutscene during it. it may be better to just not reference the specific function and say "when your animation is changed at any point above", or something to that effect

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -12442,8 +12442,11 @@ void func_8084ECA4(Player* this, PlayState* play) {
         }
     }
 
+    //! @bug If the animation is changed at any point above (such as by func_8083C0E8() or func_808322D0()), it will
```
Dragorn421, line 12445: It's more a "when" than a "if" since it happens consistently imo but whatever

Dragorn421 (approved):

Roman971 (approved):

PR #1275, opened 2022-06-12
Format script improvements and use strict name checks

Roman971:
Following recent discord discussions, I changed the script to use `clang-tidy-6.0` when it's available. Unlike more recent versions, this one finds a lot more issues with argument names, which I also fixed in this PR.
Note that I went with "6.0" instead of "6" because that's the name of the package in debian/ubuntu for some reason, but some people may have to use an alias if it doesn't line up with theirs (or just use the default `clang-tidy` even if it's not the right version).

I also added `-m32` and `-Wno-everything` to compiler options to avoid errors and warnings we don't want when running clang-tidy, and removed `-fsyntax-only` since it serves no purpose in this context.


fig02 (approved):

Dragorn421 (approved):

Roman971:
Changed this to use `readability-inconsistent-declaration-parameter-name.Strict` instead of clang-tidy 6, which seems to keep the same behavior without requiring a specific version (see the recent discord discussion for more details)

Dragorn421 (approved):
Doesn't work for me on clang-tidy-13 :sweat_smile: 

<details>

```
$ clang-tidy --version
LLVM (http://llvm.org/):
  LLVM version 13.0.0
  
  Optimized build.
  Default target: x86_64-pc-linux-gnu
  Host CPU: tigerlake
```

`git diff`:
```diff
-void CollisionCheck_SpawnShieldParticlesMetalSound(PlayState* play, Vec3f* v, Vec3f* pos);
+void CollisionCheck_SpawnShieldParticlesMetalSound(PlayState* play, Vec3f* v, Vec3f* pleaseChangeThis);
```

`./format.sh` -> does nothing

</details>

Whatever, I can live without it and eventually someone else will run it :pray:

Roman971:
That's really weird, but the example you gave you should get fixed regardless of 'Strict' or version, so it's probably just a general issue with the `readability-inconsistent-declaration-parameter-name` check?

fig02 (approved):

PR #1283, opened 2022-06-16
Doc room drawing

Dragorn421:
Example of mesh header type 2 "cull":

<details>

(DC room 0)
Imported with the blender importer for 2.79 (meshs resulting from dlists are named `opa_....`),
and creating one circle for each entry at the corresponding pos with corresponding radius (the white circles in the screenshots, named `area_....`)

<details>

python changes to the importer to spawn the white circles highlighting the entries:

```py
                    log.info('Reading %d display lists from 0x%X to 0x%X', count, start, end)
                    for j in range(start, end, 16):
                        x,y,z,r = unpack_from(">hhhh", data, j)
                        opa = unpack_from(">L", data, j+8)[0]
                        if opa:
                            self.use_transparency = False
                            self.buildDisplayList(None, [None], opa, mesh_name_format='%s_opa')
                        xlu = unpack_from(">L", data, j+12)[0]
                        if xlu:
                            self.use_transparency = True
                            self.buildDisplayList(None, [None], xlu, mesh_name_format='%s_xlu')
                        print(x,y,z,r, opa, xlu)
                        bpy.ops.mesh.primitive_cylinder_add(location=(x,-z,y), radius=r, end_fill_type="NOTHING")
                        bpy.data.objects["Cylinder"].name = "area_" + hex(opa)
```

</details>

(looking inwards from the entrance) top front left corridor and neighbouring area:
image

room between the bombable wall and the entrance corridor:
image

part of floor that is lava, seen from below
image

</details>


src/code/z_room.c
```diff
@@ -408,34 +441,37 @@ void Room_Draw1Multi(PlayState* play, Room* room, u32 flags) {
...
-                Vec3f skyboxOffset;
+                Vec3f offset;
 
                 gfx = POLY_OPA_DISP;
-                Camera_GetSkyboxOffset(&skyboxOffset, activeCam);
-                func_8009638C(&gfx, bgImage->source, bgImage->tlut, bgImage->width, bgImage->height, bgImage->fmt,
-                              bgImage->siz, bgImage->mode0, bgImage->tlutCount,
-                              (skyboxOffset.x + skyboxOffset.z) * 1.2f + skyboxOffset.y * 0.6f,
-                              skyboxOffset.y * 2.4f + (skyboxOffset.x + skyboxOffset.z) * 0.3f);
+                Camera_GetSkyboxOffset(&offset, activeCam);
```
engineer124, line 470:
```suggestion
                Camera_GetSkyboxOffset(&quakeOffset, activeCam);
```
This can be called `quakeOffset` as this function should actually be called `Camera_GetQuakeOffset`

src/code/z_room.c
```diff
@@ -354,51 +388,50 @@ void Room_Draw1Single(PlayState* play, Room* room, u32 flags) {
...
-    camDataIdx = activeCam->camDataIdx;
-    // jfifid
-    camDataIdx2 = func_80041C10(&play->colCtx, camDataIdx, BGCHECK_SCENE)[2].y;
-    if (camDataIdx2 >= 0) {
-        camDataIdx = camDataIdx2;
+    index = activeCam->camDataIdx;
+    newIndex = func_80041C10(&play->colCtx, index, BGCHECK_SCENE)[2].y;
+    if (newIndex >= 0) {
+        index = newIndex;
     }
```
engineer124, line 409: So could you explain to me what `index`/`newIndex` is being used for in this context? This is all `BgCamInfo` code, with `func_80041C10` getting the `bgCamFuncData`, and in this particular case, `[2].y` is getting what's called `jfifId`, is that accurate given the docs here? So `index` is a `bgCamDataIndex` but `newIndex` is a `jfjfId` but they're being compared and overwritten with each other. So is `jfifId` correct?

src/code/z_room.c
```diff
@@ -235,7 +263,7 @@ s32 func_80096238(void* data) {
             // "If the original buffer size isn't at least 150kb, it will be out of control."
```
EllipticEllipsis, line 263:
```suggestion
            // "If the original buffer size isn't at least 150kB, it will be out of control."
```

src/code/z_room.c
```diff
@@ -317,33 +347,37 @@ void Room_Draw1Single(PlayState* play, Room* room, u32 flags) {
...
-    meshHeader1Entry = SEGMENTED_TO_VIRTUAL(meshHeader1Single->base.entry);
-    drawBg = (flags & ROOM_DRAW_OPA) && isFixedCamera && meshHeader1Single->source && !(SREG(25) & 1);
-    drawOpa = (flags & ROOM_DRAW_OPA) && (meshHeader1Entry->opa != NULL) && !(SREG(25) & 2);
-    drawXlu = (flags & ROOM_DRAW_XLU) && (meshHeader1Entry->xlu != NULL) && !(SREG(25) & 4);
+    meshHeaderPrerenderSingle = &room->meshHeader->prerenderSingle;
+    dListsEntry = SEGMENTED_TO_VIRTUAL(meshHeaderPrerenderSingle->base.entry);
+    drawBg = (flags & ROOM_DRAW_OPA) && isFixedCamera && (meshHeaderPrerenderSingle->source != NULL) &&
+             !(R_ROOM_PREREND_NODRAW_FLAGS & (1 << 0));
+    drawOpa = (flags & ROOM_DRAW_OPA) && (dListsEntry->opa != NULL) && !(R_ROOM_PREREND_NODRAW_FLAGS & (1 << 1));
+    drawXlu = (flags & ROOM_DRAW_XLU) && (dListsEntry->xlu != NULL) && !(R_ROOM_PREREND_NODRAW_FLAGS & (1 << 2));
```
EllipticEllipsis, line 355: No defines for these flags?

include/z64scene.h
```diff
@@ -74,87 +74,91 @@ typedef struct {
...
+    /* 0x00 */ MeshHeaderPrerenderBase base;
     /* 0x08 */ u8    count;
     /* 0x0C */ BgImage* list;
-} MeshHeader1Multi; // size = 0x10
+} MeshHeaderPrerenderMulti; // size = 0x10
 
 typedef struct {
     /* 0x00 */ Vec3s pos;
-    /* 0x06 */ s16 unk_06;
+    /* 0x06 */ s16 radius;
```
EllipticEllipsis, line 132: Could this match as a `Sphere16` substruct?

include/z64scene.h
```diff
@@ -74,87 +74,91 @@ typedef struct {
...
 
 typedef struct {
     /* 0x00 */ MeshHeaderBase base;
     /* 0x01 */ u8 numEntries;
-    /* 0x04 */ MeshHeader01Entry* entries;
-    /* 0x08 */ MeshHeader01Entry* entriesEnd;
-} MeshHeader0; // size = 0x0C
+    /* 0x04 */ MeshHeaderDListsEntry* entries;
+    /* 0x08 */ MeshHeaderDListsEntry* entriesEnd;
+} MeshHeaderAll; // size = 0x0C
```
EllipticEllipsis, line 84: I think `All` is an unfortunate adjective to use since it could be interpreted as encompassing the others, rather than being a distinct type that draws the whole lot. I don't have a better suggestion yet, will think about it more.

EllipticEllipsis (commented):
Nice to finally see what these different types are. I don't really understand what the pretty pictures are supposed to show, though.

Also, this only depends on z. Does that mean these could be well offscreen and still be being rendered if they fall inside the appropriate z range, or is more culling done elsewhere?

Dragorn421:
The screenshots try to show that the dlist for each mesh_header_type_cull entry, fit the position and radius the entry indicates

----------

Indeed it's only culling what is "too far in front" of the view, or "too far behind". That does mean something "far to the side" is not culled

If there's more culling elsewhere, it's done in the DLists (DC room 0 DLists do use `gsSPCullDisplayList` commands).

It's not really culling but when discussing this topic with a focus on performance, I think it's relevant to mention the scissoring/clipping boxes (but I'm not an expert here):
http://n64devkit.square7.ch/n64man/gsp/gSPClipRatio.htm
http://n64devkit.square7.ch/pro-man/pro12/12-03.htm#01

src/code/z_room.c
Dragorn421: I don't feel like it's worth introducing flags and enums for debug features? it's usually simple enough that it's self-documenting like here

If you insist though, sure. Where to put them?

include/z64scene.h
Dragorn421: It does, good idea. btw I was still thinking in 2D and was still seeing this as a cylinder, thanks for making sphere remarks :)

src/code/z_room.c
```diff
@@ -52,170 +52,198 @@ void Room_Draw0(PlayState* play, Room* room, u32 flags) {
...
-                spA4->unk_04 = temp_f2;
-                iter = spB4;
+
+        // If the entry isn't fully before the rendered depth range
+        if (-(f32)meshHeaderCullEntry->radius < projectedPos.z) {
+
+            // Compute the depth at which this entry starts
+            entryStartZ = projectedPos.z - meshHeaderCullEntry->radius;
+
+            // If the entry isn't fully after the rendered depth range
```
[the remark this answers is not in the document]
Dragorn421: Changed to "beyond" and wrote some stuff on the `Room_DrawCullMeshes` function, is that better?

include/z64scene.h
Dragorn421: Well actually yes it matches, but it won't work until ZAPD is updated

src/code/z_room.c
EllipticEllipsis: The natural place seems to be with the rest of the room draw stuff in the scene header. I think it is worth having stuff for debug features since those are often one of the first things people want to mess with (after they give up on Player, at least).

include/z64scene.h
fig02: thought this as well when skimming the pr for the first time, fwiw

src/code/z_room.c
```diff
@@ -52,170 +52,207 @@ void Room_Draw0(PlayState* play, Room* room, u32 flags) {
...
+    /* 0x04 */ f32 startZ;
+    /* 0x08 */ struct MeshHeaderCullEntryLinked* prev;
+    /* 0x0C */ struct MeshHeaderCullEntryLinked* next;
+} MeshHeaderCullEntryLinked; // size = 0x10
+
+/**
+ * Handle room drawing for the "cull" type of mesh header.
+ *
+ * Each entry referenced by the header is attached to display lists, and a position and radius indicating the volume
+ * those display lists take.
```
EllipticEllipsis, line 83:
```suggestion
 * those display lists take up.
```
or maybe `fill`, or even `the volume that contains those display lists take` (do we usually use two words for "display list"?

src/code/z_room.c
```diff
@@ -52,170 +52,207 @@ void Room_Draw0(PlayState* play, Room* room, u32 flags) {
...
+ * The first step is Z-sorting the entries, also excluding the entries which bounding sphere is entirely before or
+ * beyond the rendered depth range.
+ * The second step is drawing the entries that are left, in ascending depth.
```
EllipticEllipsis, line 86: Nice, this is much better now. Two things
```suggestion
 * The first step Z-sorts the entries, and excludes those entries with a bounding sphere that is entirely before or
 * beyond the rendered depth range.
 * The second step is drawing the entries that are left, from farthest to nearest.
```
(could also use "without" or "outside" the rendered depth range).

"ascending depth" is a bit oxymoronic, I think this is a bit clearer although perhaps it's a little far from the z-depth idea.

src/code/z_room.c
```diff
@@ -52,170 +52,207 @@ void Room_Draw0(PlayState* play, Room* room, u32 flags) {
...
+ */
+void Room_DrawCullMeshes(PlayState* play, Room* room, u32 flags) {
```
fig02, line 88: since culling is the lack of drawing, i I think `DrawCull` is pretty unfortunate. I understand it means to draw the cull type, meshes that *can* cull, but it caught me off guard for a moment when i first saw it

dont have any suggestions other than "DrawCullType", but can understand why that isnt desireable

src/code/z_room.c
```diff
@@ -52,170 +52,207 @@ void Room_Draw0(PlayState* play, Room* room, u32 flags) {
...
-            temp_f2 = projectedPos.z - meshHeader2Entry->unk_06;
-            if (temp_f2 < play->lightCtx.fogFar) {
-                spA4->unk_00 = meshHeader2Entry;
-                spA4->unk_04 = temp_f2;
-                iter = spB4;
+
+        // If the entry isn't fully before the rendered depth range
+        if (-(f32)meshHeaderCullEntry->radius < projectedPos.z) {
+
+            // Compute the depth at which this entry starts
```
EllipticEllipsis, line 146: I think "starts" is a bit vague/not quite there, what you mean is something like "the shallowest point on the sphere in the depth buffer", but we could again probably get away with something like "nearest point".

src/code/z_room.c
```diff
@@ -52,170 +52,207 @@ void Room_Draw0(PlayState* play, Room* room, u32 flags) {
...
+        pos.z = meshHeaderCullEntry->pos.z;
         SkinMatrix_Vec3fMtxFMultXYZW(&play->viewProjectionMtxF, &pos, &projectedPos, &projectedW);
-        if (-(f32)meshHeader2Entry->unk_06 < projectedPos.z) {
-            temp_f2 = projectedPos.z - meshHeader2Entry->unk_06;
-            if (temp_f2 < play->lightCtx.fogFar) {
-                spA4->unk_00 = meshHeader2Entry;
-                spA4->unk_04 = temp_f2;
-                iter = spB4;
+
+        // If the entry isn't fully before the rendered depth range
```
EllipticEllipsis, line 143: I'd say we could sphere it up a bit more here too, maybe say "entry's bounding sphere" or something like that. Could even introduce "unculling sphere" as a term of art, I suppose, although we've struggled with that style of terminology previously. Maybe just "sphere of visibility", lol. (Essentially the idea is to distinguish the simple geometric calculational idea from the actual drawing of the complex shape within.)

src/code/z_room.c
EllipticEllipsis: `Cullable`?

src/code/z_room.c
EllipticEllipsis: Or `bounding sphere` sounds quite good, actually, although I suppose generically it doesn't have to geometrically enclose it.

Dragorn421: Yeah it isn't required to enclose the geometry but it's ofc meant to

Here is each entry in DC room 0, each blob is an entry's dlists geometry with the bounding sphere from the entry around it

image

it fits very well each entry's mesh

src/code/z_room.c
Dragorn421: I did the inanimate-whose "which" again
not sure where I learned this mistake from :thinking: 

Also it's not `from farthest to nearest`, it's the opposite, it's drawing from the nearest entry
Performance-wise, this is probably so that the rdp can ditch triangles that are obscured by geometry just by checking the depth framebuffer instead of overwriting the same pixel in the color framebuffer over and over

EllipticEllipsis: Ah right, that makes more sense. I'm more used to thinking about it like painting, where you do the sky first, then the mountains, then the happy little trees and so on...

Dragorn421: That said, it's good to draw opa dlists from nearest to furthest probably, but it would be better to draw the xlu dlists from furthest to nearest to have things look better. OoT doesn't use enough partial translucency that this really matters though

include/z64scene.h
Dragorn421: I admit I didn't make a big effort to name this one, but the only other idea I had was "Simple"/"Normal"/"Basic" :sweat: 
"Everything" seemed worse too
If anyone can come up with a suggestion it's taken

fig02: all supposed to convey that there is no conveying and all of them are drawn at all times right?
if my understanding is correct then Normal is probably best for me, but also open to other suggestions

fig02:
the only thing holding this up on my end is `MeshHeaderAll`
I think Normal is an improvement, even if we decide to change it later

include/z64scene.h
Dragorn421: Changed to "Normal" but I'm not a fan :shrug: 
wasn't a fan of "All" either anyway

If anyone has a suggestion eventually, feel free

src/code/z_room.c
Dragorn421: @engineer124 Can I get your thoughts on the latest iteration of this that I did I forgot when?

src/code/z_room.c
```diff
@@ -244,106 +287,118 @@ s32 func_80096238(void* data) {
...
 
-    gDPPipeSync(displayListHead++);
-    *displayList = displayListHead;
+    gDPPipeSync(gfx++);
+
+    *gfxP = gfx;
 }
 
-void Room_Draw1Single(PlayState* play, Room* room, u32 flags) {
+#define ROOM_PREREND_NODRAW_BACKGROUND (1 << 0)
```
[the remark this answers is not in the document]
Dragorn421: "bg" is used as a synonym for (bgcheck kind of) "collision" in a bunch of spots and I feel like it's likely to stay

but also if this shouldn't be "prerender" too there's more to change to this anyway

src/code/z_room.c
```diff
@@ -232,10 +275,10 @@ s32 func_80096238(void* data) {
             osSyncPrintf("成功…だと思う。 time = %6.3f ms \n", OS_CYCLES_TO_USEC(time) / 1000.0f);
             // "Writing back to original address from work buffer."
             osSyncPrintf("ワークバッファから元のアドレスに書き戻します。\n");
-            // "If the original buffer size isn't at least 150kb, it will be out of control."
+            // "If the original buffer size isn't at least 150kB, it will be out of control."
             osSyncPrintf("元のバッファのサイズが150キロバイト無いと暴走するでしょう。\n");
 
-            bcopy(gZBuffer, data, sizeof(gZBuffer));
+            bcopy(gZBuffer, data, sizeof(u16[SCREEN_HEIGHT][SCREEN_WIDTH]));
```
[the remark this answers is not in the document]
Dragorn421: the data is a screen height*width rgba16 image before it is the same size as the depth framebuffer, their size being the same (ofc gZBuffer needs to be at least as large) and in particular the 2d array type and dimensions being the same is "just" a detail/coincidence

src/code/z_room.c
```diff
@@ -52,170 +52,213 @@ void Room_Draw0(PlayState* play, Room* room, u32 flags) {
...
     }
 
-    meshHeader2 = &room->meshHeader->meshHeader2;
-    meshHeader2Entry = SEGMENTED_TO_VIRTUAL(meshHeader2->entries);
-    spA4 = spB8;
+    meshHeaderCullable = &room->meshHeader->cullable;
+    meshHeaderCullableEntry = SEGMENTED_TO_VIRTUAL(meshHeaderCullable->entries);
+    insert = linkedEntriesBuffer;
+
+    ASSERT(meshHeaderCullable->numEntries <= ARRAY_COUNT(linkedEntriesBuffer), "polygon2->num <= SHAPE_SORT_MAX",
```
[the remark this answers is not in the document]
Dragorn421: well of course I disagree but okay I'll change it, no big deal

include/z64scene.h
```diff
@@ -61,9 +61,9 @@ typedef struct {
 // Mesh headers
 
 typedef enum {
-    /* 0 */ MESH_HEADER_TYPE_0,
-    /* 1 */ MESH_HEADER_TYPE_1,
-    /* 2 */ MESH_HEADER_TYPE_2,
+    /* 0 */ MESH_HEADER_TYPE_NORMAL,
+    /* 1 */ MESH_HEADER_TYPE_PRERENDER,
```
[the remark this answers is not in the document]
Dragorn421: would `_prerendered_background` be okay or too long?

just "background" could be fine but it feels incomplete, as there's more data than just the background (the "depth geometry") and "background" feels too generic to me but idk

include/z64scene.h
```diff
@@ -74,87 +74,92 @@ typedef struct {
 typedef struct {
     /* 0x00 */ Gfx* opa;
     /* 0x04 */ Gfx* xlu;
-} MeshHeader01Entry; // size = 0x08
+} MeshHeaderDListsEntry; // size = 0x08
```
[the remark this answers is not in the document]
Dragorn421: `MeshHeaderDListsEntry` is used for both `MeshHeaderNormal` and `MeshHeaderPrerenderBase`

I kept "header" because I personally don't like "mesh" on its own, it sounds super generic, maybe we could replace all "meshHeader" occurrences with "roomMesh" though?

src/code/z_room.c
```diff
@@ -52,170 +52,213 @@ void Room_Draw0(PlayState* play, Room* room, u32 flags) {
...
+ *
+ * Each entry referenced by the header is attached to display lists, and a position and radius indicating the bounding
+ * sphere for the geometry drawn.
+ * The first step Z-sorts the entries, and excludes the entries with a bounding sphere that is entirely before or
+ * beyond the rendered depth range.
+ * The second step draws the entries that remain, from nearest to furthest.
+ */
+void Room_DrawCullable(PlayState* play, Room* room, u32 flags) {
+    MeshHeaderCullable* meshHeaderCullable;
+    MeshHeaderCullableEntry* meshHeaderCullableEntry;
```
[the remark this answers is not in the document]
Dragorn421: `Room_DrawCullable` has a lot of temps so idk about making some of them even more vaguely named

include/z64scene.h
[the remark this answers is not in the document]
fig02: Ive mentioned it in the past but I dont like the term mesh at all for this, personally. imo its too ambiguous and doesnt sound at all to me like the dlists for a room. mesh could very well also be referring to a collision mesh

src/code/z_room.c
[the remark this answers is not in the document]
fig02: I for one would like to see the "bg" in bgcheck go eventually, but it may not be a popular opinion. And ofc that wouldnt happen for a while anyway so its not like its relevant for discussions now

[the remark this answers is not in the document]
Roman971: Is "bg" used as a synonym for "collision" in our documentation? I don't remember seeing that and it seems wrong to me.

It's part of "bgcheck" but that's because the system is specifically about doing collision checks with "background" stuff as opposed to actors and other dynamic entities. The "collision" part was implicit in the original name, probably because "check" was considered enough to make things clear, even tho that's arguable (and probably why it became collision_bg in AC).
tl;dr I don't think "bgcheck" being a thing should prevent us from using "bg" for non-collision stuff, and that system will probably be renamed at some point anyway

src/code/z_room.c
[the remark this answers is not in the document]
Roman971: Yeah I suppose you can see it as being the size of `data` instead `gZBuffer`, since that's what it copies to. It's still a bit unfortunate to hardcode the type here but w/e

src/code/z_room.c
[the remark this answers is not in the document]
Dragorn421: Sure I don't know or want to suggest anything here. But in the mean time since currently "bg" is overloaded I don't think it's worth shortening "background" to "bg".
And for solving this particular comment on `ROOM_PREREND_NODRAW_BACKGROUND` we probably want to rethink "prerend" anyway before the other part

include/z64scene.h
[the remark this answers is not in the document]
Dragorn421: (changed "mesh header" to "room shape" but the comment on `RoomShapeDlistsEntry` having two uses still stands)

src/code/z_room.c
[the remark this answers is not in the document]
Dragorn421: I wish we could `static_assert(sizeof(gZBuffer) >= ...)` yeah

src/code/z_room.c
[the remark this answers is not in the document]
fig02: I think most would agree that they would see bg and think its bgcheck related, and I at least have actively avoided using bg in the past for that purpose.
I think its fair to say that the reader should use context to figure out that its not related, but I also accept the fact that it could be confusing. at least until things are renamed 

mainly thinking of things like "bgCamData" or other things that use bg to mean bgCheck

Dragorn421:
I'm honestly considering
`ROOM_SHAPE_TYPE_PRERENDER` -> `ROOM_SHAPE_TYPE_JPEG`
because:
- "prerender" clashes with Prerender.c,
- and also scenes like hyrule market could qualify as "prerendered" but they are `ROOM_SHAPE_TYPE_NORMAL` and the background is a skybox
- the backgrounds are litterally jpeg (see: `z_jpeg.c` among other files and the ucode `njpgdspMain`), and it's the most important thing of that room shape type

I'll change to that in the future unless there's opposition

fig02:
seems like its probably fine. tho arent the images technically JFIF not JPEG?
jpeg would obviously be the more recognizable one for most people, but yea. Maybe we can just say "image" or something instead of a specific format?

Dragorn421:
Don't quote me on it because my knowledge is a few minutes of websearch (and experimentally, the fact that MS Paint / Blender can open OoT's jpegs just fine), but afaict JFIF and JPEG are very closely related and what we call "jpeg" may as well be called "jfif"

https://en.wikipedia.org/wiki/JPEG
https://en.wikipedia.org/wiki/JPEG_File_Interchange_Format

https://stackoverflow.com/questions/1427623/are-all-jpeg-files-jfif
> JPEG = the compression algorithm (Joint Photographic Experts Group)
>
> JFIF = the file format (JPEG File Interchange Format)
>
> Effectively when you mention jpeg/jpg you are actually talking about JPEG/JFIF

fig02:
ok, then its probably fine. 
In theory one could change the image format that gets processed, and I think the intent of the poly type would still be the same (a flat image, instead of a specific format of an image). But I dont care too much either way at this point and just want to get this in lol

EllipticEllipsis:
Can't really get much more definitive than the W3C's page:
> JPEG JFIF, which is what people generally mean when they refer to "JPEG", is a file format created by the [Independent JPEG Group (IJG)](http://www.ijg.org/) for the transport of single JPEG-compressed images.

https://www.w3.org/Graphics/JPEG/

include/z64scene.h
```diff
@@ -58,103 +58,111 @@ typedef struct {
...
+    /* 0x08 */ RoomShapeDListsEntry* entriesEnd;
+} RoomShapeNormal; // size = 0x0C
 
 typedef enum {
-    /* 1 */ MESH_HEADER1_FORMAT_SINGLE = 1,
-    /* 2 */ MESH_HEADER1_FORMAT_MULTI
-} MeshHeader1Format;
+    /* 1 */ ROOM_SHAPE_IMAGE_FORMAT_SINGLE = 1,
+    /* 2 */ ROOM_SHAPE_IMAGE_FORMAT_MULTI
+} RoomShapeImageFormat;
```
fig02, line 89: i hope "image format" doesnt get confused with the usual meaning of image format, like for textures. but its prob fine

Dragorn421: oh I didn't think of that. "format" is just inherited from ancient docs so we could change it
maybe just "type" 8) ?

Dragorn421: -> `ROOM_SHAPE_IMAGE_AMOUNT_` :shrug:

fig02 (approved):

include/z64scene.h
```diff
@@ -58,103 +58,111 @@ typedef struct {
...
     /* 0x0C */ void* tlut;
     /* 0x10 */ u16   width;
     /* 0x12 */ u16   height;
     /* 0x14 */ u8    fmt;
     /* 0x15 */ u8    siz;
-    /* 0x16 */ u16   mode0;
+    /* 0x16 */ u16   tlutMode;
     /* 0x18 */ u16   tlutCount;
-} BgImage; // size = 0x1C
+} RoomShapeImageMultiBackgroundEntry; // size = 0x1C
```
Roman971, line 122: This name is too long imo, and we could probably shorten it with Bg and/or omitting some of the less important terms

src/code/z_room.c
```diff
@@ -22,19 +18,23 @@ Gfx D_801270B0[] = {
...
-void func_80095AA0(PlayState* play, Room* room, Input* arg2, UNK_TYPE arg3) {
+void func_80095AA0(PlayState* play, Room* room, Input* input, s32 arg3) {
 }
 
-void Room_Draw0(PlayState* play, Room* room, u32 flags) {
+void Room_DrawNormal(PlayState* play, Room* room, u32 flags) {
     s32 i;
-    MeshHeader0* meshHeader0;
-    MeshHeader01Entry* meshHeader0Entry;
+    RoomShapeNormal* header;
```
Roman971, line 36: Are the shape structs considered headers still? If not, a few variables should probably be updated like this one

include/z64scene.h
Dragorn421: well every word is important, "background" or "entry" may be the least important but one of them has to stay

"RoomShape": common prefix, hard to shorten
"Image" : type of the RoomShape
"Multi" : type of the RoomShapeImage (there's also "Single")

-> `RoomShapeImageMultiBgEntry`,
could be `RoomShapeImageMultiEntry` or `RoomShapeImageMultiBg` if you really want it

src/code/z_room.c
Dragorn421: I guess not, good catch

include/z64scene.h
Roman971: `RoomShapeImageMultiBgEntry` is probably good enough. It's still a bit on the long side but it should work

Tho I suggest applying the same change to the `Room_GetImageMultiBackgroundEntry` function (and we can change its argument to just `roomShape`, imo)

Roman971 (approved):

PR #1284, opened 2022-06-17
Introduce PLAY_LOG()

fig02:
play has a logging system, which when enabled, prints out a file and line number at various parts of an update cycle. can be helpful for debugging at what stage in the play lifecycle something goes wrong.

A macro is introduced here to condense the (imo) spammy code that is needed for it. This makes things easier to read, and will make it easier to ifdef it out when we eventually get to retail

### DIFF
--- a/src/code/z_play.c
+++ b/src/code/z_play.c
@@ -15,6 +15,14 @@ u64 D_801614D0[0xA00];
 
 void Play_SpawnScene(PlayState* this, s32 sceneNum, s32 spawn);
 
+// If R_ENABLE_PLAY_LOGS is enabled, the number "1" with a file and line number will be printed during
+// various phases of the Play update cycle. This can be helpful for debugging at what stage a problem may be occuring.
+// Due to the high number of logs occuring rapidly, the game will run slow with this enabled.
+#define PLAY_LOG(line)                        \
+    if (1 && R_ENABLE_PLAY_LOGS) {            \
+        LOG_NUM("1", 1, "../z_play.c", line); \
+    }
+
 void func_800BC450(PlayState* this) {
     Camera_ChangeDataIdx(GET_ACTIVE_CAM(this), this->unk_1242B - 1);
 }
@@ -158,6 +166,7 @@ void Play_Destroy(GameState* thisx) {
 
     this->state.gfxCtx->callback = NULL;
     this->state.gfxCtx->callbackParam = 0;
+
     SREG(91) = 0;
     R_PAUSE_MENU_MODE = 0;
 
@@ -300,6 +309,7 @@ void Play_Init(GameState* thisx) {
     Play_SpawnScene(
         this, gEntranceTable[((void)0, gSaveContext.entranceIndex) + ((void)0, gSaveContext.sceneSetupIndex)].scene,
         gEntranceTable[((void)0, gSaveContext.entranceIndex) + ((void)0, gSaveContext.sceneSetupIndex)].spawn);
+
     osSyncPrintf("\nSCENE_NO=%d COUNTER=%d\n", ((void)0, gSaveContext.entranceIndex), gSaveContext.sceneSetupIndex);
 
     // When entering Gerudo Valley in the right setup, trigger the GC emulator to play the ending movie.
@@ -320,10 +330,12 @@ void Play_Init(GameState* thisx) {
             gSaveContext.totalDays++;
             gSaveContext.bgsDayCount++;
             gSaveContext.dogIsLost = true;
+
             if (Inventory_ReplaceItem(this, ITEM_WEIRD_EGG, ITEM_CHICKEN) ||
                 Inventory_ReplaceItem(this, ITEM_POCKET_EGG, ITEM_POCKET_CUCCO)) {
                 Message_StartTextbox(this, 0x3066, NULL);
             }
+
             gSaveContext.nextDayTime = NEXT_TIME_DAY_SET;
         } else {
             gSaveContext.nextDayTime = NEXT_TIME_NIGHT_SET;
@@ -435,12 +447,14 @@ void Play_Update(PlayState* this) {
         HREG(81) = 0;
         osSyncPrintf("object_exchange_rom_address %u\n", gObjectTableSize);
         osSyncPrintf("RomStart RomEnd   Size\n");
+
         for (i = 0; i < gObjectTableSize; i++) {
             s32 size = gObjectTable[i].vromEnd - gObjectTable[i].vromStart;
 
             osSyncPrintf("%08x-%08x %08x(%8.3fKB)\n", gObjectTable[i].vromStart, gObjectTable[i].vromEnd, size,
                          size / 1024.0f);
         }
+
         osSyncPrintf("\n");
     }
 
@@ -477,6 +491,8 @@ void Play_Update(PlayState* this) {
         }
 
         if (this->transitionMode) {
+            // if `transitionMode` is not TRANS_MODE_OFF
+
             switch (this->transitionMode) {
                 case TRANS_MODE_SETUP:
                     if (this->transitionTrigger != TRANS_TRIGGER_END) {
@@ -550,6 +566,7 @@ void Play_Update(PlayState* this) {
                         (this->transitionCtx.transitionType == TRANS_TYPE_FADE_WHITE_CS_DELAYED) ||
                         (this->transitionCtx.transitionType == TRANS_TYPE_FADE_WHITE_INSTANT)) {
                         this->transitionCtx.setColor(&this->transitionCtx.instanceData, RGBA8(160, 160, 160, 255));
+
                         if (this->transitionCtx.setUnkColor != NULL) {
                             this->transitionCtx.setUnkColor(&this->transitionCtx.instanceData,
                                                             RGBA8(160, 160, 160, 255));
@@ -716,7 +733,6 @@ void Play_Update(PlayState* this) {
 
                 case TRANS_MODE_SANDSTORM_INIT:
                     if (this->transitionTrigger != TRANS_TRIGGER_END) {
-                        // trigger in, leaving area
                         this->envCtx.sandstormState = SANDSTORM_FILL;
                         this->transitionMode = TRANS_MODE_SANDSTORM;
                     } else {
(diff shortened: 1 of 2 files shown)


src/code/z_play.c
```diff
@@ -477,6 +491,8 @@ void Play_Update(PlayState* this) {
         }
 
         if (this->transitionMode) {
+            // if `transitionMode` is not TRANS_MODE_OFF
+
```
Roman971, line 495:
```suggestion
        if (this->transitionMode) { // != TRANS_MODE_OFF
```

src/code/z_play.c
```diff
@@ -15,6 +15,14 @@ u64 D_801614D0[0xA00];
 
 void Play_SpawnScene(PlayState* this, s32 sceneNum, s32 spawn);
 
+// If R_ENABLE_PLAY_LOGS is enabled, the number "1" with a file and line number will be printed during
+// various phases of the Play update cycle. This can be helpful for debugging at what stage a problem may be occuring.
```
Roman971, line 19: It feels a bit unnecessary to mention the "Play update cycle" here. I think I would just say something like "will be printed at various points", or simply when the macro is used.
It's also slightly incorrect since the log is used in `Play_Main` as well

src/code/z_play.c
```diff
@@ -15,6 +15,14 @@ u64 D_801614D0[0xA00];
...
+// Due to the high number of logs occuring rapidly, the game will run slow with this enabled.
+#define PLAY_LOG(line)                        \
+    if (1 && R_ENABLE_PLAY_LOGS) {            \
+        LOG_NUM("1", 1, "../z_play.c", line); \
+    }
```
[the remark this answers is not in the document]
EllipticEllipsis: I suspect that the 1 here is a debugging define, the same thing happens in other files. (MM's z_play has some really horrible functions to match because of these.)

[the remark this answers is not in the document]
Roman971: Do you have any evidence that MM's z_play is affected by this? I don't see any of these functions decompiled in MM so I haven't been able to compare them myself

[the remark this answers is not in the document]
EllipticEllipsis: Hmm, I thought we had a file in MM but maybe not. z_std_dma does this too, though.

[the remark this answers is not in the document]
EllipticEllipsis: This many do while (0)s is probably going to mess up the block-counting, too.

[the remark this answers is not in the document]
Roman971: It should be fine since we would be replacing the `if (1)` with `do while (0)`

src/code/z_play.c
```diff
@@ -15,6 +15,14 @@ u64 D_801614D0[0xA00];
 
 void Play_SpawnScene(PlayState* this, s32 sceneNum, s32 spawn);
 
+// If R_ENABLE_PLAY_LOGS is enabled, the number "1" with a file and line number will be printed during
+// various phases of the Play update cycle. This can be helpful for debugging at what stage a problem may be occuring.
+// Due to the high number of logs occuring rapidly, the game will run slow with this enabled.
```
Dragorn421, line 20: The slowdown is an implementation detail of the emulator / IS64 scripts imo (and especially the pj64 one is incredibly slow),

An easy way to test that is first 1) changing `#define gISVDbgPrnAdrs ((ISVDbg*)0xB3FF0000)` to use `0xB3FE0000`, making the game read/write to a region unrelated to the is64, then 2) removing the `if (data != ASCII_TO_U32('I', 'S', '6', '4')) {` magic check

This makes the game run exactly as if it was printing logs through the IS64, but it's using the wrong part of memory which makes the emulator not react to any of it.
And with this `R_ENABLE_PLAY_LOGS = true;` doesn't cause any slowdown

So printing doesn't really slow the game down, but it slows down emulation when the emulator supports the IS64 regs

src/code/z_play.c
Dragorn421: `Play_Main` is the "state update" function I guess, but yeah it's more common to think of update as `Play_Update`

I also wouldn't mention how this can be helpful for debugging, as this applies to nearly every single print/debug feature imo

src/code/z_play.c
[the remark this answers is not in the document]
fig02: do while 0 does indeed work to replace if 1 in these, and is more likely what it originally was imo

Retail will tell us for sure of course

src/code/z_play.c
fig02: I feel like it's worth mentioning, mainly because on the surface it may not be so clear why the number "1" is being spammed everywhere. It's not exactly a self explanatory logging system imo. 
If each print was descriptive and said what was about to happen, it'd be different story

I think saying something like "the play update cycle" is important cause of how "top level" all of this is. A log call between two functions is really a log call across entire systems being processed, so I really want to stress that importance. I know 3/43 are in main, but it's still its main purpose. But if there's another way to say that and still convey its importance, lmk

src/code/z_play.c
fig02: I did this one better and tested on console 

I am very confident I've seen a high abundance of logs slow down on console before, but this is apparently not enough, cause it runs fine

Dragorn421: Yeah of course printing logs isn't free but this isn't a lot at all
Fwiw the is64 emulator built into mupen can almost keep up with the play logs enabled

src/code/z_play.c
Dragorn421: To me this explanation on how it helps debugging Play_Main, feels the same as if you had a comment in an actor like "if you want to change how the actor behaves you can do this".

Maybe it's just the wording that feels like it's directly addressed to the reader that puts me off?
Wording what you remarked on on discord, that the logs happen at a high level, between each sub-system "update", sounds more useful/detailed to me

fig02: even in my original comment I dont see how Im addressing the reader. Its phrased in an impersonal way without acknowledging the reader at all, but maybe I am misunderstanding 

But yeah, I am trying to convey the point that the logs happen in between play sub systems being processed. My current wording is: `// This can be helpful for debugging at what point during the `PlayState` loop an issue is occurring.`
That is expecting that you have that pretext that the play state is a gamestate and is processing many systems. But Id rather make that assumption than try and be specific about the places that it gets used. Which is why I say "it can be used to.."

Dragorn421: Maybe a suggestion will clear up what I mean?
Current:
```c
// If R_ENABLE_PLAY_LOGS is enabled, the number "1" with a file and line number will be printed.
// This can be helpful for debugging at what point during the `PlayState` loop an issue is occuring.
```
->
```c
// If R_ENABLE_PLAY_LOGS is enabled, the number "1" with a file and line number will be printed at various key points in
// the play state execution, in particular between calls to various systems.
// As a result, this allows tracing what the play state execution is doing at a high level.
```

(explain what it does and why it's neat instead of just mentioning it's good for debugging (most debug features are useful, given you are in a situation that could use them))

Idk maybe another way to express how I feel about giving such tips in the source is that it feels out of place, comments should document and explain but not give an opinion or tips (for example we don't have "set this to E for 20fps" in the rom header when that's really one of the first things anyone should do)

fig02: I dont see my current wording as giving tips or advice. I see it has explaining, objectively, why it is helpful in this context. I only use the phrasing of *can be* because it depends on how its used.

For example, your comment is assuming that they *are* placed inbetween calls to systems. Which they are, but Id rather say that they can be helpful if used in that way, rather than comment on how the original devs used it directly. Its explaining its purpose for existing, and if used right, will be helpful.

I do not agree that im saying "debug feature is good", I am saying that this particular debug feature has a helpful purpose if used right.

All that said, I will try to take your comment and work it in a way that works for me

fig02:
```
// If R_ENABLE_PLAY_LOGS is enabled, the number "1" with a file and line number will be printed.
// When placed at various key points in the Play state execution, like between calls to various systems, 
// it can allow for tracing what the play state execution is doing at a high level.
```
for example, is how I would phrase this.
My point is I want to say what the macros *purpose* is, while not saying "this macro *is used* here. That kind of phrasing, to me, is like commenting on where certain functions get used. The point of the macro description here is to explain how it can be used, not how it is used.

Hopefully this wording is a suitable compromise. I think we just have different perspectives on what kind of angle these descriptions should be coming from.

Dragorn421: Sorry to argue over such a small thing, I really don't understand your point of view :sweat: , I hope it doesn't look like bad faith

> When placed at various key points

This feels weird because it makes it sound like it's up to a dev to add the line wherever, but it's already there?

Also I understand we don't want to explicitly document where something is used, but for regs/debug features in particular it seems appropriate to indicate what they seem to be for
For example just the reg being named after "play" indicates being about the play state / z_play ?

fig02: we'll just need more opinions to tiebreaker this, cause I dont think we'll reach an agreement

Roman971: I think I agree with Fig about trying to say what the macro's purpose is and not where it's used in the base code. I think I would even go as far as only stating what the macro does and the general purpose of such a log (ie. tracing back to a specific function call or section of code), without referring to play itself, but I understand if that's seen as too vague.

Regarding Fig's suggestion and the part about `When placed at various key points`, I agree it can sound a bit weird, but maybe that can be improved with a different wording like `By being placed at ...`? Tho I don't think it's a big deal either way, personally

Dragorn421 (approved):
:sweat_smile:

src/code/z_play.c
```diff
@@ -969,117 +930,78 @@ void Play_Update(PlayState* this) {
                         Play_SetViewpoint(this, this->viewpoint ^ (VIEWPOINT_LOCKED ^ VIEWPOINT_PIVOT));
                     }
                 }
+<<<<<<< HEAD
+=======
                 Play_ChangeViewpointBgCamIndex(this);
             }
+>>>>>>> 00b98027dbb1898f477aebbd9464b37c6f1a80f0
```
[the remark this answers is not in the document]
fig02: dunno how it let me commit this. my vscode had a stroke

Roman971 (approved):

PR #1288, opened 2022-06-18
Doc pass on bgcheck-waterbox properties

Dragorn421:


include/z64bgcheck.h
```diff
@@ -73,18 +71,39 @@ typedef struct {
...
+#define WATERBOX_LIGHT_INDEX_MASK 0x00001F00
+#define WATERBOX_LIGHT_INDEX(properties) \
+    (((properties) >> WATERBOX_LIGHT_INDEX_SHIFT) & (WATERBOX_LIGHT_INDEX_MASK >> WATERBOX_LIGHT_INDEX_SHIFT))
+#define WATERBOX_LIGHT_INDEX_NONE 0x1F // warns and defaults to 0
+
+#define WATERBOX_ROOM_INDEX_SHIFT 13
+#define WATERBOX_ROOM_INDEX_MASK 0x0007E000
+#define WATERBOX_ROOM_INDEX(properties) \
+    (((properties) >> WATERBOX_ROOM_INDEX_SHIFT) & (WATERBOX_ROOM_INDEX_MASK >> WATERBOX_ROOM_INDEX_SHIFT))
+#define WATERBOX_ROOM_INDEX_ALL 0x3F // value for "room index" indicating "all rooms"
```
fig02, line 91: room index? things like light index make sense to me cause it is used as an index to either light config or light settings, but this is just a room number isnt it?

include/z64bgcheck.h
```diff
@@ -73,18 +71,39 @@ typedef struct {
...
+
+#define WATERBOX_FLAG_19 (1 << 19)
+
+#define WATERBOX_PROPERTIES(bgCamIndex, lightIndex, roomIndex, setFlag19)       \
+    ((((bgCamIndex)&WATERBOX_BGCAM_INDEX_MASK) >> WATERBOX_BGCAM_INDEX_SHIFT) | \
+     (((lightIndex)&WATERBOX_LIGHT_INDEX_MASK) >> WATERBOX_LIGHT_INDEX_SHIFT) | \
+     (((lightIndex)&WATERBOX_ROOM_INDEX_MASK) >> WATERBOX_ROOM_INDEX_SHIFT) | ((setFlag19) ? WATERBOX_FLAG_19 : 0))
```
fig02, line 98: the ternary for that last one is odd to me. shouldnt it just be the value? if you pass 0 then it will be 0

include/z64bgcheck.h
Dragorn421: I didn't know room numbers were specifically called "room numbers" if that's what you mean, so change
`WATERBOX_ROOM_INDEX...` -> `WATERBOX_ROOM_NUMBER...` ?

include/z64bgcheck.h
Dragorn421: I'm envisioning `true`/`false` to be passed as `setFlag19`, that seems more in line with other enums, but I can change the last arg to just be `flags` (and it would only be used for the single flag that exists)

fig02: even if it was true or false, I dont see the purpose of the ternary. true or false will just be 0 or 1 anyway right? 
this might be some standard thing ive just never seen before, just seems weird to do extra stuff that isnt needed

include/z64bgcheck.h
fig02: id even say just _ROOM is fine tbh. for example from the actor struct:
`/* 0x003 */ s8 room; // Room number the actor is in. -1 denotes that the actor won't despawn on a room change`

but yea people generally say "room number". the field is curRoom.num for example

include/z64bgcheck.h
Dragorn421: For example `SCENE_CMD_ROOM_BEHAVIOR` takes true/false for `showInvisActors` and `disableWarpSongs`, and sets (or doesn't set) the corresponding bit?
That's what `((setFlag19) ? WATERBOX_FLAG_19 : 0)` does,
or do you prefer `(((setFlag19) & 1) << WATERBOX_FLAG_19_SHIFT)` or something?

fig02: yea I would just expect it to be set like everything else. at least ive never seen a boolean thats bitpacked like this be set with a ternary.
if its something we want to start doing then we can see how others feel about it, but afaik its not something we've ever done

fig02: also should this be shifting left? this is packing the values in right

Dragorn421: Yeah it should, duh me

fig02: id also add a space between the & on both sides

Dragorn421: The formatter made me (not) do it!
(I try to still format parts of .h files to help consistency)

Dragorn421:
Actually tested `WATERBOX_PROPERTIES` a bit this time,
```
bdan_sceneCollisionHeader_013074WaterBoxes[2] : 0x00006000 -> WATERBOX_PROPERTIES(0, 0, 3, false)
HAKAdanCH_sceneCollisionHeader_00A558WaterBoxes[9] : 0x0000220C -> WATERBOX_PROPERTIES(12, 2, 1, false)
spot00_sceneCollisionHeader_008464WaterBoxes[0] : 0x00000102 -> WATERBOX_PROPERTIES(2, 1, 0, false)
```

Roman971 (approved):

PR #1290, opened 2022-06-19
Cleanup z_debug (reg editor)

Dragorn421:


include/z64.h
```diff
@@ -82,15 +82,13 @@ typedef struct{
     /* 0x04 */ MtxF mf;
 } HorseStruct;
 
-// Game Info aka. Static Context (dbg ram start: 80210A10)
```
Roman971, line 85: I feel like keeping a mention of "Static Context" helps since it has been used a lot in the past, but I don't really mind either way

src/code/z_debug.c
```diff
@@ -3,223 +3,285 @@
...
+    { BTN_START, BTN_R },      // BREG
+    { BTN_START, BTN_A },      // dREG
+    { BTN_START, BTN_B },      // kREG
+    { BTN_START, BTN_CRIGHT }, // bREG
+
 };
 
-char regChar[] = " SOPQMYDUIZCNKXcsiWAVHGmnBdkb";
+char sRegGroupChars[REG_GROUPS] = {
+    ' ', //  REG
```
Roman971, line 64: The full string kinda makes sense to me tbh, but I guess it's debatable.
Although I would at least omit the REG name comments since the names literally come from these characters, so it's a bit redundant imo

Dragorn421: Yeah my idea is that it confirms the letters associated with each group, and it helps realize that this is where they come from
Also, formatting and symmetry with `sRegGroupInputCombos`

?

include/z64.h
Dragorn421: Is the `GameInfo` name itself definitive?

The debug string we have with `game_info` is for `gSaveContext` (which imo would be a better fit for "static context" btw) (`osSyncPrintf("\ngame_info.mode=[%d] restart_flag", ((void)0, gSaveContext.respawnFlag));`)

If I was to suggest renaming `GameInfo`, I'd probably pick `RegsContext` or something like that

Roman971: It's not supposed to be definitive, but it hasn't been important to rename since you don't actually refer to it when using regs. Right now I would probably suggest something like `GlobalRegs` or `GlobalRegsContext`, but I haven't put much thought into it

mzxrules (commented):
thanks for fixing my awful code :)

include/z64.h
fig02: regarding actually keeping the comment, I feel like in general we have been just hard committing to new names of things (for example, removing mentions of global context)
it may be confusing for people not used to the new names yet, but itll be better for cleanliness and consistency overall

as for what to actually call gameinfo/static context, i dont have too much of an opinion. naming it after regs would be good

include/z64.h
```diff
@@ -82,15 +82,13 @@ typedef struct{
...
-typedef struct {
-    /* 0x00 */ s32  regPage;   // 1 is first page
-    /* 0x04 */ s32  regGroup;  // "register" group (R, RS, RO, RP etc.)
-    /* 0x08 */ s32  regCur;    // selected register within page
-    /* 0x0C */ s32  dpadLast;
-    /* 0x10 */ s32  repeat;
-    /* 0x14 */ s16  data[REG_GROUPS * REG_PER_GROUP]; // 0xAE0 entries
+typedef struct {
+    /* 0x00 */ s32  regPage; // 0: no page selected (reg editor is not active); 1: first page; `REG_PAGES`: last page
+    /* 0x04 */ s32  regGroup; // Reg group, indexed from 0 to `REG_GROUPS`-1. Each group has its own character to identify it.
```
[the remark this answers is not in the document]
fig02: agree here, dont think you need to restate the variable name

src/code/z_debug.c
```diff
@@ -3,223 +3,285 @@
...
 }
 
-// Draw Memory Viewer
+// Draw the reg editor
 void func_80063C04(GfxPrint* printer) {
     s32 i;
-    s32 page = (gGameInfo->regPage * REG_PER_PAGE) - REG_PER_PAGE;
-    s32 regGroup = (gGameInfo->regGroup * REG_PAGES + gGameInfo->regPage) * REG_PER_PAGE - REG_PER_PAGE;
+    s32 startIndexInGroup = (gGameInfo->regPage - 1) * REGS_PER_PAGE;
+    s32 startIndexInData = ((gGameInfo->regGroup * REG_PAGES) + gGameInfo->regPage - 1) * REGS_PER_PAGE;
```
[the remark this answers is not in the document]
fig02: I think these suggestions are probably an improvement yeah, at least as someone who didnt read the code very closely these suggestions give me a better idea at a glance of what they are

include/z64.h
Dragorn421: I'm going to leave "static context" out and renaming GameInfo to another PR

fig02 (approved):

Roman971 (approved):

PR #1293, opened 2022-06-21
Doc, macros, enums for `SurfaceType.data`

Dragorn421:
See `include/z64bgcheck.h` and `src/code/z_bgcheck.c` for the main changes.
The rest of the changes are just using the defines/enums/renamed functions.

Another pass will be needed to update a whole bunch of variable names, I'd like to do that later to avoid possibly doubling this PR's size.

Note: I don't think floor type/property/effect should be definitive, I just picked different words to have something to refer them as.


fig02 (commented):
in general i think these macros could use more underscores for readability
 `SURFACETYPE0_ISHORSEBLOCKED`
 `SURFACETYPE1_CONVEYORSPEED`
 etc
 
 SURFACETYPEX_ as a prefix makes sense, but after that initial prefix i think we can add some more underscores between words

fig02 (approved):

mzxrules:
Thinking on it a bit, the surface type getter macros are an odd design choice. Are end users suppose to use the functions or the macros?

fig02:
I do think mzx has a point actually. This is a rare case where there are functions dedicated to doing these unpacking operations.
Others have already expressed interest in macro-ifying the shifting access of bit packed values (SHIFTL/R), so we can probably just use those directly in the function (later), without making specific macros for the getters?

Dragorn421:
master:
```c
u32 SurfaceType_GetConveyorSpeed(CollisionContext* colCtx, CollisionPoly* poly, s32 bgId) {
    return SurfaceType_GetData(colCtx, poly, bgId, 1) >> 18 & 7;
}
```

This PR:
```c
#define SURFACETYPE1_CONVEYOR_SPEED_SHIFT 18
#define SURFACETYPE1_CONVEYOR_SPEED_MASK 0x001C0000
#define SURFACETYPE1_CONVEYOR_SPEED(data1)            \
    (((data1) >> SURFACETYPE1_CONVEYOR_SPEED_SHIFT) & \
     (SURFACETYPE1_CONVEYOR_SPEED_MASK >> SURFACETYPE1_CONVEYOR_SPEED_SHIFT))

u32 SurfaceType_GetConveyorSpeed(CollisionContext* colCtx, CollisionPoly* poly, s32 bgId) {
    return SURFACETYPE1_CONVEYOR_SPEED(SurfaceType_GetData(colCtx, poly, bgId, 1));
}
```

Do you want (assuming it matches):
```c
u32 SurfaceType_GetConveyorSpeed(CollisionContext* colCtx, CollisionPoly* poly, s32 bgId) {
    return _SHIFTR(SurfaceType_GetData(colCtx, poly, bgId, 1), 18, 3);
}

_SHIFTL(SURFACETYPE_CONVEYOR_SPEED_SLOW, 18, 3);
```

or

```c
#define SURFACETYPE1_CONVEYOR_SPEED_SHIFT 18
#define SURFACETYPE1_CONVEYOR_SPEED_BITS 3

u32 SurfaceType_GetConveyorSpeed(CollisionContext* colCtx, CollisionPoly* poly, s32 bgId) {
    return _SHIFTR(SurfaceType_GetData(colCtx, poly, bgId, 1), SURFACETYPE1_CONVEYOR_SPEED_SHIFT, SURFACETYPE1_CONVEYOR_SPEED_BITS);
}

_SHIFTL(SURFACETYPE_CONVEYOR_SPEED_SLOW, SURFACETYPE1_CONVEYOR_SPEED_SHIFT, SURFACETYPE1_CONVEYOR_SPEED_BITS);
```

Note this PR doesn't have any `_SHIFTL` kind of macro (they would eventually be used in `SurfaceType` arrays in collision data like ```c
SurfaceType spot00_sceneCollisionHeader_008464SurfaceType[] = {
    { 0x40000000, 0x00000FC2 },
```)

**Comment by fig02** [comment]
the details on exactly how this will be done is something that tharo is figuring out and will propose for actor params. but by extension, whatever we agree on will be used for all kinds of bitpacked values id imagine

id personally say to leave as raw shifts and ands until we figure out those details

and yeah, making corresponding set macros to be used to define the data will use the same method i imagine

**Review by fig02 — APPROVED** [review]

File: include/z64bgcheck.h [hunk 1]
```diff
@@ -87,11 +87,94 @@ typedef struct {
...
+    /*  4 */ SURFACETYPE_FLOOR_TYPE_4,
+    /*  5 */ SURFACETYPE_FLOOR_TYPE_5,
+    /*  6 */ SURFACETYPE_FLOOR_TYPE_6,
+    /*  7 */ SURFACETYPE_FLOOR_TYPE_7,
+    /*  8 */ SURFACETYPE_FLOOR_TYPE_8,
+    /*  9 */ SURFACETYPE_FLOOR_TYPE_9,
+    /* 10 */ SURFACETYPE_FLOOR_TYPE_10,
+    /* 11 */ SURFACETYPE_FLOOR_TYPE_11,
+    /* 12 */ SURFACETYPE_FLOOR_TYPE_12
+} SurfaceTypeFloorType;
```
Roman971, line 104: Do all of these enums need to be prefixed with `SURFACETYPE_`? I understand that they are packed in the "surface type data", but I feel like the values themselves could make sense as their own enums. For example I think it could be fine to just have a `FLOOR_TYPE_` or `WALL_TYPE_` enum

Dragorn421: I guess it can go if it doesn't clash with anything. More opinions?

src/code/z_bgcheck.c
```diff
@@ -602,15 +616,15 @@ f32 BgCheck_RaycastFloorStatic(StaticLookup* lookup, CollisionContext* colCtx, u
 
 /**
  * Compute wall displacement on `posX` and `posZ`
- * sets `wallPolyPtr` to `poly` if `wallPolyPtr` is NULL or not a damage wall
+ * sets `wallPolyPtr` to `poly` if `wallPolyPtr` is NULL or doesn't have `surfaceData[0] & 0x08000000` set
```
mzxrules, line 619: I'm pretty confident that this particular check, for whatever reason, is checking if the surface is a damage wall.

include/z64bgcheck.h
fig02: yea I think itll be fine, they are pretty clearly denoted by their own prefix and dont need to be grouped with the variable they are packed into in this case

Roman971: Among all those I think the only ones that would need a more specific prefix are the sfx types, which could probably be `SURFACE_SFX_TYPE_`

src/code/z_bgcheck.c
Dragorn421: this particular check yes, I think it prioritizes finding a damage wall so if Link is at a corner with a safe wall and a hurty wall, the hurty wall is picked and link gets damaged

But the flag itself is used for a bit more, including for stuff on floors

fig02 (approved):

Roman971 (approved):

PR #1295, opened 2022-06-22
Document `SaveContext.gameMode`

Dragorn421:
As they have been known for a while.

It's weird though that the usage on transitions is so jank:
- the check in `Play_Init` to force `TRANS_TYPE_FADE_BLACK_SLOW` would have been better replaced with simply `gSaveContext.nextTransitionType = TRANS_TYPE_FADE_BLACK_SLOW` in `Opening_SetupTitleScreen`
- When setting `GAMEMODE_FILE_SELECT`, the hardcoded swap to file select in the transition code is only implemented for instance type transitions

I would like to name gamemode_3 (most likely credits as already known) later, along with the involved cutscenes


fig02:
no need to wait on the credits one imo, its 100% what they are. its adult talking to zelda in hyrule field (with the scene hidden so it looks like clouds) and gerudo valley. its even the same scene + setup that gets checked for the dumb credits hack :D 

`(gEntranceTable[((void)0, gSaveContext.entranceIndex)].scene == SCENE_SPOT09) && gSaveContext.sceneSetupIndex == 6`

fig02 (approved):

Roman971 (approved):

include/z64save.h
```diff
@@ -245,6 +245,13 @@ typedef enum {
     /* 3 */ SUNSSONG_SPECIAL // time does not advance, but signals the song was played. used for freezing redeads
 } SunsSongState;
 
+typedef enum {
+    /* 0 */ GAMEMODE_NORMAL,
+    /* 1 */ GAMEMODE_TITLE_SCREEN,
+    /* 2 */ GAMEMODE_FILE_SELECT, // Note: only instance type transitions swap to file select
```
EllipticEllipsis, line 251: What's an instance type transition?

src/overlays/actors/ovl_En_Light/z_en_light.c
```diff
@@ -49,7 +49,7 @@ void EnLight_Init(Actor* thisx, PlayState* play) {
     EnLight* this = (EnLight*)thisx;
     s16 yOffset;
 
-    if (gSaveContext.gameMode == 3) {
+    if (gSaveContext.gameMode == GAMEMODE_END_CREDITS) {
         // special case for the credits
```
EllipticEllipsis, line 53: Comment probably no longer necessary.

include/z64save.h
Dragorn421: (copying from discord)

transitions start by setting `transitionTrigger`, `nextEntranceIndex`, `transitionType`, for example:
https://github.com/zeldaret/oot/blob/e16779cc338c53b91387a28e3e97cd6f5f036a66/src/overlays/actors/ovl_Boss_Ganon/z_boss_ganon.c#L1498-L1500

they start being handled in z_play with the check on `transitionTrigger`:
https://github.com/zeldaret/oot/blob/e16779cc338c53b91387a28e3e97cd6f5f036a66/src/code/z_play.c#L481-L483

then a bunch of lines below, the transition type is passed to `Play_SetupTransition`
https://github.com/zeldaret/oot/blob/e16779cc338c53b91387a28e3e97cd6f5f036a66/src/code/z_play.c#L531

In `Play_SetupTransition`,
For instance type transitions, `transitionMode` isn't changed so stays `TRANS_MODE_SETUP`
for non-instance types, `transitionMode` is set accordingly

when `Play_SetupTransition` returns, the following block breaks the switch for non-instance modes
https://github.com/zeldaret/oot/blob/e16779cc338c53b91387a28e3e97cd6f5f036a66/src/code/z_play.c#L536-L539
but instance modes continue

eventually
https://github.com/zeldaret/oot/blob/e16779cc338c53b91387a28e3e97cd6f5f036a66/src/code/z_play.c#L630-L639
is reached as part of processing instance modes and that's where file select is switched to

PR #1298, opened 2022-06-23
Fix misc 15

Dragorn421:


src/code/z_actor.c
```diff
@@ -2182,10 +2181,10 @@ void Actor_Draw(PlayState* play, Actor* actor) {
     Lights_Draw(lights, play->state.gfxCtx);
 
     if (actor->flags & ACTOR_FLAG_12) {
-        Matrix_SetTranslateRotateYXZ(
-            actor->world.pos.x + play->mainCamera.skyboxOffset.x,
-            actor->world.pos.y + (f32)((actor->shape.yOffset * actor->scale.y) + play->mainCamera.skyboxOffset.y),
-            actor->world.pos.z + play->mainCamera.skyboxOffset.z, &actor->shape.rot);
+        Matrix_SetTranslateRotateYXZ(actor->world.pos.x + play->mainCamera.skyboxOffset.x,
+                                     actor->world.pos.y +
+                                         ((actor->shape.yOffset * actor->scale.y) + play->mainCamera.skyboxOffset.y),
```
EllipticEllipsis, line 2186: This looks like the addends are just in the wrong order in the first place to me.

src/code/z_bgcheck.c
```diff
@@ -4171,17 +4171,17 @@ u32 SurfaceType_IsWallDamage(CollisionContext* colCtx, CollisionPoly* poly, s32
 /**
  * Zora's Domain WaterBox in King Zora's Room
  */
-WaterBox zdWaterBox = { -348, 877, -1746, 553, 780, 0x2104 };
+WaterBox sZdWaterBox = { -348, 877, -1746, 553, 780, 0x2104 };
```
EllipticEllipsis, line 4174: I think this ought to be ZD if we must abbreviate it like this, since it's 2 words. But it's used so little I think it'd be fine to go the route of spelling it out in full.

src/code/z_bgcheck.c
```diff
@@ -4191,10 +4191,10 @@ f32 zdWaterBoxMaxZ = -967.0f;
 s32 WaterBox_GetSurface1(PlayState* play, CollisionContext* colCtx, f32 x, f32 z, f32* ySurface,
```
EllipticEllipsis, line 4191: Might as well mention the special case in this function in the description, too.

src/code/z_actor.c
Dragorn421: It doesn't seem to match written some other way but idk, float regalloc

src/code/z_bgcheck.c
Dragorn421: idk, the hardcoded stuff is right there taking 90% of  the body

There's so much hardcoded stuff everywhere it seems more damaging than worth it to comment on every single function that has this kind of thing?

src/code/z_message_PAL.c
```diff
@@ -3172,7 +3172,7 @@ void Message_Update(PlayState* play) {
                     R_TEXTBOX_TEXHEIGHT = 512;
                 } else {
                     Message_GrowTextbox(msgCtx);
-                    Audio_PlaySoundIfNotInCutscene(0);
+                    Audio_PlaySoundIfNotInCutscene(NA_SE_PL_WALK_GROUND - SFX_FLAG);
```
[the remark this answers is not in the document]
Dragorn421: hey it's the same jank block of code with the `R_SCENE_CAM_TYPE` stuff :PepeGa:

I guess it does nothing because of `- SFX_FLAG`, it plays sound for one frame (notice `msgMode = MSGMODE_TEXT_BOX_GROWING` below), idk if it can be heard at all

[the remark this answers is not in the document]
engineer124: There's seems to be a lot of weirdness when it comes to sfxId = 0. In `z_message` alone, there's plenty of instances of `Audio_PlaySoundGeneral(0, ...)`, and there's an early return for if the `sfxId = 0`:
https://github.com/engineer124/oot/blob/eb462b298e54b402066dced6ad2629c1514f188e/src/code/code_800F7260.c#L233-L235

I'm hesitant to say it was intended as `NA_SE_PL_WALK_GROUND` and would leave this as 0 until it can be worked out exactly what happens with `sfxId = 0`

src/code/z_actor.c
EllipticEllipsis: Yep, looks like it really is that stupid, how annoying.

src/code/z_bgcheck.c
EllipticEllipsis: And in the Doxygen?

src/code/z_message_PAL.c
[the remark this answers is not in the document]
Dragorn421: Oh really, weird
Can I slap a TODO here? like `// TODO: this may be NA_SE_PL_WALK_GROUND - SFX_FLAG, or not, investigate sfxId=0`

src/code/z_bgcheck.c
Dragorn421: I'm overall unsure about the purpose of doxygen generated docs, ngl, what would someone be doing that they don't have the code under their eyes?

In particular this hardcoded thing seems more like a gotcha for mapmaking and I don't expect mapmakers to look at this particular function, neither in source nor in doxygen

src/code/z_message_PAL.c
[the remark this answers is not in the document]
engineer124: Sure

fig02 (approved):

Roman971 (approved):

PR #1299, opened 2022-06-24
[Doc] actor En_Skb (Stalchild)

krm01:
I tried to be consistent with naming conventions I've come across in other actors, but this is my first real doc here so I'm sure some names/choices are not ideal (I suspect the `actionState` part will be contentious)

### DIFF
--- a/src/overlays/actors/ovl_En_Skb/z_en_skb.c
+++ b/src/overlays/actors/ovl_En_Skb/z_en_skb.c
@@ -250,7 +250,7 @@ void EnSkb_SetupWalkForward(EnSkb* this) {
     Animation_Change(&this->skelAnime, &gStalchildWalkingAnim, 0.96000004f, 0.0f,
                      Animation_GetLastFrame(&gStalchildWalkingAnim), ANIMMODE_LOOP, -4.0f);
     this->actionState = ENSKB_ACTION_WALKING;
-    this->headlessWalkDirectionOffset = 0;
+    this->headlessWalkYawOffset = 0;
     this->actor.speedXZ = this->actor.scale.y * 160.0f;
     EnSkb_SetupAction(this, EnSkb_WalkForward);
 }
@@ -262,9 +262,9 @@ void EnSkb_WalkForward(EnSkb* this, PlayState* play) {
     Player* player = GET_PLAYER(play);
 
     if ((this->bodybreakState != 0) && ((play->gameplayFrames & 0xF) == 0)) {
-        this->headlessWalkDirectionOffset = Rand_CenteredFloat(50000.0f);
+        this->headlessWalkYawOffset = Rand_CenteredFloat(50000.0f);
     }
-    Math_SmoothStepToS(&this->actor.shape.rot.y, (this->actor.yawTowardsPlayer + this->headlessWalkDirectionOffset), 1, 0x2EE, 0);
+    Math_SmoothStepToS(&this->actor.shape.rot.y, (this->actor.yawTowardsPlayer + this->headlessWalkYawOffset), 1, 0x2EE, 0);
     this->actor.world.rot.y = this->actor.shape.rot.y;
     thisKeyFrame = this->skelAnime.curFrame;
     SkelAnime_Update(&this->skelAnime);
--- a/src/overlays/actors/ovl_En_Skb/z_en_skb.c
+++ b/src/overlays/actors/ovl_En_Skb/z_en_skb.c
@@ -228,7 +228,7 @@ void EnSkb_SetupDespawn(EnSkb* this) {
     Animation_Change(&this->skelAnime, &gStalchildUncurlingAnim, -1.0f,
                      Animation_GetLastFrame(&gStalchildUncurlingAnim), 0.0f, ANIMMODE_ONCE, -4.0f);
     this->actionState = ENSKB_ACTION_BURIED;
-    this->enableATCollision = 0;
+    this->enableATCollision = false;
     this->actor.flags &= ~ACTOR_FLAG_0;
     this->actor.speedXZ = 0.0f;
     Audio_PlayActorSound2(&this->actor, NA_SE_EN_AKINDONUTS_HIDE);
@@ -309,9 +309,9 @@ void EnSkb_Attack(EnSkb* this, PlayState* play) {
     frameData = this->skelAnime.curFrame;
     if (frameData == 3) {
         Audio_PlayActorSound2(&this->actor, NA_SE_EN_STALKID_ATTACK);
-        this->enableATCollision = 1;
+        this->enableATCollision = true;
     } else if (frameData == 6) {
-        this->enableATCollision = 0;
+        this->enableATCollision = false;
     }
     if (this->collider.base.atFlags & AT_BOUNCED) {
         this->collider.base.atFlags &= ~(AT_HIT | AT_BOUNCED);
@@ -326,7 +326,7 @@ void EnSkb_SetupRecoil(EnSkb* this) {
                      ANIMMODE_ONCE_INTERP, 0.0f);
     this->collider.base.atFlags &= ~AT_BOUNCED;
     this->actionState = ENSKB_ACTION_RECOILING;
-    this->enableATCollision = 0;
+    this->enableATCollision = false;
     EnSkb_SetupAction(this, EnSkb_Recoil);
 }
 
@@ -341,7 +341,7 @@ void EnSkb_SetupStunned(EnSkb* this) {
         this->actor.speedXZ = 0.0f;
     }
     Audio_PlayActorSound2(&this->actor, NA_SE_EN_GOMA_JR_FREEZE);
-    this->enableATCollision = 0;
+    this->enableATCollision = false;
     this->actionState = ENSKB_ACTION_STUNNED;
     EnSkb_SetupAction(this, EnSkb_Stunned);
 }
@@ -443,7 +443,7 @@ void EnSkb_CheckDamage(EnSkb* this, PlayState* play) {
     if ((this->actionState != ENSKB_ACTION_DYING) && (this->actor.bgCheckFlags & (BGCHECKFLAG_WATER | BGCHECKFLAG_WATER_TOUCH)) &&
         (this->actor.yDistToWater >= 40.0f)) {
         this->actor.colChkInfo.health = 0;
-        this->enableATCollision = 0;
+        this->enableATCollision = false;
         EnSkb_SetupDeath(this, play);
     } else if (this->actionState >= ENSKB_ACTION_ATTACKING) {
         if (this->collider.base.acFlags & AC_HIT) {
@@ -451,7 +451,7 @@ void EnSkb_CheckDamage(EnSkb* this, PlayState* play) {
             if (this->actor.colChkInfo.damageEffect != 6) {
                 this->lastDamageEffect = this->actor.colChkInfo.damageEffect;
                 Actor_SetDropFlag(&this->actor, &this->collider.elements[1].info, true);
-                this->enableATCollision = 0;
+                this->enableATCollision = false;
                 if (this->actor.colChkInfo.damageEffect == 1) {
                     if (this->actionState != ENSKB_ACTION_STUNNED) {
                         Actor_SetColorFilter(&this->actor, 0, 0x78, 0, 0x50);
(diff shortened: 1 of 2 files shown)


src/overlays/actors/ovl_En_Skb/z_en_skb.c
```diff
@@ -187,26 +187,26 @@ void EnSkb_Destroy(Actor* thisx, PlayState* play) {
     Collider_DestroyJntSph(play, &this->collider);
 }
 
-void func_80AFCD60(EnSkb* this) {
+void EnSkb_Main(EnSkb* this) {
```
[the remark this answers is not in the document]
krm01: Yea this is kind of the function every other state transition goes back to by default unless it dies. that is `Main` to me.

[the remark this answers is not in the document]
mzxrules: Except it isn't, the entry point is always `RiseFromGround`.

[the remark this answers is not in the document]
krm01: Sure, if you'd prefer a different name for it let me know what it should be and I'll change it. I was just answering your question.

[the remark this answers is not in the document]
fig02: you can name this around "choose action" or "decide action"
but yea def wouldnt use main. in this codebase main is mostly an overarching term that encompasses both the update and drawing process (for systems that are split up that way anyway, which actors are)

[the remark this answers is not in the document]
krm01: Oh I gotcha - okay sounds good to me. Fixed in 9ebffc7

src/overlays/actors/ovl_En_Skb/z_en_skb.c
```diff
@@ -245,25 +246,25 @@ void func_80AFCFF0(EnSkb* this, PlayState* play) {
...
     f32 playSpeed;
     Player* player = GET_PLAYER(play);
 
-    if ((this->unk_283 != 0) && ((play->gameplayFrames & 0xF) == 0)) {
-        this->unk_288 = Rand_CenteredFloat(50000.0f);
+    if ((this->bodybreakState != 0) && ((play->gameplayFrames & 0xF) == 0)) {
+        this->headlessWalkDirectionOffset = Rand_CenteredFloat(50000.0f);
     }
-    Math_SmoothStepToS(&this->actor.shape.rot.y, (this->actor.yawTowardsPlayer + this->unk_288), 1, 0x2EE, 0);
+    Math_SmoothStepToS(&this->actor.shape.rot.y, (this->actor.yawTowardsPlayer + this->headlessWalkDirectionOffset), 1, 0x2EE, 0);
```
Dragorn421, line 267: this file needs formatting

We use `clang-format` (preferably `clang-format-11`) and `clang-tidy` for automatic code formatting
The formatter can be run on all .c files with `./format.sh` (script at the repo's root)
or on a single file `./format.sh src/overlays/actors/ovl_En_Skb/z_en_skb.c`

Dragorn421, line 267: imo `headlessWalkDirectionOffset` -> `headlessWalkYawOffset` to help saying it's an angle (imo "direction" is more often a vector)

src/overlays/actors/ovl_En_Skb/z_en_skb.c
```diff
@@ -498,19 +499,19 @@ void EnSkb_Update(Actor* thisx, PlayState* play) {
...
+    EnSkb_CheckDamage(this, play);
     Actor_MoveForward(&this->actor);
     Actor_UpdateBgCheckInfo(play, &this->actor, 15.0f, 30.0f, 60.0f,
                             UPDBGCHECKINFO_FLAG_0 | UPDBGCHECKINFO_FLAG_2 | UPDBGCHECKINFO_FLAG_3 |
                                 UPDBGCHECKINFO_FLAG_4);
     this->actionFunc(this, play);
     this->actor.focus.pos = this->actor.world.pos;
     this->actor.focus.pos.y += (3000.0f * this->actor.scale.y);
-    if (this->unk_281 != 0) {
+    if (this->hitboxActive != 0) {
```
Dragorn421, line 510: We avoid calling colliders "hitbox" because it's too vague for OoT's various colliders

`enableAT` would probably be okay

Also this member seems to be a boolean: you can replace the `= 0`, `= 1` with `= false`, `= true` and this condition can just be `if (this->enableAT)` (it usually doesn't break matching)

src/overlays/actors/ovl_En_Skb/z_en_skb.c
```diff
@@ -534,7 +535,7 @@ s32 EnSkb_OverrideLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Vec3f* p
         } else {
             *dList = NULL;
         }
-    } else if ((limbIndex == 12) && ((this->unk_283 & 2) != 0)) {
+    } else if ((limbIndex == 12) && ((this->bodybreakState & 2) != 0)) { // jaw, don't draw if headless
```
Dragorn421, line 538: I'm not sure about the name for `bodybreakState`,

1) it's flags, not a single state value
2) it's not just for the BodyBreak system, it's also for the actor's "own" usage

I don't have a suggestion right now

Dragorn421 (commented):
Nice job!

src/overlays/actors/ovl_En_Skb/z_en_skb.h
```diff
@@ -8,17 +8,26 @@ struct EnSkb;
...
 typedef void (*EnSkbActionFunc)(struct EnSkb*, PlayState*);
 
+
+#define ACTION_STATE_SPAWNING 0
+#define ACTION_STATE_DYING 1
+#define ACTION_STATE_DAMAGED 2
+#define ACTION_STATE_ATTACKING 3
+#define ACTION_STATE_WALKING 4
+#define ACTION_STATE_RECOILING 5
+#define ACTION_STATE_STUNNED 6
```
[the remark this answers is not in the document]
krm01: So I originally had this as an enum but it wasn't matching, idk what I was doing wrong. That's the only reason I changed to flags. I'll try again with enum and the name changes.

src/overlays/actors/ovl_En_Skb/z_en_skb.c
```diff
@@ -468,26 +469,26 @@ void func_80AFD968(EnSkb* this, PlayState* play) {
...
+                    if (this->bodybreakState == 0) {
                         if ((this->actor.colChkInfo.damageEffect == 0xD) ||
                             ((this->actor.colChkInfo.damageEffect == 0xE) &&
                              ((player->meleeWeaponAnimation >= PLAYER_MWA_RIGHT_SLASH_1H &&
                                player->meleeWeaponAnimation <= PLAYER_MWA_LEFT_COMBO_2H) ||
                               (player->meleeWeaponAnimation == PLAYER_MWA_BACKSLASH_RIGHT ||
                                player->meleeWeaponAnimation == PLAYER_MWA_BACKSLASH_LEFT)))) {
                             BodyBreak_Alloc(&this->bodyBreak, 2, play);
-                            this->unk_283 = 1;
+                            this->bodybreakState = 1; // prepare for decapitation
```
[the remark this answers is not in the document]
krm01: Oops yeah I didn't mean to leave that in lol will tweak

src/overlays/actors/ovl_En_Skb/z_en_skb.c
krm01: Yeah this is one of the things I am least happy about... it's usage is very weird to me, but certainly related to bodybreak stuff. I wanted to give names to the values they use, but I honestly couldn't come up with any sensible naming scheme.

0 is "nothing special"
1 is "prepare for becoming headless"
2 is "currently headless"
4 is "prepare for falling apart completely"
8 is "done falling apart"

as best I can tell.

src/overlays/actors/ovl_En_Skb/z_en_skb.c
```diff
@@ -219,22 +220,22 @@ void func_80AFCE5C(EnSkb* this, PlayState* play) {
...
     }
 }
 
-void func_80AFCF48(EnSkb* this) {
+void EnSkb_SetupDespawn(EnSkb* this) {
     Animation_Change(&this->skelAnime, &gStalchildUncurlingAnim, -1.0f,
                      Animation_GetLastFrame(&gStalchildUncurlingAnim), 0.0f, ANIMMODE_ONCE, -4.0f);
-    this->unk_280 = 0;
-    this->unk_281 = 0;
+    this->actionState = ACTION_STATE_SPAWNING;
```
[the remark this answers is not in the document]
krm01: That's true, I like `buried`, will update.

src/overlays/actors/ovl_En_Skb/z_en_skb.c
krm01: Agree, changed.

src/overlays/actors/ovl_En_Skb/z_en_skb.h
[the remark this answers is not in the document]
EllipticEllipsis: It probably doesn't need to be in the header anyway, I doubt anything else wants to know about it.

Dragorn421 (approved):

Roman971 (approved):
Looks good to me overall 👍 
The only thing I would comment on is `bodybreakState` like Dragorn mentioned, but I don't really have a suggestion either (aside maybe from using "flags")

src/overlays/actors/ovl_En_Skb/z_en_skb.c
```diff
@@ -142,8 +149,8 @@ static InitChainEntry sInitChain[] = {
 
 void EnSkb_Init(Actor* thisx, PlayState* play) {
     EnSkb* this = (EnSkb*)thisx;
-    s16 paramOffsetBody;
     s16 paramOffsetArm;
+    s16 paramOffsetBody;
```
fig02, line 153: i know this wasnt your change, but imo theres no need to say in the temp name that it came from params. `armOffset` and `bodyOffset` seem fine

src/overlays/actors/ovl_En_Skb/z_en_skb.c
```diff
@@ -4,28 +4,35 @@
...
 
+typedef enum {
+    ENSKB_ACTION_BURIED,
+    ENSKB_ACTION_DYING,
+    ENSKB_ACTION_DAMAGED,
+    ENSKB_ACTION_ATTACKING,
+    ENSKB_ACTION_WALKING,
+    ENSKB_ACTION_RECOILING,
+    ENSKB_ACTION_STUNNED
+} EnSkbActionState;
```
fig02, line 15: the term "action state" has always felt pretty redundant to me. To me, an action is a very specific kind of state, so saying both is not necessary

In the armos I used the term "behavior" to describe this kind of variable. That is, it describes the general behavior of the actor without being tied to a specific action/action function.
Behavior might not be the best term for this, but its a pattern that is somewhat common among enemies and itd be good to make it consistent

src/overlays/actors/ovl_En_Skb/z_en_skb.c
```diff
@@ -525,7 +533,7 @@ s32 EnSkb_OverrideLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Vec3f* p
     s16 pad[2];
 
     if (limbIndex == 11) {
-        if ((this->unk_283 & 2) == 0) {
+        if ((this->bodybreakState & 2) == 0) { // head limb, head is still attached
```
fig02, line 536: like previously mentioned by others, I dont think this should named after BodyBreak when it isnt directly related to the BodyBreak system. Its more like it works in parallel with the body break system. all BodyBreak does is spawn the parts as they fall off. But this is more about controlling what gets drawn on the model that still exists

`breakFlags` or `bodyFlags` are fine for me, but using both terms together implies direct relationship with the system named BodyBreak

fig02 (commented):
sorry for the late review on this one as well

src/overlays/actors/ovl_En_Skb/z_en_skb.c
krm01: that's fair, it looks like Armos is currently the only one following this but I am all for alignment on it going forward, I'll take consistency over 100%-correct any day. changed

src/overlays/actors/ovl_En_Skb/z_en_skb.c
krm01: after most recent review, going with `breakFlags`

src/overlays/actors/ovl_En_Skb/z_en_skb.c
fig02: yea behavior is def not common. I dont think we have a decided on term for this just yet. was moreso just putting my thoughts out there on it. we will prob have mroe discussion and be consistent about it soon

fig02 (approved):

PR #1300, opened 2022-06-24
Rename "Scene Setup Index" to "Scene Layer"

fig02:
This is a potentially controversial change, depending on how attached one is to the classic modding name.

This term is heavily inspired by (more like directly taken from) the Wind Waker and Twilight Princess engine. Over there they have a very similar concept of what we call "setups" (it even gets incremented by 1 at night too!)

I think the term is perfect for what it describes. The scene itself can be seen as the "base" and then different layers can be applied which change up what actors are loaded and where, and what cutscenes will play and such.

--

An enum was added for the basic child/adult and day/night layers. The expectation is that individual scenes will define their own layer enums to describe what they are, since they depend on that context.

Some scenes use layers 0-3 for other purposes, which is why those specific cases in z_play were not replaced to use the enum. They will need the scene specific enum whenever it is created


src/code/z_demo.c
```diff
@@ -151,14 +151,14 @@ void func_800645A0(PlayState* play, CutsceneContext* csCtx) {
     Input* input = &play->state.input[0];
 
     if (CHECK_BTN_ALL(input->press.button, BTN_DLEFT) && (csCtx->state == CS_STATE_IDLE) &&
-        (gSaveContext.sceneSetupIndex >= 4)) {
+        (gSaveContext.sceneLayer > SCENE_LAYER_ADULT_NIGHT)) {
```
Dragorn421, line 154: Would it be worth having a `SCENE_LAYER_CUTSCENES_FIRST` (?) name for a 4 here, instead of relying on `SCENE_LAYER_ADULT_NIGHT` being the last non-cutscene layer?

fig02: I had the same thought yeah. Just wasnt sure what an elegant thing to call it would be, or if it was even worth.

Dragorn421: I'm starting to think it would be worth it, it would be better to understand what those checks are, and since all such checks have been `< 4`-like it seems to have been the natural thing to do so far

It could also be used in the main layer computation
`gSaveContext.sceneLayer = 4 + (gSaveContext.cutsceneIndex & 0xF);`

fig02: fwiw not all of them were like that, but a majority of them were. it was whatever the person decompiling felt like putting :D 
but yeah, im just not sure what to call it

fig02: `#define SCENE_LAYER_CUTSCENE_BASE SCENE_LAYER_ADULT_NIGHT + 1`
idk, maybe others will have more ideas

EllipticEllipsis: Well, I was thinking you'd whack a `!` in front of it. but yes, fundamentally it comes down to what makes more sense in context.

fig02: for some reason thought it wouldnt match like that, but it seems to so far. will try out the change and see how it looks

fig02: I like the way it turned out, pushed change

fig02: think i would probably prefer a name like `SCENE_LAYER_IS_CUTSCENE`, but i can also understand wanting to keep it like the other IS_ macros
in general i find `IS_` unfortunate cause it doesnt indicate a clear system that the macro pertains to

include/z64save.h
```diff
@@ -252,6 +252,16 @@ typedef enum {
     /* 3 */ GAMEMODE_END_CREDITS
 } GameMode;
 
+typedef enum {
+    /* 0 */ SCENE_LAYER_CHILD_DAY,
+    /* 1 */ SCENE_LAYER_CHILD_NIGHT,
+    /* 2 */ SCENE_LAYER_ADULT_DAY,
+    /* 3 */ SCENE_LAYER_ADULT_NIGHT    
+} SceneLayer;
```
mzxrules, line 260:
```suggestion
typedef enum {
    /* 0 */ SCENE_LAYER_CHILD_DAY,
    /* 1 */ SCENE_LAYER_CHILD_NIGHT,
    /* 2 */ SCENE_LAYER_ADULT_DAY,
    /* 3 */ SCENE_LAYER_ADULT_NIGHT
    /* 4 */ SCENE_LAYER_CUTSCENE_0,
    /* 5 */ SCENE_LAYER_CUTSCENE_1,
    /* 6 */ SCENE_LAYER_CUTSCENE_2,
    /* 7 */ SCENE_LAYER_CUTSCENE_3,
    /* 8 */ SCENE_LAYER_CUTSCENE_4,
    /* 9 */ SCENE_LAYER_CUTSCENE_5,
    /* A */ SCENE_LAYER_CUTSCENE_6,
    /* B */ SCENE_LAYER_CUTSCENE_7,
    /* C */ SCENE_LAYER_CUTSCENE_8,
} SceneLayer;

#define SCENE_LAYER_HYRULE_FIELD_CHILD 0
#define SCENE_LAYER_HYRULE_FIELD_CHILD_OCARINA_OF_TIME 1
#define SCENE_LAYER_KOKIRI_FOREST_ADULT_HOSTILE 2
#define SCENE_LAYER_KOKIRI_FOREST_ADULT_PEACEFUL 3
```
It seems wrong to introduce this enumeration but then still use raw values anyway.

Additionally day/night, child/adult is true for most cases, but there is a special case for Hyrule Field and Kokiri Forest where this pattern is overridden. 

In Hyrule Field, 0 and 1 chosen depending on whether the Ocarina of Time can be obtained, where 0 is the "normal" scene setup for Child, and 1 is loaded whenever the player is to obtain the Ocarina of Time. 

In Kokiri Forest, 2 and 3 are chosen depending on whether the Forest Temple is completed, with 2 being the forest with enemies, and 3 being the forest with everyone safe and the deku sprout spawned.

include/z64save.h
```diff
@@ -252,6 +252,16 @@ typedef enum {
...
+
+#define SCENE_LAYER_CUTSCENE_START (SCENE_LAYER_ADULT_NIGHT + 1)
```
mzxrules, line 262:
```suggestion
```
We can replace this with `SCENE_LAYER_CUTSCENE_0` instead if we opt to use it.

include/z64save.h
fig02: The cutscene layer enums will come from the individual scenes. General purpose defines for each cutscene dont make much sense, when you can have scene specific ones that can be named more precisely for their purpose.

I am aware of the special cases where the pattern is overridden, that is why they do not use this enum currently. They will use the hyrule field and kokiri forest enums respectively, when they are made

include/z64save.h
```diff
@@ -252,6 +252,16 @@ typedef enum {
...
+#define IS_CUTSCENE_LAYER (gSaveContext.sceneLayer >= SCENE_LAYER_CUTSCENE_START)
```
mzxrules, line 263: looking through, it seems that more often than not you're negating `IS_CUTSCENE_LAYER`, which makes me think it should be expressed as the inverse instead.

fig02: there are 46 `IS_CUTSCENE_LAYER` and 37 `!IS_CUTSCENE_LAYER`.

fig02: never mind to my previous comment, did the search wrong.
IS_NIGHT is also more commonly negated, but we keep it in that form. I think it makes more sense to keep it as the positive rather than the negative

include/z64save.h
fig02: take for example, kakariko
an enum like such would live in the spot01 header (whever they actually get committed), and can be used whenever the layer is being checked in kakariko specific contexts
```
typedef enum {
    /* 4 */ SPOT01_LAYER_NOCTURNE_PART1 = SCENE_LAYER_CUTSCENE_START
    /* 5 */ SPOT01_LAYER_NOCTURNE_PART2
    /* 6 */ SPOT01_LAYER_WELL_DRAINING
    ...
} KakarikoSceneLayer;
```

mzxrules: would you also have scene specific cutscene indexes?

fig02: I am almost thinking that cutscene indexes should be in terms of layers, something like `GET_CS_INDEX(SPOT01_LAYER_WELL_DRAINING)`

but i havent fleshed out the idea and determined if it makes sense yet

include/z64save.h
Dragorn421, line 262: I'm unsure if this is better over
```suggestion
#define SCENE_LAYER_CUTSCENE_START 4
```
at all

Also I need to think more about it but I may like `SCENE_LAYER_CUTSCENE_FIRST` more, may be dumb but "cutscene start" sounds too much like "start a cutscene", but again idk

include/z64save.h
Dragorn421, line 263: I feel like `IS_CUTSCENE_LAYER(gSaveContext.sceneLayer)` (pass the layer as argument) would be more readable, and there are conditional statements where `gSaveContext.sceneLayer` is explicit anyway, so not having it in some looks inconsistent?

Which also makes me wonder if a macro is worth it at all

include/z64save.h
fig02: I think its worth it from a shiftability point of view, if someone wants to add more "base/non cutscene" layers, for whatever reason
not saying its likely, but may as well make it less hardcoded

I am leaning toward just putting it in the enum though, sort of like the conventional "_MAX" entries, but it has a slightly different purpose

include/z64save.h
fig02: i think it makes sense for it to be explicit when its checking for a specific layer, and not when its not. It is actually pretty helpful to search `gSaveContext.sceneLayer` and see the special cases that arent generalized to "is or is not a cutscene"

dont see a purpose in the arg being supplied personally. its like IS_DAY or IS_NIGHT which is making a more general statement and abstracting the variable used to determine that

fig02: to elaborate, I see the macro as a general query of "is the current scene layer a cutscene?"
not
"is this layer a cutscene"

Dragorn421: but with `IS_DAY` you only have a few usage of `gSaveContext.nightFlag` (for setting it and indexing things), while `gSaveContext.sceneLayer` and the macro hiding it are used in the same context

fig02: the context is different tho. when you see the variable accessed directly, it concerns code about a very specific layer. when this macro is used it is a generalization about them

include/z64save.h
Dragorn421: yeah my point is i don't think it's more shiftable as `(SCENE_LAYER_ADULT_NIGHT + 1)` because entries past adult_night wouldn't shift the value. so this may seem shiftable when it's not
I'm not a fan of putting it in the enum but maybe it would be fine since the enum documents general layer behaviour / expectations. And it would bury this issue for good

mzxrules: To add to that, there is a use case for having a proper `SCENE_LAYER_MAX = 21` for `SCENE_LAYER, since there's a special cutscene index (0xFFFD) used for certain cutscenes that don't require a "scene layer"

fig02: I dont think it makes sense to define a value like that in terms of layers when it doesnt really involve them at all. Its a magic number as far as im concerned, just exists to satisfy the `gSaveContext.cutsceneIndex >= 0xFFF0` check for a cutscene script to be processed

mzxrules: it's also one more than the last valid cutscene index, and is special cased to not play a cutscene if the value is set.

src/code/z_play.c
```diff
@@ -291,41 +291,42 @@ void Play_Init(GameState* thisx) {
...
+        gSaveContext.sceneLayer = SCENE_LAYER_ADULT_NIGHT;
     }
 
-    tempSetupIndex = gSaveContext.sceneSetupIndex;
+    // save the base scene layer (before accounting for the special cases below) to use later for the transition type
+    baseSceneLayer = gSaveContext.sceneLayer;
+
     if ((gEntranceTable[((void)0, gSaveContext.entranceIndex)].scene == SCENE_SPOT00) && !LINK_IS_ADULT &&
-        gSaveContext.sceneSetupIndex < 4) {
+        !IS_CUTSCENE_LAYER) {
```
Dragorn421, line 309: Should we still use the generic check for cutscene layers when it's scene-specific? (here, hyrule field)

src/code/z_play.c
```diff
@@ -504,16 +505,16 @@ void Play_Update(PlayState* this) {
             switch (this->transitionMode) {
                 case TRANS_MODE_SETUP:
                     if (this->transitionTrigger != TRANS_TRIGGER_END) {
-                        s16 sceneSetupIndex = 0;
+                        s16 sceneLayer = 0;
```
Dragorn421, line 508: child day layer?

(also bruh I didn't know the "continue bgm" flag was only checked on that layer's entrance entry)

src/code/z_scene.c
```diff
@@ -411,10 +411,10 @@ void Scene_CommandAlternateHeaderList(PlayState* play, SceneCmd* cmd) {
 
     osSyncPrintf("\n[ZU]sceneset age    =[%X]", ((void)0, gSaveContext.linkAge));
     osSyncPrintf("\n[ZU]sceneset time   =[%X]", ((void)0, gSaveContext.cutsceneIndex));
-    osSyncPrintf("\n[ZU]sceneset counter=[%X]", ((void)0, gSaveContext.sceneSetupIndex));
+    osSyncPrintf("\n[ZU]sceneset counter=[%X]", ((void)0, gSaveContext.sceneLayer));
 
-    if (gSaveContext.sceneSetupIndex != 0) {
-        altHeader = ((SceneCmd**)SEGMENTED_TO_VIRTUAL(cmd->altHeaders.data))[gSaveContext.sceneSetupIndex - 1];
+    if (gSaveContext.sceneLayer != SCENE_LAYER_CHILD_DAY) {
+        altHeader = ((SceneCmd**)SEGMENTED_TO_VIRTUAL(cmd->altHeaders.data))[gSaveContext.sceneLayer - 1];
```
Dragorn421, line 417: I would be inclined to do
```suggestion
        altHeader = ((SceneCmd**)SEGMENTED_TO_VIRTUAL(cmd->altHeaders.data))[gSaveContext.sceneLayer - SCENE_LAYER_CHILD_NIGHT];
```

src/code/z_play.c
fig02: I saw this as just being an initilizer for the variable and not actually child day, but i think youre right. because of this, the continue bgm flag only matters when applied to the child day layer or a cutscene lol

I think thats probably not intentional? but that seems to be what the case is.
This probably needs a note in entrance_table.h or something

src/code/z_play.c
fig02: id say so, this is still checking if its going to be a cutscene. then if its not, it uses values 0 - 3 for custom purposes

src/code/z_scene.c
```diff
@@ -426,7 +426,7 @@ void Scene_CommandAlternateHeaderList(PlayState* play, SceneCmd* cmd) {
             osSyncPrintf("\nげぼはっ！ 指定されたデータがないでええっす！");
 
             if (gSaveContext.sceneLayer == SCENE_LAYER_ADULT_NIGHT) {
-                altHeader = ((SceneCmd**)SEGMENTED_TO_VIRTUAL(cmd->altHeaders.data))[gSaveContext.sceneLayer - 2];
+                altHeader = ((SceneCmd**)SEGMENTED_TO_VIRTUAL(cmd->altHeaders.data))[gSaveContext.sceneLayer - SCENE_LAYER_ADULT_DAY];
```
mzxrules, line 429: This is just wrong, as it only works in this case by coincidence. Imagine if we added a new `SCENE_LAYER_2`, `SCENE_LAYER_3`, and `SCENE_LAYER_ADULT_DAY` and `SCENE_LAYER_ADULT_NIGHT` were shifted to indexes 4 and 5 for w/e reason. The alt header index calculation would still return 1, mapping to `SCENE_LAYER_2` instead of the intended `SCENE_LAYER_ADULT_DAY`.

src/code/z_scene.c
```diff
@@ -414,7 +414,7 @@ void Scene_CommandAlternateHeaderList(PlayState* play, SceneCmd* cmd) {
     osSyncPrintf("\n[ZU]sceneset counter=[%X]", ((void)0, gSaveContext.sceneLayer));
 
     if (gSaveContext.sceneLayer != SCENE_LAYER_CHILD_DAY) {
-        altHeader = ((SceneCmd**)SEGMENTED_TO_VIRTUAL(cmd->altHeaders.data))[gSaveContext.sceneLayer - 1];
+        altHeader = ((SceneCmd**)SEGMENTED_TO_VIRTUAL(cmd->altHeaders.data))[gSaveContext.sceneLayer - SCENE_LAYER_CHILD_NIGHT];
```
mzxrules, line 417: I don't particularly like introducing the enum here without better documenting how this function behaves, because I feel that adding the enum here obfuscates the logic.

Arguably using the enum in this particular spot is incorrect, since it doesn't make sense as to why subtracting `SCENE_LAYER_CHILD_NIGHT` is necessary to access e.g. `SCENE_LAYERP_CUTSCENE_0`, which this section does.

src/code/z_scene.c
Dragorn421:
```suggestion
                altHeader = ((SceneCmd**)SEGMENTED_TO_VIRTUAL(cmd->altHeaders.data))[gSaveContext.sceneLayer - SCENE_LAYER_CHILD_NIGHT - (SCENE_LAYER_ADULT_DAY - SCENE_LAYER_CHILD_NIGHT)];
```
may be more correct,
`- SCENE_LAYER_CHILD_NIGHT` for the sceneLayer to alternate header index,
`- (SCENE_LAYER_ADULT_DAY - SCENE_LAYER_CHILD_NIGHT)` for changing the start index from child_night to adult_day

but tbh this part of the code looks jank and I don't think there's much to do to salvage it

src/code/z_scene.c
Dragorn421: It's very clear how this function behaves imo. The child day layer is default, and others are "alternate" starting from child_night (hence the child_night usage here)

I thought about this issue and it could be addressed with `#define SCENE_LAYER_ALTERNATE_HEADER_START SCENE_LAYER_CHILD_NIGHT` though I don't particularly support that solution.

imo `SCENE_LAYER_CHILD_NIGHT` simply maps to the first alternate header (if there are alternate headers) so the subtraction here changes the index from being a layer index to an alternate header index

src/code/z_scene.c
Dragorn421: `- (SCENE_LAYER_ADULT_DAY - SCENE_LAYER_CHILD_NIGHT)` could also be `- SCENE_LAYER_ADULT_NIGHT + SCENE_LAYER_ADULT_DAY` since the whole purpose of this block is to use adult_day instead of adult_night

src/code/z_scene.c
mzxrules: but is there any reason to bind that 1 to `SCENE_LAYER_CHILD_NIGHT`? I'd argue no, considering that the alt header system exists in MM, but the concept of `SCENE_LAYER_CHILD_NIGHT` is completely gone.

Dragorn421: I have 0 idea how MM works but this is OoT

and I can't answer to why, it's just what the engine does

fig02: the reason is as dragorn said, its the first of the alternate headers. 
that being said, it may be more worthwhile to define it as the default + 1. This fact will be true no matter how many base "non" cutscene layers there are 

it is true that in mm there is only 1 "gameplay" layer and the rest are cutscenes. In both contexts, the important bit here is that its 1 more than the default layer

Dragorn421:
Two more mentions of ""scene setup"":
```c
#define SCENE_CAM_TYPE_SHOOTING_GALLERY 0x50 // Unreferenced in code, and set only by the main setup of the shooting gallery scene
```
```c
// When entering Gerudo Valley in the right setup, trigger the GC emulator to play the ending movie.
```

Dragorn421 (approved):

src/code/z_scene.c
Roman971: I think I agree with mzx here and we should probably not use the enum values here. To me this condition and array access is intended as "if not the first layer, get the alt header for it", where the first layer is necessarily 0 and the other layers are >= 1. So I think it makes a lot more sense as `if (layer != 0) get alt header of index (layer - 1)`, and there is no reason to bind child day/night to this, since it should be 0 and 1 regardless

src/code/z_scene.c
Roman971: I would go with `- SCENE_LAYER_ADULT_NIGHT + SCENE_LAYER_ADULT_DAY` like Dragon suggested, since the point is to get the alt header for `SCENE_LAYER_ADULT_DAY` and this is right after a `layer == SCENE_LAYER_ADULT_NIGHT` (so it should arguably just be `[SCENE_LAYER_ADULT_DAY]`)

Dragorn421 (approved):

Roman971 (approved):

PR #1301, opened 2022-06-25
Use intptr types in a few specific boot files

Roman971:
Updated 4 boot files to use intptr types: `z_std_dma.c`, `yaz0.c`, `idle.c` and `stackcheck.c`. I can PR these files separately if needed, but I felt the changes were small enough to be fine as a first chunk. I also added `stdint.h` from #1175, which includes `intptr_t` and `uintptr_t` as well as the standard 32 bit defines.

My goal is to work through each file/system and use intptr types wherever necessary, mostly to make the codebase more portable and potentially allow running compilation checks as 64 bit.
However, it's worth noting that I'm only focusing on `intptr_t`/`uintptr_t` and not touching `size_t` or `ptrdiff_t` stuff at all for now, since intptr types are a lot more important and I think it's easier to deal with them on their own (especially with the whole signed `size_t` issue).


Roman971:
Pushed some changes to use pointers in stackcheck, and I also fixed the return type for one of the functions in there

fig02 (approved):

Dragorn421 (approved):

PR #1137, opened 2022-02-10
automated script to build ROM

rav3ndust:
Made a little script that will handle all the steps outlined in the readme to build rom. helpful for my process, hope it comes in handy for saving steps for others. :)

PR #1305, opened 2022-06-27
Created script files for building docker

noremacskich:
Moved the build and running steps from the docker readme into scripts that that can be run on command, without having to memorize the commands themselves

Also, updated the Docker steps with some tips, and have it directly reference the new scripts

### DIFF
--- a/docs/BUILDING_DOCKER.md
+++ b/docs/BUILDING_DOCKER.md
@@ -6,21 +6,40 @@ To use Docker, you'll need either Docker Desktop or Docker Toolbox installed and
 
 You'll also need to prepare a local version of the project with a copied base ROM (see steps [2](../README.md#2-clone-the-repository) and [3](../README.md#3-prepare-a-base-rom) of the Linux instructions).
 
+**_Suggestion:_**  To avoid using sudo, look into "docker rootless" on google
+
 ## 2. Create the Docker image
 
 From inside your local project, run the following command:
 
 ```bash
+sh setup_docker.sh
+```
+
+Or if you are using a different environment
+
+```bash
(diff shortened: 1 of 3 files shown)


Dragorn421:
Thanks for your contribution!

The top level (root folder) is already quite cluttered in the repo, and I think I speak for everyone when I say we want to avoid adding more stuff there

I personally think it's enough to have the commands in the .md file only, like they currently are, especially since they're one-liners.

(for reference there was a similar (?) suggestion in the past for the ubuntu build
- #1137)

The other changes in `BUILDING_DOCKER.md` don't seem very meaty, but maybe someone else will have a different opinion.

noremacskich:
Fair enough, one thought that I had was to move the Dockerfile and docker-compose file, along with the two scripts I created into a docker folder.

I did manage to get that up and running.  The problem with it was it felt a little off, something didn't sit right with me.  The big one being that you need to be in the docker folder in order to build the container, but to run it you could basically be anywhere, as long as you could hit the run script.

So, I dug around a little bit on docker norms, and typically you'd want Dockerfile and docker-compose in the root of the project.  Also figured out why you need to be in the folder when you run the script when reading up on it, it's just a built in limitation of docker.

If we are okay with an atypical layout, I could redo those changes.

Otherwise,  I could move the two scripts into the docker folder, kinda like I was originally wanting to do.

Though, if the norm is that the documentation is good enough, then I could also live with that, I'll just keep the changes to my local fork.

Up to you guys though

fig02:
Thank you for your suggestion, but the current Docker instructions we have currently are good enough for how little they're used

PR #1307, opened 2022-06-28
Some doc on env light settings

Dragorn421:
- `EnvLightSettings`/`LightContext`/`EnvironmentContext` : "fog far" -> "z far" (fog far is always 1000 in OoT i.e. at z far)
- `EnvLightSettings.fogNear` -> `blendRateAndFogNear` and macros (6 bits are `blendRate`, 10 are `fogNear`)
- Add `LIGHTCTX_FOGNEAR_MAX`, `LIGHTCTX_ZFAR_MAX`
- Remove `LightSettings`, alias to `EnvLightSettings` with typedef for ZAPD compatibility
- other minor changes: hexadecimal -> decimal, name two temps used as weights in `Environment_Update`


include/z64environment.h
```diff
@@ -132,17 +132,34 @@ typedef struct {
...
+    /* 0x00 */ u8 ambientColor[3];
+    /* 0x03 */ s8 light1Dir[3];
+    /* 0x06 */ u8 light1Color[3];
+    /* 0x09 */ s8 light2Dir[3];
+    /* 0x0C */ u8 light2Color[3];
+    /* 0x0F */ u8 fogColor[3];
+    /* 0x12 */ s16 fogNear;
+    /* 0x14 */ s16 zFar;
+} CurrentEnvLightSettings; // size = 0x16
+
```
fig02, line 162: it was a conscious choice to not make a new struct, when the only small difference was the blend rate being packed into fogNear.
We can talk about it again, but just saying that it was done on purpose

include/z64light.h
```diff
@@ -43,12 +43,15 @@ typedef struct LightNode {
     /* 0x8 */ struct LightNode* next;
 } LightNode; // size = 0xC
 
+#define LIGHTCTX_FOGNEAR_MAX 996
+#define LIGHTCTX_ZFAR_MAX 12800
+
```
fig02, line 48: no need for ctx in the name imo, Light_ or Lights_ for the system name is identifying enough

include/z64environment.h
Dragorn421: Afaict there is no constraint for `EnvLightSettings` and `CurrentEnvLightSettings` to be the same layout, but I agree they were most likely the same struct initially

Still this seemed to me like a better idea than either
- using `blendRateAndFogNear` as just `fogNear` (not bitpacked) for `EnvironmentContext.lightSettings`
- using `fogNear` as `blendRateAndFogNear` (undocumented bitpacked besides macros and comments) for the scene's `EnvLightSettings` array

fig02: I would rather it be the same structure and document the blend rate better personally

include/z64light.h
Dragorn421: The prefix could be `ENV_` too? the values are in `LightContext` but they're computed in `Environment_Update` 😅

include/z64environment.h
Dragorn421: I actually think the blend rate is more important since it defines how quickly the light settings transition from one settings entry to another (when the env system is doing settings blending), which affects all members and not just fog near.
The effect of tweaking the blend rate seems more immediately noticeable to me than the fog distance

fig02: im not saying that its less important, but overall the fields main purpose is fogNear. Its only the blend rate temporarily.
We will probably just need more opinions, cause I still would like it to be how I originally had it

EllipticEllipsis: This sounds like a union to me.

include/z64environment.h
```diff
@@ -132,17 +132,34 @@ typedef struct {
...
     /* 0x09 */ s8 light2Dir[3];
     /* 0x0C */ u8 light2Color[3];
     /* 0x0F */ u8 fogColor[3];
-    /* 0x12 */ s16 fogNear;
-    /* 0x14 */ s16 fogFar;
+    /* 0x12 */ s16 blendRateAndFogNear;
+    /* 0x14 */ s16 zFar;
 } EnvLightSettings; // size = 0x16
 
+typedef EnvLightSettings LightSettings;
```
Thar0, line 150: It seems appropriate to add a comment explaining what this is for, suggestion:
```suggestion
// For ZAPD compatibility, TODO remove with a ZAPD update
typedef EnvLightSettings LightSettings;
```

include/z64environment.h
Roman971: My opinion is pretty split on this, but in this case I think it makes sense to have separate structs since there isn't any struct copying or anything like that happening, and the combined field only exists in the "scene" struct. It's a bit unfortunate to have an extra struct just for this difference, but it does allow us to name the field properly for the "current" settings, which is the struct used in most places (and externally)

fig02: I still dont fully agree, but I will concede for the sake of moving forward

include/z64environment.h
```diff
@@ -132,17 +132,36 @@ typedef struct {
...
     /* 0x00 */ u8 ambientColor[3];
     /* 0x03 */ s8 light1Dir[3];
     /* 0x06 */ u8 light1Color[3];
     /* 0x09 */ s8 light2Dir[3];
     /* 0x0C */ u8 light2Color[3];
     /* 0x0F */ u8 fogColor[3];
-    /* 0x12 */ s16 fogNear;
-    /* 0x14 */ s16 fogFar;
+    /* 0x12 */ s16 blendRateAndFogNear;
+    /* 0x14 */ s16 zFar;
```
fig02, line 147: I think we should more explicitly document the nuance that fog far is always going to be max, so the same as zFar so changing one is effectively changing the other

I say to add a brief description here (or in the live version of the struct i guess?), cause if I was new to this codebase id be confused by "near and far" being next to each other but they are for different things

include/z64environment.h
Dragorn421: I could add a comment near `CurrentEnvLightSettings` like 

> Nearly the same struct as `EnvLightSettings`, but `CurrentEnvLightSettings.fogNear` is not a bitfield like `blendRateAndFogNear` is

?

fig02: Im not sure tbh. On one hand Im fine with it, on the other im wondering if that would just make things more confusing. may be best to just committing to them being different things and letting the reader work out that small difference if they choose to

include/z64environment.h
Dragorn421: The effective values end up in `LightContext`, where I have some comments already, so I'm not sure where to write anything now

fig02: I think the environment context is the more user-facing system/struct and is where people would go to read about this first

fig02 (approved):

include/z64environment.h
```diff
@@ -132,17 +132,36 @@ typedef struct {
...
 typedef struct {
     /* 0x00 */ u8 ambientColor[3];
     /* 0x03 */ s8 light1Dir[3];
     /* 0x06 */ u8 light1Color[3];
     /* 0x09 */ s8 light2Dir[3];
     /* 0x0C */ u8 light2Color[3];
     /* 0x0F */ u8 fogColor[3];
-    /* 0x12 */ s16 fogNear;
-    /* 0x14 */ s16 fogFar;
+    /* 0x12 */ s16 blendRateAndFogNear; // The blendRate determines how fast the current light settings fade to the new one (under LIGHT_MODE_SETTINGS, otherwise unused). The fogNear is in 0-1000 range (0: starts immediately, 1000: no fog), but is clamped to ENV_FOGNEAR_MAX. The fogFar is almost always 1000 (ends at max depth, set by zFar).
```
Roman971, line 146: This comment goes way over our line width limit (ie. 120 characters), so I think it should be split in multiple lines, or shortened to fit but I doubt that's an option here

Dragorn421: I can put the part on fogFar somewhere else but not sure where. Maybe on zFar with some rewording

It would still be too long, and last time this was raised iirc @fig02 didn't want to break comments in structs over several lines?

Roman971: Not sure how it should be handled but I think this is the only part holding the PR so we should probably do it out one way or another

fig02: I have been meaning to suggest some changes to the comments, but haven't had a moment to write it out.

In short, I think the differences between `EnvLightSettings` and `CurrentEnvLightSettings` are pretty nuanced and need a bit more of a larger explanation, not just a short sentence next to the struct member. imo it warrants a larger comment near the two structs explaining their differences in purpose and the weird quirk about the bitpacked blend rate

When i have a moment to put some thought into it, I can suggest more specific wording

Dragorn421:
(afaik this PR is "Needs discussion" because of a comment in a header file that goes over the line length limit)

include/z64environment.h
fig02: sorry this took almost 2 months for me to look at, was just on the backburner for a while lmao

here is my attempt at reorganizing/rewording a bit <https://github.com/fig02/oot/blob/comments/include/z64environment.h#L139>

to try and explain my thought process:
- I moved the "current" struct above the settings struct that holds data that lives in the scene. The "current" lights are more relevant in terms of the light system and what you see in the game. The light settings in the scene can be see as more of a static version that is derived from that

- the "current" struct has explanations for fogNear and zFar (easier imo to have those explanations separate from the confusing `blendRateAndFogNear` variable )

- a longer explanation of blend rate added between the 2 structs, along with explaining why these two very similar structs are separate

Dragorn421: Looks good to me

fig02 (approved):

Roman971 (approved):

PR #1310, opened 2022-06-30
object_cne Docs

hensldm:
Realized I never PR'd this after doing the same object in MM.


fig02 (approved):
in general i think we could use a more intuitive prefix than their original cne, but that can be done at a later time

Roman971 (approved):

PR #1312, opened 2022-07-02
Cleanup libultra controller routines

Thar0:


include/ultra64/controller.h
```diff
@@ -79,6 +95,26 @@
...
 
+// Accesory detection
+#define CONT_ADDR_DETECT    0x8000
+// Rumble
+#define CONT_ADDR_RUMBLE    0xC000
+// Controller Pak
+// Transfer Pak
+#define CONT_ADDR_GB_POWER  0x8000 // Same as the detection address, but semantically different
+#define CONT_ADDR_GB_BANK   0xA000
+#define CONT_ADDR_GB_STATUS 0xB000
```
fig02, line 106: some line breaks will prob help declutter this, with the comments added in

include/ultra64/controller.h
```diff
@@ -3,6 +3,21 @@
 
 #include "message.h"
 
+/**
+ * Controller channel
+ * Each game controller channel has 4 error bits that are defined in bit 6-7 of
+ * the Rx and Tx data size area bytes. Programmers need to clear these bits
+ * when setting the Tx/Rx size area values for a channel
+ */
+#define CHNL_ERR_NORESP		0x80	/* Bit 7 (Rx): No response error */
```
Dragorn421, line 12: please use spaces for indentation instead of tabs :fearful:

Dragorn421 (commented):
here are a few unimportant things

include/ultra64/controller.h
Thar0: Another one for the SGI fax 📠

fig02 (approved):

src/libultra/io/contramread.c
```diff
@@ -4,51 +4,53 @@
 
 s32 __osPfsLastChannel = -1;
 
+#define READFORMAT(ptr) ((__OSContRamReadFormat*)(ptr))
```
Roman971, line 7: Could this be defined in a common header maybe? It's defined in 3 different files right now which is a bit unfortunate

Roman971 (commented):
I also don't agree with the removal of some parens around conditionals (if anything I think we should add them while doing cleanups)

Roman971 (approved):

PR #1316, opened 2022-07-06
Better AudioMgr_StopAllSfx match

hensldm:
Removes the `(u32)` cast by just using a pointer iterator.

### DIFF
--- a/src/code/code_800C3C20.c
+++ b/src/code/code_800C3C20.c
@@ -5,9 +5,9 @@ u8 sSfxBankIds[] = {
 };
 
 void AudioMgr_StopAllSfx(void) {
-    u32 i;
+    volatile u8* bank;
 
-    for (i = 0; (u32)(i < ARRAY_COUNT(sSfxBankIds)); i++) {
-        Audio_StopSfxByBank(sSfxBankIds[i]);
+    for (bank = &sSfxBankIds[0]; bank < (sSfxBankIds + ARRAY_COUNT(sSfxBankIds)); bank++) {
+        Audio_StopSfxByBank(*bank);
     }
 }


src/code/code_800C3C20.c
```diff
@@ -5,9 +5,9 @@ u8 sSfxBankIds[] = {
 };
 
 void AudioMgr_StopAllSfx(void) {
-    u32 i;
+    volatile u8* bank;
```
hensldm, line 8: Note, I added volatile because it was needed for the similar function in MM. It is NOT needed here, so if it is preferred I can remove it.

src/code/code_800C3C20.c
```diff
@@ -5,9 +5,9 @@ u8 sSfxBankIds[] = {
 };
 
 void AudioMgr_StopAllSfx(void) {
-    u32 i;
+    volatile u8* bankIdPtr;
 
-    for (i = 0; (u32)(i < ARRAY_COUNT(sSfxBankIds)); i++) {
-        Audio_StopSfxByBank(sSfxBankIds[i]);
+    for (bankIdPtr = &sSfxBankIds[0]; bankIdPtr < (sSfxBankIds + ARRAY_COUNT(sSfxBankIds)); bankIdPtr++) {
```
engineer124, line 10:
```suggestion
    for (bankIdPtr = &sSfxBankIds[0]; bankIdPtr < (sSfxBankIds + sizeof(sSfxBankIds)); bankIdPtr++) {
```
Since this is pointer math, I think `sizeof` makes more sense. `ARRAY_COUNT` only matches because `sSfxBankIds` is size `u8`. If `sSfxBankIds` was switched to a `u32`, `ARRAY_COUNT` would break but `sizeof` would still match

hensldm: It should still work. Pointer arithmetic includes the size of the underlying element and adds accordingly. This is why pointer iterators work with just `++`.

hensldm: `sizeof` actually gives the wrong result when changing the type.
For example with sizeof at a type of `u16` I get that the end of the array would be at `sSfxBankIds+0x1c` which is 14 `u16` elements.
If I use ARRAY_COUNT I get `sSfxBankIds+0xe` which is the same 7 `u16` elements.

engineer124: Oh right yeah you're right

src/code/code_800C3C20.c
```diff
@@ -5,9 +5,9 @@ u8 sSfxBankIds[] = {
 };
 
 void AudioMgr_StopAllSfx(void) {
-    u32 i;
+    volatile u8* bankIdPtr;
```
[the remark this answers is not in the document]
fig02: it was needed for mm which suggests it will be needed for our retail as well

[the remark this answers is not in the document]
Dragorn421: Apparently having `bankIdPtr` as `volatile` or not has no effect on this implementation of this function (i.e. OoT's implementation, which is different from MM's, MM has an inner `if` https://decomp.me/scratch/fxTS7 )

Since this function is probably no different in other OoT versions, (i.e. previous versions since OoT MQ debug is one of the most "recents"), I think it won't ever need `volatile` and we can remove it at least for now (instead of carrying it forever :p)

Dragorn421 (approved):

Roman971 (approved):

fig02 (approved):

PR #1319, opened 2022-07-09
Second try

danielddyrseth:
Yes


fig02:
please see my comments on the previous pr, you cannot change the logic of the code because it breaks matching. That is not within the scope of this project.
Also, this code isnt valid.

danielddyrseth:
I understand.

PR #1322, opened 2022-07-11
BossDodongo_OverrideLimbDraw Better match?

hensldm:
Posted this in the Discord, but this seems like a better match to me, and also looks like this matches retail. 
Note case 7/8 are very flexible and can be basically anything above 6, as well as the if statement can be basically anything as long as it is complex enough that IDO doesn't optimize too much.

### DIFF
--- a/src/overlays/actors/ovl_Boss_Dodongo/z_boss_dodongo.c
+++ b/src/overlays/actors/ovl_Boss_Dodongo/z_boss_dodongo.c
@@ -1093,7 +1093,7 @@ s32 BossDodongo_OverrideLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Ve
 
         CLOSE_DISPS(play->state.gfxCtx, "../z_boss_dodongo.c", 3826);
     }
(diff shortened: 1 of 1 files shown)


src/overlays/actors/ovl_Boss_Dodongo/z_boss_dodongo.c
```diff
@@ -1088,7 +1093,6 @@ s32 BossDodongo_OverrideLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Ve
 
         CLOSE_DISPS(play->state.gfxCtx, "../z_boss_dodongo.c", 3826);
     }
-    { s32 pad; } // Required to match
     return 1;
```
EllipticEllipsis, line 1096:
```suggestion
    return true;
```

EllipticEllipsis (approved):
Works on retail too, at least the CPO.

fig02 (approved):

Roman971 (approved):

PR #1324, opened 2022-07-16
Fix audio strings with -use_readwrite_const and -signed

Roman971:
`-use_readwrite_const` is added to `code_800F7260.c` and `code_800F9280.c`, which makes all their strings end up in .data, and allows us to use strings literals in these files (see the old mess at the top of `code_800F7260.c`).
`-signed` is also added to `code_800EC960.c` because `AudioDebug_ScrPrt` uses signed chars for its string manipulations, and this flag tells IDO to treat chars as s8 instead of u8.

Note that I had to move the data previously in `code_800F7260.c` and `code_800F9280.c` to a new file, because the data needs to be after all the strings in these files. So it either has to be near the end of `code_800F9280.c`, or in a following file like this which seems more sensible.
This file is called `audio_external_data.c` atm but that's just a temporary name and I can change it if needed.


Makefile
```diff
@@ -199,6 +199,10 @@ build/src/code/fmodf.o: OPTFLAGS := -g
 build/src/code/__osMemset.o: OPTFLAGS := -g
 build/src/code/__osMemmove.o: OPTFLAGS := -g
 
+build/src/code/code_800EC960.o: CFLAGS += -signed
+build/src/code/code_800F7260.o: CFLAGS += -use_readwrite_const
+build/src/code/code_800F9280.o: CFLAGS += -use_readwrite_const
```
Dragorn421, line 204: Can you add comments on what these do, and why they're needed?
(basically what you wrote for the PR description)

Roman971: Done (assuming the wording is what you had in mind)

fig02 (approved):

PR #1328, opened 2022-07-19
Fix RaycastFloor flags

mzxrules:
Resolves a minor issue where afew RaycastFloor functions were tied to the wrong flags


src/code/z_bgcheck.c
```diff
@@ -24,6 +24,14 @@
 #define COLPOLY_IGNORE_ENTITY (1 << 1)
 #define COLPOLY_IGNORE_PROJECTILES (1 << 2)
 
+// floor raycast flags (frcFlags)
+#define BGCHECK_FRC_0 0
```
Roman971, line 28: We usually don't have a flag define for 0, and this one isn't even used so I think we should just omit it
```suggestion
```

src/code/z_bgcheck.c
```diff
@@ -24,6 +24,14 @@
...
+#define BGCHECK_FRC_1 (1 << 0)
+#define BGCHECK_FRC_2 (1 << 1)
+#define BGCHECK_FRC_4 (1 << 2)
+#define BGCHECK_FRC_8 (1 << 3)
+#define BGCHECK_FRC_10 (1 << 4)
```
Roman971, line 33: We've been using the bit for default flag names so I suggest doing the same here (ie. 1, 2, 3, 4, etc instead of 1, 2, 4, 8)

Dragorn421 (commented):
+1 on rcFlags -> raycastFlags
also if "floor raycast" means "downwards raycast" maybe "raycast down" would be better than "floor raycast" (afaik all raycasts are downwards but I'm not sure (`DynaRaycast` -> `DynaRaycastDown` eventually?))

fig02:
agree with the above. "raycast floor check wall" doesnt make much sense at all. if the first floor is meant to convey raycast down, then that sounds better

src/code/z_bgcheck.c
```diff
@@ -527,10 +560,10 @@ void StaticLookup_AddPoly(StaticLookup* lookup, CollisionContext* colCtx, Collis
  * Locates the closest static poly directly underneath `pos`, starting at list `ssList`
  * returns yIntersect of the closest poly, or `yIntersectMin`
  * stores the pointer of the closest poly to `outPoly`
- * if (flags & 1), ignore polys with a normal.y < 0 (from vertical walls to ceilings)
+ * if BGCHECK_GROUND_CHECK_ON, ignore polys with a normal.y < 0 (from vertical walls to ceilings)
```
Dragorn421, line 563:
```suggestion
 * if groundChk & BGCHECK_GROUND_CHECK_ON, ignore polys with a normal.y < 0 (from vertical walls to ceilings)
```

mzxrules: feels unnecessary since `BGCHECK_GROUND_CHECK_ON` on its own expresses itself as being a groundChk flag in the name

Dragorn421: Well the function has a gazillion arguments

fig02: I dont think you necessarily need to show the anding operation in the comment, but it can say something like "if `BGCHECK_GROUND_CHECK_ON` is passed" or "if `BGCHECK_GROUND_CHECK_ON` is set" to make it read more like a sentence 

but yea I do think it should be clear just by association of the name what its referring to.
But in that same vein, `groundChk` should just be `groundCheck` with the whole word, its only two characters and it will better match the define names

Dragorn421 (approved):

src/code/z_bgcheck.c
mzxrules: it isn't 2 characters, it's 20, and it would introduce an inconsistent variable names with some abbreviated some not. Unless perhaps you want to eliminate the `chk` abbr altogether which in that case it would be about 262 characters in bgcheck alone.

fig02: obviously i meant 2 characters per usage of the word... and I dont think the total amount of characters it adds is really relevant, when all that really matters is how long individual lines get

its already inconsistent with the enum name, so pick your posion. I think the abbreviating is not necessary and not helping to solve any particular problem, it would be fine as the whole word

fig02 (approved):
we can settle the check vs chk discussion later, bgcheck needs lots of work anyway

src/code/z_bgcheck.c
mzxrules: The bigger picture point for me and this PR is that `chk` is an established abbreviation in the code base at the moment, primarily in bgcheck, camera, Math3D, and inherited from event_chk_inf. If groundChk should just be groundCheck, then arguably all chks should be checks, but I'd rather avoid doing something like that in this PR since it falls outside of the scope of what I want to do with this PR. 

Either way, `groundChk` and `GROUND_CHECK` are equivalent to myself at least.

Roman971 (commented):
Just a few minor suggestions but otherwise it should be good to go

Roman971 (approved):

PR #1331, opened 2022-07-26
Rewrite format script in python to improve speed with multiprocessing

Roman971:
I essentially just moved the script to a python equivalent with multiprocessing (`-j`), along with the usual benefits that come with argparse and a proper language. In theory the multiprocessing could have been done in bash (utilizing something like GNU `parallel` or `xargs`), but it also required a much more involved process for `clang-tidy` so I figured it was easier to deal with it in python.

About running `clang-tidy` in parallel, the code has a comment to explain this, but basically the issue is it can make changes in files #include'd so it's not safe to run in parallel. The solution used here is to run it with `--export-fixes` so it doesn't apply fixes on its own at first, then we use `clang-apply-replacements` to apply the exported fixes at once. It's a bit annoying to do since we need to manage a temp directory and all, but it works out in the end. This also means `clang-apply-replacements` is required when running the script with `-j` to have parallelism, but it's the only option we have afaik.

I also included `clang-tidy-11` and `clang-apply-replacements-11` as possible options with these changes for ease of use, but binaries without a version are still allowed ofc. Otherwise the script should behave mostly the same as it used to in the bash version.

To give an idea of the improvements, here is how it performs with `-j16` on my machine:
```
$ time ./format.py -j16
Formatting files with 16 jobs
Running clang-format...
Running clang-tidy...
Adding missing final new lines...
Done formatting files.

real    0m7.305s
user    1m25.608s
sys     0m5.268s
```
vs the existing script:
```
$ time ./format.sh
Formatting C files. This will take a bit
Running clang-format...
Running clang-tidy...
Adding missing final new lines...
Done formatting all files.

real    0m46.886s
user    0m37.545s
sys     0m1.548s
```


fig02 (approved):

hensldm:
I've been using this in MM to test it out, I really appreciate the speedup. However, there are times when I wish I could just run `clang-format` without `clang-tidy`. Thoughts on adding an optional argument to just run `clang-format` without `clang-tidy`?

Roman971:
Could you explain the use case for running the script with only `clang-format`?
I feel like this script should always run everything since it's intended to be used once in a while, usually before comitting changes, and it's pretty fast regardless now (when using `-j`).

If the idea is to actively use it while making changes, I recommend integrating clang-format in your IDE (eg. with the vscode extension), which you can even automate when saving a file for example.

hensldm:
Oh right, I forgot about the integration method. That would be better for my use case. Nevermind on the additional arg then.

Dragorn421 (approved):
Thanks a lot for this!

(I can shave one more second off by splitting the source files into `2*nb_jobs` chunks instead of `nb_jobs` chunks but it's not super significant and probably computer-specific so whatever :sweat_smile: )

Some notes that don't matter:
- `'` and `"` is the same in python, using `'` for single characters and `"` for strings is a bad habit from C :smiling_imp: 
- I would pass the arguments to subprocess as lists instead of single strings with space-separated arguments but whatever
- silencing stdout/stderr would be nice to have an option to not do that to facilitate debugging if ever needed

PR #1332, opened 2022-07-28
Improvements to Video Interface related functions and data

Thar0:
Largely concerns the VI functions in libultra and `z_vimode.c`.
Also some improvements to `devmgr.c` since I replaced `OSMgrArgs` structure with the original `OSDevMgr`.

I couldn't think of short enough names for the last two `ViModeEditState`s, suggestions welcome.


fig02 (approved):
the actual content of this code is way over my head so I cant weigh in on any names, but it looks fine to me from what I can tell

include/z64.h
```diff
@@ -1764,20 +1764,27 @@ typedef struct {
     /* 0x10 */ s16 unk_10;
 } JpegDecoderState; // size = 0x14
 
+typedef enum {
+    VI_MODE_EDIT_STATE_INACTIVE,
+    VI_MODE_EDIT_STATE_ACTIVE,
+    VI_MODE_EDIT_STATE_2, // active, more adjustments
+    VI_MODE_EDIT_STATE_3  // active, more adjustments, print comparison with NTSC LAN1 mode
+} ViModeEditState;
```
engineer124, line 1772:
```suggestion
typedef enum {
    /* 0 */ VI_MODE_EDIT_STATE_INACTIVE,
    /* 1 */ VI_MODE_EDIT_STATE_ACTIVE,
    /* 2 */ VI_MODE_EDIT_STATE_2, // active, more adjustments
    /* 3 */ VI_MODE_EDIT_STATE_3  // active, more adjustments, print comparison with NTSC LAN1 mode
} ViModeEditState;
```
Just a small cleanup suggestion

EllipticEllipsis (commented):
Nice, this one's been in the back of my mind for a while so it's nice to finally see what it does.

src/code/game.c
```diff
@@ -243,14 +243,14 @@ void GameState_Update(GameState* gameState) {
     func_800C4344(gameState);
 
     if (SREG(63) == 1u) {
```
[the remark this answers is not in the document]
fig02: i was able to remove one of the U's from game.c in something else im working on. they should all be double checked

src/code/z_vimode.c
```diff
@@ -171,33 +203,33 @@ void ViMode_Save(ViMode* viMode) {
...
 
-    viMode->viModeBase = SREG(48);
-    viMode->viWidth = SREG(49) & ~3;
-    viMode->viHeight = SREG(50);
-    viMode->unk_64 = SREG(51);
-    viMode->unk_60 = SREG(52);
-    viMode->unk_5C = SREG(53);
-    viMode->unk_58 = SREG(54);
+    viMode->editState = R_VI_MODE_STATE;
+    viMode->viWidth = R_VI_MODE_WIDTH & ~3;
```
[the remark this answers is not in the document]
Thar0: Yeah this seems like something of a bug. I'm not entirely sure why the width is adjusted in the way it is, there's no documentation stating the video width is restricted to particular multiples. I can make a comment but I'm unsure if I should mark it with a bug tag?

src/code/z_vimode.c
```diff
@@ -22,137 +23,168 @@ void ViMode_LogPrint(OSViMode* osViMode) {
...
+ * @param type          Identifying type for the OSViMode
+ * @param tvType        TV Type: NTSC, PAL, MPAL or FPAL
+ * @param loRes         Boolean: true = low resolution, false = high resolution.
+ *                      Corresponds to "L" or "H" in libultra VI mode names
+ * @param antialiasOff  Boolean: true = point-sampling, false = anti-aliasing.
+ *                      Corresponds to "P" or "A" in libultra VI mode names
+ * @param modeN         Boolean: controls interlacing mode, different based on resolution.
+ *                      Corresponds to "N" or "F" in libultra VI mode names
+ * @param fb16Bit       Bolean: true = 16-bit framebuffer, false = 32-bit framebuffer.
+ *                      Corresponds to "1" or "2" in libultra VI mode names
```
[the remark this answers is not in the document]
Thar0: This is mentioned in the N64 programming manual section 30.1.3

src/code/game.c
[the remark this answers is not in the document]
Thar0: I tried to remove them from this function however I was unsuccessful

src/code/z_vimode.c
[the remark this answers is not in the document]
EllipticEllipsis: I'd say it is a bug, although it's annoying that we can't tell what it's meant to do.

Dragorn421 (commented):
Partial review

Are the VI regs defines like `#define HSYNC` original work or where do they come from?

Thar0:
The VI reg defines are from libreultra & ultralib, not sure if they existed anywhere prior to libreultra

src/libultra/io/visetspecial.c
```diff
@@ -1,38 +1,55 @@
...
+/**
+ * Configures VI "special features" to be applied on the next context swap.
+ *
+ * "Special features" refer to the mode bits in the Video Interface control register that enable effects such as gamma
+ * correction, gamma dither, dither filtering, anti-aliasing filtering and divot filtering. Configuring the same
+ * setting ON and OFF in the same call will result in OFF taking precedence.
+ *
+ * Any unrecognized bits will be ignored. Note that this is very intentional as in early revisions of retail N64
+ * hardware setting bit 5 in the `features` field of OSViContext may cause physical damage to the console once it is
+ * fed to VI_CONTROL_REG on next context swap.
```
EllipticEllipsis, line 13: You might as well refer to the place in the docs that mentions this here, I think.

EllipticEllipsis (commented):
I think you may have missed the variable in idle.c that seems to be the vimode type (`D_80013960`) (also tells us what arg3 of Sched_Init is, maybe?)

(Also, for some reason on N64 that variable is data instead of bss.)

src/libultra/io/visetspecial.c
Thar0: There is no manual page that discusses this afaik

fig02 (approved):

EllipticEllipsis (approved):

src/libultra/io/vimodepallan1.c
```diff
@@ -1,32 +1,46 @@
...
     },
     { // fldRegs
       {
-          0x280,    // origin
-          0x400,    // yScale
-          0x5F0239, // vStart
-          0x9026B,  // vBurst
-          2,        // vIntr
+          // [0]
+          ORIGIN(640),         // origin
```
Dragorn421, line 32: based on usage in `__osViSwapContext` this isn't directly the origin reg value (thankfully), merely an offset (which I have no idea what its purpose is), so it probably shouldn't use `ORIGIN()`

also inspired from `ViMode_Configure`, 640 is 320\*2 (width\*2) ?

src/libultra/io/viswapcontext.c
```diff
@@ -1,48 +1,49 @@
...
     }
-    if (viNext->state & 0x80) {
-        viNext->y.scale = (viNext->y.offset << 0x10) & 0x3FF0000;
-        origin = osVirtualToPhysical(viNext->buffer);
+    if (viNext->state & VI_STATE_FADE) {
+        viNext->y.scale = (viNext->y.offset << VI_SUBPIXEL_SH) & (VI_2_10_FPART_MASK << VI_SUBPIXEL_SH);
+        origin = osVirtualToPhysical(viNext->framep);
     }
+
     HW_REG(VI_ORIGIN_REG, u32) = origin;
```
Dragorn421, line 47: `ORIGIN(origin)` here or where origin is set above?

src/libultra/io/devmgr.c
```diff
@@ -1,97 +1,100 @@
...
+            info = &ioMesg->piHandle->transferInfo;
+            blockInfo = &info->block[info->blockNum];
+
+            info->sectorNum = -1;
+            if (info->transferMode != 3) {
+                blockInfo->dramAddr = (void*)((u32)blockInfo->dramAddr - blockInfo->sectorSize);
             }
 
-            phi_s2 = ((transfer->transferMode == 2) && (ioMesg->piHandle->transferInfo.cmdType == 0)) ? 1 : 0;
+            messageSend = ((info->transferMode == 2) && (ioMesg->piHandle->transferInfo.cmdType == 0)) ? true : false;
```
Dragorn421, line 28: honestly `messageSend` as a boolean is more confusing to me than a 0/1 or is there something I'm missing

include/ultra64/rcp.h
```diff
@@ -39,6 +39,24 @@
...
+#define VI_CTRL_DIVOT_ON            0x00010 // 4: default = on
+#define VI_CTRL_SERRATE_ON          0x00040 // 6: on if interlaced
+#define VI_CTRL_ANTIALIAS_MASK      0x00300 // [9:8] anti-alias mode
+#define VI_CTRL_ANTIALIAS_MODE_0    0x00000 // Bit [9:8] anti-alias mode: AA enabled, resampling enabled, always fetch extra lines
+#define VI_CTRL_ANTIALIAS_MODE_1    0x00100 // Bit [9:8] anti-alias mode: AA enabled, resampling enabled, fetch extra lines as-needed
+#define VI_CTRL_ANTIALIAS_MODE_2    0x00200 // Bit [9:8] anti-alias mode: AA disabled, resampling enabled, operate as if everything is covered
+#define VI_CTRL_ANTIALIAS_MODE_3    0x00300 // Bit [9:8] anti-alias mode: AA disabled, resampling disabled, replicate pixels
+#define VI_CTRL_PIXEL_ADV_MASK      0x01000 // [15:12] pixel advance mode
+#define VI_CTRL_PIXEL_ADV_1         0x01000 // Bit [15:12] pixel advance mode: ?
+#define VI_CTRL_PIXEL_ADV_2         0x02000 // Bit [15:12] pixel advance mode: ?
```
Dragorn421, line 56: n64brew only mentions 0b11 and the code only mentions 0b11 so probably these two can go?

include/ultra64/rcp.h
```diff
@@ -39,6 +39,24 @@
...
+#define VI_CTRL_PIXEL_ADV_3         0x03000 // Bit [15:12] pixel advance mode: ?
+#define VI_CTRL_DITHER_FILTER_ON    0x10000 // 16: dither-filter mode
```
Dragorn421, line 58: according to current n64brew docs this is "dedither" not "dither" :shrug:

include/ultra64/rcp.h
```diff
@@ -136,4 +154,21 @@
...
+/**
+ * External device info
+ */
+#define DEVICE_TYPE_CART    0   /* ROM cartridge */
+#define DEVICE_TYPE_BULK    1   /* ROM bulk */
+#define DEVICE_TYPE_64DD    2   /* 64 Disk Drive */
+#define DEVICE_TYPE_SRAM    3   /* SRAM */
+/* 4-6 are reserved */
+#define DEVICE_TYPE_INIT    7   /* initial value */
+/* 8-14 are reserved */
```
Dragorn421, line 172: Only thing I found on this:

> rcp.h
> Changed 8-14 for DEVICE_TYPE to reserve. 

(in "Release Notes OS2.0K" `n64man/relnote_k/index.htm`)

do you know more lol

src/libultra/io/visetspecial.c
```diff
@@ -1,38 +1,55 @@
...
-        __osViNext->features &= ~OS_VI_DIVOT;
+        __osViNext->features &= ~VI_CTRL_DIVOT_ON;
     }
+
     if (func & OS_VI_DITHER_FILTER_ON) {
-        __osViNext->features |= OS_VI_DITHER_FILTER;
-        __osViNext->features &= ~(OS_VI_UNK200 | OS_VI_UNK100);
+        // If the dither filter is enabled, also force anti-alias mode 0 irrespective of the current OSViMode.
+        __osViNext->features |= VI_CTRL_DITHER_FILTER_ON;
+        __osViNext->features &= ~VI_CTRL_ANTIALIAS_MASK;
```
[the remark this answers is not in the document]
EllipticEllipsis: But is that actually an improvement? We know that frequently they set flags one at a time, this just makes it harder to read what's going on.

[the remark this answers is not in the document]
Dragorn421: I'm not sure what you mean
like,
```c
        __osViNext->features &= ~VI_CTRL_ANTIALIAS_MASK;
        __osViNext->features |= VI_CTRL_ANTIALIAS_MODE_0;
```
doesn't match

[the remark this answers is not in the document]
EllipticEllipsis: I obviously don't understand what you're intending with this, then.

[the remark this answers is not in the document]
Dragorn421: Well the goal of this line is to set the "antialias mode" to `VI_CTRL_ANTIALIAS_MODE_0`, so for searchability/shiftability purposes (even though this relies on N64 hardware so it's not like there's a point in practice) it would be better to reference that name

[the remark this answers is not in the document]
EllipticEllipsis: Hmm, seems like this doesn't actually make sense anyway since it's using it like an enum from the looks of it: reading the description, 3 is not actually 1 | 2 even though it looks like it?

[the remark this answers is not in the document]
Dragorn421: Yeah VI_CTRL_ANTIALIAS_MODE_0/1/2/3 is an enum indeed

src/libultra/io/vimodepallan1.c
Thar0: `ORIGIN()` is for `OSViFieldRegs::origin`, not `VI_ORIGIN_REG`

If we wanted we could make `ORIGIN` be a macro in terms of width, bitdepth and target resolution:
`#define ORIGIN(width, fb16Bit, loRes) ((width) * 2 * ((fb16Bit) ? 1 : 2) * ((loRes) ? 1 : 2)))`
If it were just in terms of `width` it would not work for other VI modes that are not included in oot, breaking compatibility. I'm not sure if it's worth the added complexity

src/libultra/io/devmgr.c
Thar0: I can see that it can be a bit confusing since it was written weirdly, (`messageSend != true`) but I don't think that disqualifies it from being a boolean?

include/ultra64/rcp.h
Thar0: These other modes do exist to some capacity, afaik, so I think it's better to maintain a complete set of defines.
Also notably, the iQue player needs different pixel advance settings or the output is corrupted

include/ultra64/rcp.h
Thar0: Right, what it means by "dither filter" is a filter that smooths out dithering applied by earlier stages rather than a filter that performs dithering. Unfortunately, that's just how SGI named it.

include/ultra64/rcp.h
Thar0: The comments are just lifted straight from OS2.0L rcp.h

src/libultra/io/devmgr.c
Dragorn421: well usually we still boolify `== true`-using stuff when it's otherwise mostly used as a bool, here the only use is `== true`, but whatever

My comment is more that, what does "messageSend == true" mean? the message will be / has been sent ? idk it just looks like a "exit loop" flag to me and for some reason using true/false makes it harder to read

but if it's just me whatever it's not like this is important

include/ultra64/rcp.h
Dragorn421: why only the two low bits then?

src/libultra/io/vimodepallan1.c
Dragorn421: It would be good to comment on the macros what their purpose is then (if not use prefixes), since "origin" is also a vi reg and the macros look just like the vi reg macros family WIDTH, HSYNC etc

include/ultra64/rcp.h
Dragorn421: I forgot/didn't notice some of these are original names okay

include/ultra64/rcp.h
Dragorn421: so rcp.htm in the manual is out-of-date and I guess others too
I'll flee those in the future

include/ultra64/rcp.h
Thar0: Because I copied them from ultralib and never noticed it wasn't complete 😎 
I've changed it to one macro with an argument to generate all values.

src/libultra/io/visetspecial.c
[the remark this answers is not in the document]
Thar0: I'm reluctant to do this since it's very possible this won't match other libultra versions. (I'm also concerned about matching ultralib, particularly debug builds)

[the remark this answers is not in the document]
Dragorn421: Ok
`// VI_CTRL_ANTIALIAS_MODE_0` ? (don't have to)

src/code/z_vimode.c
```diff
@@ -1,158 +1,190 @@
...
+    s32 hiRes;
+    s32 antialiasOn;
+    s32 modeF;
+    s32 fb32Bit;
+    s32 hiResDeflicker; // deflickered interlacing
+    s32 hiResInterlaced;
+    s32 loResDeinterlaced;
+    s32 loResInterlaced;
+    s32 modeLAN1; // L=(lo res) A=(antialias)     N=(deinterlace)        1=(16-bit)
+    s32 modeLPN2; // L=(lo res) P=(point-sampled) N=(normal interlacing) 2=(32-bit)
```
Dragorn421, line 58: "N=(normal interlacing)" contradicts `loResDeinterlaced = loRes && modeN` below?

Dragorn421 (approved):
Good to merge!

src/code/z_vimode.c
```diff
@@ -1,158 +1,190 @@
...
+        // Anti-aliased, fetch extra lines as-needed
+        viMode->customViMode.comRegs.ctrl |= VI_CTRL_ANTIALIAS_MODE_1;
+    } else if (modeLPN2 | modeHPN2) {
+        // Point-sampled, resampling disabled
+        viMode->customViMode.comRegs.ctrl |= VI_CTRL_ANTIALIAS_MODE_3;
     } else {
-        viMode->customViMode.comRegs.ctrl |= 0;
+        if (antialiasOff) {
+            // Point-sampled, resampling enabled
+            viMode->customViMode.comRegs.ctrl |= VI_CTRL_ANTIALIAS_MODE_2;
```
AngheloAlf, line 102: Any reason to not merge this into an `else if` like it was?

Thar0: The conditions being checked are of a different character (full mode vs only AA on/off), I think it's a bit clearer to see what's going on this way

PR #1334, opened 2022-07-30
Backport some Z_DLF changes from MM

louist103:
Bring over the macros and intptr_t used in MM. OOT still needs to keep this structure since this file was likely refactored in MM.

### DIFF
--- a/include/macros.h
+++ b/include/macros.h
@@ -199,4 +199,10 @@ extern GraphicsContext* __gfxCtx;
                        ((height)-1) << G_TEXTURE_IMAGE_FRAC);                                                          \
     } while (0)
 
+#define OVERLAY_RELOCATION_OFFSET(overlayEntry) \
+    ((uintptr_t)((overlayEntry)->vramStart) - (uintptr_t)((overlayEntry)->loadedRamAddr))
+#define VRAM_PTR_SIZE(entry) ((uintptr_t)((entry)->vramEnd) - (uintptr_t)((entry)->vramStart))
+
+#define RELOCATE_ADDR(addr, vRamStart, allocu32) ((addr) - (vRamStart) + (allocu32))
+
 #endif
--- a/src/code/z_DLF.c
+++ b/src/code/z_DLF.c
@@ -20,42 +20,41 @@ void Overlay_LoadGameState(GameStateOverlay* overlayEntry) {
 
         osSyncPrintf(VT_FGCOL(GREEN));
         osSyncPrintf("OVL(d):Seg:%08x-%08x Ram:%08x-%08x Off:%08x %s\n", overlayEntry->vramStart, overlayEntry->vramEnd,
-                     overlayEntry->loadedRamAddr,
-                     (u32)overlayEntry->loadedRamAddr + (u32)overlayEntry->vramEnd - (u32)overlayEntry->vramStart,
-                     (u32)overlayEntry->vramStart - (u32)overlayEntry->loadedRamAddr, "");
+                     overlayEntry->loadedRamAddr, (uintptr_t)overlayEntry->loadedRamAddr + VRAM_PTR_SIZE(overlayEntry),
+                     OVERLAY_RELOCATION_OFFSET(overlayEntry), "");
         osSyncPrintf(VT_RST);
 
         if (overlayEntry->unk_14 != NULL) {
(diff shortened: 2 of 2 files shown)


include/macros.h
```diff
@@ -199,4 +199,10 @@ extern GraphicsContext* __gfxCtx;
                        ((height)-1) << G_TEXTURE_IMAGE_FRAC);                                                          \
     } while (0)
 
+#define OVERLAY_RELOCATION_OFFSET(overlayEntry) \
+    ((uintptr_t)((overlayEntry)->vramStart) - (uintptr_t)((overlayEntry)->loadedRamAddr))
+#define VRAM_PTR_SIZE(entry) ((uintptr_t)((entry)->vramEnd) - (uintptr_t)((entry)->vramStart))
+
+#define RELOCATE_ADDR(addr, vRamStart, allocu32) ((addr) - (vRamStart) + (allocu32))
```
fig02, line 206: are these macros needed? I feel like its just hiding implementation details in a function where you would want to see these details up front

Roman971: Here is my opinion after reading the discussion in discord:
In general I agree with the others mentioning how these macros hide details in an already internal function. More specifically, I think `OVERLAY_RELOCATION_OFFSET` could just be replaced with a comment like Dragorn suggested, `VRAM_PTR_SIZE` feels unnecessary since doing `end - start` with these casts is very straightforward, and `RELOCATE_ADDR` isn't used so we probably don't need it (at least in oot)

fig02: @hensldm wanted to try out using the same macro for both cases, instead of having two seperate ones. if that works out id probably be ok with a compromise to just have the one macro.
But overall I do still feel that its hiding stuff that we dont need to hide

hensldm: Oh right, thanks for the reminder.
So it doesn't look like I can combine them into 1 macro, loadfragment in MM doesn't like the `intptr_t` cast, but it seems necessary in `DLF` and the other common usages. 
However I can make `RELOCATE_ADDR` usable in both places (again it doesn't apply here in relocation.c yet as the function got refactored around MM pal time, and so the macro doesn't really work, though I do think it is very much an improvement in the old version). I've also added an `UNRELOCATE_ADDR` for the second function, though not particularly fond of the name.
https://github.com/zeldaret/oot/compare/master...hensldm:oot:dlf

Dragorn421:
These macros are barely used in OoT, I'm sure they benefit MM (which got refactored code afaik?) but it just doesn't seem relevant in OoT. Of the three macros added, one is unused and another is used exactly once.
The third macro added is imo not worth the slight abstraction.

Can the macros get more use in OoT or is this their full potential?

Dragorn421:
I'm considering if this PR should be closed, as concern has been voiced about the usefulness of these macros, and hasn't met any reassuring arguments.
I'm also under the impression the changes are partial, as further changes were mentioned but never came, and it looks like the macros may be used more extensively (using MM as reference). Given it has been months without updates, I also consider this PR inactive.

TLDR: Close this PR? because: changes are questionable / changes are partial / PR is inactive

--------------

Furthermore, personally, I went to look at MM's usage of these macros to get an idea anyway of the "fuller potential" I referred to and it didn't convince me

I find it questionable to use the same macro on different types
(see: `OVERLAY_RELOCATION_OFFSET` and `VRAM_PTR_SIZE`, this isn't suggested in this PR but since the goal is MM consistency I assume it would be argued for eventually?),

as well as macroifying some operations that are
1) (as mentioned before) very simple (just a subtraction) and
2) done elsewhere in other forms (especially `VRAM_PTR_SIZE`'s "end - start", e.g. in `Load2_LoadOverlay`)

TLDR: more questioning of the macros

hensldm:
I've been mulling this over for the past couple days.

Generally I am a big fan of macros, but as pointed out the macros are actually used for different types (I did not notice until now) which does seem questionable to me as well.

As for `RELOCATE_ADDR`, which I liked at first, I've been having second thoughts, mainly because of the refactor to the function in question that makes it most valuable (note N64 OoT should have MM's version so this might come up again). I was able to make a more general version that could be used elsewhere (though still not in the one function), but I don't really know. It is just a simple subtraction so perhaps just not worth it (though I do think it helps make clearer what is actually getting relocated to RAM).

Dragorn421:
Closing (because: see above)

PR #1336, opened 2022-07-30
More documentation for padmgr.c

Thar0:
Part 2 of controller related documentation.

Also fixes some things in `fault.h` and `irqmgr.h`.


src/code/padmgr.c
```diff
@@ -1,20 +1,73 @@
...
+ * Libultra osCont* routines        : Interface for building commands and safely using the Serial Interface
+ *  |
+ * Serial Interface                 : Hardware unit for sending joybus commands and receiving data via DMA
+ *  |
+ * PIF                              : Forwards joybus commands and receives response data from the devices
+ *  |---¬---¬---¬-------¬
+ *  1   2   3   4       5           : The joybus devices plugged into the four controller ports or on the cartridge
+ *
+ * Joybus communication is handled on another thread as polling and receiving controller data is a slow process; the
+ * N64 programming manual quotes 2 milliseconds as the expected value from calling `osContStartReadData` to receiving
```
Dragorn421, line 27:
```suggestion
 * N64 programming manual quotes 2 milliseconds as the expected delay from calling `osContStartReadData` to receiving
```
?

also could you quote a specific part of the manual? (where)

src/code/padmgr.c
```diff
@@ -1,20 +1,73 @@
...
+ * it becomes ambiguous as to which DMA has completed, so a locking system is required to arbitrate access to the SI.
+ *
+ * Once the task requiring the serial event queue is complete, it should be released with a call to
+ * `PadMgr_ReleaseSerialEventQueue`.
+ *
+ * If another process tries to acquire the event queue, the current thread will be blocked until the event queue is
+ * released. Note the possibility for a double lock, if the thread that already holds the serial event queue attempts
+ * to acquire it again it will block forever.
+ *
+ * @return The message queue to which SI interrupt events are posted to.
```
Dragorn421, line 55:
```suggestion
 * @return The message queue to which SI interrupt events are posted.
```
one to too many (monkahm)

src/code/padmgr.c
```diff
@@ -316,35 +439,43 @@ void PadMgr_HandlePreNMI(PadMgr* padMgr) {
...
 
-void PadMgr_RequestPadData(PadMgr* padMgr, Input* inputs, s32 mode) {
+/**
+ * Fetches the most recently polled inputs from padmgr
+ *
+ * @param padMgr        PadMgr instance
+ * @param[out] inputs   Array of 4 Input to copy inputs into
+ * @param gamePoll      True if polling inputs for the game state at a lower speed than once per VI
+ */
+void PadMgr_RequestPadData(PadMgr* padMgr, Input inputs[MAXCONTROLLERS], s32 gameRequest) {
```
Dragorn421, line 449: gamePoll / gameRequest

also I didn't expect that explanation so I may be back!

Dragorn421 (commented):
few non-important things

src/code/padmgr.c
Thar0: This is definitely the most mysterious part for me as well so I hope you will give it some extra thought at some point 😄

src/code/fault.c
```diff
@@ -352,18 +352,18 @@ void Fault_Sleep(u32 msec) {
...
 
-void PadMgr_RequestPadData(Input* input, s32 mode);
+void PadMgr_RequestPadData(Input inputs[MAXCONTROLLERS], s32 gameRequest);
 
-void Fault_PadCallback(Input* input) {
+void Fault_PadCallback(Input inputs[MAXCONTROLLERS]) {
     //! @bug This function is not called correctly, it is missing a leading PadMgr* argument. This
     //! renders the crash screen unusable.
     //! In Majora's Mask, PadMgr functions were changed to not require this argument, and this was
     //! likely just not addressed when backporting.
```
EllipticEllipsis, line 361: This should probably be AVOID_UB'd (along with the prototype).

src/code/mempak.c
```diff
@@ -40,7 +40,7 @@ s32 Mempak_FindFile(s32 controllerNb, char start, char end) {
     u32 bit = 1;
     s32 flag = 0;
 
-    mq = PadMgr_LockSerialMesgQueue(&gPadMgr);
+    mq = PadMgr_AcquireSerialEventQueue(&gPadMgr);
 
     for (idx = start; idx <= end; idx++) {
         sMempakExtName[0] = idx - 0x27;
```
EllipticEllipsis, line 46: Would it make sense to use a char literal here too?

src/code/padmgr.c
```diff
@@ -1,20 +1,73 @@
...
+ *
+ * When a DMA to/from PIF RAM completes, an SI interrupt is generated to notify the process that the DMA has completed
+ * and a message is posted to the serial event queue. If multiple processes are trying to use the SI at the same time
+ * it becomes ambiguous as to which DMA has completed, so a locking system is required to arbitrate access to the SI.
+ *
+ * Once the task requiring the serial event queue is complete, it should be released with a call to
+ * `PadMgr_ReleaseSerialEventQueue`.
+ *
+ * If another process tries to acquire the event queue, the current thread will be blocked until the event queue is
+ * released. Note the possibility for a double lock, if the thread that already holds the serial event queue attempts
```
EllipticEllipsis, line 52: Could use `@note` for this note to make it a bit more prominent.

src/code/padmgr.c
```diff
@@ -1,20 +1,73 @@
...
+
+/**
+ * Acquires exclusive access to the serial event queue.
+ *
+ * When a DMA to/from PIF RAM completes, an SI interrupt is generated to notify the process that the DMA has completed
+ * and a message is posted to the serial event queue. If multiple processes are trying to use the SI at the same time
+ * it becomes ambiguous as to which DMA has completed, so a locking system is required to arbitrate access to the SI.
+ *
+ * Once the task requiring the serial event queue is complete, it should be released with a call to
+ * `PadMgr_ReleaseSerialEventQueue`.
```
EllipticEllipsis, line 49:
```suggestion
 * `PadMgr_ReleaseSerialEventQueue()`.
```
and it should make the link to the function automatically.

src/code/padmgr.c
```diff
@@ -23,69 +76,94 @@ OSMesgQueue* PadMgr_LockSerialMesgQueue(PadMgr* padMgr) {
...
+            if (padMgr->padStatus[i].status & CONT_CARD_ON) {
+                if (padMgr->pakType[i] == one) { // CONT_PAK_RUMBLE
+                    if (padMgr->rumbleEnable[i]) {
+                        if (padMgr->rumbleTimer[i] < 3) {
                             // clang-format off
                             if (1) {} osSyncPrintf(VT_FGCOL(YELLOW));
                             // clang-format on
 
-                            // "Vibration pack jumble jumble"?
+                            // "Rumble pack brrr"
```
EllipticEllipsis, line 148: Looks like this also has info on which controller is rumbling, `コン` being short for "controller"
```suggestion
                            // "Controller `i+1` Rumble pack brrr"
```
or similar.

src/code/padmgr.c
```diff
@@ -23,69 +76,94 @@ OSMesgQueue* PadMgr_LockSerialMesgQueue(PadMgr* padMgr) {
...
                             osSyncPrintf("padmgr: %dコン: %s\n", i + 1, "振動パック ぶるぶるぶるぶる");
                             osSyncPrintf(VT_RST);
 
-                            if (__osMotorAccess(&padMgr->pfs[i], temp) != 0) {
-                                padMgr->pakType[i] = 0;
+                            // This should be the osMotorStart macro
```
EllipticEllipsis, line 152: Might be worth saying why it cannot be. Indeed, maybe name the variable in a way that better reflects how it's used in place of the enum.

src/code/padmgr.c
```diff
@@ -23,69 +76,94 @@ OSMesgQueue* PadMgr_LockSerialMesgQueue(PadMgr* padMgr) {
...
+                            if (__osMotorAccess(&padMgr->rumblePfs[i], one) != 0) {
+                                padMgr->pakType[i] = CONT_PAK_NONE;
                                 osSyncPrintf(VT_FGCOL(YELLOW));
                                 // "A communication error has occurred with the vibration pack"
```
EllipticEllipsis, line 156: This also mentions the controller number.

src/code/padmgr.c
```diff
@@ -94,164 +172,191 @@ void PadMgr_RumbleControl(PadMgr* padMgr) {
...
                 if (padMgr->ctrlrIsConnected[i]) {
+                    // If we get no response, consider the controller disconnected
                     padMgr->ctrlrIsConnected[i] = false;
-                    padMgr->pakType[i] = 0;
-                    padMgr->rumbleCounter[i] = 0xFF;
+                    padMgr->pakType[i] = CONT_PAK_NONE;
+                    padMgr->rumbleTimer[i] = 0xFF;
                     osSyncPrintf(VT_FGCOL(YELLOW));
                     // "Do not respond"?
                     osSyncPrintf("padmgr: %dコン: %s\n", i + 1, "応答しません");
```
EllipticEllipsis, line 348: Looks more like "not responding", there's some examples of usage of this phrase here: https://context.reverso.net/%E7%BF%BB%E8%A8%B3/%E6%97%A5%E6%9C%AC%E8%AA%9E-%E8%8B%B1%E8%AA%9E/%E5%BF%9C%E7%AD%94%E3%81%97%E3%81%BE%E3%81%9B%E3%82%93

src/code/padmgr.c
```diff
@@ -94,164 +172,191 @@ void PadMgr_RumbleControl(PadMgr* padMgr) {
...
                 input->cur.stick_y = 0;
-                input->cur.errno = padnow1->errno;
+                input->cur.errno = pad->errno;
                 if (padMgr->ctrlrIsConnected[i]) {
+                    // If we get no response, consider the controller disconnected
                     padMgr->ctrlrIsConnected[i] = false;
-                    padMgr->pakType[i] = 0;
-                    padMgr->rumbleCounter[i] = 0xFF;
+                    padMgr->pakType[i] = CONT_PAK_NONE;
+                    padMgr->rumbleTimer[i] = 0xFF;
```
EllipticEllipsis, line 345: 255, or is FF treated specially?

src/code/padmgr.c
```diff
@@ -94,164 +172,191 @@ void PadMgr_RumbleControl(PadMgr* padMgr) {
...
                 }
                 break;
-            case 4:
+            case (CHNL_ERR_OVERRUN >> 4):
+                // Overrun error, reuse previous inputs
                 input->cur = input->prev;
                 LOG_NUM("this->Key_switch[i]", padMgr->ctrlrIsConnected[i], "../padmgr.c", 380);
                 osSyncPrintf(VT_FGCOL(YELLOW));
                 // "Overrun error occurred"
                 osSyncPrintf("padmgr: %dコン: %s\n", i + 1, "オーバーランエラーが発生");
```
EllipticEllipsis, line 332: This patters is so frequent in this file I suspect it's a macro. Otherwise there's not a lot of reason to separate off the error message itself.

EllipticEllipsis (commented):
Nice job, few small things.

src/code/mempak.c
Thar0: It does make sense to use a char literal. I have follow-up documentation for mempak ready which fixes this once this PR is in

fig02 (approved):
definitely read this over at some point and forgot to hit the green button

src/code/padmgr.c
```diff
@@ -23,235 +84,260 @@ OSMesgQueue* PadMgr_LockSerialMesgQueue(PadMgr* padMgr) {
...
 void PadMgr_RumbleReset(PadMgr* padMgr) {
-    padMgr->rumbleOffFrames = 3;
+    padMgr->rumbleOffTimer = 3;
 }
 
-void PadMgr_RumbleSetSingle(PadMgr* padMgr, u32 ctrlr, u32 rumble) {
-    padMgr->rumbleEnable[ctrlr] = rumble;
-    padMgr->rumbleOnFrames = 240;
+/**
+ * Enables or disables rumble on controller port `port` for 240 VI,
```
Dragorn421, line 263: well technically the timer is global and the same for all rumble paks? but I guess the details are largely irrelevant

src/code/padmgr.c
```diff
@@ -23,235 +84,260 @@ OSMesgQueue* PadMgr_LockSerialMesgQueue(PadMgr* padMgr) {
...
+    padMgr->rumbleEnable[port] = rumble;
+    padMgr->rumbleOnTimer = 240;
 }
 
-void PadMgr_RumbleSet(PadMgr* padMgr, u8* ctrlrRumbles) {
+/**
+ * Enables or disables rumble on all controller ports for 240 VI,
+ * ~4 seconds at 60 VI/sec and ~4.8 seconds at 50 VI/sec
+ *
+ * @param enable    Array of 4 u8 containing either true or false to enable or disable rumble for that controller
```
Dragorn421, line 275:
```suggestion
 * @param enable    Array of u8 containing either true or false to enable or disable rumble for each controller
```
may as well not hardcode the 4 from `MAXCONTROLLERS` in a comment

src/code/padmgr.c
```diff
@@ -23,235 +84,260 @@ OSMesgQueue* PadMgr_LockSerialMesgQueue(PadMgr* padMgr) {
...
                     padMgr->ctrlrIsConnected[i] = true;
-                    osSyncPrintf(VT_FGCOL(YELLOW));
-                    osSyncPrintf("padmgr: %dコン: %s\n", i + 1, "認識しました"); // "Recognized"
-                    osSyncPrintf(VT_RST);
+                    // "Recognized"
+                    PADMGR_LOG(i, "認識しました");
                 }
                 break;
-            case 4:
+            case (CHNL_ERR_OVERRUN >> 4):
```
Dragorn421, line 311: where do these defines come from?
os_cont.h http://n64devkit.square7.ch/header/os_cont.htm has
```c
/* controller errors */
#define CONT_NO_RESPONSE_ERROR          0x8
#define CONT_OVERRUN_ERROR              0x4
#ifdef _HW_VERSION_1
#define CONT_FRAME_ERROR                0x2
#define CONT_COLLISION_ERROR            0x1
#endif 
```
but if it wasn't for seeing this (`CHNL_ERR_OVERRUN`) I would have never tied those defines to `OSContPad.errno`

src/code/padmgr.c
```diff
@@ -263,29 +349,43 @@ void PadMgr_ProcessInputs(PadMgr* padMgr) {
...
     s32 i;
-    OSMesgQueue* serialEventQueue = PadMgr_LockSerialMesgQueue(padMgr);
+    OSMesgQueue* serialEventQueue = PadMgr_AcquireSerialEventQueue(padMgr);
     u32 mask;
 
+    // Begin reading controller data
     osContStartReadData(serialEventQueue);
+
+    // Execute retrace callback
     if (padMgr->retraceCallback) {
```
Dragorn421, line 361: not yours but
```suggestion
    if (padMgr->retraceCallback != NULL) {
```

src/code/padmgr.c
```diff
@@ -298,15 +398,19 @@ void PadMgr_HandleRetraceMsg(PadMgr* padMgr) {
...
+        --padMgr->rumbleOffTimer;
         PadMgr_RumbleStop(padMgr);
-    } else if (padMgr->rumbleOnFrames == 0) {
+    } else if (padMgr->rumbleOnTimer == 0) {
+        // If the rumble on timer is inactive, no rumble
         PadMgr_RumbleStop(padMgr);
     } else if (!padMgr->preNMIShutdown) {
-        PadMgr_RumbleControl(padMgr);
-        --padMgr->rumbleOnFrames;
+        // If not in PreNMI, update rumble
```
Dragorn421, line 411:
```suggestion
        // If not past PreNMI, update rumble
```
given that it's a single point in time, right (I get what "in prenmi" would mean but exactitude y'know)

src/code/padmgr.c
Dragorn421: I guess technically commenting that "if this is called once per VI, then passing gameRequest=true or false changes nothing", is correct
(besides the first ever call due to `inputOut` possibly holding garbage in the gameRequest=false path, but let's ignore that)

(note idk if "changes nothing is correct, as this particular piece of code:
```c
buttonDiff = inputOut->prev.button ^ inputOut->cur.button;
inputOut->press.button = inputOut->cur.button & buttonDiff;
inputOut->rel.button = inputOut->prev.button & buttonDiff;
```
idk if it outputs the same thing when run once in `PadMgr_RequestPadData` or several times in `PadMgr_UpdateInputs` with more than one inputs)

----------

But I think this is better explained by just seeing this boolean as a "start game input frame", since it resets the accumulators,
and the use case (it's true only right before gamestate_update()) seems to comfort that

And other calls use false since VIs may have elapsed since, so they don't want to ruin the accumulators (and also they'd want accurate press/rel data since the last time they called).

At least what is sure to me is that `gameRequest=true` should be a single function's job so this argument may be called with that in mind, like `mainCall` or `startInputFrame` idk

src/code/padmgr.c
Thar0: It's one of the secrets of rcp.h

src/code/padmgr.c
```diff
@@ -1,20 +1,81 @@
...
+        osSyncPrintf(VT_FGCOL(YELLOW));                                  \
+        /* padmgr: Controller %d: %s */                                  \
+        osSyncPrintf("padmgr: %dコン: %s\n", (controllerNo) + 1, (msg)); \
+        osSyncPrintf(VT_RST);                                            \
+    }(void)0
+
+#define LOG_SEVERITY_NOLOG 0
+#define LOG_SEVERITY_INFO 1
+#define LOG_SEVERITY_ERROR 2
+#define LOG_SEVERITY_VERBOSE 3
```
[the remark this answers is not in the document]
Thar0: They're inferred from other files' `LOG_SEVERITY` macros, I want to unify them all eventually but this isn't the place to do that

src/code/padmgr.c
```diff
@@ -1,20 +1,81 @@
...
+
+s32 gPadMgrLogSeverity = LOG_SEVERITY_INFO;
```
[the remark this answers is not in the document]
Thar0: For similar reasons to above I'll hold off on this change, since all other `*LogSeverity` variables are currently `g`-prefixed. We can look into changing them all at once some other time.

src/code/padmgr.c
```diff
@@ -23,235 +84,260 @@ OSMesgQueue* PadMgr_LockSerialMesgQueue(PadMgr* padMgr) {
...
-    static u32 errcnt = 0;
-    static u32 frame;
-    s32 temp = 1;
+/**
+ * Activates the rumble pak for all controllers it is enabled on, stops it for all controllers it is disabled on and
+ * attempts to initialize it for a controller if it is not already initialized.
+ */
+void PadMgr_UpdateRumble(PadMgr* padMgr) {
+    static u32 sRumbleErrorCount = 0; // original name: "errcnt"
+    static u32 sRumbleUpdateCounter;
```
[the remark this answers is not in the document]
Thar0: afaik the C standard requires that static variables be initialized to 0 if not specified, and this is even before considering that almost everyone initializes bss to 0 anyway, so I think it should be clear enough?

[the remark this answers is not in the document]
Dragorn421: I didn't know static meant uninitialized -> 0 so alright

src/code/padmgr.c
Dragorn421: I swear I grep'ed and didn't find it before. found it now thanks :sweat_smile:

src/code/padmgr.c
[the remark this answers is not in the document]
EllipticEllipsis: https://en.cppreference.com/w/c/language/initialization#Implicit_initialization

Dragorn421 (approved):

src/code/code_800EC960.c
```diff
@@ -1267,7 +1267,8 @@ OcarinaNote sScarecrowsLongSongSecondNote;
 u8 sAudioHasMalonBgm;
 f32 sAudioMalonBgmDist;
 
-void PadMgr_RequestPadData(PadMgr* padmgr, Input* inputs, s32 mode);
+void PadMgr_RequestPadData(PadMgr* padMgr, Input inputs[MAXCONTROLLERS], s32 gameRequest);
```
Roman971, line 1270: We generally don't use the array notation in arguments, and I personally think we should continue to avoid it. This was discussed recently in discord and I explained why I think it's bad, so it might be simpler to just read that for my opinion (see <https://discord.com/channels/688807550715560050/688851976746041391/997577684446093364> or search for "array argument" in discord)

src/code/padmgr.c
```diff
@@ -298,53 +399,64 @@ void PadMgr_HandleRetraceMsg(PadMgr* padMgr) {
...
-    padMgr->preNMIShutdown = true;
+    padMgr->isResetting = true;
     PadMgr_RumbleReset(padMgr);
 }
 
-void PadMgr_RequestPadData(PadMgr* padMgr, Input* inputs, s32 mode) {
+/**
+ * Fetches the most recently polled inputs from padmgr
+ *
+ * @param[out] inputs   Array of Input to copy inputs into
```
Roman971, line 427: Iirc last time [out] (and [in]) were discussed, the consensus was to not use them, so we should probably omit it here as well
```suggestion
 * @param inputs   Array of Input to copy inputs into
```

fig02 (approved):

Dragorn421 (approved):

Roman971 (approved):

PR #1340, opened 2022-07-31
More documentation for PreRender.c

Thar0:
Turns out the pause menu effect is a software implementation of the Video Interface hardware component's anti-aliasing and divot filters

A few functions are still unnamed, I couldn't think of names that are both good and reasonably short. I have described them in comments however.


fig02 (approved):
another one where I cant really weigh in on the content of names, gfx stuff is a bit too low level for me. I like the new function names from the perspective of z_play tho 

looks good to me from a high level

EllipticEllipsis (commented):
Nice job on this, a lot more readable. I have a few questions/suggestions.

src/code/PreRender.c
```diff
@@ -1,6 +1,16 @@
+/**
+ * @file PreRender.c
+ *
+ * This file implements various routines important to framebuffer effects, such as RDP accelerated color and depth
+ * buffer copies and coverage drawing. Also contains software implementations of the Video Interface anti-aliasing and
```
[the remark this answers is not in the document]
Thar0: I'm not sure exactly how standard it is, it seems to be at least somewhat used in other graphics contexts like OpenGL but not very much and not really explained.. The N64 programming manual describes it very briefly but reasonably well to get an intuition on it http://n64devkit.square7.ch/pro-man/pro15/15-03.htm

src/code/PreRender.c
```diff
@@ -298,178 +414,253 @@ void func_800C213C(PreRender* this, Gfx** gfxp) {
...
+ * @param y     Center pixel y
+ */
+void PreRender_AntiAliasFilter(PreRender* this, s32 x, s32 y) {
     s32 i;
     s32 j;
-    s32 buffA[3 * 5];
+    s32 buffCvg[3 * 5];
     s32 buffR[3 * 5];
     s32 buffG[3 * 5];
     s32 buffB[3 * 5];
```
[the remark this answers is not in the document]
Thar0: I don't think it's really that worthwhile since it's such a specific algorithm and it's all in the one function so it's all pretty local already

src/code/PreRender.c
```diff
@@ -33,165 +43,222 @@ void PreRender_Destroy(PreRender* this) {
...
 }
 
-void func_800C0F28(PreRender* this, Gfx** gfxp, void* buf, void* bufSave) {
+/**
+ * Copies RGBA16 image `buf` to `bufSave`
+ *
+ * @param this      PreRender instance
+ * @param gfxp      Display list pointer
+ * @param buf       Image to copy from
+ * @param bufSave   Buffer to copy to
```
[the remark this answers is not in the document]
Thar0: I went with `img` and `imgDst` to be a little more specific

src/code/PreRender.c
```diff
@@ -254,40 +338,69 @@ void func_800C1F20(PreRender* this, Gfx** gfxp) {
...
+/**
+ * Draws the coverage of the current framebuffer `this->fbuf` to an I8 image at `this->cvgSave`. Overwrites
+ * `this->fbuf` in the process.
+ */
+void PreRender_DrawCoverage(PreRender* this, Gfx** gfxp) {
+    PreRender_DrawCoverage16b(this, gfxp);
     LogUtils_CheckNullPointer("this->cvg_save", this->cvgSave, "../PreRender.c", 532);
     if (this->cvgSave != NULL) {
-        func_800C1B24(this, gfxp, this->fbuf, this->cvgSave);
+        PreRender_DrawCoverage8b(this, gfxp, this->fbuf, this->cvgSave);
```
Dragorn421, line 384: So, to recap

- `PreRender_DrawCoverage16b`
  - with `G_RM_VISCVG`
  - overwrites the rgba16 `this->width`*`this->height` framebuffer
  - at `this-fbuf`
  - with its coverage, stored into the color channels (red=green=blue=coverage , over 5 bits)
- `PreRender_DrawCoverage8b`
  - loads the rgba16 `this->width`*`this->height` framebuffer/image (same thing here since coverage doesn't matter anymore)
  - at `this-fbuf`
  - as a ia16 `this->width`*`this->height` image
  - then with `G_RM_OPA_CI2` and combiner=texel0
  - draws it to `cvgSave` as a i8 framebuffer

To summarize further,
- `PreRender_DrawCoverage16b` overwrites `this->fbuf` with its coverage (rgba16)
- `PreRender_DrawCoverage8b` converts the rgba16-stored coverage into i8

The rgba16->ia16->i8 step is interesting, I guess it does `i8 = (r << 3) | (g >> 2) = (r << 3) | (r >> 2)` (r=g=b=coverage)
(which looks a lot like `buffR[i] = (pxIn.r << 3) | (pxIn.r >> 2);` in `PreRender_AntiAliasFilter` and angrylion's rgba16 to rgba32 conversion https://github.com/ata4/angrylion-rdp-plus/blob/6660e93acc94506933238c9a593494684d0d3bde/src/core/n64video/rdp/tmem.c#L2063 )

anyway I would rename the two functions `PreRender_DrawCoverage` is split into like, maybe,
- `PreRender_DrawCoverage16b` -> `PreRender_FbufToCoverage`
- `PreRender_DrawCoverage8b` -> `PreRender_CoverageRgba16ToI8`
(idk really)

src/code/PreRender.c
```diff
@@ -254,40 +355,71 @@ void func_800C1F20(PreRender* this, Gfx** gfxp) {
     }
 }
 
-void func_800C1FA4(PreRender* this, Gfx** gfxp) {
+/**
+ * Fetches the coverage of the current framebuffer into an image of the same format as the currentcolor image, storing
```
Dragorn421, line 359:
```suggestion
 * Fetches the coverage of the current framebuffer into an image of the same format as the current color image, storing
```
😅

Dragorn421 (commented):
Now the coverage things look even more crazy

Afaict the coverage-get is only used to be fed into the software divot filter, do you know what goal this achieves?

Dragorn421 (approved):
I choose to believe whoever wrote this had a part in designing the RDP or took their daily coffee with the team that did

fig02 (approved):

include/functions.h
```diff
@@ -1512,22 +1512,22 @@ void PreRender_SetValuesSave(PreRender* this, u32 width, u32 height, void* fbuf,
...
+void PreRender_CopyImageRegionImpl(PreRender* this, Gfx** gfxp);
+void func_800C170C(PreRender* this, Gfx** gfxp, void* buf, void* bufSave, u32 r, u32 g, u32 b, u32 a);
 void func_800C1AE8(PreRender* this, Gfx** gfxp, void* fbuf, void* fbufSave);
-void func_800C1B24(PreRender* this, Gfx** gfxp, void* fbuf, void* cvgSave);
-void func_800C1E9C(PreRender* this, Gfx** gfxp);
-void func_800C1F20(PreRender* this, Gfx** gfxp);
-void func_800C1FA4(PreRender* this, Gfx** gfxp);
-void func_800C20B4(PreRender* this, Gfx** gfxp);
-void func_800C2118(PreRender* this, Gfx** gfxp);
+void PreRender_DrawCoverage8b(PreRender* this, Gfx** gfxp, void* img, void* cvgDst);
```
Roman971, line 1519: I can't find this function anywhere in the C file, is it supposed to be here?
Same comment for `PreRender_DrawCoverage16b` and possible others

Also feel free to remove prototypes from this file if they are internal and not meant to be called externally, which I assume is the case for most functions in this system

src/code/PreRender.c
```diff
@@ -298,178 +430,260 @@ void func_800C213C(PreRender* this, Gfx** gfxp) {
...
+ * Applies the Video Interface divot filter to an image.
+ *
+ * This filter removes "divots" in an anti-aliased image, single-pixel holes created when many boundary edges all
+ * occupy a single pixel. The algorithm removes these by sliding a 3-pixel-wide window across each row of pixels and
+ * replaces the center pixel color with the median pixel color in the window.
+ *
+ * This is a software implementation of the same algorithm used in the Video Interface hardware when OS_VI_DIVOT_ON is
+ * set in the VI Control Register.
+ *
+ * @param this  PreRender instance
```
Roman971, line 684: Imo we could just omit the `@param` for simple cases like this
```suggestion
```

src/code/z_play.c
```diff
@@ -1237,10 +1237,10 @@ void Play_Draw(PlayState* this) {
 
                     this->pauseBgPreRender.fbuf = gfxCtx->curFrameBuffer;
                     this->pauseBgPreRender.fbufSave = (u16*)gZBuffer;
-                    func_800C1F20(&this->pauseBgPreRender, &sp70);
+                    PreRender_SaveFramebuffer(&this->pauseBgPreRender, &sp70);
```
Roman971, line 1240: More of a side note but at some point we really need to decide if we're using "framebuffer" or "frame buffer", and use it consistently.. (iirc it was supposed to be the latter)

Thar0: Yeah definitely, it's worthy of note that libultra VI functions use framebuffer/Framebuffer so if we want to be 100% consistent we would have to use the same

fig02 (approved):

Roman971 (approved):

PR #1346, opened 2022-08-12
Document some Player-related things

EllipticEllipsis:
Vague title for what ended up being a rather less focussed PR than initially planned.
- Document 2 seldom-used Player flags, `PLAYER_STATE2_FREEZE_WITH_GRAVITY` and `PLAYER_STATE2_DRAW_REFLECTION`; I expect we will have to rethink the name of the former a bit, but it's only used by Twinrova so it's a bit tricky to deduce anything general about it.
- Named `Player_StopHorizontalMovement`, and cleaned up a few arg names where other functions tell us what they are. Some of the functions are going to be quite troublesome to name, and I did not aim to be complete here.
- Documented various things linked by the vague theme of "losing items":
  - The function that deletes Master Sword. Did not name this since it is a cutscene action/cue interpreter and they're presumably all going to be named in the same way.
  - The function that burns a Deku Shield, `Player_TryBurnShield`
  - The last parts of the functions in Like Likes that remove shields and tunics
  - The functions in `code_80097A00`, some of which had confusing names or argument names. Clarified the appropriate enum usage a bit: we can't really use `*` as a wildcard when everywhere else in the file it's used to mean pointer.
- Took the opportunity to finish documenting 3 actors: ItemShield, EnRr, EnIceHono, and mark a well-known bug in the first one (and label the troublesome matrices in zl2/zl3 consistently).
- Add offset comments to the cutscene arrays: the animation/other type handlers I did in dec since they're only referenced from the heterogenous arrays, but these arrays I left in hex since they refer to quantities currently written in hex in most other places in the codebase, and the purpose of this is specifically to aid documentation now rather than be permanent.

And possibly something else I elided or forgot that doesn't fit into one of the above.


src/code/code_80097A00.c
```diff
@@ -224,30 +225,43 @@ u8 gItemSlots[] = {
...
+    gSaveContext.equips.equipment |= value << gEquipShifts[type];
 }
 
-u8 Inventory_DeleteEquipment(PlayState* play, s16 equipment) {
+/**
+ * Deletes the currently-equipped equipment of type `equipmentType`
+ * 
+ * @param equipmentType Type of equipment (sword/shield/tunic/boots) to delete (uses `EquipmentType` enum).
+ * 
+ * @return u8 deleted item
```
Dragorn421, line 244: Having the type seems redundant?

src/code/code_80097A00.c
```diff
@@ -259,7 +273,13 @@ u8 Inventory_DeleteEquipment(PlayState* play, s16 equipment) {
...
 }
 
-void Inventory_ChangeUpgrade(s16 upgrade, s16 value) {
-    gSaveContext.inventory.upgrades &= gUpgradeNegMasks[upgrade];
-    gSaveContext.inventory.upgrades |= value << gUpgradeShifts[upgrade];
+/**
+ * Changes the current upgrade
+ * 
+ * @param type Type of upgrade (uses `UpgradeType` enum).
+ * @param value Level to change upgrade to (plain number, valid range depends on `upgrade`).
```
Dragorn421, line 280:
```suggestion
 * @param upgradeType Type of upgrade (uses `UpgradeType` enum).
 * @param value Level to change upgrade to (plain number, valid range depends on `upgradeType`).
```
Also I'm not a fan of writing things like "upgradeType Type of upgrade", it's just filling the screen with redundancy?

src/code/z_sram.c
```diff
@@ -308,11 +308,11 @@ static s16 sDungeonEntrances[] = {
...
  *  Copy save currently on the buffer to Save Context and complete various tasks to open the save.
  *  This includes:
- *  - Set proper entrance depending on where the game was saved
- *  - If health is less than 3 hearts, give 3 hearts
- *  - If either scarecrow song is set, copy them from save context to the proper location
- *  - Handle a case where the player saved and quit after zelda cutscene but didnt get the song
- *  - Give and equip master sword if player is adult and doesnt have kokiri sword (bug?)
+ *  - Set proper entrance depending on where the game was saved.
+ *  - If health is less than 3 hearts, give 3 hearts.
+ *  - If either Scarecrow's Song is set, copy it from SaveContext to the proper location.
```
Dragorn421, line 313: may want to reword that sentence better, it's trying to cover the two scarecrow songs

src/overlays/actors/ovl_En_Ice_Hono/z_en_ice_hono.c
```diff
@@ -151,35 +151,37 @@ void EnIceHono_InitSmallFlame(Actor* thisx, PlayState* play) {
 
 void EnIceHono_Init(Actor* thisx, PlayState* play) {
     EnIceHono* this = (EnIceHono*)thisx;
-    s16 params = this->actor.params;
+    s16 params = this->actor.params; // Required for regalloc
```
Dragorn421, line 154: is there a point introducing these comments at this point?

src/overlays/actors/ovl_En_Zl2/z_en_zl2.c
```diff
@@ -435,15 +435,15 @@ void func_80B4F230(EnZl2* this, s16 arg1, s32 arg2) {
 s32 func_80B4F45C(PlayState* play, s32 limbIndex, Gfx** dList, Vec3f* pos, Vec3s* rot, void* thisx, Gfx** gfx) {
     s32 pad;
     EnZl2* this = (EnZl2*)thisx;
-    Mtx* sp74;
+    Mtx* mtxs;
```
Dragorn421, line 438: `matrices`?

src/overlays/actors/ovl_Item_Shield/z_item_shield.c
```diff
@@ -62,24 +65,24 @@ void ItemShield_Init(Actor* thisx, PlayState* play) {
...
-    this->unk_19C = 0;
+    this->stateFlags = 0;
 
     switch (this->actor.params) {
-        case 0:
+        case ITEMSHIELD_TYPE_COLLECTIBLE:
             ActorShape_Init(&this->actor.shape, 1400.0f, NULL, 0.0f);
             this->actor.shape.rot.x = 0x4000;
-            ItemShield_SetupAction(this, func_80B86BC8);
+            ItemShield_SetupAction(this, ItemShield_Collectable_Wait);
```
Dragorn421, line 74: collectible/collectable probably should be consistent?

src/overlays/actors/ovl_Item_Shield/z_item_shield.h
```diff
@@ -4,18 +4,26 @@
 #include "ultra64.h"
 #include "global.h"
 
+#define ITEMSHIELD_FLAME_COUNT 8
+
+typedef enum {
+    /* 0 */ ITEMSHIELD_TYPE_COLLECTIBLE,
+    /* 0 */ ITEMSHIELD_TYPE_BURNING
```
Dragorn421, line 11:
```suggestion
    /* 1 */ ITEMSHIELD_TYPE_BURNING
```
oopsie 8)

Dragorn421 (commented):
This is just 600 lines changed, the vast majority trivially so, and z_player is already overwhelming NotLikeThis

src/code/code_80097A00.c
EllipticEllipsis: The main point of this is to say which enums to use.

src/overlays/actors/ovl_En_Ice_Hono/z_en_ice_hono.c
fig02: yeah imo this is not needed

EllipticEllipsis: I hope everyone noticed that this temp is not actually used.

src/code/code_80097A00.c
```diff
@@ -224,30 +225,43 @@ u8 gItemSlots[] = {
...
-void Inventory_ChangeEquipment(s16 equipment, u16 value) {
-    gSaveContext.equips.equipment &= gEquipNegMasks[equipment];
-    gSaveContext.equips.equipment |= value << gEquipShifts[equipment];
+/**
+ * Change currently-equipped item
+ * 
+ * @param type Type of equipment to change. Use `EquipmentType` enum
+ * @param value Value of equipment to change. Use the appropriate `EquipValue____` enum
+ */
+void Inventory_ChangeEquipment(s16 type, u16 value) {
```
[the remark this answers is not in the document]
EllipticEllipsis: I don't see why you'd change one but not the other, would you want `equipmentValue` too?

src/code/code_80097A00.c
Dragorn421: Yeah so `@param upgradeType uses `UpgradeType` enum.` is enough?

src/overlays/actors/ovl_En_Ice_Hono/z_en_ice_hono.c
Dragorn421: I did not :sweat_smile: 
Now I would say either remove the comment anyway, or actually put an explanation since I assume we don't know for sure that's what the original source had?

fig02: I would prefer `// unused but required to match`
reason being that we are moving toward this being less of a decomp codebase and more of a "useable" codebase, so the nitty gritty details like "regalloc" start to become less important

src/code/code_80097A00.c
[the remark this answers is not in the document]
Dragorn421: well, elsewhere `EquipmentType` values are `equipmentType`
`value` is just value usually but `equipValue` (I don't think that's used anywhere currently) would work too yeah

src/overlays/actors/ovl_Item_Shield/z_item_shield.c
EllipticEllipsis: Does anyone care which? Apparently both are valid.

fig02: to me "collectible" is the noun, the thing that is collectable (collect-able)
thats prob just something ive made up in my head and not actually standard though lol

Dragorn421: "collectable" is only an adjective so probably go with "collectible"?

EllipticEllipsis: OED only lists "-able", with the note
> Forms:  Also (now chiefly U.S.) -ible.

EllipticEllipsis: Although interestingly the noun form apparently originates from the US.

EllipticEllipsis: Well Item00 and elsewhere uses "ible", so I'll use that for the sake of standardisation.

src/code/code_80097A00.c
EllipticEllipsis: The other thing is, it may sound redundant in certain cases, but sometimes (arguably most of the other arguments in here) there are usage details that are worth specifying, and the simplest way to do that is to just list all the params where the slightest confusion may arise.

src/code/z_sram.c
EllipticEllipsis: `- Copy any set Scarecrow's Songs from SaveContext to the proper location.`?

src/code/code_80097A00.c
EllipticEllipsis: Interestingly this is what the default Doxygen generator does.

src/code/z_sram.c
Dragorn421: I like the first one

src/code/code_80097A00.c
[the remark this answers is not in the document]
Roman971: `type` seems appropriate here since the function is specific to the equipment and named as such

include/functions.h
```diff
@@ -1172,9 +1172,9 @@ void Room_Draw(PlayState* play, Room* room, u32 flags);
 void func_80097534(PlayState* play, RoomContext* roomCtx);
 void Sample_Destroy(GameState* thisx);
 void Sample_Init(GameState* thisx);
-void Inventory_ChangeEquipment(s16 equipment, u16 value);
-u8 Inventory_DeleteEquipment(PlayState* play, s16 equipment);
-void Inventory_ChangeUpgrade(s16 upgrade, s16 value);
+void Inventory_ChangeEquipment(s16 equipmentType, u16 equipValue);
+u8 Inventory_DeleteEquippedItem(PlayState* play, s16 equipmentType);
```
fig02, line 1176: wondering why the wording change on this function
It is true that it removes from your current equips if you have it equipped, but it also removes if from your inventory.
Id say that `Inventory_DeleteEquipment` makes more sense alongside `Inventory_ChangeEquipment` as well. and having the word "item" in it is prob more confusing and makes it sound like it takes an item id

src/overlays/actors/ovl_En_Blkobj/z_en_blkobj.c
```diff
@@ -77,13 +77,16 @@ void EnBlkobj_Destroy(Actor* thisx, PlayState* play) {
     DynaPoly_DeleteBgActor(play, &play->colCtx.dyna, this->dyna.bgId);
 }
 
+/**
+ * Draw Player's reflection until it is close enough to start the fight
+ */
 void EnBlkobj_Wait(EnBlkobj* this, PlayState* play) {
```
fig02, line 83: feel like this comment makes it seem like this actor is doing the drawing of the reflection. dunno if we can change that phrasing to make it clear that its just telling link to do it

src/code/z_sram.c
```diff
@@ -448,9 +448,13 @@ void Sram_OpenSave(SramContext* sramCtx) {
         }
     }
 
-    if (LINK_AGE_IN_YEARS == YEARS_ADULT && !CHECK_OWNED_EQUIP(EQUIP_TYPE_SWORD, EQUIP_INV_SWORD_MASTER)) {
+    //! Restore Master Sword to inventory and re-equip it if missing. Patch for one way of getting the "swordless Link"
+    //! glitch
```
fig02, line 452: we can probably put the note about the patch on its own line, and phrase it as "This is a patch for one way...". also why the exclamation marks? we really only do that for bugs

include/functions.h
EllipticEllipsis: The key thing I wanted to get across is that it will delete the currently-equipped thing of that type, which the original name didn't (and which is also not how `ChangeEquipment` works). I don't think they're analogous if only because the arguments are not the same.

src/overlays/actors/ovl_Item_Shield/z_item_shield.c
```diff
@@ -217,11 +239,15 @@ void ItemShield_Update(Actor* thisx, PlayState* play) {
...
     ItemShield* this = (ItemShield*)thisx;
 
-    if (!(this->unk_19C & 2)) {
+    if (!(this->stateFlags & ITEMSHIELD_STATE_INVISIBLE)) {
         OPEN_DISPS(play->state.gfxCtx, "../z_item_shield.c", 457);
         Gfx_SetupDL_25Opa(play->state.gfxCtx);
         gSPMatrix(POLY_OPA_DISP++, Matrix_NewMtx(play->state.gfxCtx, "../z_item_shield.c", 460),
                   G_MTX_NOPUSH | G_MTX_LOAD | G_MTX_MODELVIEW);
+        //! @bug This displaylist will attempt to draw a displaylist from segment 12, but this actor never sets segment
+        //! 12, so it will draw whatever was last set in segment 12. Therefore when that was not a displaylist, it is
```
fig02, line 248: can possibly reword to remove "segment 12" being repeated 3 times in a row.
"This display list references segment 12, which the player actor usually sets up properly before drawing his shield. However when drawn in isolation by this actor, that segment was not set properly. If an actor that processes before this one sets segment 12 to something other than a display list, a crash is likely to occur. A notable example..."

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -3722,15 +3724,15 @@ s32 func_80838144(s32 arg0) {
...
 }
 
-s32 func_8083816C(s32 arg0) {
-    return (arg0 == FLOOR_TYPE_4) || (arg0 == FLOOR_TYPE_7) || (arg0 == FLOOR_TYPE_12);
+s32 func_8083816C(s32 floorType) {
+    return (floorType == FLOOR_TYPE_4) || (floorType == FLOOR_TYPE_7) || (floorType == FLOOR_TYPE_12);
 }
 
-void func_8083819C(Player* this, PlayState* play) {
+void Player_TryBurnShield(Player* this, PlayState* play) {
```
fig02, line 3731: would say deku shield in the name. may be obvious to most that deku is the only one that can burn, but yea

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -14127,6 +14148,10 @@ void func_80852608(PlayState* play, Player* this, CsCmdActorAction* arg2) {
     func_808322D0(play, this, &gPlayerAnim_002838);
 }
 
+/**
+ * Unequip Master Sword and remove from inventory. Since this only checks on frame 10 of the animation, if the animation
+ * never reaches this frame (such as if stuck in a "superslide" state), the sword will not be removed.
+ */
```
fig02, line 14154: not sure about mentioning the "bug" of the superslide state in the function comment. seems more appropriate at the check for the frame

src/overlays/actors/ovl_En_Blkobj/z_en_blkobj.c
fig02: "Wait for player to be in range to start the fight. Until then, signal that he should draw his own reflection", or something

include/functions.h
Dragorn421: Besides change/delete, what's the difference between `Inventory_ChangeEquipment` and `Inventory_DeleteEquippedItem`? Both deal with the currently equiped item

Note your comment on `Inventory_ChangeEquipment` even says "Change currently-equipped item."
while `Inventory_DeleteEquippedItem` has "Deletes the currently-equipped equipment"
It just seems natural to use either interchangeably

src/overlays/actors/ovl_Item_Shield/z_item_shield.c
```diff
@@ -138,38 +147,48 @@ void func_80B86BC8(ItemShield* this, PlayState* play) {
...
+    // Spawn flames at random locations until despawn
+    for (i = 0; i < ITEMSHIELD_FLAME_COUNT; i++) {
+        type = ITEMSHIELD_FLAME_TYPE_COUNT - this->reverseflameTypeIndices[i];
+        sFlamePos.x = this->flameBasePoss[i].x;
+        sFlamePos.y = this->flameBasePoss[i].y + (this->actor.shape.yOffset * 0.01f) + (sFlameScales[type] * -10.0f * 0.2f);
+        sFlamePos.z = this->flameBasePoss[i].z;
+        EffectSsFireTail_SpawnFlame(play, &this->actor, &sFlamePos, sFlameScales[type] * 0.2f, -1,
+                                    sFlameIntensities[type]);
+
+        // cycle through flame types and rechoose random positions until about to disappear
```
Dragorn421, line 176:
```suggestion
        // cycle through flame types and pick new random positions until shield is about to disappear
```

src/overlays/actors/ovl_Item_Shield/z_item_shield.c
```diff
@@ -62,24 +65,24 @@ void ItemShield_Init(Actor* thisx, PlayState* play) {
...
-            this->unk_19C |= 2;
-            for (i = 0; i < 8; i++) {
-                this->unk_19E[i] = 1 + 2 * i;
-                this->unk_1A8[i].x = Rand_CenteredFloat(10.0f);
-                this->unk_1A8[i].y = Rand_CenteredFloat(10.0f);
-                this->unk_1A8[i].z = Rand_CenteredFloat(10.0f);
+            ItemShield_SetupAction(this, ItemShield_Burning_FlyFromPlayer);
+            this->stateFlags |= ITEMSHIELD_STATE_INVISIBLE;
+            for (i = 0; i < ITEMSHIELD_FLAME_COUNT; i++) {
+                this->reverseflameTypeIndices[i] = 1 + 2 * i;
```
Dragorn421, line 82:
```suggestion
                this->flameStates[i] = 1 + 2 * i;
```
```suggestion
                this->flameTimers[i] = 1 + 2 * i;
```

include/z64player.h
```diff
@@ -431,7 +431,7 @@ typedef struct {
 #define PLAYER_STATE2_12 (1 << 12)
 #define PLAYER_STATE2_13 (1 << 13)
 #define PLAYER_STATE2_14 (1 << 14)
-#define PLAYER_STATE2_15 (1 << 15)
+#define PLAYER_STATE2_FREEZE_WITH_GRAVITY (1 << 15) // Replaces a large block in the middle of Player_UpdateCommon with only gravity and bgCheck updating (only used by Twinrova's freezing effects)
```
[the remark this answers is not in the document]
fig02: Only twinrova ends up using this yeah, but there's nothing twinrova specific about it. Can think of many other places someone would want to use this (it should honestly be used in most cases where actor.freezTtimer is used)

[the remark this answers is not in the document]
Dragorn421: I mean, for now we don't know exactly what this skips because this chunk of code is huge and player is mostly undocumented, and also I have the impression this is a Morithack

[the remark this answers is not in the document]
fig02: You dont really need to know the details of what the large update block does, cause all that matters is what its replaced with. It can be boiled down to: 
- stop horizontal movement
- call move forward (allow falling to the ground if airborne. no horiz movement cause of the above)
- update bgcheck info so he doesnt fall through the floor.

This is objectively a better method of freezing link because it allows some important aspects of his actor to update even tho the "main logic" will not. If this was used instead of freezeTimer in some instances it would even prevent SRM and related bugs from being able to manifest.

I dont doubt it was added by morita for his use case, but it doesnt mean it has to be named after that as if it only makes sense in that context. Its pretty general and useful outside of that.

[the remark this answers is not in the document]
fig02: Not to mention this state/code exists in mm and is used by multiple things. its broader than just a twinrova thing

fig02:
taking over this pr due to author inactivity

most existing review has been addressed except for a couple things that probably need more discussion

include/z64item.h
```diff
@@ -15,7 +15,7 @@ typedef enum {
     /* 4 */ EQUIP_TYPE_MAX
 } EquipmentType;
 
-// `EquipInv*` enums are for Inventory.equipment (for example used in the `CHECK_OWNED_EQUIP` macro)
+// `EquipInv____` enums are for Inventory.equipment (for example used in the `CHECK_OWNED_EQUIP` macro)
```
fig02, line 18: will say I am not the biggest fan of the `*` notation that has been used alot lately, mainly cause it looks like pointers in the context of a C codebase lol

But also dont think `____` should be used in place of it

src/overlays/actors/ovl_Item_Shield/z_item_shield.c
```diff
@@ -138,38 +147,51 @@ void func_80B86BC8(ItemShield* this, PlayState* play) {
...
+            this->flameTypes[i]--;
         } else if (this->timer > 16) {
-            this->unk_19E[i] = 15;
-            this->unk_1A8[i].x = Rand_CenteredFloat(15.0f);
-            this->unk_1A8[i].y = Rand_CenteredFloat(10.0f);
-            this->unk_1A8[i].z = Rand_CenteredFloat(15.0f);
+            this->flameTypes[i] = ITEMSHIELD_FLAME_TYPE_COUNT;
+            this->flameBasePoss[i].x = Rand_CenteredFloat(15.0f);
+            this->flameBasePoss[i].y = Rand_CenteredFloat(10.0f);
+            this->flameBasePoss[i].z = Rand_CenteredFloat(15.0f);
```
engineer124, line 186: the plural s on `Pos` looks funny. `this->flameBasePos` is probably fine, but could also do `this->flameBasePosList` or `this->flameBasePosArr`?

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -9953,6 +9956,7 @@ s32 func_80847A78(Player* this) {
 
 static Vec3f D_80854798 = { 0.0f, 18.0f, 0.0f };
 
+// Player_UpdateBgCheck
 void func_80847BA0(PlayState* play, Player* this) {
```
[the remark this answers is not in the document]
fig02: this function does *alot*. updating bgcheck info is the first half of the function, and the rest is bgcheck adjacent things. idk if I like just "update bg check", and would probably prefer just removing the comment until we come up with a better name (unless we do so now)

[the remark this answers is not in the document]
fig02: documenting all of the unks in this function is probably a pr of its own (though its pretty doable to do now, so i can attempt it soon)
so i will just remove the comment for now

src/overlays/actors/ovl_Item_Shield/z_item_shield.c
fig02: went with something a bit different, see how that looks

fig02: actually, its more of an offset than anything
the effect handles the "attaching to body parts" part (in a draw function no less)

fig02: this kinda rabbit holes into a question of "should the effect refer to this as an offset everywhere, and break the normal pattern of pos/vel/accel"

maybe just leave it as pos for now and let an effects focused pass answer this question

engineer124: Yeah, that's fine with me

include/z64item.h
Dragorn421: Well wildcards are a lot more often `*` than not so yeah idk https://en.wikipedia.org/wiki/Wildcard_character

`EquipInvXxx`, `EquipInvXXX`, `EquipInv***`, `EquipInv...`, `EquipInv[...]`, `EquipInv.*`, `EquipInv%` ...

include/z64save.h
```diff
@@ -62,13 +62,13 @@ typedef enum {
...
     /* 0x04 */ u8 cButtonSlots[3];
-    /* 0x08 */ u16 equipment; // a mask where each nibble corresponds to a type of equipment `EquipmentType`, and each nibble is a piece `EquipValue*`
+    /* 0x08 */ u16 equipment; // a mask where each nibble carries data for one type of equipment (`EquipmentType` enum), and the contents of that nibble is one piece of that type of equipment (`EquipValue____` enums)
 } ItemEquips; // size = 0x0A
 
 typedef struct {
     /* 0x00 */ u8 items[24];
     /* 0x18 */ s8 ammo[16];
-    /* 0x28 */ u16 equipment; // a mask where each nibble corresponds to a type of equipment `EquipmentType`, and each bit to an owned piece `EquipInv*`
+    /* 0x28 */ u16 equipment; // a mask where each nibble carries data for one type of equipment (`EquipmentType` enum), and each bit to a single owned piece (`EquipInv____` enums)
```
Dragorn421, line 71:
```suggestion
    /* 0x28 */ u16 equipment; // a mask where each nibble carries data for one type of equipment (`EquipmentType` enum), and each bit to a single owned piece (`EquipInvSword`, `EquipInvShield`, `EquipInvTunic`, `EquipInvBoots` enums)
```

include/z64save.h
```diff
@@ -62,13 +62,13 @@ typedef enum {
 typedef struct {
     /* 0x00 */ u8 buttonItems[4];
     /* 0x04 */ u8 cButtonSlots[3];
-    /* 0x08 */ u16 equipment; // a mask where each nibble corresponds to a type of equipment `EquipmentType`, and each nibble is a piece `EquipValue*`
+    /* 0x08 */ u16 equipment; // a mask where each nibble carries data for one type of equipment (`EquipmentType` enum), and the contents of that nibble is one piece of that type of equipment (`EquipValue____` enums)
```
Dragorn421, line 65:
```suggestion
    /* 0x08 */ u16 equipment; // a mask where each nibble carries data for one type of equipment (`EquipmentType` enum), and the contents of that nibble is one piece of that type of equipment (`EquipValueSword`, `EquipValueShield`, `EquipValueTunic`, `EquipValueBoots` enums)
```

include/functions.h
Dragorn421: either
```suggestion
void Inventory_ChangeEquipment(s16 equipmentType, u16 equipValue);
u8 Inventory_DeleteEquipment(PlayState* play, s16 equipmentType);
```
or
```suggestion
void Inventory_ChangeEquippedItem(s16 equipmentType, u16 equipValue);
u8 Inventory_DeleteEquippedItem(PlayState* play, s16 equipmentType);
```

src/overlays/actors/ovl_Item_Shield/z_item_shield.c
```diff
@@ -138,38 +147,51 @@ void func_80B86BC8(ItemShield* this, PlayState* play) {
...
+
+        // cycle through flame types and pick new random positions until the shield is about to disappear
+        if (this->flameTypes[i] != 0) {
+            this->flameTypes[i]--;
         } else if (this->timer > 16) {
-            this->unk_19E[i] = 15;
-            this->unk_1A8[i].x = Rand_CenteredFloat(15.0f);
-            this->unk_1A8[i].y = Rand_CenteredFloat(10.0f);
-            this->unk_1A8[i].z = Rand_CenteredFloat(15.0f);
+            this->flameTypes[i] = ITEMSHIELD_FLAME_TYPE_COUNT;
```
Dragorn421, line 183: This means `ITEMSHIELD_FLAME_TYPE_COUNT` is a flame type
and all values down to 0 are flame types
so there would be `ITEMSHIELD_FLAME_TYPE_COUNT+1` flame type values
i.e. `ITEMSHIELD_FLAME_TYPE_COUNT` is misnamed, idk how to fix but it could be 16 instead and use `ITEMSHIELD_FLAME_TYPE_COUNT-1` here

also then `sFlameScales[ITEMSHIELD_FLAME_TYPE_COUNT]` and `sFlameIntensities[ITEMSHIELD_FLAME_TYPE_COUNT]`

mzxrules: Had a quick look at what's going on. 

this->flameTypes essentially stores the remaining "life" for 8 different flames. It starts by initializing the 8 flames to the odds (1, 3, ..., 15) based on the number of flames. Then, each call to `ItemShield_Burning_Combust` draws the flame by computing the current animation frame (`type = 15 -  this->flameTypes`), and then calling EffectSsFireTail_SpawnFlame. Finally, if the "life" count is greater than zero, it gets decremented, otherwise a new flame is "spawned" with a life of 15 as long as there is enough time remaining.

So I think `this->flameTypes` -> `this->flameLife`, `type` -> `animFrame`, `ITEMSHIELD_FLAME_TYPE_COUNT` -> ITEMSHIELD_FLAME_LIFE_MAX` might be the way to go here, and I think a comment at line 82 is warranted to point out the weird coupling of `ITEMSHIELD_FLAME_COUNT` to the flame life.

fig02:

PR #1343, opened 2022-08-02
Name scenes (scene enum, entrance enum, various identifiers)

Dragorn421:
Name scenes. Doesn't change file names (e.g. in the spec or assets), only enums and such

- ~~In `include/tables/scene_table.h`, comment a name that would be eventually used for the enum and possibly the file (for example `DEKU_TREE`, `MARKET_ENTRANCE_DAY`, `DEATH_MOUNTAIN_TRAIL`)~~
- ~~In each map's xml, comment a prefix to use for the assets symbols (for example `gDekuTree`, `gMarketEntranceDay`, `gDMT`)~~

~~The goal is to formally agree on names for the maps instead of introducing bits at a time at other occasions, like naming random symbols in maps or just writing a sentence in plain English about a particular map's behavior.~~

~~I didn't name all scenes because I'm not sure what some should be named. They could be filled in later or by another PR, if we go with commenting names at all~~

PR #1348, opened 2022-08-13
Document Talon (En_Ta)

luryus:
Fully Documented Talon / ovl_En_Ta. Also renamed some Talon-related EVENTCHKINF macros, which caused changes in quite a few other places.

~~Still a draft PR, because there are still a few things left to rename (TALON_STATE_FLAG_3, ENTA_OTHER).~~


src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -10,26 +10,52 @@
...
+#define TALON_STATE_FLAG_CUCCO_GAME_START_EVENT_TRIGGERED (1 << 5)
+#define TALON_STATE_FLAG_UNUSED (1 << 6)
+#define TALON_STATE_FLAG_FLOOR_CAMERA_ACTIVE (1 << 7)
+#define TALON_STATE_FLAG_RAISING_HANDS (1 << 8)
+#define TALON_STATE_FLAG_RESTORE_BGM_ON_DESTROY (1 << 9)
+
+#define TALON_FACE_REACTION_SET 24
+
+#define EVENTINF_CUCCO_GAME_FINISHED EVENTINF_HORSES_0A
+#define EVENTINF_CUCCO_GAME_WON EVENTINF_HORSES_08
```
Dragorn421, line 27: You can just rename `EVENTINF_HORSES_08` and `EVENTINF_HORSES_0A`

The idea with the common `horses_` part is because all the `0x00-0x0F` flags are logically tied by statements like `gSaveContext.eventInf[EVENTINF_HORSES_INDEX] = 0;` which clears all the flags at once (not sure yet what the purpose is)

So idk maybe `EVENTINF_HORSES_CUCCO_GAME_WON` and such :smile:

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -10,26 +10,52 @@
...
+
+typedef enum { TALON_EYE_INDEX_OPEN, TALON_EYE_INDEX_HALF, TALON_EYE_INDEX_CLOSED } TalonEyeIndex;
```
Dragorn421, line 29: We prevent the formatter from putting such enums on single lines by having comments like `/* 0 */` before each (one example among many: `EnDaikuTalkState`), and in general we do that on all enums

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -63,33 +89,42 @@ static ColliderCylinderInit sCylinderInit = {
     { 30, 40, 0, { 0, 0, 0 } },
 };
 
-void func_80B13AA0(EnTa* this, EnTaActionFunc arg1, EnTaUnkFunc arg2) {
+void EnTa_SetActionFunctions(EnTa* this, EnTaActionFunc arg1, EnTaBlinkFunc arg2) {
```
Dragorn421, line 92: In general this kind of function we call `_SetupAction` (see the many others in the repo), though this one is different since it also has that blink function, but I think it's still fine
```suggestion
void EnTa_SetupAction(EnTa* this, EnTaActionFunc arg1, EnTaBlinkFunc arg2) {
```

Oh also `EnTaBlinkFunc` is the anim func actually?

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -63,33 +89,42 @@ static ColliderCylinderInit sCylinderInit = {
...
-    if (GET_EVENTINF(EVENTINF_HORSES_0A)) {
-        if (GET_EVENTINF(EVENTINF_HORSES_08)) {
-            if (GET_ITEMGETINF(ITEMGETINF_02)) {
+    // Check if cucco game was just finished
+    if (GET_EVENTINF(EVENTINF_CUCCO_GAME_FINISHED)) {
+        // Check if the game was won
+        if (GET_EVENTINF(EVENTINF_CUCCO_GAME_WON)) {
+            if (GET_ITEMGETINF(ITEMGETINF_TALON_BOTTLE)) {
+                // "Why don't you come to work here, on this ranch"
                 this->actor.textId = 0x2088;
```
Dragorn421, line 106: We had such comments on text ids before (long ago), but it's not something we currently do anymore

I think in these cases (text ids in this file) comments like `// first time winning` would be good instead

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -107,80 +142,83 @@ void EnTa_Init(Actor* thisx, PlayState* play2) {
...
-                func_80B13AA0(this, func_80B14D98, func_80B167C0);
-                this->eyeIndex = 0;
+                EnTa_SetActionFunctions(this, EnTa_IdleBackAtRanch, EnTa_AnimLoopCurrent);
+                this->eyeIndex = TALON_EYE_INDEX_OPEN;
                 Animation_PlayOnce(&this->skelAnime, &gTalonStandAnim);
                 this->currentAnimation = &gTalonStandAnim;
             }
             break;
         default:
+            // "Other Talon" == child Talon
```
Dragorn421, line 195: maybe...
```suggestion
        default: // child era Talon
            // "Other Talon"
```

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -107,80 +142,83 @@ void EnTa_Init(Actor* thisx, PlayState* play2) {
...
-                    func_80B13AA0(this, func_80B14634, func_80B167FC);
-                    this->eyeIndex = 2;
+                    EnTa_SetActionFunctions(this, EnTa_IdleAsleepInCastle, EnTa_AnimSleeping);
+                    this->eyeIndex = TALON_EYE_INDEX_CLOSED;
                     Animation_PlayOnce(&this->skelAnime, &gTalonSleepAnim);
                     this->currentAnimation = &gTalonSleepAnim;
                     this->actor.shape.shadowScale = 54.0f;
                 }
-            } else if (play->sceneId == SCENE_SOUKO) {
+            } else if (play->sceneId == SCENE_SOUKO) { // Talon's house
```
Dragorn421, line 212: We avoid commenting on the scene checks as the plan is to name them something more sensible than the debug names eventually (work started in #1343 )

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -537,89 +596,93 @@ void func_80B14E28(EnTa* this, PlayState* play) {
...
-        func_80B13AAC(this, play);
+        EnTa_SetActionFuncWithBackToSleepAnimation(this, EnTa_IdleSittingInLonLonHouse);
+        EnTa_SetTextForTalkInLonLonHouse(this, play);
     }
-    this->unk_2E0 |= 1;
+    this->stateFlags |= TALON_STATE_FLAG_TRACKING_PLAYER;
 }
 
-s32 func_80B150AC(EnTa* this, PlayState* play, s32 idx) {
+bool EnTa_IsPlayerHoldingSuperCucco(EnTa* this, PlayState* play, s32 cuccoIdx) {
```
Dragorn421, line 630: We don't use `bool`, because many booleans are typed by the game as not-int, so for consistency reasons we just don't use it at all

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -940,193 +1030,210 @@ void func_80B161C0(EnTa* this, PlayState* play) {
...
-        if (nextEyeIndex >= blinkTimer) {
-            this->eyeIndex = 0;
-            if (this->unk_2CE > 0) {
-                this->unk_2CE--;
+        // If the eyes were already closed (nextEyeIndex >= 3),
+        // set the eyes open and either set a new blink timer.
+        // If just woken up (rapidBlinks > 0), blink a few times
+        // in quick succession before starting the normal blink
+        // cycle.
+        if (nextEyeIndex >= 3) {
```
Dragorn421, line 1176: you can add a `_max` value to the enum as the last/max value and
```suggestion
        if (nextEyeIndex >= TALON_EYE_INDEX_MAX) {
```

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -1138,46 +1245,57 @@ void EnTa_Update(Actor* thisx, PlayState* play) {
...
-    if (!(this->unk_2E0 & 4)) {
-        this->unk_2B0(this);
+    if (!(this->stateFlags & TALON_STATE_FLAG_SUPPRESS_BLINK)) {
+        this->blinkFunc(this);
     }
 
-    if (this->unk_2E0 & 1) {
-        Actor_TrackPlayer(play, &this->actor, &this->unk_2D4, &this->unk_2DA, this->actor.focus.pos);
+    if (this->stateFlags & TALON_STATE_FLAG_TRACKING_PLAYER) {
+        Actor_TrackPlayer(play, &this->actor, &this->headRotStep, &this->torsoRot, this->actor.focus.pos);
```
Dragorn421, line 1256: why is one "step" and not the other?
(I think `headRotStep` can just be `headRot`)

Dragorn421 (commented):
Very nice job :)

src/overlays/actors/ovl_En_Ta/z_en_ta.c
luryus: Those flags are used in other minigames, too: 08 in Gerudo archery and 0A in Malon's race. So renaming them globally with these names would not be that great as they are also used for other purposes

src/overlays/actors/ovl_En_Ta/z_en_ta.c
luryus: Ok, I will add those!

src/overlays/actors/ovl_En_Ta/z_en_ta.c
luryus: I had the anim and blink function types mixed up in some places, I noticed this earlier but I guess I forgot to push the fixes 😄

src/overlays/actors/ovl_En_Ta/z_en_ta.c
luryus: I thought I renamed both fields to include `step`... apparently not

src/overlays/actors/ovl_En_Ta/z_en_ta.c
Dragorn421: Oh, I didn't notice that. Then I wouldn't name the flags at all yet, since these bits pertain to the same memory regardless of the actor they're used in I don't think different names should be used to designate them
but you could add a comment on the `EVENTINF_HORSES_08`/`0A` defines about their use for Talon's cucco minigame

include/z64save.h
```diff
@@ -354,13 +354,13 @@ typedef enum {
...
 #define EVENTCHKINF_69 0x69
-#define EVENTCHKINF_6A 0x6A
+#define EVENTCHKINF_TALON_WOKEN_IN_KAKARIKO 0x6A
 
 // 0x6B
-#define EVENTCHKINF_6B_INDEX 6
-#define EVENTCHKINF_6B_SHIFT 11
-#define EVENTCHKINF_6B_MASK (1 << EVENTCHKINF_6B_SHIFT)
-#define EVENTCHKINF_6B ((EVENTCHKINF_6B_INDEX << 4) | EVENTCHKINF_6B_SHIFT)
+#define EVENTCHKINF_TALON_RETURNED_FROM_KAKARIKO_INDEX 6
```
Yanis42, line 360: (afaik Talon never goes back to Lon-Lon after you woke him up as adult)

luryus: He does, if you talk to him after getting epona, he will move to the stable at Lon Lon

src/overlays/actors/ovl_En_Ta/z_en_ta.c
luryus: I added the comments (and tried to see how the flags are used in the other minigames), but still left these defines in place until the horse flags are renamed

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -10,26 +10,58 @@
 
 #define FLAGS (ACTOR_FLAG_0 | ACTOR_FLAG_3)
 
+#define TALON_STATE_FLAG_TRACKING_PLAYER (1 << 0)
+#define TALON_STATE_FLAG_GIVING_MILK_REFILL (1 << 1)
+#define TALON_STATE_FLAG_SUPPRESS_BLINK (1 << 2)
+// This has no effect, see EnTa_OverrideLimbDraw for details
+#define TALON_STATE_FLAG_SUPPRESS_ROCKING_ANIM (1 << 3)
```
luryus, line 17: I renamed this flag too, suggestions for better names are welcome :)

RUBenGAMArrarodRiguEZ-ToMtOm (approved):

include/z64save.h
```diff
@@ -354,13 +354,13 @@ typedef enum {
...
+#define EVENTCHKINF_TALON_RETURNED_FROM_KAKARIKO_SHIFT 11
+#define EVENTCHKINF_TALON_RETURNED_FROM_KAKARIKO_MASK (1 << EVENTCHKINF_TALON_RETURNED_FROM_KAKARIKO_SHIFT)
+#define EVENTCHKINF_TALON_RETURNED_FROM_KAKARIKO 0x6B
```
Dragorn421, line 363:
```suggestion
#define EVENTCHKINF_TALON_RETURNED_FROM_KAKARIKO ((EVENTCHKINF_TALON_RETURNED_FROM_KAKARIKO_INDEX << 4) | EVENTCHKINF_TALON_RETURNED_FROM_KAKARIKO_SHIFT)
```
the idea with this kind of thing is that for moving the flag you would then only have to touch the index, shift, or both, instead of possibly this define and index/shift going out of sync, which would most likely cause bugs

but I'm biased in defending this approach because I came up with this so if someone else thinks we shouldn't bother, feel free to say

Dragorn421 (commented):
<details>

<summary>
Graph of the flow between functions in this actor. Nice job!
</summary>

noname gv

With http://www.gson.org/egypt/
```sh
CFLAGS=-fdump-rtl-expand COMPILER=gcc make -j32
egypt build/src/overlays/actors/ovl_En_Ta/z_en_ta.c.237r.expand --callees AudioMgr_Init --omit EnTa_AnimLoopCurrent,EnTa_AnimRunToEnd | dot -Tsvg -O
```

</details>

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -788,148 +861,153 @@ void func_80B1585C(EnTa* this, PlayState* play) {
...
+        EnTa_SetTextForTalkInLonLonHouse(this, play);
     }
-    this->unk_2E0 |= 1;
+    this->stateFlags |= TALON_STATE_FLAG_TRACKING_PLAYER;
 }
 
-void func_80B15E80(EnTa* this, PlayState* play) {
+void EnTa_GiveItemInLonLonHouse(EnTa* this, PlayState* play) {
+    // Player sets itself as the parent after func_8002F434 (get item)
+    // is called. So if parent is set, the item was already given.
```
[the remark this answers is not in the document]
luryus: Yeah, I added this comment after I found out that the parent check somehow relates to getting the item, but I totally agree that it's best to not make any assumptions about the details here. I'll remove the comment

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -788,148 +861,153 @@ void func_80B1585C(EnTa* this, PlayState* play) {
...
     }
-    this->unk_2E0 |= 1;
+    this->stateFlags |= TALON_STATE_FLAG_TRACKING_PLAYER;
 }
 
-void func_80B15CC8(EnTa* this, PlayState* play) {
-    if (this->unk_2E0 & 0x10) {
-        func_80B13AA0(this, func_80B15BF8, func_80B16938);
-        this->unk_2E0 &= ~0x10;
+void EnTa_StartingCuccoGameAnimation1(EnTa* this, PlayState* play) {
```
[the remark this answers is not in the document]
luryus: Good point! I also changed all  `EnTa_RunAwayAnim*` -> `EnTa_RunAway*`

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -738,47 +800,57 @@ void func_80B154FC(EnTa* this, PlayState* play) {
...
                 Math_SmoothStepToS(&this->superCuccos[i]->actor.world.rot.y, i * -10000 - 3000, 2, 0x800, 0x100);
                 this->superCuccos[i]->actor.shape.rot.y = this->superCuccos[i]->actor.world.rot.y;
             }
         }
-    } else if (this->unk_2CC == 35) {
+    } else if (this->timer == 35) {
+        // At this point, prepare each super cucco for flight
         for (i = 0; i < ARRAY_COUNT(this->superCuccos); i++) {
-            this->unk_2C4[i] = (s32)(Rand_CenteredFloat(6.0f) + 10.0f);
+            // Set a 10-16 frame zero-gravity time for each super cucco
```
Dragorn421, line 837: It's not 10-16 actually, `Rand_CenteredFloat(6)` returns a random number in (-3,3)

with the truncation I think it means the timer can be 7-12 ?

include/z64save.h
```diff
@@ -354,13 +354,13 @@ typedef enum {
...
 
 // 0x6B
-#define EVENTCHKINF_6B_INDEX 6
-#define EVENTCHKINF_6B_SHIFT 11
-#define EVENTCHKINF_6B_MASK (1 << EVENTCHKINF_6B_SHIFT)
-#define EVENTCHKINF_6B ((EVENTCHKINF_6B_INDEX << 4) | EVENTCHKINF_6B_SHIFT)
+#define EVENTCHKINF_TALON_RETURNED_FROM_KAKARIKO_INDEX 6
+#define EVENTCHKINF_TALON_RETURNED_FROM_KAKARIKO_SHIFT 11
+#define EVENTCHKINF_TALON_RETURNED_FROM_KAKARIKO_MASK (1 << EVENTCHKINF_TALON_RETURNED_FROM_KAKARIKO_SHIFT)
+#define EVENTCHKINF_TALON_RETURNED_FROM_KAKARIKO ((EVENTCHKINF_TALON_RETURNED_FROM_KAKARIKO_INDEX << 4) | EVENTCHKINF_TALON_RETURNED_FROM_KAKARIKO_SHIFT)
```
mzxrules, line 363: what the hell is this nonsense

include/z64save.h
mzxrules: i'm of the opinion that the index/shift macros are annoyingly tedious, and the effort would be better spent by making a custom getter and check macros for this special case. Either something more generic like

    #define GETVAL_EVENTCHKINF(flag) (gSaveContext.eventChkInf[(flag) >> 4])
    #define CHECKVAL_EVENTCHKINF(v, flag) ((v) & (1 << ((flag) & 0xF)))

Dragorn421: Well this is outside the scope of this PR, especially since they're a new contributor. Feel free to bring up this discussion on the discord

luryus: I don't really have any opinion about this; maybe it would be clearer if z64save.h only had the flags as simple indices, because most of the flags follow that format already, and `En_Hy` handled the shifting and stuff with the macros suggested by @mzxrules. 
But yeah, that should probably be done later when someone documents `En_Hy` (afaik that's the only place where `_INDEX` and `_MASK` are used).

Should I restore `EVENTCHKINF_TALON_RETURNED_FROM_KAKARIKO` to just be `0x6B`?

Dragorn421: Well for now and for consistency this should stay _index and _shift since other flags do that

If this is changed down the line everything will be changed at once

Dragorn421 (approved):

src/overlays/actors/ovl_En_Ta/z_en_ta.h
```diff
@@ -9,31 +9,60 @@
 struct EnTa;
 
 typedef void (*EnTaActionFunc)(struct EnTa*, PlayState*);
-typedef void (*EnTaUnkFunc)(struct EnTa*);
+typedef void (*EnTaBlinkFunc)(struct EnTa*);
+typedef void (*EnTaAnimFunc)(struct EnTa*);
+
+typedef enum {
+    /* 0 */ ENTA_OTHER, // Child era
```
hensldm, line 16:
```suggestion
    /* 0 */ ENTA_CHILD_ERA,
```
Any reason not to call it the Child era type?

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -107,80 +146,83 @@ void EnTa_Init(Actor* thisx, PlayState* play2) {
...
-                this->eyeIndex = 2;
+                EnTa_SetupAction(this, EnTa_IdleAsleepInKakariko, EnTa_AnimSleeping);
+                this->eyeIndex = TALON_EYE_INDEX_CLOSED;
                 Animation_PlayOnce(&this->skelAnime, &gTalonSleepAnim);
                 this->currentAnimation = &gTalonSleepAnim;
                 this->actor.shape.shadowScale = 54.0f;
             }
             break;
-        case 2:
+        case ENTA_RETURNED_FROM_KAKARIKO:
```
hensldm, line 181: It is a little nicer to have an empty line between `break`s and the next case to show they are separate (vs fall-through cases where the new line is ommited).

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -941,192 +1016,204 @@ void func_80B161C0(EnTa* this, PlayState* play) {
...
+            this->blinkFunc = EnTa_BlinkWaitUntilNext;
         } else {
             this->eyeIndex = nextEyeIndex;
             this->blinkTimer = 1;
         }
     }
 }
 
-void func_80B167C0(EnTa* this) {
+void EnTa_AnimLoopCurrent(EnTa* this) {
```
hensldm, line 1173: Not sure about `Loop` since `Animation_PlayOnce` is what actually gets called. Perhaps `AnimPlayCurrent`, or possibly just `AnimNormal` to follow the style of the others where this is just the normal animation that doesn't do anything special.

luryus: The point here was that when this is set as the current `animFunc`, current animation will be looped (`Animation_PlayOnce` is called again as soon as the animation ends). As opposed to `EnTa_AnimRunToEnd` that does nothing after the current animation finishes.

I think `PlayCurrent` or `Normal` would not describe this enough?

fig02: I would use "repeat" or some term that gets that same meaning across without actually using the term "loop". Cause as mentioned, loop is a different type of playback method that is a part of the animation system (which they should have used here, but unfortunately did not)

luryus: That makes sense, I'll change this to `repeat`

src/overlays/actors/ovl_En_Ta/z_en_ta.h
luryus: This came from a printf that translated to "other Talon", but it's not actually used anywhere. EnTa_Init only has cases for the other two types and the default case. Now that I checked this, child era Talon actors actually use `0xFFFF` as the param, not zero. 

But yeah, I think changing this to CHILD_ERA is fine as that describes the other states better.

hensldm (approved):
LGTM

Dragorn421 (approved):

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -252,77 +299,78 @@ s32 func_80B142F4(EnTa* this, PlayState* play, u16 textId) {
...
-void func_80B143D4(EnTa* this, PlayState* play) {
+void EnTa_SleepTalkInLonLonHouse(EnTa* this, PlayState* play) {
     if (Actor_TextboxIsClosing(&this->actor, play)) {
-        func_80B13AA0(this, func_80B146F8, func_80B167FC);
+        EnTa_SetupAction(this, EnTa_IdleAsleepInLonLonHouse, EnTa_AnimSleeping);
     }
 }
 
-void func_80B14410(EnTa* this) {
+void EnTa_SetWokenFlagsAndActionFunctions(EnTa* this) {
```
fig02, line 320: maybe `SetupWokenUp`? would want this to be more about what it means for talons action as opposed to "setting flags and action function", if that makes sense

fig02: think the "woken" functions can also just be phrased as "awake"
so, this function can be like `EnTa_SetupAwake` which will then pass you to `EnTa_IdleAwakeInCastle` or `EnTa_IdleAwakeInKakariko`, same goes for the other "woken" functions. Think it just makes more sense as "talon is currently awake" rather than "talon has been awoken"

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -392,132 +443,137 @@ void func_80B14818(EnTa* this, PlayState* play) {
...
 }
 
-void func_80B14D98(EnTa* this, PlayState* play) {
-    if (func_80B142F4(this, play, 0x2055)) {
-        func_80B13AA0(this, func_80B14D4C, func_80B167C0);
+void EnTa_IdleBackAtRanch(EnTa* this, PlayState* play) {
+    if (EnTa_TalkToPlayer(this, play, 0x2055)) {
+        EnTa_SetupAction(this, EnTa_TalkBackAtRanch, EnTa_AnimRepeatCurrent);
     }
 }
```
fig02, line 564: do these need to say "back" I think just at ranch is probably fine no?
its more of a lore detail that he was sent back to the ranch after being somewhere else, but yea idk

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -537,89 +593,92 @@ void func_80B14E28(EnTa* this, PlayState* play) {
...
     Play_ChangeCameraStatus(play, this->returnToCamId, CAM_STAT_ACTIVE);
     Play_ClearCamera(play, this->subCamId);
 }
 
-void func_80B14F20(EnTa* this, EnTaActionFunc arg1) {
-    func_80B13AA0(this, arg1, func_80B16854);
-    this->eyeIndex = 2;
+void EnTa_SetActionFuncWithBackToSleepAnimation(EnTa* this, EnTaActionFunc actionFunc) {
+    EnTa_SetupAction(this, actionFunc, EnTa_AnimSitSleeping);
+    this->eyeIndex = TALON_EYE_INDEX_CLOSED;
```
fig02, line 603: maybe for these two we can say `EnTa_SetupActionWithSleepAnim`, and `EnTa_SetupActionWithWakeUpAnim`
this behaves very much like a normal "setup action" wrapper we see in other actors, so I dont think we need to say we are "setting the action func" in the name directly

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -788,148 +866,153 @@ void func_80B1585C(EnTa* this, PlayState* play) {
...
         Animation_Change(&this->skelAnime, &gTalonSitHandsUpAnim, 1.0f, 1.0f,
                          Animation_GetLastFrame(&gTalonSitHandsUpAnim), ANIMMODE_ONCE, 0.0f);
-        this->unk_2CC = 50;
-        func_80088B34(0x1E);
-        func_800F5ACC(NA_BGM_TIMED_MINI_GAME);
-        this->unk_2E0 |= 0x200;
+        this->timer = 50;
+
+        func_80088B34(30);                     // Set the minigame timer for 30 seconds
+        func_800F5ACC(NA_BGM_TIMED_MINI_GAME); // Start minigame music
```
fig02, line 889: imo no need to comment on these, only cause it will just be something to remove later when the functions are properly named. I agree it helps with understanding short term, but thinking longer term this will be self explanatory

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -941,192 +1024,205 @@ void func_80B161C0(EnTa* this, PlayState* play) {
...
                     break;
-                case 0x208B:
-                    func_80B14FAC(this, func_80B15FE8);
+                case 0x208B: // Play cucco game or buy milk
+                    EnTa_SetActionFuncWithWakeUpAnimation(this, EnTa_WaitBuyMilkOrPlayCuccoGameResponse);
                     break;
                 default:
-                    func_80B14FAC(this, func_80B16364);
+                    EnTa_SetActionFuncWithWakeUpAnimation(this, EnTa_ContinueTalkInLonLonHouse);
                     break;
```
fig02, line 1113: we tend to add line breaks inbetween cases with breaks, it makes it easier to read. you may choose to do that now, or if not we will probably do a larger cleanup of this in the whole codebase at some point

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -941,192 +1024,205 @@ void func_80B161C0(EnTa* this, PlayState* play) {
...
-                this->actionFunc = func_80B1642C;
+            case 0x2088: // Later win
+                this->actionFunc = EnTa_TalkAfterCuccoGameWon;
                 break;
         }
+        // Disable immediate automatic talk
         this->actor.flags &= ~ACTOR_FLAG_16;
     } else {
+        // Enable immediate automatic talk
         this->actor.flags |= ACTOR_FLAG_16;
```
fig02, line 1137: same reasoning as my previous comment, I think this can just left uncommented on for the short term (hopefully we get actor flag docs in this year copium)

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -1138,46 +1234,56 @@ void EnTa_Update(Actor* thisx, PlayState* play) {
...
 
-    if (this->unk_2E0 & 0x8) {
-        this->unk_2E0 &= ~0x8;
-    } else if ((limbIndex == 8) || (limbIndex == 10) || (limbIndex == 13)) {
+    // Rocking/wobbling animation for the torso and arms
+    if (this->stateFlags & TALON_STATE_FLAG_SUPPRESS_ROCKING_ANIM) {
+        // TALON_STATE_FLAG_SUPPRESS_ROCKING_ANIM might have been supposed to prevent
+        // the rocking in some situations, (e.g. while sleeping).
+
+        // But because this function is run first for the root limb (limbIndex 1),
```
fig02, line 1280: really small thing but imo dont need a line break between those two

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -1188,15 +1294,15 @@ s32 EnTa_OverrideLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Vec3f* po
 }
 
 void EnTa_PostLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Vec3s* rot, void* thisx) {
-    static Vec3f D_80B16E7C = {
+    static Vec3f headOffset = {
         1100.0f,
         1000.0f,
         0.0f,
     };
```
fig02, line 1301: if you remove the comma on the last line and re run the formatter it will put it all in one line which looks a bit nicer in this context

src/overlays/actors/ovl_En_Ta/z_en_ta.c
fig02: I think at the very least, if we are going with this strategy of giving these flags aliases depending on context, they should be defined right next to the original definition of the flag, as opposed to locally in the actor. That makes it pretty hard to track things down imo.

fig02 (commented):
very sorry for the late review on this, been getting back into the swing of things after a few weeks off

src/overlays/actors/ovl_En_Ta/z_en_ta.c
Dragorn421: copium

src/overlays/actors/ovl_En_Ta/z_en_ta.c
luryus: Personally I prefer the aliases, because the alternatives (naming the multi-purpose flags something like `EVENTINF_TALON_CUCCO_GAME_WON_OR_GERUDO_ARCHERY_WHATEVER` or just using a generic name with comments explaining its use) are worse and more difficult to read.

I will move these aliases to z64save.h. Or I can just revert these to the original names if the aliases are not something that you want to introduce right now

fig02: Our handling and even the naming scheme of these flags is not something thats very established yet. So it will take time for us to settle on something we like

It has its pros and its cons, the aliases make the code nicer to read, but also make it easier for one to not realize that the flags are shared (and there are many bugs in the game because of accidental shared flag usage, so it would be nice to help prevent future users of the codebase from making the same mistakes)

But yea im fine with an alias for now as long as they are defined in a centralized place

fig02 (approved):
apologizes again for how long this took to get in. its a pretty big actor with alot to review, and also as mentioned, a few of us took a bit of a break from reviewing things for a few weeks lol
great job!

src/overlays/actors/ovl_En_Ta/z_en_ta.c
```diff
@@ -252,77 +295,78 @@ s32 func_80B142F4(EnTa* this, PlayState* play, u16 textId) {
...
     } else {
-        func_80B13AA0(this, func_80B14CAC, func_80B167C0);
-        SET_EVENTCHKINF(EVENTCHKINF_6A);
+        EnTa_SetupAction(this, EnTa_IdleAwakeInKakariko, EnTa_AnimRepeatCurrent);
+        SET_EVENTCHKINF(EVENTCHKINF_TALON_WOKEN_IN_KAKARIKO);
     }
 }
 
-void func_80B1448C(EnTa* this, PlayState* play) {
+void EnTa_TalkAfterAwakeUp(EnTa* this, PlayState* play) {
```
Dragorn421, line 326: "awake up" :sweat_smile:

fig02: lmfao 
maybe it can be TalkAwake, as opposed to trying to phrase it "talk after doing something"?

luryus: Great search-replace here... 😅 

Naming this is a bit tricky: this is just an intermediate state in the wake-up sequence, so TalkAwake would be confusing with the "fully awake" states already named `TalkAwakeIn.*` imo. The sequence goes something like this:

- `EnTa_WakeUp`: plays wake up animation, sets up `EnTa_TalkWakingUp`
- `EnTa_TalkWakingUp`: Talking eyes closed, but already standing up. Sets up `EnTa_TalkAfterWokenUp`
- `EnTa_TalkAfterWokenUp`: Talking standing up, eyes half-open. Sets up the "fully awake" actions (`EnTa_SetupAwake`)

The previous name `EnTa_TalkAfterWokenUp` was not great either. `TalkWakingUp2`? `TalkAlmostAwake`? 😄

Dragorn421: I have no idea
after a lot of time thinking about it, huh...
`EnTa_TalkWakingUp` -> `EnTa_TalkWokenUp`
`EnTa_TalkAfterWokenUp` -> `EnTa_TalkWakingUp`
:shrug:

fig02: tbh WakingUp1 and WakingUp2 kinda sounds better to me than trying to give them unique descriptions. it sounds like its part of the same process of waking up to me

PR #1349, opened 2022-08-15
TwoHeadArena and TwoHeadGfxArena docs

Thar0:
In preparation for upcoming `graph.c` docs

Also introduces an `ALIGNOF` macro for abstracting the various different ways supported by compilers for getting the byte alignment of a type.


src/code/TwoHeadArena.c
```diff
@@ -1,140 +1,135 @@
...
 
 void* THA_GetTail(TwoHeadArena* tha) {
     return tha->tail;
 }
 
-void* THA_AllocStart(TwoHeadArena* tha, u32 size) {
+/**
+ * Allocates to the head of the Two Head Arena. The allocation will not have any alignment guarantees.
+ */
+void* THA_AllocStart(TwoHeadArena* tha, size_t size) {
```
fig02, line 29: i havent done a thorough review at all, just during my initial skim:
I read this as "start the allocation" even tho i know the intention is start and end/front and back
we should prob use head and tail? since we're already using those terms in the functions above

fig02: I also realize it was already named this, just never noticed before i guess

Thar0: I wasn't too sure what sort of direction to take names in so if you spot anything else that wasn't renamed but probably should be that'd be great

src/code/TwoHeadArena.c
```diff
@@ -1,140 +1,135 @@
+/**
+ * @file TwoHeadArena.c
+ *
+ * This file implements a simple double-ended stack allocator for general purpose.
+ *
+ * A double-ended stack allocator accepts allocations at either the "head" or "tail" of it's allotted memory region.
```
Dragorn421, line 6:
```suggestion
 * A double-ended stack allocator accepts allocations at either the "head" or "tail" of its allotted memory region.
```

src/code/TwoHeadArena.c
```diff
@@ -1,140 +1,135 @@
...
     void* start = tha->head;
 
-    tha->head = (void*)((u32)tha->head + size);
+    tha->head = (u8*)tha->head + size;
     return start;
 }
 
-void* THA_AllocStart1(TwoHeadArena* tha) {
-    return THA_AllocStart(tha, 1);
+void* THA_AllocHead1(TwoHeadArena* tha) {
```
Dragorn421, line 36: My first impression from the name was this is a duplicate of AllocHead, maybe `AllocHeadByte` or `AllocByteHead` would be clearer? (Kinda like allocGfx and others on thga)

src/code/TwoHeadArena.c
```diff
@@ -1,140 +1,135 @@
...
+    return THA_AllocHead(tha, 1);
 }
 
-void* THA_AllocEnd(TwoHeadArena* tha, u32 size) {
-    u32 mask;
+/**
+ * Allocates to the tail end of the Two Head Arena. The allocation will be aligned based on the size of the allocation.
+ * For all allocations greater than 16-bytes, they will be aligned to 16-bytes. Otherwise, the alignment will be the
```
Dragorn421, line 42:
```suggestion
 * All allocations of 16-bytes or more, will be aligned to 16-bytes. Otherwise, the alignment will be the
```
(idk)

src/code/TwoHeadArena.c
```diff
@@ -1,140 +1,135 @@
...
     } else if (size == 2 || size == 6 || size == 10 || size == 12 || size == 14) {
-        mask = ~1;
+        // Align 2 for multiples of 2
+        mask = ALIGN_MASK(2);
+    } else if (size >= 0x10) {
+        // Align 0x10 for allocations greater than 0x10
+        mask = ALIGN_MASK(0x10);
     } else {
-        mask = (size >= 0x10) ? ~0xF : 0;
+        //! @bug if size is 1 the computation below will give NULL. The mask for this case would be
```
Dragorn421, line 62:
```suggestion
        //! @bug if size is less than 16 and odd or zero the computation below will give NULL. The mask for this case would be
```
I think. Not sure on the wording but my point is it also happens in those cases?

src/code/TwoHeadArena.c
```diff
@@ -1,140 +1,135 @@
...
     return tha->tail;
 }
 
-void* THA_AllocEndAlign16(TwoHeadArena* tha, u32 size) {
-    u32 mask = ~0xF;
+/**
+ * Allocates to the tail end of the Two Head Arena with guaranteed 16-byte alignment.
+ */
+void* THA_AllocTailAlign16(TwoHeadArena* tha, size_t size) {
+    uintptr_t mask = ALIGN_MASK(0x10);
```
Dragorn421, line 75: I find it weird to have 16 in the name and 0x10 in the code, but it's not a big deal and I understand why we either don't want 0x10 in the function name, or "hide" that 16 has a particular bit layout

src/code/TwoHeadArena.c
```diff
@@ -1,140 +1,135 @@
...
+ * Gets the remaining size of the Two Head Arena
+ *
+ * @return Remaining size. A negative number indicates an overflow.
+ */
+s32 THA_GetRemaining(TwoHeadArena* tha) {
+    return (s32)((u8*)tha->tail - (u8*)tha->head);
 }
 
+/**
+ * Returns true if the Two Head Arena has overflown, false otherwise
```
Dragorn421, line 106:
```suggestion
 * @return true if the Two Head Arena has overflowed, false otherwise
```

src/code/TwoHeadArena.c
```diff
@@ -1,140 +1,135 @@
...
+    tha->head = tha->start;
+    tha->tail = (u8*)tha->start + tha->size;
 }
 
-void THA_Ct(TwoHeadArena* tha, void* ptr, u32 size) {
-    tha->bufp = ptr;
+/**
+ * Creates a new Two Head Arena at `start` with available size `size`
+ */
+void THA_New(TwoHeadArena* tha, void* start, size_t size) {
```
Dragorn421, line 124: Should we call this Init ?

Dragorn421 (commented):
Great job, thanks a lot for cleaning this up 🚀

Is the union between THA and THGA-anonymous-struct "safe"? (is it well defined that it corresponds in this case to "casting pointers"?)

Is it worth mentioning somewhere that all the things like `POLY_OPA_DISP++` are effectively dword allocations on thga heads?

src/code/TwoHeadArena.c
mzxrules: drop the comma after `or more`

src/code/TwoHeadArena.c
mzxrules: seems that way, though I question whether a size less than 1 should be considered a valid argument for the function.

src/code/TwoHeadArena.c
```diff
@@ -1,140 +1,135 @@
+/**
+ * @file TwoHeadArena.c
+ *
+ * This file implements a simple double-ended stack allocator for general purpose.
+ *
```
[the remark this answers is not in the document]
Dragorn421:
```suggestion
 *
 * This file implements a simple general purpose double-ended stack allocator.
 *
```
? 5Head
or "generic" :shrug:

src/code/TwoHeadGfxArena.c
```diff
@@ -0,0 +1,102 @@
+/**
+ * @file TwoHeadGfxArena.c
+ *
+ * This file implements a particular use of the double-ended stack allocator from TwoHeadArena.c for graphics data.
+ *
+ * Display list commands are allocated from the head while other graphics data such as matrices and vertices are
+ * allocated from the tail end.
+ *
```
Dragorn421, line 8: There's whitespace at the end of this line and somewhere else (run formatter)

Dragorn421 (commented):
Good for me besides open comments (`THA_New` vs `THA_Init`, fig's, format)

include/tha.h
```diff
@@ -0,0 +1,28 @@
...
+#define THA_H
+
+#include "ultra64.h"
+#include "alignment.h"
+
+typedef struct {
+    /* 0x0000 */ size_t size;
+    /* 0x0004 */ void*  start;
+    /* 0x0008 */ void*  head;
+    /* 0x000C */ void*  tail;
```
hensldm, line 11:
```suggestion
    /* 0x00 */ size_t size;
    /* 0x04 */ void*  start;
    /* 0x08 */ void*  head;
    /* 0x0C */ void*  tail;
```

src/code/TwoHeadArena.c
hensldm: I think `Init` makes sense, but also don't mind it staying `New` as well.

hensldm (approved):
1 small nit

src/code/speed_meter.c
```diff
@@ -186,25 +186,26 @@ void SpeedMeter_DrawAllocEntries(SpeedMeter* meter, GraphicsContext* gfxCtx, Gam
     }
 
     thga = (TwoHeadGfxArena*)&state->tha;
-    SpeedMeter_InitAllocEntry(&entry, thga->size, thga->size - THA_GetSize((TwoHeadArena*)thga),
+    //! @bug THA_GetRemaining call should be THGA_GetRemaining like the others below, harmless as-is
+    SpeedMeter_InitAllocEntry(&entry, thga->size, thga->size - THA_GetRemaining(&thga->tha),
```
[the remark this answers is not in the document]
Thar0: I argue that it is a bug since suppose the functionality of `THGA_GetRemaining` were extended in any way, this could be a harmful difference. I think it's useful to point this out as a bug comment

Dragorn421 (approved):

fig02 (approved):

PR #1345, opened 2022-08-10
Document Ganon objects

tom-overton:
This PR documents:
- object_ganon
- object_ganon2
- object_ganon_anime1
- object_ganon_anime2
- object_ganon_anime3

It also updates various files to use limb enums as expected (this is a perfect case for putting the limb enum in the object, btw).

Thanks, as always, to @emilybrooks for helping me with some tricky Japanese translations.

PR #1353, opened 2022-08-20
Document object_gndd (Young Ganondorf)

tom-overton:
A companion piece to #1345. I documented the younger Ganondorf's object as well, since many textures were similar or identical to the older Ganondorf.


fig02 (approved):

Roman971 (approved):
Some of these names are too long imo (eg. `gYoungGanondorfHorsebackMagicChargeUpStartAnim`), but I guess we will just have to consider using shorter prefixes or omitting some terms

tom-overton:
I suppose I could omit "magic" from that one, since he doesn't charge anything else up. What do you think, @Roman971 ?

fig02:
it could possibly be mistaken for the horse "charging", like running away/charging, dunno
its probably fine to omit it, but yeah. overall I think longer anim names arent a *huge* issue, cause they are usually only referenced a few times

tom-overton:
Ah, yeah, I didn't think about that, but you're right. And while Phantom Ganon is a completely separate object, he *does* charge at you on horseback. I suppose I'll keep the existing name for now.

PR #1354, opened 2022-08-21
`GameInfo` -> `RegEditor`

Dragorn421:
Follow-up to #1290
Also name some regs-related functions with the `Regs_` "system name"

The names may benefit discussions, comments are welcome


Dragorn421:
I would agree 100% if the actual regs' data wasn't stored inside `DebugRegs` (`DebugRegs.data`)

Roman suggested `GlobalRegs` in the past, idk about "global" since their values are most often initialized on the state initialization (instead of persisting, which I'd kinda expect but idk) but maybe that would be good, or use another similar adjective to prefix `Regs`
(or just `Regs` but meh)

Roman971:
The reason I suggested `GlobalRegs` is because these regs are accessible and usable from anywhere so they are effectively global, and they contain all sort of information that may have an impact on the global behavior of the game.
There are probably other good options, but tbh I don't think the name is very important since the struct itself is never used externally. `RegsContext` would be good enough as well for that reason, imo

I also don't see any issue with using "context" when it makes sense. It's a pretty standard term used in many programming languages, librairies and systems.

fig02 (approved):

Roman971 (approved):

fig02:
will merge tomorrow, incase there are anymore thoughts on these names

src/code/game.c
```diff
@@ -58,7 +58,7 @@ void GameState_SetFBFilter(Gfx** gfx) {
     *gfx = gfxP;
 }
 
-void func_800C4344(GameState* gameState) {
+void GameState_UpdateRegs(GameState* gameState) {
```
mzxrules, line 61: UpdateRegs doesn't really capture the purpose of this function, though I imagine it's hard to give a good name for this function since it performs various seemingly unrelated data logging tasks and then also updates the RegEditor.

mzxrules:
For me, my intuition is that it doesn't feel right to call the struct a Context, but I find it hard to make a convincing argument as I don't have a clear picture of what justifies/eliminates naming something a Context. Or another way to think of it, what justifies elevating state (i.e. a struct) to the term of context.

Additionally, the thought occurs to me that the struct should really be named `RegEditor`, with the justification being that none of the struct's members are ever directly accessed outside of the z_debug.c scope.

src/code/game.c
Roman971: Yeah this function seems to do various debug things, including logging stuff and setting the log verbosity on a bunch of systems, so maybe it should be named around that (eg. `GameState_DebugUpdate`)

Dragorn421: I reverted the name back to `func_800C4344`, there are other debug functions to name, may as well do them later

Dragorn421:
> none of the struct's members are ever directly accessed outside of the z_debug.c scope

What about the `*REG` macros?

fig02:
i imagine by directly accessed he means not hidden by a macro

RegEditor is fine with me personally, since that is what all of the things in this struct are. stuff that make up the editor. but i am also fine with context, so whichever we have more of a consensus on

Dragorn421:
Is it really okay to name RegEditor a struct that holds data that is functionnally needed for the game's normal (non-debug) operation?

fig02:
I think its more of a mistake on their part that they left critical non-debug values in regs. Its entire purpose is to be displayed in the editor and editable by the controller.

But yea, Im still fine with context

Dragorn421:
Needs solving:
`gGameInfo`'s new name: `gRegsContext` vs `gRegEditor`

AngheloAlf:
I vote for `gRegEditor`, kinda tired of seeing `Context` everywhere

fig02:
do note that the name for this specific thing doesnt matter too much in regards to seeing "context" everywhere, cause all accesses are behind a macro. but i do understand your point in general

Im personally fine with `gRegEditor`. I think its fair to feel iffy about it, but it makes sense to me that the structure is the data itself (all the regs) and the other moving parts of the editor that make it work as a whole

Dragorn421:
So, `GameInfo* gGameInfo;` -> `RegEditor* gRegEditor;` ?

fig02:
i guess so, 3 people in this thread including myself have said they are fine with it.
If there is anyone who doesnt agree with the name nows the time to speak up

Dragorn421:
Now `gRegEditor`

image

fig02 (approved):

engineer124 (approved):
lgtm

Roman971:
I think `RegsContext` (or `GlobalRegs`) is much more accurate than `RegEditor` since the struct also includes the actual regs data which is relevant regardless of the editor, including in retail roms where the editor can't be used at all.
Tho again it's not too big of a deal since the name isn't used directly when accessing regs, but I feel like we might as well go for something accurate if we're renaming it

fig02:
being editable by the editor is the main point of it existing at all. They did leave important values in there even in retail yeah, but it doesnt change that its a part of the overall reg editor system imo

hensldm (approved):

PR #1356, opened 2022-08-21
Lightweight trigonometry doc

Dragorn421:
Very simple documentation clearing up what the various trigonometry functions do / take in / output (units like radians/binang and ranges)


src/code/code_800FCE80.c
```diff
@@ -132,6 +140,11 @@ f32 Math_FAtanF(f32 x) {
     }
 }
 
+/**
+ * @param y
+ * @param x
```
fig02, line 145: feel like these param comments can just be omitted especially cause they arent adding any explanation or additional context. goes for most of these functions

fig02 (approved):

src/code/sys_math_atan.c
```diff
@@ -77,24 +77,32 @@ static u16 sATan2Tbl[] = {
...
+            ret = sAtan2Tbl[tblIdx];
         }
     }
 
     return ret;
 }
 
+/**
+ * @return angle from (1,0) to (x,y) as binang, in [-0x8000,0x7FFF] range
+ */
```
mzxrules, line 105: This might deserve an additional warning to note that the argument order is swapped from how atan2 is typically defined in other languages. Notably, the equivalent function in MM even swaps the argument order back to y,x.

Also, you have the same (1,0) thing.

src/code/sys_math_atan.c
```diff
@@ -107,29 +115,33 @@ s16 Math_Atan2S(f32 x, f32 y) {
...
         }
     }
-    return ret;
+
+    return (s16)ret;
 }
 
+/**
+ * @return angle from (1,0) to (x,y) in radians, in [-pi,pi) range
+ */
```
mzxrules, line 144: same as above

src/code/code_800FCE80.c
```diff
@@ -132,6 +139,9 @@ f32 Math_FAtanF(f32 x) {
     }
 }
 
+/**
+ * @return angle from (1,0) to (x,y) in radians, in (-pi,pi] range
```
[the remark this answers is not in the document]
Dragorn421: Sorry for the late response, I must have missed the notification

No it's not a typo, it's refering to the unit circle with (1,0) being at 1 on the x axis (or maybe more simply the angle from the vector (1,0) to the vector (x,y))

The range is `(-pi;pi]` as in, it could be `[0;2pi)` instead (or maybe the angle could be unoriented and it'd be `[0;pi]`), I'm not sure what to say
It's trying to clear up the unit used (radians), it being oriented, and the range of the principal value used here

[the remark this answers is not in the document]
mzxrules: ah, I didn't consider that (1,0) would be an orientation vector. I usually conceptualize atan2 as being the tool you use to find the angle between two points, so I interpreted `angle from (1,0)` to be expressing the point of origin for calculations.

As for the range, was more just asking for clarification for why the range of return values is expressed as `(-pi, pi]`, but I've worked it out to be correct.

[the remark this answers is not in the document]
Dragorn421: So is this fine then?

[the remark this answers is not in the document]
mzxrules: i would suggest making it clearer that (1,0) isn't a point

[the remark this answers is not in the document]
Dragorn421: like
```suggestion
 * @return angle from vector (1,0) to (x,y) in radians, in (-pi,pi] range
```
?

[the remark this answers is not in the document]
mzxrules: yea, think that's good enough

fig02:
(mzx never approved this but he said it was fine in discord. so this needs either 1 more lead or 1 more contributor for merge)

src/libultra/gu/sinf.c
```diff
@@ -14,6 +14,10 @@ static const du pilo = { 0x3E6110B4, 0x611A6263 };
 
 static const fu zero = { 0x00000000 };
 
+/**
+ * @param x radians
```
hensldm, line 18: For some the arg is `x` while others it is `angle` (at first I thought it was `x` if it was radians and `angle` for binary angles, but that is also not the case). Feel like these could/should be standardized.

src/libultra/gu/sins.c
```diff
@@ -2,6 +2,10 @@
 
 #include "sintable.inc.c"
 
+/**
+ * @param x binang
+ * @return sin(x)*0x7FFF
```
hensldm, line 7: I'll admit, I get where the `*0x7FFF` is coming from, but it seems kind of strange and did take me a little bit to think it through. I like what you did in `Math_GetAtan2Tbl`, so perhaps using `@return sin(x) as binang` instead?

Dragorn421: sin() doesn't return a binang :sunglasses: 
this is just a sin() value but since it's an integer value, it's scaled to go from `-short_max` to `short_max`

src/libultra/gu/sinf.c
Dragorn421: Thoughts on `Math_CosF(f32 rad)`, `Math_CosS(s16 binang)` then? sounds like a good idea

hensldm: Fine by me.

src/libultra/gu/sins.c
hensldm: Oh right, `sin` doesn't return an angle 🤦. I'm fine with this than.

src/code/code_800FCE80.c
```diff
@@ -132,6 +139,9 @@ f32 Math_FAtanF(f32 x) {
     }
 }
 
+/**
+ * @return angle from vector (1,0) to (x,y) in radians, in (-pi,pi] range
```
hensldm, line 143: Thoughts on maybe
`@return angle between (x,y) and the x-axis in radians, in (pi,pi] range`?

Dragorn421: I'm not convinced idk
The orientation ((1,0) to (x,y)) is less clear and "x axis" isn't explicitly oriented

But suggestions on phrasing this are welcome, mine fits how I look at it but there are probably people out there who wouldn't get what this means

hensldm: Perhaps just swapping `to (x,y)` and `from vector (1,0)`, like 
`@return angle to (x,y) from vector (1,0) in radians, in (-pi,pi] range`
cause at the moment, when I first read it, I imagine it is describing the vector that starts at `(1,0)` and points to `(x,y`) which I'm pretty sure is not what you want. This would clear that up as really only `(1,0)` describes the vector.

Dragorn421: Maybe adding "around (0,0)" somewhere could help with that?

hensldm: Possibly, I would need to see an example. (tbh this isn't too big a deal and might just be a me problem).

fig02: anons suggestion of `@return angle to (x,y) from vector (1,0) in radians, in (-pi,pi] range`
makes it easier for me to follow along, at least

Dragorn421:
```suggestion
 * @return angle to (x,y) around (0,0) from vector (1,0) in radians, in (-pi,pi] range
```
```suggestion
 * @return angle to (x,y) from vector (1,0) around (0,0) in radians, in (-pi,pi] range
```

hensldm: I like the second one better :)

fig02: yea thats fine with me as well

fig02 (approved):

hensldm (approved):
LGTM :)

PR #1357, opened 2022-08-21
Docker fixes

luryus:
Noticed a few issues when setting up my build environment last week:

- The Docker image build fails because Ubuntu 21.10 is EOL and its apt repos are unavailable
  - FIxed by updating the base image to Ubuntu 22.04
- With the updated Ubuntu image, ZAPD build is breaks due to a git error. The new Ubuntu image ships with a newer version of Git that has the directory ownership security fix (https://github.blog/2022-04-12-git-security-vulnerability-announced/).
  - Fixed by marking `/oot` as a safe directory during the image build
- There's no dockerignore, so the entire repo is copied to the image build context. That takes a long time.
  - Fixed by adding a `.dockerignore` that ignores everything (building the Docker image requires nothing from the repo).


Dragorn421 (approved):
:see_no_evil: (I mean I don't know Docker so I can't really speak for these changes but it seems reasonable)

fig02 (approved):

Roman971 (approved):

PR #1359, opened 2022-08-21
Generic actor params getters

Thar0:
Creates generic params getter macros that can be built upon to create specific macros for extracting values stored in the params field.
There are unfortunately quite a number of distinct patterns that must be met, so there are 6 different getters:
 - Mask then shift
 - Shift then mask
 - No mask, only shift
 - No shift, only mask
 - Shift then mask to a bit position other than 0
 - Mask then shift to a bit position other than 0

The last two are very rare, are they worth having?

There is also one case that I could not turn into a macro, it was a mask by 0x33 (0b110011). Since this bit range is not continuous and all of the getter macros take a number of bits rather than the mask itself, this mask could not be constructed.


Dragorn421:
Bump on the comment

> Do we still want to use those new macros for single bit checks (flags)?

which was in reaction to

```diff
-    invisble = this->actor.params & SPAWN_INVISIBLE;
+    invisble = PARAMS_GET_NOSHIFT(this->actor.params, 15, 1); // SPAWN_INVISIBLE
```

Personally I'm waiting on solving that to continue reviewing as this comes up often-ish

(also many conflicts in here... I hope it's manageable :fearful:)

src/overlays/actors/ovl_En_Floormas/z_en_floormas.c
```diff
@@ -133,7 +133,7 @@ void EnFloormas_Init(Actor* thisx, PlayState* play2) {
     Collider_SetCylinder(play, &this->collider, &this->actor, &sCylinderInit);
     CollisionCheck_SetInfo(&this->actor.colChkInfo, &sDamageTable, &sColChkInfoInit);
     this->zOffset = -1600;
-    invisble = this->actor.params & SPAWN_INVISIBLE;
+    invisble = PARAMS_GET_NOSHIFT(this->actor.params, 15, 1); // SPAWN_INVISIBLE
```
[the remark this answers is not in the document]
hensldm: I think they still make sense for single bit checks when they are a part of params (they are still bitpacked).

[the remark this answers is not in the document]
Dragorn421: So to take an actual example from the PR
```c
#define ENDOOR_PARAMS_DOUBLE_DOOR_FLAG 0x40
#define ENDOOR_IS_DOUBLE_DOOR(thisx) ((thisx)->params & ENDOOR_PARAMS_DOUBLE_DOOR_FLAG)
```
becomes
```c
#define ENDOOR_PARAMS_DOUBLE_DOOR_FLAG          PARAMS_MAKE_MASK(6, 1)
#define ENDOOR_IS_DOUBLE_DOOR(thisx)            PARAMS_GET_NOSHIFT((thisx)->params, 6, 1)
```

I guess that's fine? I find it weird that a single bit check ends up abstracted as the more complicated "get one-bit-wide mask without shifting" but it's fine.
I can go with that.

src/overlays/actors/ovl_En_Go2/z_en_go2.c
```diff
@@ -32,13 +32,13 @@ INFTABLE_10F - (not on cloud modding)
 INFTABLE_11E - Bomb bag upgrade obtained from rolling Goron
 
 EnGo
-pathIndex: this->actor.params & 0xF
-Goron: this->actor.params & 0xF0
+pathIndex: PARAMS_GET2(this->actor.params, 0, 4)
```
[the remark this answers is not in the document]
Thar0: idk it was there before I started, if it's not supposed to be there I can remove it or it can be removed separately?

src/overlays/actors/ovl_En_Floormas/z_en_floormas.c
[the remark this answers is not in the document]
Thar0: imo all params accesses should aim to go through macros as much as possible, possibly besides weird one-off access patterns

hensldm:
Is it still the plan for actor's to make their own macros, with descriptive names, that just build off these ones?

include/z64actor.h
```diff
@@ -581,4 +578,43 @@ typedef struct {
...
+    ((1 << (n)) - 1)
+
+// Extracts the `n`-bit value at position `s` in `p`, shifts then masks
+// No possibility of sign extension
+#define PARAMS_GET(p, s, n) \
+    (((p) >> (s)) & NBITS_TO_MASK(n))
+
+// Extracts the `n`-bit value at position `s` in `p`, masks then shifts
+// Possibility of sign extension
+#define PARAMS_GET2(p, s, n) \
```
hensldm, line 593: More so just a thought, not how good or bad it really is. Wonder if it would be better to name this `PARAMS_GET_SIGN` instead to show it has the possibly to sign extend?

fig02:
Not that this dictates what we do, but Ive been learning about tww/tp lately, and they ended up taking the generic approach in those games:
```c
inline u32 fopAcM_GetParamBit(void* ac, u8 shift, u8 bit) {
    return (fopAcM_GetParam(ac) >> shift) & ((1 << bit) - 1);
}
```
(also, they actually called them param(s) in those games which I did not know. pretty neat)

Thar0:
The intention I had going into this is that these generic params macros introduced in this PR would be used to build specific macros for individual actors as they are documented.

There are already some examples of this in this PR, from actors that had already received some params documentation prior. For example `z_obj_switch` defines:
```c
#define OBJSWITCH_TYPE(thisx) PARAMS_GET((thisx)->params, 0, 3)
#define OBJSWITCH_SUBTYPE(thisx) PARAMS_GET((thisx)->params, 4, 3)
#define OBJSWITCH_SWITCH_FLAG(thisx) PARAMS_GET((thisx)->params, 8, 6)
#define OBJSWITCH_FROZEN(thisx) PARAMS_GET((thisx)->params, 7, 1)
```
And then uses the specific macros.

fig02:
(on the topic of that example, Id like to standardize having `_GET_` in the name of the specific macro, which would match the generic name as well)

overall I am okay with that approach though

include/z64actor.h
Thar0: I was thinking of naming the variants `GET_SIGNED` and `GET_UNSIGNED`, or `GET_S` and `GET_U` and keep the comment to keep them relatively short.

fig02: im fine with the whole word, personally

include/z64actor.h
```diff
@@ -581,4 +578,43 @@ typedef struct {
...
+// Converts a number of bits to a bitmask, helper for params macros
+// e.g. 3 becomes 0b111 (7)
+#define NBITS_TO_MASK(n) \
+    ((1 << (n)) - 1)
+
+// Extracts the `n`-bit value at position `s` in `p`, shifts then masks
+// No possibility of sign extension
+#define PARAMS_GET(p, s, n) \
+    (((p) >> (s)) & NBITS_TO_MASK(n))
+
```
fig02, line 590: feel like we can prob afford the input vars to have full names, just so theres no ambiguity. incase someone forgets the order of the args they can hover over it and be reminded very easily 

at least, id like if `p` was the whole word "params". I think its fine if s and n are single letters as they commonly are in things like gfx macros

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.c
```diff
@@ -10,10 +10,11 @@
...
 
-#define OBJSWITCH_TYPE(thisx) ((thisx)->params & 7)
-#define OBJSWITCH_SUBTYPE(thisx) (((thisx)->params >> 4) & 7)
-#define OBJSWITCH_SWITCH_FLAG(thisx) (((thisx)->params >> 8) & 0x3F)
-#define OBJSWITCH_FROZEN(thisx) (((thisx)->params >> 7) & 1)
+#define OBJSWITCH_TYPE(thisx) PARAMS_GET((thisx)->params, 0, 3)
+#define OBJSWITCH_SUBTYPE(thisx) PARAMS_GET((thisx)->params, 4, 3)
+#define OBJSWITCH_SWITCH_FLAG(thisx) PARAMS_GET((thisx)->params, 8, 6)
+#define OBJSWITCH_FROZEN(thisx) PARAMS_GET((thisx)->params, 7, 1)
+
```
fig02, line 17: if we can, for any already existing actor-specific macros, would like them to say `_GET_` in them so we can standardize that now

include/z64actor.h
hensldm: Yeah whole word seems fine, especially since most if not all usages of these are going to be in headers anyways behind other macros.

include/z64actor.h
Dragorn421: I wouldn't mind if the arguments were named the same as in `_SHIFTL(v, s, w)` actually

include/z64actor.h
Dragorn421: I'd prefer using signed/unsigned over calling one "2"

and would be fine with either `GET_SIGNED`/`GET_UNSIGNED` or `GETS`/`GETU` or `GET_S`/`GET_U`

src/overlays/actors/ovl_Obj_Switch/z_obj_switch.c
Dragorn421: what about booleans like "frozen" here, should we still use `_GET_` or `_IS_` ?
another example is `ENDOOR_IS_DOUBLE_DOOR`, should it be `ENDOOR_GET_IS_DOUBLE_DOOR` or?

fig02: id say it depends on how its named. if you imagine the "set" macro that actually sets the params in the actor list:
`OBJSWITCH_PARAMS(type, subtype, switchFlag, isFrozen)`
I think it should be `_GET_<NAME>`, whatever the variable name is

fig02:
I feel like its about time to get this PR in. It will help greatly with actor documentation.
Ive gone ahead and made all requested changes, with the permission of tharo. My changes summarized below:
- All existing *specific* params macros are now of the form `<actor>_GET_<param-name>`.
- `PARAMS_GET` -> `PARAMS_GET_U`
- `PARAMS_GET2` -> `PARAMS_GET_S`
- some uneeded comments were removed

This change was not requested above in review, but I made it on my own (in communication with tharo):
- The macro formerly named `PARAMS_GET_S` has been removed. It was a special case used in one actor
- `PARAMS_GET2_S` has been removed. It was a special case used in one actor
I dont feel the generalized macro system needs to support this one-off use cases. It keeps the generalized system more clear to only need to worry about `PARAMS_GET_S` and `PARAMS_GET_U`

I tried to split my commits to make it easier to review.

krm01 (approved):
LGTM, I was kinda against this until I realized the authors have already done this for every actor, and it wasn't just a prescription for future docs to have to use. since its already been implemented though, LGTM!

fig02 (approved):
This approval is slightly biased since I completed the most recent set of changes. But I'm moreso approving the overall concept and initial implementation :)

src/overlays/actors/ovl_Bg_Mori_Hineri/z_bg_mori_hineri.h
```diff
@@ -6,6 +6,9 @@
 
 struct BgMoriHineri;
 
+// Due to a unique access pattern, this param cannot use the generic "PARAMS_GET_U" macros
+#define TWISTED_HALLWAY_GET_PARAM_15(thisx) (((thisx)->params & (NBITS_TO_MASK(1) << (15))) >> ((15) - (1)))
```
mzxrules, line 10: I hate how overwritten this is.

fig02: Its written in the same form as the generic macros. The point is to easily let you see the bit width and position.

fig02: Its also 1 of 2 cases where a custom actor-specific actor macro is used. Imo we can figure out these edge cases another time, like when the actor is documented

mzxrules: Personally I don't find that the generic macro easy to follow logically in it's raw form. Probably more important though is that the macro isn't a proper params getter; the bit gets extracted but then you have a secondary operation "baked in" of the value being shifted into the bit 1 spot instead of bit 0 like the majority of params getters. This is an important detail that is being obfuscated by the macro, and makes it hard to follow the subsequent logic. 

In this unusual instance, I feel `PARAMS_GET_NOSHIFT` would be the more appropriate way to extract the value. The right shift needs to be exposed to the user.

hensldm: Fwiw I agree with mxz here. PARAMS_GET_NOSHIFT seems better so you can explicitly show the unique shift (probably the same for en_ishi as well which seems to have the same structure). However also agree we could wait until the actors are documented.

hensldm (approved):

src/overlays/actors/ovl_Bg_Mori_Hineri/z_bg_mori_hineri.h
fig02: can either of you write out how it would be used with NOSHIFT? I would be willing to change it now

hensldm: `#define TWISTED_HALLWAY_GET_PARAM_15(thisx) (PARAMS_GET_NOSHIFT((thisx)->params, 15, 1) >> 14)`
Oh wait en_ishi is different. Don't think NOSHIFT works. My bad.

hensldm: en_ishi does work like `PARAMS_GET_NOSHIFT(PARAMS_GET_NOMASK((thisx)->params, 12 - 2), 2, 4)` but they may be a bit too much.

mzxrules: I would do something more like: 

    #define TWISTED_HALLWAY_GET_TYPE_NOSHIFT(x) PARAMS_GET_NOSHIFT(x, 15, 1)
    ...
    this->dyna.actor.params = TWISTED_HALLWAY_GET_TYPE_NOSHIFT(this->dyna.actor.params) >> 14;

Dragorn421: I suggest just reverting this (to main), and we can indeed figure it out (or not) later, but not let it stop the whole thing

Dragorn421 (approved):
My only problem with this PR is `PARAMS_GET_NOSHIFT(params, N, 1)`, I think those should generally just be simple flag checks like `& FLAG` with `#define FLAG (1 << N)`. (when it's the case that it's a flag check: sometimes it's not and `PARAMS_GET_NOSHIFT` is appropriate)

But this is at most with two dozen of actors, so not many, and it can be changed later. I'll try to document those relevant actors myself after this is merged in order to have a clearer view :)

Dragorn421:
how "merge ready" is this?

fig02:
imo this is ready to go, like you mention we can discuss and change some of the edge cases as we document the actors they belong to

Dragorn421:
merging when CI catches up

PR #1362, opened 2022-08-24
Document INF Table

CharlesAverill:
90% of these were documented by digging through textIds, but a few (really just the Ruto Jabu Jabu flags) were pulled from [CloudModding](https://wiki.cloudmodding.com/oot/Save_Format#inf_table), which is pretty accurate for most things but is deserving of more discretion.

I expect some nitpicks on wording for a few of these, just did what I thought was the most readable.


CharlesAverill:
Also looks like a ton of files got touched by the formatter, should I take those out? Didn't mess with the camera or any of that stuff

Dragorn421:
The formatting changes are probably because you are using an "old" version of clang-format, we more-or-less standardized on clang-format-11
So yeah it would be better to not include them, and/or maybe install a newer clang-format

CharlesAverill:
Unstaged unrelated files and added WIP tags to uncertain flags

Dragorn421:
I can't depart from feeling like the flags should be named as part of documenting the individual actors

It would probably be fine to name the flags based on

- previous documentation (CloudModding) despite it being occasionally unreliable,
- and guesses based on associated text ids,

but

- it may omit important details from the usage,
- and actors may later be documented around those names without reconsidering them.

Also reviewing this is super blind, basically read through all the NPC actors or just don't review besides typos in the names :sweat_smile: 

----------

Personally that's why I've been avoiding reviewing this PR, I don't think it would be a good move to merge it.
Good to keep around for reference though.

So all in all I would close it. More thoughts welcome

fig02:
100% agree. I think your notes on what you have so far are a good starting point and should be saved somewhere (keep this branch up even if the PR is closed, for example), but I think we should focus on NPC actor docs before attempting to name these things.

I also know charles youve been busy with other irl things, and a PR like this will need alot of conversation and maintenance before it gets close to being mergeable. 

So, I think waiting for a bit until we have more information to make better decisions, and some more time/motivation from everyone to be able to review and upkeep this will go a long way.
If its alright with you I think the best move is to close the PR.

CharlesAverill:
Totally agree with y’all. I want to come back to the community next summer
when school winds down, I don’t see the possibility of rejoining before
then.

fig02:
sounds good, thank you!

PR #1364, opened 2022-08-25
Doc speed_meter timers and misc. low-ish level stuff

Dragorn421:
- Name timers related to speed_meter
- `GameState.isPreNMIStateRunning`
- Miscellaneous stuff


src/code/graph.c
```diff
@@ -166,36 +174,41 @@ void Graph_TaskSet00(GraphicsContext* gfxCtx) {
...
         gfxCtx->callback(gfxCtx, gfxCtx->callbackParam);
     }
 
-    time = osGetTime();
-    if (D_8016A550 != 0) {
-        D_8016A558 = (D_8016A558 + time) - D_8016A550;
-        D_8016A550 = time;
+    timeNow = osGetTime();
+    if (gAudioThreadUpdateTimeStart != 0) {
+        // The audio thread audio update is running on the audio thread
```
Roman971, line 202: I feel like this would be better as simply `The audio thread update is running` or `Audio update is running on the audio thread`

include/z64.h
```diff
@@ -1065,7 +1065,7 @@ typedef struct GameState {
     /* 0x84 */ GameAlloc alloc;
     /* 0x98 */ u32 running;
     /* 0x9C */ u32 frames;
-    /* 0xA0 */ u32 unk_A0;
+    /* 0xA0 */ u32 isPreNMIStateRunning;
```
Roman971, line 1068: I wonder if we couldn't use a simpler name like `inPreNMI` or `preNMIRunning`/`preNMIActive`

src/code/speed_meter.c
```diff
@@ -1,25 +1,60 @@
...
-volatile OSTime D_8016A550;
-volatile OSTime D_8016A558;
-volatile OSTime gRSPAudioTotalTime;
-volatile OSTime gRSPGFXTotalTime;
-volatile OSTime gRSPOtherTotalTime;
+/**
+ * How much time the audio update on the audio thread (`func_800E4FE0`) took in total, between scheduling the last two
+ * graphics tasks.
+ */
+volatile OSTime gAudioThreadUpdateTimeTotalPerGfxTask;
```
Roman971, line 8: Shouldn't this be `gAudioThreadUpdateTimeTotal` to be consistent with the other timers?

src/code/speed_meter.c
```diff
@@ -1,25 +1,60 @@
...
+/**
+ * How much time elapsed between scheduling the previous graphics task and the current one being ready (the previous
+ * task not necessarily being finished yet), without the amount of time spent on the audio update in the audio thread.
+ */
+volatile OSTime gGfxTaskSentToNextReadyMinusAudioThreadUpdateTime;
```
Roman971, line 13: Imo the "without the amount of time spent on audio" should either be left out of the name or reworded to something that's not a full description in the name itself (tho I don't have a proper suggestion for that atm)

include/z64.h
Dragorn421: Dropping "state" isn't possible imo, otherwise it'd be referencing PreNMI (the period between the PreNMI hardware interrupt and actual reset/NMI) instead of the state named PreNMI

`inPreNMIState` or just `isPreNMIState` may work?

src/code/speed_meter.c
Dragorn421: Yeah the name is horrible I agree, but it's very important though that the "time on audio" is not accounted in that timer as it's a big chunk,
image

Each chunk (1/4th) of the top blue-green-red-magenta bar represents the time for a VI retrace (see `SpeedMeter_DrawTimeEntries`) i.e. 1/50 seconds in PAL

first timer bar (red) is `gAudioThreadUpdateTimeTotalPerGfxTask`
second timer bar (yellow) is `gGfxTaskSentToNextReadyMinusAudioThreadUpdateTime`
(see `sSpeedMeterTimeEntryArray`)

you can see the red bar is not negligible (well this is emulator but I expect hardware to be similar at least in ratios)

Note: more or less, `gGraphUpdatePeriod = gGfxTaskSentToNextReadyMinusAudioThreadUpdateTime + gAudioThreadUpdateTimeTotalPerGfxTask + (waiting for task done and framebuffer swap)`

src/code/graph.c
Dragorn421: I think I wanted to try make the "concurrency" (it's not really, only one thread is ever running at the same time) aspect clearer but yeah it can be simplified to `The audio thread update is running`

src/code/speed_meter.c
Dragorn421: maybe? but the time period this is measured over is not the same, this is from gfxTask completed to next gfxTask completed (which also means it may accumulate to high values if for example the gfx buffers overlap/overflow and the game skips frames, but that may be an oversight), the others are from `Graph_Update` end to next `Graph_Update` end

include/z64.h
Roman971: I was relying on the fact that this is a gameState field to omit the "state" part (eg. `gameState->inPreNMI`), but if you feel that's not clear enough then yeah we could go with `inPreNMIState`

src/code/speed_meter.c
Roman971: Hmm I think I understand, it's probably fine as is then, tho I wonder why they did it that way since it kinda defeats the purpose of showing all these timers together if they don't measure things in the same time range

src/code/graph.c
```diff
@@ -166,36 +174,41 @@ void Graph_TaskSet00(GraphicsContext* gfxCtx) {
...
-    time = osGetTime();
-    if (D_8016A550 != 0) {
-        D_8016A558 = (D_8016A558 + time) - D_8016A550;
-        D_8016A550 = time;
+    timeNow = osGetTime();
+    if (gAudioThreadUpdateTimeStart != 0) {
+        // The audio thread update is running
+        // Add the time already spent to the accumulator and leave the rest for the next cycle
+
+        gAudioThreadUpdateTimeAcc = gAudioThreadUpdateTimeAcc + timeNow - gAudioThreadUpdateTimeStart;
```
Roman971, line 205: Looks like this can be written with `+=` if the pad variable at the top is also removed. I suggest doing it since it makes nicer and consistent with the other `gAudioThreadUpdateTimeAcc +=` in audio
```suggestion
        gAudioThreadUpdateTimeAcc += timeNow - gAudioThreadUpdateTimeStart;
```

src/code/speed_meter.c
```diff
@@ -1,25 +1,60 @@
...
+volatile OSTime gAudioThreadUpdateTimeAcc;
+// Accumulator for `gRSPAudioTimeTotal`
+volatile OSTime gRSPAudioTimeAcc;
+// Accumulator for `gRSPGfxTimeTotal`.
+volatile OSTime gRSPGfxTimeAcc;
+volatile OSTime gRSPOtherTimeAcc;
 volatile OSTime D_8016A578;
-volatile OSTime gRDPTotalTime;
+// Accumulator for `gRDPTimeTotal`
+volatile OSTime gRDPTimeAcc;
```
fig02, line 47: I would put some line breaks through all of this just to make it clear what comment belongs to what
(same comment for the 2 variables with block comments in graph.c)

src/code/speed_meter.c
fig02: I am not a fan of the super verbose names like this, but I do not have any suggestions, so I think we just go with this and improve it at some point if we can

Roman971: Yeah I guess we can do that tho it's a bit unfortunate

hensldm (approved):
Seems fine to me. Don't have any suggestions on possible shorter names unfortunately.

fig02:
made a pr addressing my comment from above https://github.com/Dragorn421/oot/pull/2

fig02 (approved):

Roman971 (approved):
Still really not a fan of `gGfxTaskSentToNextReadyMinusAudioThreadUpdateTime` but I guess we'll have to live with it until a better name is found

PR #1367, opened 2022-08-29
`stickRel` -> `stickAdj`

Dragorn421:
This PR renames names using "rel" in the "adjusted" sense, by replacing "rel" with "adj":

-----------

`Input.rel.stick_x`/y stores the "adjusted" x/y stick position, which is the current x/y stick position but clamped and with a "deadzone" (see `PadUtils_UpdateRelXY`)

"adjusted" comes from this very old comment in the `Input` struct:
https://github.com/zeldaret/oot/blob/0d52a4aebdd40562cc8653dc95e2e07528019c9b/include/padmgr.h#L10

"rel" can be misleading as it could stand for "released" (which isn't "adjusted") or "relative" (to what?)

-----------

Changes:

- `PauseContext.stickRelX`/Y -> `PauseContext.stickAdjX`/Y
- `FileSelectState.stickRelX`/Y -> `FileSelectState.stickAdjX`/Y
- `sOcarinaInputStickRel` -> `sOcarinaInputStickAdj`
- `relStickX` -> `stickAdjX` in `KaleidoScope_UpdatePrompt`

-------------

idk if "adj" is the best choice or is worth mentioning at all, comments are welcome

### DIFF
--- a/include/z64.h
+++ b/include/z64.h
@@ -778,8 +778,8 @@ typedef struct {
     /* 0x0208 */ u16    alpha;
     /* 0x020A */ s16    offsetY;
     /* 0x020C */ char   unk_20C[0x08];
-    /* 0x0214 */ s16    stickRelX;
-    /* 0x0216 */ s16    stickRelY;
+    /* 0x0214 */ s16    stickAdjX;
+    /* 0x0216 */ s16    stickAdjY;
     /* 0x0218 */ s16    cursorPoint[5]; // "cursor_point"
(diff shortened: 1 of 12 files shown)


EllipticEllipsis:
We could do a union with note about usage, `released` for buttons and `adjusted` for stick.

fig02 (approved):
mistakenly thought this pr was renaming the input struct member to adj, not just the related variables that are isolated to the stick values. so disregard previous comments, tho we should begin thinking about what we want to rename it to

Roman971 (approved):

PR #1368, opened 2022-08-29
`R_PAUSE_MENU_MODE` -> `R_PAUSE_BG_PRERENDER_STATE` and docs

Dragorn421:


include/regs.h
```diff
@@ -49,13 +49,14 @@
...
 #define R_UPDATE_RATE                            SREG(30)
+#define R_GRAPH_TASKSET00_FLAGS                  SREG(33)
 #define R_ENABLE_AUDIO_DBG                       SREG(36)
 #define R_FB_FILTER_TYPE                         SREG(80)
 #define R_FB_FILTER_PRIM_COLOR(c)                SREG(81 + (c))
 #define R_FB_FILTER_A                            SREG(84)
 #define R_FB_FILTER_ENV_COLOR(c)                 SREG(85 + (c))
 #define R_ENABLE_FB_FILTER                       SREG(88)
-#define R_PAUSE_MENU_MODE                        SREG(94)
+#define R_PAUSE_BG_PRERENDER_STATE               SREG(94) // `PauseBgPreRenderState`
```
Roman971, line 59: I don't really understand the reasoning behind this rename. `R_PAUSE_MENU_MODE` handles more than just the prerender state, since it actually governs how and when the pause menu gets loaded and drawn on the screen (partially at least), and it would exist in some form even if the prerender wasn't a thing.
Some of its values should probably be named around the prerender steps, or documented with them, but I don't think it makes sense to name the reg entirely around the prerender part.

Dragorn421: Everywhere it is used, `R_PAUSE_BG_PRERENDER_STATE` is about the state of the pause menu's background prerender. Mainly, in z_play all the states refer to the step making the prerender is at. In other spots it's about whether the pause menu may draw, which relies entirely on the prerender being done

All values of `R_PAUSE_BG_PRERENDER_STATE` are tied to the state of the pause background prerender so changing the reg's name seemed natural

Note the actual pause state `PauseContext.state` is something else (and a way bigger enum)

include/z64.h
```diff
@@ -701,6 +701,14 @@ typedef struct {
     }                   restrictions;
 } InterfaceContext; // size = 0x270
 
+typedef enum {
+    /* 0 */ PAUSE_BG_PRERENDER_OFF, // Inactive, do nothing.
+    /* 1 */ PAUSE_BG_PRERENDER_DRAW, // The current frame is only drawn for the purpose of serving as the pause background.
```
Thar0, line 706: When reading the code I got confused between this being the `DONE` state, as in I thought `DRAW` here meant "draw the pause background".
I think something like `SETUP` would work here and be less ambiguous.

include/z64.h
```diff
@@ -701,6 +701,14 @@ typedef struct {
...
+    /* 2 */ PAUSE_BG_PRERENDER_FILTER, // The previous frame was PAUSE_BG_PRERENDER_DRAW, now apply prerender filters.
```
Thar0, line 707: From a modding perspective it may help to be slightly more general in naming this state? I suggest `PROCESS` instead of `FILTER` since a filter is just one way in which the background could be processed

include/regs.h
fig02: Yeah in my mind `pauseCtx.state` has always conflicted with `R_PAUSE_MENU_MODE` in name. I am currently feeling like it would be beneficial to be specific about what its for.

In my opinion the only usage of the mode that can be argued is NOT for the bg processing is `R_PAUSE_MENU_MODE >= 3` which says that the pause menu is active and can draw. But as dragorn has said above, you can also look at it from the viewpoint of "bg processing is done, so pause menu can draw now".

include/z64.h
fig02: agreed here. in this state `SREG(33)` will decrement the frame buffer index causing the screen to "freeze" and it deliberately  skips the "draw overlay elements" block so that the hud wont show up in the "screenshot". I think its fair to say that this is the setup or "setting up the frame" for the bg before it starts processing

include/z64.h
fig02: also agree here. as a practical example, a mod tharo and I work on replaced the filters here with some blurring instead. 

not that every name has to be mod-inclusive, but I think this is a case where being a little bit more generic doesnt hurt anything, processing still tells you what it does (and is arguably a better state-like verb name)

include/regs.h
fig02: ` since it actually governs how and when the pause menu gets loaded and drawn on the screen (partially at least), and it would exist in some form even if the prerender wasn't a thing`

as far as I can tell, overlay loading is determined only by `pauseCtx.state`. It does control when it can draw, but as mentioned above its really only delaying drawing until the pause bg is done. 
Its purpose for existing is to control the flow of drawing in Play_Draw in relation to the background processing. So that it can for example exclude the hud in its screenshot. If the bg didnt exist i believe everything could be simplified to be controlled by the state variable

fig02: as a closing note on my spiel, I will note that we do know its original name does agree with calling it a generic mode: `osSyncPrintf("PR_KAREIDOSCOPE_MODE=%d\n", R_PAUSE_MENU_MODE);`
But yea even still I think being more specific is still beneficial here

include/z64.h
Dragorn421: I'll change to setup, it is indeed less ambiguous
Is the comment fine?

include/regs.h
Dragorn421: At the very least this isn't a "mode" that can change at will to have things behave differently, it describes what step the prerender is at (so "state"/...)

It's not really about UI (besides allowing to draw some as part of drawing the pause screen at all) so I wouldn't name it with the current "menu"

that leaves `R_PAUSE_STATE`, which is too generic, and knowing this deals (at 100% afaict) with drawing the prerender background well -> `R_PAUSE_BG_PRERENDER_STATE`

src/code/graph.c
```diff
@@ -222,8 +222,8 @@ void Graph_TaskSet00(GraphicsContext* gfxCtx) {
 
     scTask->next = NULL;
     scTask->flags = OS_SC_NEEDS_RSP | OS_SC_NEEDS_RDP | OS_SC_SWAPBUFFER | OS_SC_LAST_TASK;
-    if (SREG(33) & 1) {
-        SREG(33) &= ~1;
+    if (R_GRAPH_TASKSET00_FLAGS & 1) {
+        R_GRAPH_TASKSET00_FLAGS &= ~1;
         scTask->flags &= ~OS_SC_SWAPBUFFER;
         gfxCtx->fbIdx--;
     }
```
fig02, line 229: was hoping this reg would make its purpose clear that it is stopping the frame buffer indexing from incrementing and freezing the screen. "task set 00" means nothing to me personally

I do understand that it is a set of flags because of how its set and checked, but it pretty unfortunate lol

include/regs.h
fig02: Worth noting as well that MM adds a similar variable for its pictobox prerender that aligns with the pause menu one really closely. They both are pretty clearly about the state of the prerender as its being taken and processed

Dragorn421: Right, I've thought about using "kaleido" as a term encompassing both the pause menu and the game over thing, but it would end up just using kaleido for everything with not much actually being specific to pause or gameover

It's just really weird that these two things are not separated... And since gameover is comparatively very simple I'd rather just call everything "pause" yeah but this is debatable

src/code/graph.c
Dragorn421: I would be fine naming the reg after its only flag, but I just don't know what to name it

To me the current name is temporary, it's just for the sake of not being a random REG()

hensldm: Can understand not wanting to do it in this PR, but the `1` could probably use a define which might help with this situation.

hensldm (approved):
Just one comment/suggestion. Overall seems good to me.

src/code/graph.c
Dragorn421: I would like this PR to get in so yeah I'd rather not add stuff to it now, months later
I'll add it to my todolist though

Btw this PR is a 80 lines diff, my nearing-100% documentation of pause menu is a 6000 lines diff right now...

Dragorn421:
(I'm waiting on this to PR more fragments of #1366 pause menu docs)

src/code/z_play.c
```diff
@@ -1115,19 +1115,25 @@ void Play_Draw(PlayState* this) {
         } else {
             PreRender_SetValues(&this->pauseBgPreRender, SCREEN_WIDTH, SCREEN_HEIGHT, gfxCtx->curFrameBuffer, gZBuffer);
 
-            if (R_PAUSE_MENU_MODE == 2) {
+            if (R_PAUSE_BG_PRERENDER_STATE == PAUSE_BG_PRERENDER_PROCESS) {
+                // Wait for the previous frame's DList to be processed,
+                // so that `pauseBgPreRender.fbufSave` and `pauseBgPreRender.cvgSave` are filled with the appropriate
```
fig02, line 1120: personally think you can just say the whole phrase "display list", but dont care that much
happens in one other spot as well, if you do decide to change it

fig02 (approved):

Thar0 (approved):

PR #1369, opened 2022-08-29
Replace player animation names with original names derived from OoT3D

tom-overton:
Ocarina of Time 3D contains two archives that are completely unused: `zelda_link_boy_ultra.zar` and `zelda_link_child_ultra.zar`. These contain glitchy versions of Link's original N64 models, but more importantly, they contain 561 named `.anb` files. These files are the original animation data from `link_animetion`, just converted to little-endian. In other words, they tell us the original names of the vast majority of player animations.

I wrote a script that compared the original data in `link_animetion` to every `.anb` file in these archives, and from this script's output I created this spreadsheet that maps player animation offsets within `gameplay_keep` to their original name: https://docs.google.com/spreadsheets/d/1wPDbW4goyQkZkJgPqwM9m-EcSZrEh_3kXtfIwV3V7zo/edit#gid=1949976453

There seemed to be some interest in using these names in the short-term, so I made this PR. It updates all but 12 player animation names within `gameplay_keep` to use their original name. It also updates `link_animetion` to use these names as well, just with `_Data` stapled to the end.


fig02 (approved):

Dragorn421 (approved):
:smiley:

Roman971 (approved):

PR #1374, opened 2022-09-01
Document Framebuffer Filters, VisCvg and VisZbuf

EllipticEllipsis:
Big thanks to @Dragorn421 and @Thar0 for helping me understand how these two files worked, it's been a fascinating crash course in RenderModes.

I named them what Dragorn and I agreed in Discord a few weeks ago: https://discord.com/channels/688807550715560050/688851317593997489/1010581711110803558 ; do scroll down for some very useful images and analysis by Dragorn. Would consider something a bit less abbreviated, but it does fit with the overall scheme of `VisMono`.

Docs are rather less extensive than `VisMono`, but I think I covered all the unusual and/or interesting parts. Might have over-emphasised the issue with filter 4, but not sure where people are actually going to read about the filters.


src/code/z_viscvg.c
```diff
@@ -1,70 +1,100 @@
-#include "global.h"
+/**
+ * @file z_viscvg.c
+ * Description: Visualise Coverage (how much of the neighbourhood of a pixel is covered by a primitive) in various ways.
```
Dragorn421, line 3:
```suggestion
 * Description: Visualise Coverage (how much of a pixel is covered by a primitive) in various ways.
```
Also idk how much detail we want in these files, as it's not supposed to be a tutorial to N64 graphics, but I also think these are low level and technical enough it's okay to go into details

With that said if you want to explain coverage I think it's worth mentioning its main (only?) use is for antialiasing (for use by the antialiasing and divot filters) (linking to the relevant Prerender docs may be a good idea)
(i.e. if you disable antialiasing coverage is always full everywhere)

src/code/z_viscvg.c
```diff
@@ -1,70 +1,100 @@
...
+ *
+ * To understand this file, it is helpful to remember that A_MEM is essentially synonymous with coverage, and that
+ * `GBL_c1/2(p, a, m, b)` are usually `(p * a + m * b) / (a + b)`.
+ *
+ * Coverage is maximum when not on an edge, on an edge it is smaller, and since coverage is treated as an alpha edges
+ * show up as darker in all of the available modes.
+ *
+ * We abbreviate coverage to "cvg"; "pixel RGB" is the original colour the pixel had before the filter is applied.
```
Dragorn421, line 11:
```suggestion
 * Coverage is abbreviated to "cvg"; "pixel RGB" is the original colour the pixel had before the filter is applied.
```

src/code/z_viscvg.c
```diff
@@ -1,70 +1,100 @@
-#include "global.h"
+/**
+ * @file z_viscvg.c
+ * Description: Visualise Coverage (how much of the neighbourhood of a pixel is covered by a primitive) in various ways.
+ *
+ * To understand this file, it is helpful to remember that A_MEM is essentially synonymous with coverage, and that
+ * `GBL_c1/2(p, a, m, b)` are usually `(p * a + m * b) / (a + b)`.
+ *
+ * Coverage is maximum when not on an edge, on an edge it is smaller, and since coverage is treated as an alpha edges
```
Dragorn421, line 8: It is correct to say coverage is full outside of edges, but it's not necessarily low at edges.
It often is, but imagine an edge drawn almost perfectly in between two columns of pixels, both sides will have full coverage

Maybe instead of going into details you could rephrase like "coverage is typically low at edges" (or "often" instead of "typically" idk)

(btw I think "full" is the proper term for refering to the coverage being at its maximum, but you may want to check the manual on that)

src/code/z_viscvg.c
```diff
@@ -1,70 +1,100 @@
...
-Gfx D_8012AC40[] = {
+/**
+ * Draws coverage and RGB of pixels
+ */
+Gfx sCoverageRGBDL[] = {
     gsDPSetOtherMode(G_AD_PATTERN | G_CD_MAGICSQ | G_CK_NONE | G_TC_CONV | G_TF_POINT | G_TT_NONE | G_TL_TILE |
                          G_TD_CLAMP | G_TP_NONE | G_CYC_1CYCLE | G_PM_NPRIMITIVE,
                      G_AC_NONE | G_ZS_PRIM | IM_RD | CVG_DST_CLAMP | ZMODE_OPA | FORCE_BL |
                          GBL_c1(G_BL_CLR_IN, G_BL_0, G_BL_CLR_MEM, G_BL_A_MEM) |
                          GBL_c2(G_BL_CLR_IN, G_BL_0, G_BL_CLR_MEM, G_BL_A_MEM)),
```
Dragorn421, line 54: Can't this and the one below be replaced with G_RM_VISCVG ?

src/code/z_viscvg.c
```diff
@@ -1,70 +1,100 @@
...
 
-Gfx D_8012AC00[] = {
+/**
+ * Draws only coverage: does not retain any of the original pixel RGB, primColor is used as background colour.
+ */
+Gfx sCoverageOnlyDL[] = {
     gsDPSetOtherMode(G_AD_PATTERN | G_CD_MAGICSQ | G_CK_NONE | G_TC_CONV | G_TF_POINT | G_TT_NONE | G_TL_TILE |
                          G_TD_CLAMP | G_TP_NONE | G_CYC_1CYCLE | G_PM_NPRIMITIVE,
                      G_AC_NONE | G_ZS_PRIM | G_RM_VISCVG | G_RM_VISCVG2),
+    // (blendColor) * (cvg)
```
Dragorn421, line 23: I don't think we should start commenting on every blender/combiner usage, if anywhere this could go in gbi.h near the rendermode (note tharo is reformatting it so I wouldn't touch it for now for conflict reasons) (also we need to talk about what we can do to gbi.h comments-wise but that's completely unrelated to the PR)

src/code/z_vismono.c
```diff
@@ -7,7 +9,7 @@
 // Height of the fragments the color frame buffer (CFB) is split into.
 // It is the maximum amount of lines such that all rgba16 SCREEN_WIDTH-long lines fit into
 // the half of tmem (0x800 bytes) dedicated to color-indexed data.
```
Dragorn421, line 11:
```suggestion
// the half of tmem dedicated to color-indexed data.
```

Dragorn421 (commented):
📺 🔍 👀

src/code/z_viscvg.c
EllipticEllipsis: No, because
```
#define RM_VISCVG(clk) \
    IM_RD | FORCE_BL |     \
    GBL_c##clk(G_BL_CLR_IN, G_BL_0, G_BL_CLR_BL, G_BL_A_MEM)
```
uses blendColor, not mem.

src/code/z_viscvg.c
EllipticEllipsis: In general I'd agree, I'd regard this as an exception since here it's pretty significant to what the functions in this file do.

src/code/z_viscvg.c
EllipticEllipsis: I had a go, I thought I had reasonable terminology but the manual tripped me up again.

src/code/z_viscvg.c
```diff
@@ -1,70 +1,103 @@
...
+ * - Transparently blends primColor with original frame, creating uniform fog. The "cloud surface" RenderMode is used to
+ * preserve the coverage for the second stage.
+ * - Second half is the same as sCoverageRGBDL's, i.e. coverage and RGB of pixels
+ */
+Gfx sCoverageRGBPrimFogDL[] = {
     gsDPSetCombineMode(G_CC_PRIMITIVE, G_CC_PRIMITIVE),
     gsDPSetOtherMode(G_AD_NOTPATTERN | G_CD_DISABLE | G_CK_NONE | G_TC_CONV | G_TF_POINT | G_TT_NONE | G_TL_TILE |
                          G_TD_CLAMP | G_TP_NONE | G_CYC_1CYCLE | G_PM_NPRIMITIVE,
                      G_AC_NONE | G_ZS_PRIM | G_RM_CLD_SURF | G_RM_CLD_SURF2),
+    // (fog RGB) * (cvg) + (pixel RGB) * (1 - cvg).
```
Dragorn421, line 75: `RM_CLD_SURF` is `G_BL_CLR_IN, G_BL_A_IN, G_BL_CLR_MEM, G_BL_1MA`,

notice `p` is `G_BL_CLR_IN`, that's the combined rgb
also notice `a` is `G_BL_A_IN`, that's the combined alpha
(i.e. for both prim rgb / alpha because `gsDPSetCombineMode(G_CC_PRIMITIVE, G_CC_PRIMITIVE)` above)

basically this interpolates between the current frame buffer and a solid prim color, with prim alpha as the ratio

I guess I also wouldn't use "fog" to describe this as "fog" is something very specific in N64 graphics and this is unrelated

src/code/z_viscvg.c
```diff
@@ -1,70 +1,103 @@
...
+ * Coverage is roughly how much of a pixel is covered by a primitive; it is used for antialiasing, see PreRender.c and
+ * §15 of the programming manual for details.
+ *
+ * To understand this file, it is helpful to remember that A_MEM is essentially synonymous with coverage, and that
+ * `GBL_c1/2(p, a, m, b)` are usually `(p * a + m * b) / (a + b)`.
+ *
+ * Coverage is full when not on an edge, while on an edge it is usually lower, and since coverage is treated as an
+ * alpha value, edges with lower coverage will show up as darker than interiors in all of the available modes.
+ *
+ * Coverage is abbreviated to "cvg"; "pixel RGB" is the original colour the pixel had before the filter is applied.
```
Dragorn421, line 14: Well in what is currently named `sCoverageRGBPrimFogDL` "pixel RGB" at the second blender is already not the colour from the before-filter framebuffer, it has been lerped with primColor

Maybe just use "FB RGB" as in "frame buffer rgb"?

include/z64.h
```diff
@@ -1805,13 +1815,12 @@ typedef struct {
     /* 0x14 */ Gfx* dList;
 } VisMono; // size = 0x18
 
-// Vis...
 typedef struct {
     /* 0x00 */ u32 useRgba;
     /* 0x04 */ u32 setScissor;
     /* 0x08 */ Color_RGBA8_u32 primColor;
     /* 0x08 */ Color_RGBA8_u32 envColor;
```
engineer124, line 1822:
```suggestion
    /* 0x08 */ Color_RGBA8_u32 primColor;
    /* 0x0C */ Color_RGBA8_u32 envColor;
```

src/code/z_vismono.c
```diff
@@ -1,13 +1,15 @@
...
  * Color frame buffer effect to desaturate the colors.
  */
 
 #include "global.h"
 
 // Height of the fragments the color frame buffer (CFB) is split into.
 // It is the maximum amount of lines such that all rgba16 SCREEN_WIDTH-long lines fit into
-// the half of tmem (0x800 bytes) dedicated to color-indexed data.
-#define VISMONO_CFBFRAG_HEIGHT (0x800 / (SCREEN_WIDTH * G_IM_SIZ_16b_BYTES))
+// the half of tmem dedicated to color-indexed data.
```
engineer124, line 11:
```suggestion
// the half of TMEM dedicated to color-indexed data.
```
For consistency

src/code/z_viszbuf.c
```diff
@@ -0,0 +1,98 @@
...
+ * The red increases monotonically with the depth. The significant visible oscillation is the green component, because
+ * it rolls over every time the second-most-significant bit of the mantissa increments. The blue component oscillates
+ * too rapidly to be particularly visible (it rolls over when the 7th-most-significant bit increments). The alpha is
+ * again ignored by the RenderMode.
+ */
+
+#include "global.h"
+
+// Height of the fragments the z-buffer is split into.
+// It is the maximum amount of lines such that all rgba16 SCREEN_WIDTH-long lines fit into tmem.
```
engineer124, line 33:
```suggestion
// It is the maximum amount of lines such that all rgba16 SCREEN_WIDTH-long lines fit into TMEM.
```
Here too

src/code/z_viscvg.c
EllipticEllipsis:
```c

/**
 * Two stage filtering:
 *
 * 1. Apply a uniform colour filter by transparently blending primColor with original frame. The "cloud surface"
 * RenderMode is used to preserve the coverage for the second stage.
 * 2. Second half is the same as sCoverageRGBDL's, i.e. previous colour * cvg
 */
Gfx sCoverageRGBPrimFogDL[] = {
    gsDPSetCombineMode(G_CC_PRIMITIVE, G_CC_PRIMITIVE),
    gsDPSetOtherMode(G_AD_NOTPATTERN | G_CD_DISABLE | G_CK_NONE | G_TC_CONV | G_TF_POINT | G_TT_NONE | G_TL_TILE |
                         G_TD_CLAMP | G_TP_NONE | G_CYC_1CYCLE | G_PM_NPRIMITIVE,
                     G_AC_NONE | G_ZS_PRIM | G_RM_CLD_SURF | G_RM_CLD_SURF2),
    // stage 1 color = (primColor RGB) * (cvg) + (FB RGB) * (1 - cvg)
    gsDPFillRectangle(0, 0, SCREEN_WIDTH - 1, SCREEN_HEIGHT - 1),

    gsDPSetOtherMode(G_AD_PATTERN | G_CD_MAGICSQ | G_CK_NONE | G_TC_CONV | G_TF_POINT | G_TT_NONE | G_TL_TILE |
                         G_TD_CLAMP | G_TP_NONE | G_CYC_1CYCLE | G_PM_NPRIMITIVE,
                     G_AC_NONE | G_ZS_PRIM | IM_RD | CVG_DST_CLAMP | ZMODE_OPA | FORCE_BL |
                         GBL_c1(G_BL_CLR_IN, G_BL_0, G_BL_CLR_MEM, G_BL_A_MEM) |
                         GBL_c2(G_BL_CLR_IN, G_BL_0, G_BL_CLR_MEM, G_BL_A_MEM)),
    // final colour = (stage 1's RGB) * (cvg)
    gsDPFillRectangle(0, 0, SCREEN_WIDTH - 1, SCREEN_HEIGHT - 1),
    gsSPEndDisplayList(),
};
```
Something like this?

EllipticEllipsis: I did think about something like `current` or `new` or `previous` but don't really want to get into temporal things since it's a single frame.

Dragorn421: Not sure what you refer to with current/new/prev? The text you wrote does use "previous"

"Stage 1" doesn't use coverage, only combined alpha (i.e. prim alpha)

src/code/z_viscvg.c
fig02: also colour -> color, since we use the latter pretty much everywhere else.
a few should be fixed up in a seperate cleanup pr or something 
image

src/code/z_viscvg.c
EllipticEllipsis: Yeah, it's tricky to describe the output of the first blend. I've had another go, see what you think.

src/code/z_viscvg.c
```diff
@@ -75,20 +109,29 @@ void func_800ACE98(struct_801664F0* this, Gfx** gfxp) {
...
+            gDPSetColor(gfx++, G_SETPRIMCOLOR, this->primColor.rgba);
+            gSPDisplayList(gfx++, sCoverageRGBPrimFogDL);
             break;
-        case 3:
-            gDPSetColor(gfx++, G_SETBLENDCOLOR, this->color.rgba);
-            gSPDisplayList(gfx++, D_8012AC00);
+
+        case FB_FILTER_CVG:
+            // Set background color for G_RM_VISCVG
+            gDPSetColor(gfx++, G_SETBLENDCOLOR, this->primColor.rgba);
```
Dragorn421, line 124: `VisCvg.primColor` is used as prim/blend/fog color so the previous name "color" is probably better

src/code/z_viscvg.c
```diff
@@ -75,20 +109,29 @@ void func_800ACE98(struct_801664F0* this, Gfx** gfxp) {
...
+            gSPDisplayList(gfx++, sCoverageOnlyDL);
+            break;
+
+        case FB_FILTER_CVG_RGB_FOG:
+            // Set fogging color for custom RenderMode, needs to be close to 0 to not overflow
```
Dragorn421, line 129:
```suggestion
            // Set fog color for custom RenderMode, needs to be close to 0 to not overflow
```
the name of the color register is "fog color", typically used for the fog effect but not necessarily, so for refering to it it's probably best to stick to its name

src/code/z_viscvg.c
```diff
@@ -75,20 +109,29 @@ void func_800ACE98(struct_801664F0* this, Gfx** gfxp) {
...
+            gSPDisplayList(gfx++, sCoverageRGBDL);
             break;
-        case 2:
-            gDPSetColor(gfx++, G_SETPRIMCOLOR, this->color.rgba);
-            gSPDisplayList(gfx++, D_8012AC58);
+
+        case FB_FILTER_CVG_RGB_PRIMFOG:
+            // Set primitive color for uniform fogging in custom RenderMode
+            gDPSetColor(gfx++, G_SETPRIMCOLOR, this->primColor.rgba);
+            gSPDisplayList(gfx++, sCoverageRGBPrimFogDL);
```
Dragorn421, line 119: I'm still not a fan of using "fog" at a low level when fog isn't involved

`PRIMOVERLAY` could be an idea maybe, or `primcoat`, `primcover` (thanks synonyms dictionaries)

src/code/z_viszbuf.c
```diff
@@ -0,0 +1,98 @@
...
+ * useRgba:
+ * - false will produce a monotonic fade from primColor to envColor as depth increases
+ * - true produces vibrant almost-periodic-looking bands.
+ *
+ * In both cases this occurs because of the format the depth information takes: it is 18-bit, and is a positive
+ * floating-point number with
+ *     bbb      mmmmmmmmmmm dd|dd
+ *     exponent mantissa    dz value (only first 16 bits visible to CPU)
+ *
+ * IA:
```
Dragorn421, line 14:
```suggestion
 * When reading the z-buffer as IA16:
```
or something similar

IA*16* matters of course
and I think it's worth making it clearer the z-buffer is being read/loaded/interpreted? as an image

src/code/z_viszbuf.c
```diff
@@ -0,0 +1,98 @@
...
+ *     bbbmmmmm mmmmmmdd
+ *     iiiiiiii aaaaaaaa
+ * Since positive floating-point numbers have the same ordering as their binary/hex representation, increasing
```
Dragorn421, line 17:
```suggestion
 * Since floating-point numbers following this unsigned format have the same ordering as their binary/hex representation, increasing
```
my main point here being that you may very well come up with an unorthodox floating point format where the binary representation is inconvenient, so this technically isn't a statement that is in general true

src/code/z_viszbuf.c
```diff
@@ -0,0 +1,98 @@
...
+ * the depth also increases the representation and hence the color used to combine primColor and envColor. The alpha is
```
Dragorn421, line 18:
```suggestion
 * the depth also increases the intensity in the IA16 representation and hence the color used to combine primColor and envColor. The alpha is
```
(IA stands for Intensity Alpha btw (I think))
this is more of a nitpick that I think can help readers understand the point

src/code/z_viscvg.c
EllipticEllipsis: I did it for consistency with the other two Vis files, don't think it improves it to name it `color`, especially since there's an envColor too.

src/code/z_viscvg.c
EllipticEllipsis: In lighting this would be called a "gel", I think `// Set primitive color for uniform color filter in custom RenderMode` is probably fine

Dragorn421: Sure

Dragorn421: (the comment was also on the enum name and dlist name though)

src/code/z_viscvg.c
Dragorn421: Well it's simply not a primColor in some cases. You can name it otherColor if you want to make it stand out from envColor? Or mainColor ...?

EllipticEllipsis: Come to think of it
```c
typedef struct {
    /* 0x00 */ u32 type;
    /* 0x04 */ u32 setScissor;
    /* 0x08 */ Color_RGBA8_u32 primColor;
    /* 0x0C */ Color_RGBA8_u32 envColor; // unused
} VisCvg; // size = 0x10

typedef struct {
    /* 0x00 */ u32 unk_00;
    /* 0x04 */ u32 setScissor;
    /* 0x08 */ Color_RGBA8_u32 primColor;
    /* 0x0C */ Color_RGBA8_u32 envColor;
    /* 0x10 */ u16* tlut;
    /* 0x14 */ Gfx* dList;
} VisMono; // size = 0x18

typedef struct {
    /* 0x00 */ u32 useRgba;
    /* 0x04 */ u32 setScissor;
    /* 0x08 */ Color_RGBA8_u32 primColor;
    /* 0x0C */ Color_RGBA8_u32 envColor;
} VisZbuf; // size = 0x10
```
These look like they have a common substruct.

EllipticEllipsis: It's also set from the `PRIM_COLOR` registers:
```c
        sVisCvg.primColor.r = R_FB_FILTER_PRIM_COLOR(0);
        sVisCvg.primColor.g = R_FB_FILTER_PRIM_COLOR(1);
        sVisCvg.primColor.b = R_FB_FILTER_PRIM_COLOR(2);
```
Might want to consider changing to `PRIMCOLOR` to match the gbi convention too.

src/code/z_viscvg.c
```diff
@@ -75,20 +109,29 @@ void func_800ACE98(struct_801664F0* this, Gfx** gfxp) {
...
-        case 1:
-            gSPDisplayList(gfx++, D_8012AC40);
+        case FB_FILTER_CVG_RGB:
+            gSPDisplayList(gfx++, sCoverageRGBDL);
             break;
-        case 2:
-            gDPSetColor(gfx++, G_SETPRIMCOLOR, this->color.rgba);
-            gSPDisplayList(gfx++, D_8012AC58);
+
+        case FB_FILTER_CVG_RGB_UNIFORM:
```
Roman971, line 116: Looks like this needs to be renamed to `FB_FILTER_CVG_RGB_PRIMFOG` (it causes the build to fail currently)

include/z64.h
```diff
@@ -1805,13 +1815,12 @@ typedef struct {
...
 
-// Vis...
 typedef struct {
     /* 0x00 */ u32 useRgba;
     /* 0x04 */ u32 setScissor;
     /* 0x08 */ Color_RGBA8_u32 primColor;
-    /* 0x08 */ Color_RGBA8_u32 envColor;
-} struct_80166500; // size = 0x10
+    /* 0x0C */ Color_RGBA8_u32 envColor;
+} VisZbuf; // size = 0x10
```
Roman971, line 1823: Should this be `VisZBuf` since we use "z-buffer" and not the single word variant? (eg. we have `gZBuffer` instead of `gZbuffer`)

src/code/z_viszbuf.c
```diff
@@ -0,0 +1,112 @@
...
+    this->envColor.b = 0;
+}
+
+void VisZbuf_Destroy(VisZbuf* this) {
+}
+
+void VisZbuf_Draw(VisZbuf* this, Gfx** gfxp) {
+    Gfx* gfx = *gfxp;
+    u16* zbufFrag = D_0E000000;
+    s32 fmt;// = (this->useRgba == false) ? G_IM_FMT_IA : G_IM_FMT_RGBA;
```
Roman971, line 67: What's the reasoning behind not using the ternary or at least the `== false` here?
Also this looks like it may need a formatter run?

EllipticEllipsis:
Oh, I mentioned in Discord but not here: I pushed a branch that had an idea of putting these three files as one system: https://github.com/EllipticEllipsis/oot/commit/57c739cefc5da02f103ccc035ffbdde0ce214adb
That's where the idea of removing `false` comparison comes in, unfortunately there isn't a way to totally remove the pseudoboolean but I think it looks a bit more coherent.

include/z64.h
EllipticEllipsis: Ah, I think I was using `prerender` as reference which consistently uses `zbuf`, I can change if capital b is preferred.

Roman971: Afaik we agreed to use z-buffer and thus capital B everywhere, tho it's not very consistent yet (similar situation with frame buffer which is still treated as a single word in a few places atm)

hensldm:
> Oh, I mentioned in Discord but not here: I pushed a branch that had an idea of putting these three files as one system: [EllipticEllipsis@57c739c](https://github.com/EllipticEllipsis/oot/commit/57c739cefc5da02f103ccc035ffbdde0ce214adb) That's where the idea of removing `false` comparison comes in, unfortunately there isn't a way to totally remove the pseudoboolean but I think it looks a bit more coherent.

Seems good to me.

fig02:

PR #1263, opened 2022-06-08
Some cleanup/doc on the item icons

Dragorn421:
Main things:
- `gItemIcons` in `code_80097A00.c`: format and some comments
- defines/macros in `z64interface.h` to replace magic numbers refering to `icon_item_static`/`icon_item_24_static` assets ("item icons" and "quest icons")

PR #1376, opened 2022-09-04
Update item names

Dragorn421:
Splitting from #1263 as that other PR is all over the place


fig02 (approved):

include/z64item.h
```diff
@@ -283,24 +283,24 @@ typedef enum {
...
     /* 0x96 */ ITEM_BOMBCHUS_5,
     /* 0x97 */ ITEM_BOMBCHUS_20,
-    /* 0x98 */ ITEM_STICK_UPGRADE_20,
-    /* 0x99 */ ITEM_STICK_UPGRADE_30,
-    /* 0x9A */ ITEM_NUT_UPGRADE_30,
-    /* 0x9B */ ITEM_NUT_UPGRADE_40,
+    /* 0x98 */ ITEM_DEKU_STICK_UPGRADE_20,
+    /* 0x99 */ ITEM_DEKU_STICK_UPGRADE_30,
+    /* 0x9A */ ITEM_DEKU_NUT_UPGRADE_30,
+    /* 0x9B */ ITEM_DEKU_NUT_UPGRADE_40,
```
EllipticEllipsis, line 303: Maybe this should be `CAPACITY` instead of `UPGRADE`. And the same for the quivers/bomb bags: it's not 30/40/50 of them!

fig02: plenty of other things rely on this upgrade naming scheme, so it wouldnt be so simple to change its name. `UpgradeType` enum, `Inventory_ChangeUpgrade`, `gSaveContext.inventory.upgrades` etc

include/z64item.h
```diff
@@ -195,18 +195,18 @@ typedef enum {
...
-    /* 0x36 */ ITEM_EYEDROPS,
+    /* 0x35 */ ITEM_EYEBALL_FROG,
+    /* 0x36 */ ITEM_EYE_DROPS,
     /* 0x37 */ ITEM_CLAIM_CHECK,
-    /* 0x38 */ ITEM_BOW_ARROW_FIRE,
-    /* 0x39 */ ITEM_BOW_ARROW_ICE,
-    /* 0x3A */ ITEM_BOW_ARROW_LIGHT,
+    /* 0x38 */ ITEM_BOW_FIRE,
+    /* 0x39 */ ITEM_BOW_ICE,
+    /* 0x3A */ ITEM_BOW_LIGHT,
```
mzxrules, line 206: This change is kind of weird since it's not a fire bow item, it's e.g. a Bow + Fire Arrows, hence BOW_ARROW_FIRE

include/z64item.h
```diff
@@ -257,16 +257,16 @@ typedef enum {
...
     /* 0x6F */ ITEM_STONE_OF_AGONY,
-    /* 0x70 */ ITEM_GERUDO_CARD,
+    /* 0x70 */ ITEM_GERUDOS_CARD,
     /* 0x71 */ ITEM_SKULL_TOKEN,
     /* 0x72 */ ITEM_HEART_CONTAINER,
     /* 0x73 */ ITEM_HEART_PIECE,
-    /* 0x74 */ ITEM_KEY_BOSS,
-    /* 0x75 */ ITEM_COMPASS,
+    /* 0x74 */ ITEM_DUNGEON_BOSS_KEY,
+    /* 0x75 */ ITEM_DUNGEON_COMPASS,
```
mzxrules, line 265: Don't feel it necessary to prefix these with DUNGEON. The map item is actually called Dungeon Map

include/z64item.h
Dragorn421: I feel like `ITEM_BOW_ARROW_FIRE` makes it too similar to `ITEM_ARROW_FIRE` and it could be confused for the fire arrows item

include/z64item.h
```diff
@@ -145,46 +145,46 @@ typedef enum {
...
+    /* 0x1B */ ITEM_BOTTLE_RUTOS_LETTER,
+    /* 0x1C */ ITEM_BOTTLE_BLUE_FIRE,
+    /* 0x1D */ ITEM_BOTTLE_BUG,
+    /* 0x1E */ ITEM_BOTTLE_BIG_POE,
+    /* 0x1F */ ITEM_BOTTLE_MILK_HALF,
+    /* 0x20 */ ITEM_BOTTLE_POE,
     /* 0x21 */ ITEM_WEIRD_EGG,
-    /* 0x22 */ ITEM_CHICKEN,
-    /* 0x23 */ ITEM_LETTER_ZELDA,
+    /* 0x22 */ ITEM_CUCCO,
```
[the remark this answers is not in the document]
Dragorn421: text id `0x3066` calls both `ITEM_CUCCO` and `ITEM_POCKET_CUCCO` a "chicken"
the gi model for those two is the same
pause text does say "chicken" while pocket cucco is "pocket cucco"

idk anyone has more thoughts on this? I'm okay with "chicken" too especially if that's how the name calls it apparently

note for myself: If we decide to go with "chicken" `gCuccoItemName*Tex` -> chicken

include/z64item.h
Dragorn421: The prefix is meant to highlight that the three dungeon items boss key/compass/map are in a "group", for example in `Item_Give`:
https://github.com/zeldaret/oot/blob/6b77ddba5ae0941c10c58260c44270dfcb88e182/src/code/z_parameter.c#L1426-L1428
so the group of "dungeon items"

at least that was the intention when I opened #1263 but I've since realized the item enum is a lot more sketchy than that so I'm not so big on this anymore. If someone else also thinks this is better without `dungeon_` I'm fine removing it

But I also think it's okay to have "dungeon"? (not as terrible as `ITEM_SPIRITUAL_KOKIRI_EMERALD` at least)

include/z64item.h
Dragorn421: well since `ITEM_BOMB_BAG_40` doesn't have "upgrade" why not have `ITEM_DEKU_NUT_CAPACITY_40` in theory, but I don't think it's much of an improvement

idk I don't think we can have something truly satisfying on the "upgrades items", and current master looks okay on that front

include/z64item.h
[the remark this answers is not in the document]
mzxrules: I'd ask around discord for what everyone else thinks. GI model should definitely stay Cucco (assuming it's already named as such) and so maybe that leads us toward keeping ITEM_CUCCO for internal code consistency.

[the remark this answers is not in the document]
krm01: I like cucco but I can't back it up with an argument. "chicken" just feels... wrong.

include/z64item.h
Dragorn421: Is this considered resolved?
or can we at least leave this to another PR/discussion (that I wouldn't open/start, feel free to do so Elliptic though)

Dragorn421:
Summary of what needs resolved:
- `ITEM_CUCCO` vs `ITEM_CHICKEN`
- `ITEM_BOW_FIRE` vs `ITEM_BOW_ARROW_FIRE` (and other arrow types)
- `ITEM_DUNGEON_BOSS_KEY`, `ITEM_DUNGEON_COMPASS` vs `ITEM_BOSS_KEY`, `ITEM_COMPASS`

See comments for details

include/z64item.h
```diff
@@ -145,46 +145,46 @@ typedef enum {
...
     /* 0x09 */ ITEM_BOMBCHU,
     /* 0x0A */ ITEM_HOOKSHOT,
     /* 0x0B */ ITEM_LONGSHOT,
     /* 0x0C */ ITEM_ARROW_ICE,
     /* 0x0D */ ITEM_FARORES_WIND,
     /* 0x0E */ ITEM_BOOMERANG,
-    /* 0x0F */ ITEM_LENS,
-    /* 0x10 */ ITEM_BEAN,
+    /* 0x0F */ ITEM_LENS_OF_TRUTH,
+    /* 0x10 */ ITEM_MAGIC_BEANS,
```
Roman971, line 164: Could you explain the reasoning behind making this one plural when all the other "ammo"-type items are singular? (eg. stick or bomb)

include/z64item.h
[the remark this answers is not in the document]
Roman971: If "cucco" is supposed to be the name and "chicken" is just a mistranslation, then it makes sense to me to with cucco, but either name is probably okay

include/z64item.h
Dragorn421: It wasn't a conscious decision, I guess just like you say "use deku stick here" you naturally say "use magic beans here" and not "use magic bean here"

Note I must not be alone since `icon_item_static` also has it as plural `gMagicBeansIconTex`

But I'm fine changing all those to singular "magic bean", it would indeed make more sense

Roman971:
What's stopping us from renaming `GetItemID` and `GetItemDrawID` enum values along with those? Imo we should definitely keep these enums in sync and it would make a lot more sense if they were all discussed and renamed together.
Same comment for the names in the quest and slot enums since those are very closely related to the item enum (eg. `ITEM_STICK` -> `ITEM_DEKU_STICK` should also lead to `SLOT_STICK` -> `SLOT_DEKU_STICK`)

include/z64item.h
EllipticEllipsis: Well that's just what the items are called, see the textures in `item_name_static`.

Dragorn421:
I didn't know about the slot_ enum, done

for gi/gid enums I forgot, so I now did it

include/z64item.h
```diff
@@ -257,16 +257,16 @@ typedef enum {
     /* 0x6D */ ITEM_GORON_RUBY,
     /* 0x6E */ ITEM_ZORA_SAPPHIRE,
     /* 0x6F */ ITEM_STONE_OF_AGONY,
-    /* 0x70 */ ITEM_GERUDO_CARD,
+    /* 0x70 */ ITEM_GERUDOS_CARD,
```
Roman971, line 260: Looks like `QUEST_GERUDO_CARD` should also be updated, and possibly other quest enum values

include/z64item.h
Dragorn421: Bump: Is this considered resolved?
I'll assume yes in like 7 days or so

include/z64item.h
Dragorn421: Bump: Is this considered resolved?
I'll assume yes in like 7 days or so

include/z64item.h
Dragorn421: Bump: Is this considered resolved?
I'll assume yes in like 7 days or so

include/z64item.h
hensldm: singular does seem more consistent with the other ammos.

include/z64item.h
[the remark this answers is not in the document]
hensldm: Agree with Roman. Either seems fine to me.

hensldm (approved):
LGTM :)

include/z64item.h
Dragorn421: I'll change to singular

include/z64item.h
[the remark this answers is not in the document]
Dragorn421: I'm starting to lean towards being conservative and revert to master's `ITEM_CHICKEN` for this one

I'll leave it as "cucco" for now but if someone insists on using "chicken" I'll revert to "chicken" and open another PR for cleaning the rest of cucco->chicken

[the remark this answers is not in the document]
fig02: I do wish they translated the alarm clock part somehow into english, it makes alot of sense for what it is lol (just the thing that wakes up talon)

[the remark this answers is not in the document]
Dragorn421: -> chicken

Dragorn421:
Recent changes:
"magic beans" -> "magic bean"
"cucco" -> "chicken" (the one for waking up Talon, not pocket cucco)

fig02 (approved):

Roman971 (approved):

PR #1378, opened 2022-09-06
[Doc] actor En_Bombf (Bombflower) - just the last unk

krm01:
I arrived at this name after working with bomb flowers for a mod, where I wanted them to not ignite as soon as they're picked up - but instead wait for some other event. It was shockingly easy to do, and led me to believe that perhaps that is what the devs originally planned as well, as using this variable as a "lightable" state was almost all I had to do for the mod with barely any other changes.

purely speculation on my part so understand if this would be better left as unk_200 until future discoveries.

edit: changed name to ~`isActive`~`isEnabled` after discussion

### DIFF
--- a/src/overlays/actors/ovl_En_Bombf/z_en_bombf.c
+++ b/src/overlays/actors/ovl_En_Bombf/z_en_bombf.c
@@ -90,7 +90,7 @@ void EnBombf_Init(Actor* thisx, PlayState* play) {
     EnBombf* this = (EnBombf*)thisx;
 
     Actor_SetScale(thisx, 0.01f);
-    this->isActive = true;
+    this->isEnabled = true;
     Collider_InitCylinder(play, &this->bombCollider);
     Collider_InitJntSph(play, &this->explosionCollider);
     Collider_SetCylinder(play, &this->bombCollider, thisx, &sCylinderInit);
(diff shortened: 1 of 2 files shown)


fig02:
i dont think i follow the reasoning for calling it "lighteable"

the facts we have are: 
- it being true means the timer can tick down
- it being true means it can flash, spawn the visual effects, move on to explode etc

it gets set to true unconditionally on init. it also gets set to true by the bomb instance spawned if it explodes in the flower.
you can theoretically set it to false on your own in some other actor to make it "freeze". 
Some actors do this for regular bombs, but they do that by manipulating the timer specifically.

So given this, I think `isActive` is more appropriate if we are to give it a name

fig02:
I guess `isActive` can be confused for "is currently ticking down"
but `isLightable` to me makes it sound like it can be ignited by a stick on fire

so we need a way to convey "not frozen" that also doesnt say "is currently ticking" lol

krm01:
Yea I see lol.. perhaps `isFuseLit` then? my reasoning was in making a frozen (as in, ice-cold) bomb flower that doesn't auto-ignite when you pick it up, I did this by having a param type that set this value to false in init. The only reason I don't like `isFuseLit` as much is because for the flower version of the actor, it feels weird to be setting that to true in init, as it definitely is not lit. However perhaps it's just irrelevant for the flower version.

image

fig02:
yea no `fuseLit` has the same problem as `isActive`. we need to convey that its able to tick down *if its active* but it doesnt say anything about that lit/ticking state. pretty tough lol

Dragorn421:
I read the discussion here and on Discord, would `isFuseEnabled` help clear up that `isEnabled` doesn't control ability to pick up and move the bomb flower? I think it'd work for both "can be lit up" and "is fuse ticking"

Roman971:
`isFuseEnabled` also makes the most sense to me, tho `isEnabled` is probably fine as well with the comment

Dragorn421:
@fig02 @krm01 thoughts on `isFuseEnabled` ?

krm01:
changed to `isFuseEnabled`, missed the github notification on this one!

fig02 (approved):

Dragorn421 (approved):
:bomb:
:heavy_check_mark: :point_right: :boom: 
:x: :point_right: :snowflake:

jpburnett (approved):

PR #1379, opened 2022-09-06
Swap DynaPolyActor and BgActor types

mzxrules:
Something I've wanted to do for a while, finally did it. This pr also standardizes the InitBgActor functions that appear across actors


Dragorn421:
I don't necessarily disagree with switching the two terms (I don't know what the reasoning is), however (assuming we want this to be done) I feel like we may as well do it after confirming the DynaPoly and Bg system names will stick around forever, or choosing their new names?

fig02:
yeah I think 1) it would be good to detail your reasoning here
and 2) agree that we should only go through the trouble of making this switch after we agree on how we are going to name the collision systems in general

mzxrules:
Primary motivation for the swap is that the struct currently named DynaPolyActor stores no state directly related to the DynaPoly system itself, whereas the struct named BgActor stores everything needed to calculate and directly access dyna polys. The secondary motivation was to limit the concepts of static/dynamic polys to the bgcheck systems since actors don't need to access this information, but it seems unavoidable to some degree.

Dragorn421:
I think I can understand why BgActor <-> DynaPolyActor

I think we should agree to keeping "bgcheck" or "bg something" forever beforehand if we commit to this though, since this strenghtens the use of "bg" as bgcheck-collision (in particular `this->bg`)

Also while we're at it the inner-system-struct (`BgActor` in master, `DynaPolyActor` in this) probably should be renamed further to drop the "Actor" suffix, it would make things less confusing. Or renamed in some way it doesn't look like it extends the `Actor` struct

Dragorn421:
Adding the "Needs discussion" label, not really directly about the PR itself but more about the need to (re)name the bgcheck/dynapoly system(s) first, before carrying out such changes

fig02:
yea currently im of the opinion that we should wait a little bit in general with this change till we have a more clear direction for the names of things. the pr will likely just sit here for a while building up millions of conflicts while we figure that out, and it seems better to just do the find and replace when we're ready. so id say we should close it for now

fig02:
closing for now with the intention of bringing this discussion up at a later time. 
when we have a better idea of the direction the names of things in this system will go, we will be able to make a better informed decision

PR #1383, opened 2022-09-27
Various minor fixes

Roman971:
Another batch of minor fixes (mostly straightfoward stuff I noted or noticed recently):
* Swap `REGION_US` and `REGION_JP` to match country codes
* Fix a few missing `EXCH_ITEM` enum values
* Remove unnecessary casts on some alloc function calls
* Fix a `//` (double slash) in the spec file
* Swap top and bottom args in `StackCheck_Init` to match how it's called
* Remove some unnecessary `&` in array references

Also ran the formatter which fixed an unrelated issue from a recent merge

### DIFF
--- a/src/boot/z_locale.c
+++ b/src/boot/z_locale.c
@@ -10,10 +10,10 @@ void Locale_Init(void) {
 
     switch (sCartInfo.countryCode) {
         case 'J': // "NTSC-U (North America)"
-            gCurrentRegion = REGION_US;
+            gCurrentRegion = REGION_JP;
             break;
         case 'E': // "NTSC-J (Japan)"
-            gCurrentRegion = REGION_JP;
+            gCurrentRegion = REGION_US;
             break;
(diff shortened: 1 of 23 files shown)


src/boot/z_locale.c
```diff
@@ -10,10 +10,10 @@ void Locale_Init(void) {
 
     switch (sCartInfo.countryCode) {
         case 'J': // "NTSC-U (North America)"
-            gCurrentRegion = REGION_US;
+            gCurrentRegion = REGION_JP;
             break;
         case 'E': // "NTSC-J (Japan)"
-            gCurrentRegion = REGION_JP;
+            gCurrentRegion = REGION_US;
```
EllipticEllipsis, line 16: Looks like the comments need fixing too.

EllipticEllipsis (approved):

fig02 (approved):

Thar0 (approved):

PR #1384, opened 2022-09-28
Document bongo cutscene bug

fig02:
https://youtu.be/ASanPHPPMQM


Dragorn421 (approved):
Nice find, that's funny

I think the 546 value was chosen because the second bounce normally happens at that frame, and the dev(s) didn't think / testers didn't find that there was a way to delay that happening by delaying Link's descent

Though it's weird they didn't just make the timer (a lot) more lenient, since the `else { this->timer = 546; }` seems to be there to account for the player bouncing early/earlier and advancing the timer on second bounce (if the eventchkinf flag isn't set) :shrug:

EDIT: oh and when merging, make sure the comment isn't "document bug" please :sweat_smile:

fig02:
> oh and when merging, make sure the message/comment isn't the commit's "document bug" please
because we squash on all merges, personal commit messages dont matter

Roman971 (approved):

tom-overton (approved):

AngheloAlf (approved):
Great video

PR #1385, opened 2022-09-29
Use intptr types in more code files

Roman971:
Another chunk of intptr type updates (following #1301 and #1325). This one has a bit more changes, but a lot of them were straightforward u32 -> uintptr_t changes for rom addresses. I also updated `gSegments` and its corresponding macros/variables to use uintptr, as well as the code in `main.c` and `sys_cfb.c`.

Note: `ucode_disas.c` should also be updated with the `gSegments` change, but there are a lot more types to change in there so I think it should get its own pass later.

### DIFF
--- a/src/code/sys_cfb.c
+++ b/src/code/sys_cfb.c
@@ -1,11 +1,11 @@
 #include "global.h"
 
-u32 sSysCfbFbPtr[2];
-u32 sSysCfbEnd;
+uintptr_t sSysCfbFbPtr[2];
+uintptr_t sSysCfbEnd;
 
 void SysCfb_Init(s32 n64dd) {
     u32 screenSize;
-    u32 tmpFbEnd;
+    uintptr_t tmpFbEnd;
 
     if (osMemSize >= 0x800000) {
         // "8MB or more memory is installed"
@@ -42,13 +42,13 @@ void SysCfb_Reset(void) {
     sSysCfbEnd = 0;
 }
 
-u32 SysCfb_GetFbPtr(s32 idx) {
+uintptr_t SysCfb_GetFbPtr(s32 idx) {
     if (idx < 2) {
         return sSysCfbFbPtr[idx];
     }
     return 0;
 }
 
-u32 SysCfb_GetFbEnd(void) {
+uintptr_t SysCfb_GetFbEnd(void) {
     return sSysCfbEnd;
 }
--- a/src/code/sys_cfb.c
+++ b/src/code/sys_cfb.c
@@ -42,13 +42,13 @@ void SysCfb_Reset(void) {
     sSysCfbEnd = 0;
 }
 
-uintptr_t SysCfb_GetFbPtr(s32 idx) {
+void* SysCfb_GetFbPtr(s32 idx) {
     if (idx < 2) {
-        return sSysCfbFbPtr[idx];
+        return (void*)sSysCfbFbPtr[idx];
     }
-    return 0;
+    return NULL;
 }
 
-uintptr_t SysCfb_GetFbEnd(void) {
-    return sSysCfbEnd;
+void* SysCfb_GetFbEnd(void) {
+    return (void*)sSysCfbEnd;
 }
--- a/src/code/z_construct.c
+++ b/src/code/z_construct.c
@@ -27,7 +27,7 @@ void Interface_Init(PlayState* play) {
         interfaceCtx->cDownAlpha = interfaceCtx->cRightAlpha = interfaceCtx->healthAlpha = interfaceCtx->startAlpha =
             interfaceCtx->magicAlpha = 0;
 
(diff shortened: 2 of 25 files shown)


src/code/z_construct.c
```diff
@@ -83,30 +83,30 @@ void Interface_Init(PlayState* play) {
 
     if (gSaveContext.equips.buttonItems[0] < 0xF0) {
         DmaMgr_SendRequest1(interfaceCtx->iconItemSegment,
-                            (u32)_icon_item_staticSegmentRomStart + gSaveContext.equips.buttonItems[0] * 0x1000, 0x1000,
-                            "../z_construct.c", 198);
+                            (uintptr_t)_icon_item_staticSegmentRomStart + gSaveContext.equips.buttonItems[0] * 0x1000,
```
EllipticEllipsis, line 86: At some point the 0x1000s in here should get a define, although that's probably a PR on its own.

src/code/z_parameter.c
```diff
@@ -1277,7 +1277,8 @@ void Interface_LoadItemIcon1(PlayState* play, u16 button) {
 
     osCreateMesgQueue(&interfaceCtx->loadQueue, &interfaceCtx->loadMsg, 1);
     DmaMgr_SendRequest2(&interfaceCtx->dmaRequest_160, interfaceCtx->iconItemSegment + button * 0x1000,
-                        (u32)_icon_item_staticSegmentRomStart + (gSaveContext.equips.buttonItems[button] * 0x1000),
+                        (uintptr_t)_icon_item_staticSegmentRomStart +
+                            (gSaveContext.equips.buttonItems[button] * 0x1000),
                         0x1000, 0, &interfaceCtx->loadQueue, NULL, "../z_parameter.c", 1171);
```
EllipticEllipsis, line 1282: And of course here too.

EllipticEllipsis (approved):

src/code/sys_cfb.c
```diff
@@ -1,11 +1,11 @@
 #include "global.h"
 
-u32 sSysCfbFbPtr[2];
-u32 sSysCfbEnd;
+uintptr_t sSysCfbFbPtr[2];
+uintptr_t sSysCfbEnd;
```
Thar0, line 4: Can these be changed to pointer types instead?

Roman971: I tried to do that but I couldn't get anything satisfying and it had a few issues, unfortunately.

`sSysCfbEnd` doesn't work well as a pointer because of the operations done in `SysCfb_Init`, not without a ton of casts at least

`sSysCfbFbPtr` could theoretically be a pointer with a few extra casts, but it wouldn't make much sense to use pointers for one and not the other, and it doesn't feel better overall tbh

src/code/sys_cfb.c
```diff
@@ -42,13 +42,13 @@ void SysCfb_Reset(void) {
     sSysCfbEnd = 0;
 }
 
-u32 SysCfb_GetFbPtr(s32 idx) {
+uintptr_t SysCfb_GetFbPtr(s32 idx) {
     if (idx < 2) {
```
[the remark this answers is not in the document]
Roman971: I think we try to use `index` everywhere now, but we're still in the process of updating the codebase for that.
We should probably not mix this PR with other changes like these but I'll make a note of it so we don't forget.

src/code/z_construct.c
Dragorn421: #1263

Dragorn421 (approved):
Mantra for reviewing: *u32 vrom is uintptr_t because symbols are `u8[]` for abusing the linker*, *u32 rom is uintptr_t because it's the same as vrom as far as compilation goes*

fig02 (approved):
Im approving to say that I am ok with these changes being merged and dont see any surface level issues, but I am still not knowledgeable enough in pointer types in general to be able to review the correctness of their usage.

Because of this I think we should get another contributor review and approval (@tharo if your review above has been resolved possibly)

src/code/sys_cfb.c
Thar0: I think at the very least the function return types should be pointer types, in general it feels very wrong to rely on intptr types where pointer types are usable.
I agree that `sSysCfbEnd` itself being typed as a pointer would introduce a great deal of casts due to the masking and arithmetic done with/on it but as far as the external interface is concerned it should be treated as a pointer. I don't think some extra casts in sys_cfb outweigh this.

I think `debugHeapStart` in `main.c` also stands to gain from being typed as a pointer, yes it will need a cast for `PHYS_TO_K0(0x600000) - debugHeapStart` but highlighting this line as "weird" is more useful than casting between uintptr_t and void* elsewhere in the function

Roman971: Pushed a commit that changes the `sys_cfb` functions to return pointer types and makes `debugHeapStart` a pointer in `main.c`

Thar0 (approved):

Dragorn421 (approved):

PR #1386, opened 2022-10-01
Skip compiling C-included files during the build

Roman971:
Changed the makefile so it doesn't include `.inc.c` files in the files that end up compiled as `.o`, and renamed all C-included files in the repo to `.inc.c` except asset files since those need special handling and there is no way to specify the name of the C file generated by ZAPD right now (not without also affecting the name of the header at least).

This helps with build times since there is no need to compile C-included files as their `.o` is never used (and generally can't be used because it would result in name conflicts with the file that includes it as C). That is going to be particularly relevant once we start splitting asset files in different sections or by invidual assets.
And as a nice bonus, it also makes it clear from the name that a file is included as C and not used in the spec.


src/code/z_actor.c
```diff
@@ -3963,7 +3963,7 @@ s32 func_80035124(Actor* actor, PlayState* play) {
     return ret;
 }
 
-#include "z_cheap_proc.c"
```
EllipticEllipsis, line 3966: Unfortunate.

src/libultra/gu/sins.c
```diff
@@ -1,5 +1,6 @@
 #include "ultra64.h"
-#include "sintable.c"
+
+#include "sintable.inc.c"
```
EllipticEllipsis, line 3: This should be `sintable.h` anyway. And if you do it for sins, might as well do the same for atan.

Roman971: It's a file that contains data so I'm not really following why it should be a `.h`

As for atan do you mean that `sys_math_atan.c` should have its table split as well? I didn't work on any of these files so I don't know why they were handled differently, but I assume there was a reason

EllipticEllipsis: That's how it is in libultra: https://github.com/decompals/ultralib/blob/main/src/gu/sins.c

Yes, the atan table in sys_math_atan. The only difference really seems to be that it's a game library file rather than a libultra one.

EllipticEllipsis: (This is definitely correct since the source is on the CDs.)

Roman971: I see, well I don't think it's right to make it a `.h` even if that's how it was originally, but I'm fine with putting the table in the file itself like it is for sys_math_atan (or we can make a separate c file for the atan table if we prefer)

fig02 (approved):
i dont have any opinion on the location of the table being discussed above

src/libultra/gu/sins.c
AngheloAlf: Even if the original table was a `.h` I think we shouldn't put it on a `.h`.
I think putting it on a file itself would be better tho

Dragorn421: I vote to let Tharo move it to a .h eventually in his relentless efforts to clean-up libultra

AngheloAlf: @tharo ultralib wen

Thar0: As usual my thoughts are that for libultra we should try to be as compliant as possible, however I think in this case we can leave it for later

Thar0 (approved):

Dragorn421 (approved):

PR #1387, opened 2022-10-01
Fix format.py segfaults and other issues when applying fixes

Roman971:
Removed the `--style-config` option when running `clang-apply-replacements` because it can cause segfaults and it's not actually needed in our use case. Also fixed `clang-tidy` options to not include `--fix` when running with multiprocessing.

This should make the script more consistent in general and avoid the issues it had on some systems (hopefully all of them).

### DIFF
--- a/format.py
+++ b/format.py
@@ -21,7 +21,11 @@ CLANG_VER = 11
 FORMAT_OPTS = "-i -style=file"
 
 # Clang-Tidy options (see .clang-tidy for checks enabled)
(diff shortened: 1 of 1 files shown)


Dragorn421 (approved):
`clang-apply-replacements` doesn't crash for me anymore :+1:

fig02 (approved):

hensldm (approved):
Copied this over for MM, and slugmond no longer cores.

PR #1389, opened 2022-10-03
Fix many missing NULLs

Roman971:
Simply found and replaced a bunch of 0s which should be NULL since they are assigned or compared to pointers

### DIFF
--- a/src/overlays/actors/ovl_En_Horse/z_en_horse.c
+++ b/src/overlays/actors/ovl_En_Horse/z_en_horse.c
@@ -2385,7 +2385,7 @@ void EnHorse_CutsceneUpdate(EnHorse* this, PlayState* play) {
         EnHorse_Freeze(this);
         return;
     }
-    if (linkCsAction != 0) {
+    if (linkCsAction != NULL) {
         csFunctionIdx = EnHorse_GetCutsceneFunctionIndex(linkCsAction->action);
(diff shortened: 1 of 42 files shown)


src/overlays/actors/ovl_En_Horse/z_en_horse.c
```diff
@@ -3207,7 +3207,7 @@ void EnHorse_UpdateBgCheckInfo(EnHorse* this, PlayState* play) {
         }
 
         dynaPoly = DynaPoly_GetActor(&play->colCtx, bgId);
-        if ((this->stateFlags & ENHORSE_FLAG_26) && ((dynaPoly && dynaPoly->actor.id != 0x108) || dynaPoly == 0)) {
+        if ((this->stateFlags & ENHORSE_FLAG_26) && ((dynaPoly && dynaPoly->actor.id != 0x108) || dynaPoly == NULL)) {
```
EllipticEllipsis, line 3210: This is probably just a single one, so could you do the id here too?

fig02 (approved):

src/overlays/actors/ovl_En_Horse/z_en_horse.c
Roman971: Good catch, I fixed it

hensldm (approved):

PR #1390, opened 2022-10-04
Document object_bw (Torch Slug)

tom-overton:
The Torch Slug is a bit strange in that the animations *only* animate its eyestalks; the rest of the body is animated via code. So I tried to name the animations to make it very clear what they did.

Also, running the formatter added a change to BossSst? So I guess I'll just include that too.


src/overlays/actors/ovl_En_Bw/z_en_bw.h
```diff
@@ -7,11 +7,27 @@
 struct EnBw;
 typedef void (*EnBwActionFunc)(struct EnBw*, PlayState*);
 
+typedef enum {
+    /*  0 */ TORCH_SLUG_LIMB_NONE,
+    /*  1 */ TORCH_SLUG_LIMB_BODY,
+    /*  2 */ TORCH_SLUG_LIMB_LEFT_EYESTALK_ROOT,
+    /*  3 */ TORCH_SLUG_LIMB_LEFT_EYESTALK_WRAPPER,
```
Dragorn421, line 14: We don't have an agreed-upon way to handle these wrapper limbs, but this is yet another variant to handle them,

afaik the repo currently has
`_ROOT1`, `_ROOT2` (BossGoma, EnGoma)
`_ROOT_ROOT`, `_ROOT` (EnZf)

Personally I'd like either of those approach over `_ROOT`, `_WRAPPER`, it seems easier to tell what the parent is

Thoughts?

fig02 (approved):
I think it would be good to agree on a standard way to handle these limbs as mentioned above, but dont think it needs to hold up this PR so am approving it now

Roman971 (approved):

src/overlays/actors/ovl_En_Bw/z_en_bw.h
tom-overton: This scheme comes from MM, but since I'm playing in OoT's sandbox, I can acquiesce. I'll switch to doing what EnZf does.

Dragorn421: I didn't know MM used this. I'll need to read why this way was chosen

engineer124 (approved):

PR #1392, opened 2022-10-05
Fix misc 17

Dragorn421:


Ticamus (approved):

src/code/z_bgcheck.c
```diff
@@ -4287,21 +4287,18 @@ s32 WaterBox_GetSurface2(PlayState* play, CollisionContext* colCtx, Vec3f* pos,
...
+                if (waterBox->xMin < pos->x && pos->x < waterBox->xMin + waterBox->xLength) {
+                    if (waterBox->zMin < pos->z && pos->z < waterBox->zMin + waterBox->zLength) {
+                        if (pos->y - surfaceChkDist < waterBox->ySurface &&
+                            waterBox->ySurface < pos->y + surfaceChkDist) {
+                            *outWaterBox = waterBox;
+                            return i;
+                        }
+                    }
+                }
+            }
```
fig02, line 4301: ngl the continues are kinda easier to read. im not totally opposed to nesting it with ifs, but yeah would be interested in other thoughts

Dragorn421: this is to make it more similar to `WaterBox_GetSurfaceImpl` and `func_800425B0` (I don't think it's a bold claim to say copypaste has been used here)

I'd be ok with continues too, just that I would like to keep the three funcs consistent

Dragorn421: or

```c
        if ((room == play->roomCtx.curRoom.num || room == WATERBOX_ROOM_ALL) &&
            !(waterBox->properties & WATERBOX_FLAG_19) && waterBox->xMin < pos->x &&
            pos->x < waterBox->xMin + waterBox->xLength && waterBox->zMin < pos->z &&
            pos->z < waterBox->zMin + waterBox->zLength && pos->y - surfaceChkDist < waterBox->ySurface &&
            waterBox->ySurface < pos->y + surfaceChkDist) {

            *outWaterBox = waterBox;
            return i;
        }
```

```c
        if (room == play->roomCtx.curRoom.num || room == WATERBOX_ROOM_ALL) {
            if (!(waterBox->properties & WATERBOX_FLAG_19)) {
                if (waterBox->xMin < pos->x && pos->x < waterBox->xMin + waterBox->xLength && waterBox->zMin < pos->z &&
                    pos->z < waterBox->zMin + waterBox->zLength && pos->y - surfaceChkDist < waterBox->ySurface &&
                    waterBox->ySurface < pos->y + surfaceChkDist) {

                    *outWaterBox = waterBox;
                    return i;
                }
            }
        }
```

...

krm01: I actually think I prefer the continues here as well, I'm sort of ashamed to admit, with the single if statement getting my second choice vote.

though fwiw I also prefer a different boolean composition style with the operators at the start of the line, so maybe my preference only makes sense with that in mind
```c
if ((room == play->roomCtx.curRoom.num || room == WATERBOX_ROOM_ALL)
    && !(waterBox->properties & WATERBOX_FLAG_19)
    && waterBox->xMin < pos->x
    && pos->x < waterBox->xMin + waterBox->xLength
    && waterBox->zMin < pos->z
    && pos->z < waterBox->zMin + waterBox->zLength
    && pos->y - surfaceChkDist < waterBox->ySurface
    && waterBox->ySurface < pos->y + surfaceChkDist
) {
```

Dragorn421: Yeah that would be my preferred formatting style as well (operators aligned to the right) but it's not what our formatter config does, if it's even capable of doing that :(

fig02 (approved):

Roman971 (approved):
I would personally write these bgcheck conditions the natural way with a single if, but it's not very important either way

fig02:
at the end of the day im ok with either, but I do find the continues just slightly easier to read
mzx brought up that mms checks have become a bit more complex and continues cant really be used, so consistency in that regard is one reason to keep it as &s, but yea

include/z64.h
```diff
@@ -966,7 +966,7 @@ typedef enum {
     /*  6 */ TRANS_TYPE_FADE_BLACK_SLOW,
     /*  7 */ TRANS_TYPE_FADE_WHITE_SLOW,
     /*  8 */ TRANS_TYPE_WIPE_FAST,
-    /*  9 */ TRANS_TYPE_FILL_WHITE2,
```
hensldm, line 969: Not sure if this got talked about, but based on https://github.com/zeldaret/oot/blob/master/src/code/z_play.c#L554-L559 this looks to have the intention of being `TRANS_TYPE_FILL_WHITE_FAST`, though because of the bug it just acts like a second copy of `TRANS_TYPE_FILL_WHITE`.

Dragorn421: No that wasn't noticed in the transition PR iirc
You are most likely right, but what do we make of this?

hensldm: Personally I like the idea of renaming it based on this intent, probably with a comment that it doesn't work because of the bug.

Dragorn421: Why not

I would like this to be done in another PR though, this one already has enough stuff and also reviews (and the waterbox stuff delayed it already)

PR #1393, opened 2022-10-05
Split `SaveContext` into sub-structs

Dragorn421:
As planned since #431 (about 2 years ago)

I am sure the increased verbosity will only have fans

The bulk of the changes are find-and-replace + formatting. See z64save.h and z_sram.c for actual changes/motivations

Note: Decomp tutorial needs further amending depending on what we end up doing


Dragorn421:
also
```c
// The use of ALIGNED8 here is just a temporary solution until the SaveContext is re-structured
ALIGNED8 SaveContext gSaveContext;
```
How does re-structuring help with this? Was there plans for a union with a u64 or something?

Dragorn421:
WIP with macros:
https://github.com/zeldaret/oot/compare/master...Dragorn421:oot:doc_savectx_struct_macros_wip2
e.g.
```c
                SET_RESPAWN_A0(RESPAWN_MODE_TOP, GET_RESPAWN_A0(RESPAWN_MODE_DOWN));
                SET_RESPAWN_A0_PLAYERPARAMS(RESPAWN_MODE_TOP, 0x06FF);
                SET_RESPAWN_A0_DATA(RESPAWN_MODE_TOP, 40);

if (GET_LANGUAGE == LANGUAGE_ENG) {

if (GET_INVENTORY_DUNGEONKEYS_A0(GET_MAPINDEX) <= 0) {
```
...

hensldm:
FWIW, I know OoT would prefer to get save context macros (especially where the `(void)0` trick is needed). At least in MM we do have everything and it doesn't seem terrible. (Might even be worth it for the long names to have more shift-ability with the save context)

fig02:
I think in an ideal world we would just do the macros, but do understand that practically it wont happen for a while. Am willing to change my stance on this matter for the sake of compromise and shiftability as mentioned

fig02 (approved):
Fully making my stance known that Ive changed opinion on this, and think the substructs should be used until we can further discuss and plan if/how we use macros 

(pr has a crazy amount of conflicts, would understand if it would be easier to just make a new one lol)

hensldm (approved):
SaveContext shift-ability soon™️?

Dragorn421:
Merged master into this branch, two things to check before merging:

> also
> 
> ```c
> // The use of ALIGNED8 here is just a temporary solution until the SaveContext is re-structured
> ALIGNED8 SaveContext gSaveContext;
> ```
> 
> How does re-structuring help with this? Was there plans for a union with a u64 or something?

(or just have this pr remove the comment idk)

and check if the decomp tutorial is updated

Besides these points I'd like to merge this

I don't like introducing macros for the save and I'm not going to purse it, though I made PoCs for it, but discussion can keep going after this is merged if needed

AngheloAlf:
I think leaving the `ALIGNED8` macro for `gSaveContext` is fine.

My opinion in the macros is to only have macros for accesses like `((void)0, blah)`, having macros for every member sounds overkill

fig02:
Is using proper structs supposed to remove the need for the alignment somehow? thats what your old comment seems to suggest

Dragorn421:
> Is using proper structs supposed to remove the need for the alignment somehow? thats what your old comment seems to suggest

It's not my comment, it's a comment which is currently in master
And I'm wondering the same thing about how is it supposed to allow remove the alignment

fig02:
sorry it was directed to anghleo, who wrote the comment originally

AngheloAlf:
Oh wtf, I have no recall of writing that comment, but git blame points to me lol.

I don't see how using proper structs could allow removing the macro, I may have misunderstood something at the time

fig02:
alrighty
Im ok with merging now and not worrying about the tutorial, but if someone wants to double check feel free

AngheloAlf (approved):
LGTM

PR #1394, opened 2022-10-05
`PlayerActionParam` -> `PlayerItemAction`

Dragorn421:
As initiated by the following MM PR: https://github.com/zeldaret/mm/pull/1069

I would have liked to rename `ACTORCAT_ITEMACTION` first but have absolutely no idea what to rename it to.


hensldm (approved):
Same as in MM, I agree with this change.

AngheloAlf (approved):
Hopefully will be able to figure out proper names for the actorcategories

fig02:
given that this is a pretty big terminology change, I think @Roman971 should review this as well even though my approval technically meets the merge requirements

(also Id like to formally call dibs on documenting the item/upper body system within player, the func_82C system etc. I did alot of research on it a while back to add an item into the game for spaceworld and feel I could document it well)

fig02 (approved):

Roman971 (approved):

PR #1398, opened 2022-10-09
Document elf_message scripting system 

fig02:
I wanted to rename the elf_message system to something else for https://github.com/zeldaret/oot/pull/1391, but it kinda scope creeped into redoing most of the docs in the file.

This PR documents the remaining unused types and moves away from the "elf message" terminology to using "hints" instead, with there being a separation of "navi hints" and "saria hints".

The command args were shuffled around a bit to hopefully make more sense as you read the "scripts". I wanted to re-do the commands a bit so that the command type was the macro name instead of the thing being checked, but it didnt exactly work out.

The unused "random" hint type is unhinged btw


include/z64hint_commands.h
```diff
@@ -0,0 +1,148 @@
...
+ * Byte pack macros
+ */
+
+#define HINT_B0(type, condType, tf)            \
+    _SHIFTL(HINT_CMD_TYPE_##type, 5, 3) |      \
+    _SHIFTL(HINT_CONDITION_##condType, 1, 4) | \
+    _SHIFTL(tf, 0, 1)
+
+#define HINT_B1(condType, data)                \
+    _SHIFTL(HINT_CONDITION_##condType, 4, 4) | \
```
Dragorn421, line 72: I know that's how it was before, but this is an opportunity to not use `##` concatenation, as we've been avoiding it in order to facilitate finding things (e.g. with search-all)

src/code/z_scene.c
```diff
@@ -231,8 +231,8 @@ void Scene_CommandSpecialFiles(PlayState* play, SceneCmd* cmd) {
         gSegments[5] = VIRTUAL_TO_PHYSICAL(play->objectCtx.status[play->objectCtx.subKeepIndex].segment);
     }
 
-    if (cmd->specialFiles.cUpElfMsgNum != 0) {
-        play->cUpElfMsgs = Play_LoadFile(play, &sNaviMsgFiles[cmd->specialFiles.cUpElfMsgNum - 1]);
+    if (cmd->specialFiles.naviHintFileId != NAVI_HINTS_NONE) {
+        play->naviHints = Play_LoadFile(play, &sNaviHintFiles[cmd->specialFiles.naviHintFileId - 1]);
```
Dragorn421, line 235: It seems more consistent/sensible to use "hints" as plural basically everywhere?
```suggestion
    if (cmd->specialFiles.naviHintsFileId != NAVI_HINTS_NONE) {
        play->naviHints = Play_LoadFile(play, &sNaviHintsFiles[cmd->specialFiles.naviHintsFileId - 1]);
```

this may apply to the z64hint_commands.h filename too, not sure

src/code/z_scene.c
```diff
@@ -498,7 +498,7 @@ void (*gSceneCmdHandlers[SCENE_CMD_ID_MAX])(PlayState*, SceneCmd*) = {
     Scene_CommandMiscSettings,        // SCENE_CMD_ID_MISC_SETTINGS
 };
 
-RomFile sNaviMsgFiles[] = {
+RomFile sNaviHintFiles[] = {
     ROM_FILE(elf_message_field),
     ROM_FILE(elf_message_ydan),
     ROM_FILE_UNSET,
```
Dragorn421, line 504: As an aside, afaik there's no evidence `sNaviHintFiles[2]` exists, and the code doesn't plan for `ROM_FILE_UNSET` anyway

So I would think `sNaviHintFiles[2]` is actually an unk (or file padding?)

src/code/z_elf_message.c
```diff
@@ -1,60 +1,68 @@
...
-    ELF_MSG_MEDALLION(CHECK, 0x66, false, ITEM_MEDALLION_FIRE),
-    ELF_MSG_MEDALLION(CHECK, 0x66, false, ITEM_MEDALLION_WATER),
-    ELF_MSG_SONG(CHECK, 0x67, false, ITEM_SONG_STORMS),
-    ELF_MSG_MEDALLION(CHECK, 0x68, false, ITEM_MEDALLION_SPIRIT),
-    ELF_MSG_MEDALLION(CHECK, 0x68, false, ITEM_MEDALLION_SHADOW),
-    ELF_MSG_END(0x69),
+#include "z64hint_commands.h"
+
+HintCmd sChildSariaHints[] = {
+    HINT_CMD_STRENGTH_UPG(SKIP, 0, false, 3),
```
Dragorn421, line 5:
```suggestion
    HINT_CMD_STRENGTH_UPG(SKIP, 0, false, 0x03),
```
(textId)

src/code/z_elf_message.c
```diff
@@ -64,111 +72,126 @@ u32 ElfMessage_CheckCondition(ElfMessage* msg) {
...
     }
 }
 
-u32 func_8006BF1C(ElfMessage** msgp) {
-    ElfMessage* msg = *msgp;
-    u32 sp44[10];
-    s32 temp1 = 0;
-    s32 temp2 = 0;
-    s32 temp3;
+u32 Hints_CheckRandomCondition(HintCmd** hintsPtr) {
```
Dragorn421, line 95:
```suggestion
u32 Hints_CheckConditionRandom(HintCmd** hintsPtr) {
```
for consistency with `Hints_CheckConditionChain`

Dragorn421 (changes requested):
Very nice :+1:

src/code/z_elf_message.c
fig02: this is number of entries to skip, like a branch. its confusing cause it usually is a text id unless the command type is SKIP lol

src/code/z_elf_message.c
fig02: I dont think the consistency is really needed here and prefer having it read like a normal sentence. its not like the two are related and need to be consistent

src/code/z_scene.c
fig02: sticking a plural in the middle there just doesnt really sound right to me. "the hint file is a file of hints" seems like the natural phrasing to me

src/code/z_scene.c
fig02: file padding sounds right, object table stuff should prob be in its own. would be good to get some other version's approval

                0x0000000080127508                sNaviMsgsFiles
                0x0000000080127520                gLinkObjectIds
                0x0000000080127524                gObjectTableSize
                0x0000000080127528                gObjectTable
                
 MM also agrees

fig02: we can investigate this further and make this change in another pr

src/code/z_elf_message.c
fig02: eh, I guess I dont really care that much. but it does bother me that "check condition chain" is the normal english sentence and "check condition random" is not. so its pick your poison with inconsistency

src/code/z_elf_message.c
```diff
@@ -1,60 +1,68 @@
...
+    HINT_CMD_MEDALLION(HINT_CMD_TYPE_CHECK, ITEM_MEDALLION_WATER, false, 0x6B),
+    HINT_CMD_MEDALLION(HINT_CMD_TYPE_CHECK, ITEM_MEDALLION_SPIRIT, false, 0x6C),
+    HINT_CMD_MEDALLION(HINT_CMD_TYPE_CHECK, ITEM_MEDALLION_SHADOW, false, 0x6C),
+    HINT_CMD_END(0x6D),
 };
 
-u32 ElfMessage_CheckCondition(ElfMessage* msg) {
-    s32 type = msg->byte0 & 0x1E;
+u32 Hints_CheckCondition(HintCmd* hints) {
+    s32 type = hints->byte0 & 0x1E;
```
Thar0, line 31: Macros relating these masks/shifts to the byte packing macros in `z64hint_commands.h` may make it easier for someone to edit the system?
Not sure if it's realistic however as there's a lot to cover here

Thar0 (changes requested):

src/code/z_elf_message.c
```diff
@@ -1,60 +1,68 @@
...
-    ELF_MSG_MEDALLION(CHECK, 0x66, false, ITEM_MEDALLION_WATER),
-    ELF_MSG_SONG(CHECK, 0x67, false, ITEM_SONG_STORMS),
-    ELF_MSG_MEDALLION(CHECK, 0x68, false, ITEM_MEDALLION_SPIRIT),
-    ELF_MSG_MEDALLION(CHECK, 0x68, false, ITEM_MEDALLION_SHADOW),
-    ELF_MSG_END(0x69),
+#include "z64hint_commands.h"
+
+HintCmd sChildSariaHints[] = {
+    HINT_CMD_STRENGTH_UPG(HINT_CMD_TYPE_SKIP, 0, false, 3),
+    HINT_CMD_FLAG(HINT_CMD_TYPE_CHECK, EVENTCHKINF_37, false, 0x61),
```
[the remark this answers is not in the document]
EllipticEllipsis: Especially since it's literally 2 files that use them.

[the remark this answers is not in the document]
fig02: beside searchability, imo the concat just makes things harder to understand. When diving into this system for the first time yesterday and tried to learn how it worked, I had to keep asking myself "where the hell is this coming from" and reverse engineer the macro constructs to find out how it works. imo its unnecessary

I dont agree that its hurting read and writeability, its a few extra characters on the type and thats it

src/code/z_elf_message.c
fig02: I agree macros should be made, but Im not exactly sure how id approach it. If someone else wants to I think itd be good.
But also, this system is kinda ass and anyone modding is probably going to rip it out like mm did, so Im not the most concerned about usability from that point of view lmao. But from a documentation standpoint, yea agree

src/code/z_elf_message.c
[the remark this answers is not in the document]
Thar0: If you want to understand how the macros work then yeah it adds a small layer of complexity, but that's so that using the macros is easier. It's also not especially difficult to see where these come from by inspecting the macro definitions, for the user that has great need to know how it's structured internally. Simplifying how the macros appear in the arrays should be of utmost priority, considering almost all macros use this field.

I wouldn't claim it hurts readability if it doesn't actually impact my reading of these scripts in practice. I think it's far messier than it was before, at any rate.
14 extra characters per line when each line is on average 72 characters is quite a lot (about 20%)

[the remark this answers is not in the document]
fig02: to use the system youre gonna need to understand how each part works anyway. when you see CHECK everywhere in the commands and go to search that on its own, its not gonna take you to the information you need. match case floods you with bgcheck defines, match whole word doesnt bring you there cause the definition is not a whole word. its just a mess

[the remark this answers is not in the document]
Thar0: Searching isn't the right tool for the job in the first place, if you use go to definition and see how the macro uses it you get there right away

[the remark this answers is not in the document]
fig02: you cant go to the definition of CHECK anyway cause intellisense isnt going to resolve what its concatenated from. I want to be able to directly go and see what check means without jumping through hoops

idk we will probably just need more opinions to tie breaker this, cause we just see this very differently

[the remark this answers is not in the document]
mzxrules: given that `HINT_CMD_TYPE_SKIP` is only used once I wonder if type should be incorporated into the macro name instead of being an argument.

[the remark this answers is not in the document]
fig02: this isnt only related to skip, its for all the command types
I wanted to make the command type be the actual macro as opposed to an arg, but the thing that is being checked against (flag/item/song, whatever) determines what the command structure is (as in, what combination of B0/B1/B packing arrangement it has)

So, its not really viable at least not in a clean way. At least as far as I know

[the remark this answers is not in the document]
Roman971: I'm generally against concatenating enum names with macros, but in cases like this I think it's fine because you would never need to search/find the full enum name unless you're looking at the internal system implementation, at which point you would need to look at and understand the macros anyway. It's very different from something like the item enum names which you need to use both in the macro and in other places in the code when using the system "externally"

tools/disasm_elf_msg.py
```diff
@@ -210,32 +210,32 @@ def disas_elfmsgs(start):
...
-                print(f"ELF_MSG_FLAG({ARG_0}, {ARG_1}, {ARG_2}, 0x{b1:02X}), /* eventChkInf[{(b1 >> 4) & 0xF}] & 0x{1 << (b1 & 0xF):X} */")
+                print(f"HINT_CMD_FLAG({ARG_0}, {ARG_1}, {ARG_2}, 0x{b1:02X}), /* eventChkInf[{(b1 >> 4) & 0xF}] & 0x{1 << (b1 & 0xF):X} */")
             assert b3 == 0
         elif condition_type == 2:
-            print(f"ELF_MSG_DUNGEON_ITEM({ARG_0}, {ARG_1}, {ARG_2}, {item_ids[b1]}),")
+            print(f"HINT_CMD_DUNGEON_ITEM({ARG_0}, {ARG_1}, {ARG_2}, {item_ids[b1]}),")
             assert b3 == 0
         elif condition_type == 4:
-            print(f"ELF_MSG_ITEM({ARG_0}, {ARG_1}, {ARG_2}, {item_ids[b1]}, {item_ids[b3]}),")
+            print(f"HINT_CMD_ITEM({ARG_0}, {ARG_1}, {ARG_2}, {item_ids[b1]}, {item_ids[b3]}),")
```
Roman971, line 221: It looks like this script needs to be updated further since the argument order was changed in the macros

src/elf_message/elf_message_field.c
```diff
@@ -1,33 +1,33 @@
...
-    ELF_MSG_FLAG(CHECK, 0x54, false, EVENTCHKINF_AA),
-    ELF_MSG_ITEM(CHECK, 0x55, true, ITEM_LENS, ITEM_NONE),
-    ELF_MSG_MEDALLION(CHECK, 0x57, false, ITEM_MEDALLION_SHADOW),
-    ELF_MSG_SONG(CHECK, 0x58, false, ITEM_SONG_REQUIEM),
-    ELF_MSG_STRENGTH_UPG(CHECK, 0x56, true, 1),
-    ELF_MSG_MEDALLION(CHECK, 0x5A, false, ITEM_MEDALLION_SPIRIT),
-    ELF_MSG_ITEM(CHECK, 0x5B, true, ITEM_ARROW_LIGHT, ITEM_NONE),
-    ELF_MSG_FLAG(CHECK, 0x5C, false, EVENTCHKINF_C3),
-    ELF_MSG_END(0x5F),
+static HintCmd sOverworldNaviHints[] = {
```
Roman971, line 4: I think it makes more sense to keep those global, even tho the symbol isn't used directly, since the data is technically used globally and having something used globally as a static variable feels wrong/confusing

include/z64scene.h
```diff
@@ -469,6 +469,14 @@ typedef enum {
 #define SCENE_CAM_TYPE_FIXED_MARKET 0x40 // Camera exhibits fixed behaviors and delays textboxes by a small amount before they start to appear
 #define SCENE_CAM_TYPE_SHOOTING_GALLERY 0x50 // Unreferenced in code, and used only by the main layer of the shooting gallery scene
 
+// navi hints
+// TODO: make ZAPD use this enum for `SCENE_CMD_SPECIAL_FILES`
+typedef enum {
+    NAVI_HINTS_NONE,
+    NAVI_HINTS_OVERWORLD,
+    NAVI_HINTS_DUNGEON
+} NaviHintFileId;
```
Roman971, line 478: Having an enum for this feels unnecessary and somewhat incorrect since it's really just an index in `sNaviHintFiles`, tho it's 1-indexed, as indicated by the - 1 in `Scene_CommandSpecialFiles`, but I guess it's debatable

src/code/z_elf_message.c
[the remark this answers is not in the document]
fig02: as a compromise I can bring it back for the cmd type which does matter for the scripts, but I dont feel its necessary to concat the enum names used internally in the macro that you cant see from the script anyway

src/code/z_elf_message.c
```diff
@@ -64,111 +72,126 @@ u32 ElfMessage_CheckCondition(ElfMessage* msg) {
...
-        *msgp += 1;
+
+        *hintsPtr += 1;
     }
 
     return false;
 }
 
-u16 ElfMessage_GetTextFromMsgs(ElfMessage* msg) {
+u16 Hints_GetTextIdFromScript(HintCmd* hints) {
```
Roman971, line 134: It's not perfect either but I think the argument here should be kept singular since it's really a pointer to the current hint being iterated over. Same comment for the other functions
```suggestion
u16 Hints_GetTextIdFromScript(HintCmd* hint) {
```

src/code/z_elf_message.c
```diff
@@ -64,111 +72,126 @@ u32 ElfMessage_CheckCondition(ElfMessage* msg) {
...
                 ASSERT(0, "0", "../z_elf_message.c", 282);
         }
-        msg++;
+
+        hints++;
     }
 }
 
-u16 ElfMessage_GetSariaText(PlayState* play) {
+u16 Hints_GetSariaTextId(PlayState* play) {
```
Roman971, line 174: Aren't saria hints provided through navi? In a way I would argue this justifies treating the system as "navi hints", which are differentiated between saria and C-Up ones, and I think this is what I was going for with the previous naming.

In general I don't know how I feel about having this system as simply "hints" since there are many other types of hints in the game, including gossip stones, some of the npc dialog, navi hovering certain objects, the indicators in the pause menu map, etc
So even if we don't go with "navi hints" I feel like we should use something more specific

fig02: elliptic and mzx both separately suggested QuestHint as a system name, which imo is not necessary but if others feel "hints" is too generic then I can do that.

I do feel strongly about keeping navis hints named as navi though. Yes sarias hints are provided through navi, but it is still very clearly hints from saria. I dont want to call the navi quest hints after the name of a button on the controller. And the button on the hud literally says navi on it anyway, they are navi hints imo

src/code/z_elf_message.c
fig02: its a pointer to the script that has many hints, singular sounds very awkward here

src/elf_message/elf_message_field.c
Dragorn421: Note this data is not in code, it's loaded separately

include/z64scene.h
fig02: the -1 is precisely my reasoning for doing this, you cant claim that the number provided to the scene command is an index when its not the value that actually indexes the list. its more of an id where 0 is none

src/code/z_elf_message.c
Dragorn421: The text ids referenced by Saria hints are Saria talking, even if it's Navi telling (btw I never knew you weren't directly speaking to Saria)

I think Hints as a system name is enough, there are other hints part of the game design but no such dedicated system. (I would be okay with another name, as you want)

src/code/z_scene.c
Dragorn421: "hint file" vs "file with hints", alright

btw with naming the system "Hints" why is all this using "navi hints"

fig02: this stuff only pertains to navis hints, ie those accessed with `Hints_GetNaviTextId`. sarias hints are seperate. I may not understand your question though

src/code/z_elf_message.c
Dragorn421: bruh got me

Could you explain that on `HINT_CMD_TYPE_SKIP` ?
like "Uses textId as the amount of commands to skip (including this one)"

src/code/z_elf_message.c
Dragorn421: "check random condition" also sounds like "check a random condition" (a randomly choosen condition) to me instead of "check the random type of condition"

but honestly, whatever

include/z64scene.h
Roman971: It's 1-indexed but it's still an index since the only thing it's used for is to index into the array

src/code/z_elf_message.c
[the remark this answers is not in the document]
Dragorn421: Would it be more bearable if `HINT_CMD_TYPE_` was shortened

otherwise I'm okay with fig's compromise

src/code/z_elf_message.c
fig02: it is a randomly chosen condition. this type is unused cause its really dumb tbh, but the idea is you have a bunch of conditions of this random type and it will choose one of them randomly and then check it. if that randomly chosen one passes then that hint is given

src/code/z_scene.c
Dragorn421: You understood correctly, I'm stupid, nevermind

src/code/z_elf_message.c
Roman971: It's only that when first entering the function, but it gets iterated over inside the function (similarly to how effects are often handled inside actors, see fishing for example). In cases like this I think it's better to go with the iterator name since it's always "correct" and avoids confusion

fig02: we'll need to get more opinions cause I dont agree personally, it reads fine to me as plural

src/code/z_elf_message.c
Dragorn421: oh okay I badly misunderstood what this does

alright whatever

include/z64scene.h
fig02: I dont agree but fair enough
I change it back to calling it an index in the code, but I think the enum is still nicer and easier to understand as dragorn has shown above

src/code/z_elf_message.c
fig02: I can, yeah
This type of thing is why I wanted to make the cmd type be the actual macro, so SKIP could get its own arg name and stuff, but I could not figure out how to make it work

src/code/z_elf_message.c
Dragorn421: I think it stays valid as a `hints` iterator throughout anyway, as long as the initial `hints` is a valid string of hint commands then all pointer values taken by `hints` point to a valid string of hint commands

include/z64scene.h
EllipticEllipsis: The actor table has enums, the object table has enums, the scene table has enums, I assume the entrance table has enums although with that who knows. I don't think being primarily an index into an array, albeit a small and little-used one, precludes it from having an enum.

Dragorn421: If that weighs anything, all modding tools use a dropdown list for this: none/overworld/dungeon

src/code/z_elf_message.c
EllipticEllipsis: One could argue that the targeting + C-Up system is equally well a hint system (and not ab initio just for enemies, but it is an actor hint system at least), although it does not (need to) use scripting. (There's also the ones on the world map screen although I can understand if people don't think they should count.) I think it's more sensible to be specific when this is concerned, is someone looking at the use going to instantly understand what it is without reading the files?

fig02: yea Im not against quest hint, I just personally dont see it as that necessary, especially when the cost is much larger enum names/function names throughout. But after this conversation I will change it to quest hint

On the topic of navi vs c up though, navi enemy descriptions is another reason to why imo it makes sense to make the seperation within this system as navi and saria. It clearly seperates who is the one giving you the hint, as opposed to the button on the controller which also activates hints from a separate system

src/elf_message/elf_message_field.c
fig02: I have no opinion on this matter personally, and will do whatever the consensus is. I inherited this change from dragorns initial PR that I build my changes on top of

src/code/z_elf_message.c
Roman971: The argument points to the current `HintCmd` being iterated over, so it makes sense to have it as `hint` or `hintCmd`

Note that some of the other functions only deal with one cmd so those are a no-brainer

include/z64scene.h
Roman971: To be clear I'm not saying it being an index prevents it from being associated with an enum, but I think this is a case where it's unnecessary as the values are never referenced in code and having names for them doesn't serve much purpose (and thus might not be worth having)

src/code/z_elf_message.c
fig02: its only `Hints_CheckCondition` that evaluates a single command without iterating.
Especially in this function you highlighted, the point is to pass in a whole script to be processed and not a specific command. And the only other two functions take HintCmd** so they are already named differently

include/z64scene.h
fig02: alright. but yeah, even if they arent referenced in code (other than 0/none), it makes the usage in scene files alot clearer so it is worth it in my opnion

src/code/z_elf_message.c
Roman971: My point is these functions may take a pointer to an array, but the argument is actually iterated over, so having it named as the array is confusing when looking at the function's execution (`hints++`, `hints--`, etc)

fig02: hints++ and hints-- makes perfect sense to me, its incrementing the pointer to the list of hints to move onto the next one.
my previous comment was meant to convey that `Hints_CheckCondition` is the only one where a single hint command is evaluated, while you said "some" as if theres multiple. This one function is the only one where I understand the argument to make it singular

fig02: would a compromise be to call it hintList or something? it at least conveys the part that I find important, which is that its a pointer to a list of them. And it would be analogus to display lists, an array of commands, that get incremented on all the time

Roman971: This is strange to me since we always name iterators singular under normal circumstances (eg. when iterating over actors or objects), so I don't really see why this case is different other than the fact that it points to the list when entering the function (but not once it starts iterating)

Also fwiw so far we've been going for arguments named singular in these cases, tho it's probably not 100% consistent

fig02: I just cannot get on board with a function designed to process a script of many hints taking a single hint as an argument. I 100% get it for `Hints_CheckCondition` and can change it there

Dragorn421: iterating over actors and objects, every entry is its single instance, the hints script is an organized string of entries that gets processed sequentially, and entries can be related to each other (like with the SKIP or unused RANDOM ones)

src/elf_message/elf_message_field.c
fig02: tho it is important to keep in mind that we treat things in other dma'd files that arent always accessed directly like scenes and objects as global. all of these should prob get the same treatment

tools/disasm_elf_msg.py
fig02: I talked with tharo about this before opening the PR. think it would probably be easier for him to do it since he wrote it and knows how it works better. And it seems that even before my changes it got a little out of date in some regards.

So I would prefer it gets updated seperate from this PR, but if others want it changed now we can work toward making that happen

include/z64scene.h
```diff
@@ -469,6 +469,14 @@ typedef enum {
 #define SCENE_CAM_TYPE_FIXED_MARKET 0x40 // Camera exhibits fixed behaviors and delays textboxes by a small amount before they start to appear
 #define SCENE_CAM_TYPE_SHOOTING_GALLERY 0x50 // Unreferenced in code, and used only by the main layer of the shooting gallery scene
 
+// navi hints
+// TODO: make ZAPD use this enum for `SCENE_CMD_SPECIAL_FILES`
+typedef enum {
+    NAVI_QUEST_HINTS_NONE,
+    NAVI_QUEST_HINTS_OVERWORLD,
+    NAVI_QUEST_HINTS_DUNGEON
+} NaviHintFileId;
```
Dragorn421, line 478:
```suggestion
} NaviQuestHintFileId;
```
if we keep this

Dragorn421 (approved):
:+1: and make a note to update `disasm_elf_msg.py` eventually

src/code/z_elf_message.c
Roman971: QuestHint seems good to me yeah, it should help distinguishing the different types of hints

src/code/z_elf_message.c
fig02: Yeah I think at this point I just want to hear more opinions other than us 3. cause we are all pretty clearly set on how we think this should be handled, and just need more insight on how others think we should continue. If there are others who think it should be `hint` instead of `hints` then I will move forward with that change, but I still dont agree with it

fig02:
`hints` arg changed to `hintCmd`, I believe that was the last debate to settle for this PR

Dragorn421 (approved):

src/code/z_elf_message.c
```diff
@@ -64,111 +73,126 @@ u32 ElfMessage_CheckCondition(ElfMessage* msg) {
...
 }
 
-u32 func_8006BF1C(ElfMessage** msgp) {
-    ElfMessage* msg = *msgp;
-    u32 sp44[10];
-    s32 temp1 = 0;
-    s32 temp2 = 0;
-    s32 temp3;
+u32 QuestHint_CheckRandomCondition(QuestHintCmd** hintCmdPtr) {
+    QuestHintCmd* hints = *hintCmdPtr;
```
Roman971, line 97:
```suggestion
    QuestHintCmd* hintCmd = *hintCmdPtr;
```

fig02: whoops

Roman971 (approved):

Thar0 (approved):

PR #1401, opened 2022-10-11
Document all non-Koume/Kotake parts of object_tw

tom-overton:
I had documented `object_wiz` in MM, and Fig noted that Wizrobe uses "the good twinrova fire". This made me curious, and upon further inspection, it turns out that Twinrova reuses a ton of assets from Wizrobe (in reality, it was the other way around, but I documented `object_wiz` first). This made me want to document `object_tw`, but @hensldm *also* wanted to document parts of `object_tw` since he did Koume and Kotake's objects in MM. We decided to split the object up, and this is my part of it; it includes all the magic effects, as well as the combined Twinrova.

I should note that the effect stuff in this object is the single-messiest thing I have ever seen in a Z64 object, and it seemingly defies good categorization. Hopefully it makes sense to someone other than myself.


hensldm (approved):
LTGM, I'll finish up the object when this gets in :) .

fig02 (approved):

Roman971 (approved):

PR #1410, opened 2022-10-16
Objects tw and tr

hensldm:
Finishes up the work started in #1401 by documenting the Kotake and Koume parts of object_tw, as well as their other object `object_tr`


assets/xml/objects/object_tw.xml
```diff
@@ -1,144 +1,154 @@
...
+        <DList Name="gTwinrovaKoumeHeadDL" Offset="0x134B8"/>
+        <DList Name="gTwinrovaKotakeIceHairDL" Offset="0x13AE8"/>
+        <DList Name="gTwinrovaKoumeFireHairDL" Offset="0x13D68"/>
+        <DList Name="gTwinrovaKotakeIceBroomHeadDL" Offset="0x13E98"/>
+        <DList Name="gTwinrovaKotakeIceBroomHeadOuterDL" Offset="0x13F98"/>
+        <DList Name="gTwinrovaKotakeFireBroomHeadDL" Offset="0x14070"/>
+        <DList Name="gTwinrovaKotakeFireBroomHeadOuterDL" Offset="0x14158"/>
+
+        <!-- Kotake DLs -->
+        <Blob Name="object_tw_Blob_015200" Size="0x60" Offset="0x15200"/>
```
tom-overton, line 110: Is this Blob a DL? Otherwise, I'd move it out of this section.

assets/xml/objects/object_tw.xml
```diff
@@ -1,144 +1,154 @@
...
-        <DList Name="object_tw_DL_0184C8" Offset="0x184C8"/>
-        <DList Name="object_tw_DL_0185E8" Offset="0x185E8"/>
-        <DList Name="object_tw_DL_018710" Offset="0x18710"/>
-        <DList Name="object_tw_DL_018830" Offset="0x18830"/>
-        <DList Name="object_tw_DL_018990" Offset="0x18990"/>
-        <DList Name="object_tw_DL_018998" Offset="0x18998"/>
-        <DList Name="object_tw_DL_0189A0" Offset="0x189A0"/>
+
+        <!-- Koume DLs -->
+        <Blob Name="object_tw_Blob_0176D0" Size="0x60" Offset="0x176D0"/>
```
tom-overton, line 133: Same thing with this one.

tom-overton (approved):
Just two minor things

src/overlays/actors/ovl_Boss_Tw/z_boss_tw.h
```diff
@@ -55,6 +55,37 @@ typedef enum {
     /* 26 */ FWORK_MAX
 } TwFwork;
 
+typedef enum {
+    /*  0 */ TWINROVA_KOTAKE_KOUME_LIMB_NONE,
+    /*  1 */ TWINROVA_KOTAKE_KOUME_LIMB_PELVIS,
+    /*  2 */ TWINROVA_KOTAKE_KOUME_LIMB_LEFT_THIGH,
+    /*  3 */ TWINROVA_KOTAKE_KOUME_LIMB_LEFT_SHIN,
+    /*  4 */ TWINROVA_KOTAKE_KOUME_LIMB_LEFT_FOOT,
```
fig02, line 63: I find the super long "system" name of this enum to be pretty unfortunate. 
perhaps `TWINROVA_SINGLE_` and `TWINROVA_COMBINED_` or something can work
my first instinct was `TWINROVA_WITCH_` or something to identify the split variant, but idk if I like that

Dragorn421: Could remove `TWINROVA_`?

fig02: it is a bit weird tho cause twinrova is the larger "actor" system name, which I think should be the same between the two

Dragorn421: Note the xml does use gKotakeKoume for some stuff

hensldm: One other reason to not remove `TWINROVA` is to help avoid naming collisions with `object_tr` stuff. I'm curious if I should split the enum, since they are technically different skeletons between the two (though I would still need to think of a better name for the assets shared between the two in the xml as Dragorn pointed out).

fig02: I was going to say they should probably be different, but within override and post limb draw, it is shared between the two. so a seperate enum wouldnt really help anyway

and yea i also think the long names in assets is unfortunate, but I dont have any good suggestions for now and wont hold it up cause of that

fig02 (approved):

Roman971 (approved):
Looks good to me

PR #1411, opened 2022-10-17
"NpcInfo" / `struct_80034A14_arg1` talkState docs

luryus:
I've been looking at documenting more NPCs and ran into this system that some NPCs use to handle talking to the player. More specifically I just renamed and cleaned up the "talkState" field that indicates the "state" of the NPC dialog, and documented the function that NPCs use to update the state.

This field has four possible values that it gets set to in the game. Shared functions in `z_actor` only care about 0 / 1, the other values are checked and used only by the NPC actors themselves.
- 0: not talking / idle
- 1: talking
- 2: "action": some NPC actors use this state to indicate some triggered action during the dialog (e.g. giving an item, running an animation or setting some flags)
- 3: "item given": used in a few places to indicate that an item was given to the player and the "get item" textbox is done

Suggestions for better names are more than welcome! 😅 

 <details>
 <summary>A list of the uses for the "action" state (2). This is likely incomplete or even incorrect, I just quickly skimmed over these trying to understand how the state gets used:</summary>
 
- EnMd: Move away from blocking a path
- EnMa3: Unused (copied from EnMa2)
- EnMa2: Unset auto-talk actor flag after playing Epona's song
- EnMa1: Give weird egg, teach Epona's song
- EnHy: Not used (only set, but never checked)
- EnZo: Not used (only set, but never checked)
- EnTk (dampe): Dig
- EnKz (King zora):
	- Give tunic
	- GIve bottle back, mweep
	- Give frog
- EnKo: Fado giving saw
- EnIn: set flags / transition in/out of riding / continue to next animation / ...
- EnGo2:
	- Biggoron:
		- Using eyedrops
		- Giving BGS, Prescription
	- Big rolling
		* End talk, continue rolling
	* Bomb flower goron
		* Continue animation
	* Link
		* Give tunic
		* "Ask about..."
- EnGo:
	* Giving out item, etc
- EnDu (Darunia): give bracelet, set cutscene mode
 </details>

### DIFF
--- a/include/functions.h
+++ b/include/functions.h
@@ -498,8 +498,8 @@ void func_8003424C(PlayState* play, Vec3f* arg1);
 void Actor_SetColorFilter(Actor* actor, s16 colorFlag, s16 colorIntensityMax, s16 xluFlag, s16 duration);
 Hilite* func_800342EC(Vec3f* object, PlayState* play);
 Hilite* func_8003435C(Vec3f* object, PlayState* play);
-s32 func_800343CC(PlayState* play, Actor* actor, s16* arg2, f32 interactRange,
-                  u16 (*unkFunc1)(PlayState*, Actor*), s16 (*unkFunc2)(PlayState*, Actor*));
+s32 Actor_NpcUpdateTalking(PlayState* play, Actor* actor, s16* talkState, f32 interactRange,
+                           ActorNpcGetTextIdFunc getTextId, ActorNpcGetTalkStateFunc getTalkState);
 s16 func_800347E8(s16 arg0);
 void func_80034A14(Actor* actor, struct_80034A14_arg1* arg1, s16 arg2, s16 arg3);
 void func_80034BA0(PlayState* play, SkelAnime* skelAnime, OverrideLimbDraw overrideLimbDraw,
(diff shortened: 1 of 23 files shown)


fig02:
thanks for starting to look into this.
I do think that there is a larger system at play here, and it should probably all be documented at once. For example at this time, I dont feel comfortable calling this struct a generic "npc info", when there is another struct intertwined in this system (struct_80116130/struct_80116130_0)

It seems like this should all be unraveled at once (`func_800344BC`/`D_80116130`/`func_800347E8`/`func_80034810`/`func_80034A14`
which if you dont want to do now thats fair. I was going to take a look at it eventually when I do a documentation pass of z_actor

luryus:
> For example at this time, I dont feel comfortable calling this struct a generic "npc info", when there is another struct intertwined in this system (struct_80116130/struct_80116130_0)

Note that I did not actually rename the struct, just the single state variable. I used "NpcInfo" in the title as that name is already used somewhere in the codebase and it's at least a better working name than `struct_80034A14_arg1` 😄

I think the talk state field is pretty clear and could be documented by itself. But I completely get your point too, and it's fine to just close this for now.

Luckily this is a pretty small set of changes and something I wanted to learn about anyway (so that I can understand and try to document the NPCs that use this), so no work really gone to waste here 😅

luryus:
I'll close this for now and document the whole struct and the related functions in a new PR later - they don't seem to be too complicated either.

PR #1415, opened 2022-10-22
More documentation for `z_std_dma.c`

Thar0:


src/code/z_map_exp.c
```diff
@@ -159,10 +159,10 @@ void Map_InitData(PlayState* play, s16 room) {
...
                          mapIndex, VREG(30));
             osSyncPrintf(VT_RST);
-            DmaMgr_SendRequest1(play->interfaceCtx.mapSegment,
-                                (uintptr_t)_map_i_staticSegmentRomStart +
-                                    ((gMapData->dgnMinimapTexIndexOffset[mapIndex] + room) * 0xFF0),
-                                0xFF0, "../z_map_exp.c", 346);
+            DmaMgr_SyncDmaRequestDebug(play->interfaceCtx.mapSegment,
+                                       (uintptr_t)_map_i_staticSegmentRomStart +
+                                           ((gMapData->dgnMinimapTexIndexOffset[mapIndex] + room) * 0xFF0),
+                                       0xFF0, "../z_map_exp.c", 346);
```
EllipticEllipsis, line 165: Is there a sensible sizeof or define to use for the 0xFF0?

src/code/z_message_PAL.c
```diff
@@ -1166,18 +1166,19 @@ void Message_LoadItemIcon(PlayState* play, u16 itemId, s16 y) {
         R_TEXTBOX_ICON_XPOS = R_TEXT_INIT_XPOS - sIconItem32XOffsets[gSaveContext.language];
         R_TEXTBOX_ICON_YPOS = y + 6;
         R_TEXTBOX_ICON_SIZE = 32;
-        DmaMgr_SendRequest1(msgCtx->textboxSegment + MESSAGE_STATIC_TEX_SIZE,
-                            (uintptr_t)_icon_item_staticSegmentRomStart + (itemId * 0x1000), 0x1000,
-                            "../z_message_PAL.c", 1473);
+        DmaMgr_SyncDmaRequestDebug(msgCtx->textboxSegment + MESSAGE_STATIC_TEX_SIZE,
+                                   (uintptr_t)_icon_item_staticSegmentRomStart + (itemId * 0x1000), 0x1000,
```
EllipticEllipsis, line 1170: Similar question for the ones in this file.

src/overlays/misc/ovl_kaleido_scope/z_kaleido_scope_PAL.c
```diff
@@ -2441,13 +2443,13 @@ void KaleidoScope_LoadDungeonMap(PlayState* play) {
...
-                        "../z_kaleido_scope_PAL.c", 3467);
+    DmaMgr_SyncDmaRequestDebug(interfaceCtx->mapSegment,
+                               (uintptr_t)_map_48x85_staticSegmentRomStart + (R_MAP_TEX_INDEX * 2040), 2040,
+                               "../z_kaleido_scope_PAL.c", 3467);
 
-    DmaMgr_SendRequest1(interfaceCtx->mapSegment + 0x800,
-                        (uintptr_t)_map_48x85_staticSegmentRomStart + ((R_MAP_TEX_INDEX + 1) * 2040), 2040,
-                        "../z_kaleido_scope_PAL.c", 3471);
+    DmaMgr_SyncDmaRequestDebug(interfaceCtx->mapSegment + 0x800,
+                               (uintptr_t)_map_48x85_staticSegmentRomStart + ((R_MAP_TEX_INDEX + 1) * 2040), 2040,
```
EllipticEllipsis, line 2451: 0x7F8's a pretty weird number but should probably still be in hex given its closeness to the 0x800.

src/boot/z_std_dma.c
```diff
@@ -197,7 +265,13 @@ void DmaMgr_Error(DmaRequest* req, const char* file, const char* errorName, cons
     Fault_AddHungupAndCrashImpl(buff1, buff2);
 }
 
-const char* DmaMgr_GetFileNameImpl(uintptr_t vrom) {
+/**
+ * Searches the filesystem for the entry containing the address `vrom`.
+ *
+ * @param vrom vrom address
```
EllipticEllipsis, line 271:
```suggestion
 * @param vrom virtual rom address
```
Might as well expand the acronym a bit for clarity rather than be repetitive.

src/boot/z_std_dma.c
```diff
@@ -217,46 +291,55 @@ const char* DmaMgr_GetFileNameImpl(uintptr_t vrom) {
...
         osSyncPrintf("DMA ROM:%08X RAM:%08X SIZE:%08X %s\n");
-        // the last arg of this print looks like it may be filename, but
-        // filename above this block does not match
     }
 
+    // Get the filename (for debugging)
     filename = DmaMgr_GetFileName(vrom);
-    iter = gDmaDataTable;
 
+    // Iterate through the DMA data table until the region containing the request vrom address is found
```
EllipticEllipsis, line 327:
```suggestion
    // Iterate through the DMA data table until the region containing the request's vrom address is found
```

src/boot/z_std_dma.c
```diff
@@ -325,19 +427,35 @@ void DmaMgr_ThreadEntry(void* arg) {
...
 }
 
-s32 DmaMgr_SendRequestImpl(DmaRequest* req, void* ram, uintptr_t vrom, u32 size, u32 unk, OSMesgQueue* queue,
-                           OSMesg msg) {
+/**
+ * Submits a DMA request to the DMA manager. For internal use only.
+ *
+ * @param req DMA request, filled out internally.
+ * @param ram Location in DRAM for data to be written.
+ * @param vrom Virtual ROM location for data to be read.
```
EllipticEllipsis, line 451: Or follow this function's more verbose style.

src/boot/z_std_dma.c
```diff
@@ -415,27 +545,50 @@ void DmaMgr_Init(void) {
...
+ * operation has completed.
+ *
+ * @param req DMA request structure, filled out internally.
+ * @param ram Location in DRAM for data to be written.
+ * @param vrom Virtual ROM location for data to be read.
+ * @param size Transfer size.
+ * @param queue Message queue to notify with `msg` once the transfer is complete.
+ * @param msg Message to send to `queue` once the transfer is complete.
+ * @param file Debug filename of caller
+ * @param line Debug line number of caller
```
EllipticEllipsis, line 576: There's inconsistency in this function documentation (and the rest of this file) in whether the descriptions end with `.`s.

EllipticEllipsis (commented):
Another interesting round of docs. Few minor things.

include/z64animation.h
```diff
@@ -10,7 +10,7 @@ struct Actor;
 struct SkelAnime;
 
 #define LINK_ANIMATION_OFFSET(addr, offset) \
-    (((u32)_link_animetionSegmentRomStart) + ((u32)addr) - ((u32)_link_animetionSegmentStart) + ((u32)offset))
+    (((uintptr_t)_link_animetionSegmentRomStart) + ((uintptr_t)(addr)) - ((uintptr_t)_link_animetionSegmentStart) + ((uintptr_t)(offset)))
```
hensldm, line 13:
```suggestion
    (((uintptr_t)_link_animetionSegmentRomStart) + ((uintptr_t)(addr)) - ((uintptr_t)_link_animetionSegmentStart) + ((u32)(offset)))
```
Not sure I'd consider an offset a pointer so `uintptr_t` doesn't seem appropriate. Also it seems to match fine in its one use without a cast on the offset maybe just remove it?

src/boot/z_std_dma.c
```diff
@@ -367,14 +486,23 @@ s32 DmaMgr_SendRequestImpl(DmaRequest* req, void* ram, uintptr_t vrom, u32 size,
...
+s32 DmaMgr_SyncDmaRequest(void* ram, uintptr_t vrom, size_t size) {
     DmaRequest req;
     OSMesgQueue queue;
     OSMesg msg;
     s32 ret;
 
     osCreateMesgQueue(&queue, &msg, 1);
-    ret = DmaMgr_SendRequestImpl(&req, ram, vrom, size, 0, &queue, NULL);
+    ret = DmaMgr_SendRequest(&req, ram, vrom, size, 0, &queue, NULL);
     if (ret == -1) {
```
hensldm, line 506: Not sure if it is worth mentioning the `DmaMgr_SendRequest` cannot return `-1`. The return 0 was confusing at first due to this.

src/boot/z_std_dma.c
```diff
@@ -217,46 +291,55 @@ const char* DmaMgr_GetFileNameImpl(uintptr_t vrom) {
...
+    const char* ret = DmaMgr_FindFileName(vrom);
 
     if (ret == NULL) {
         return "(unknown)";
     }
 
-    if (DmaMgr_CompareName(ret, "kanji") == 0 || DmaMgr_CompareName(ret, "link_animetion") == 0) {
+    if (DmaMgr_StrCmp(ret, "kanji") == 0 || DmaMgr_StrCmp(ret, "link_animetion") == 0) {
+        // Refuse kanji and link_animetion files. This may be related to these files being too large to be loaded all
+        // at once, however a NULL filename does not prevent them from being loaded.
```
Dragorn421, line 302: since

> a NULL filename does not prevent them from being loaded.

, I feel like

> Refuse kanji and link_animetion files.

is misleading. I'd suggest just removing that "refuse ... files" sentence, it's pretty clear what this does

src/boot/z_std_dma.c
```diff
@@ -217,46 +291,55 @@ const char* DmaMgr_GetFileNameImpl(uintptr_t vrom) {
...
-    u32 romSize;
+    size_t romSize;
     u8 found = false;
     DmaEntry* iter;
     const char* filename;
 
     if (0) {
-        // the string is defined in .rodata but not used, suggesting
-        // a debug print is here but optimized out in some way
+        // The string is defined in .rodata but not used, suggesting a debug print is here but wa optimized out in some
```
Dragorn421, line 319:
```suggestion
        // The string is defined in .rodata but not used, suggesting a debug print is here but was optimized out in some
```

src/boot/z_std_dma.c
```diff
@@ -367,14 +486,23 @@ s32 DmaMgr_SendRequestImpl(DmaRequest* req, void* ram, uintptr_t vrom, u32 size,
     return 0;
 }
 
-s32 DmaMgr_SendRequest0(void* ram, uintptr_t vrom, u32 size) {
+/**
+ * Submit a synchronous DMA request. This will block the current thread until the requested transfer is complete. Data
+ * is immediately available once the current thread resumes execution.
```
Dragorn421, line 491:
```suggestion
 * is immediately available when the function returns.
```
details about the thread resuming are technically correct but probably useless to the end user imo

Dragorn421 (commented):
Really nice job

And good catch about the fault handler used in `DmaMgr_Init` before code is loaded :mag_right:

src/overlays/misc/ovl_kaleido_scope/z_kaleido_map_PAL.c
```diff
@@ -710,7 +710,8 @@ void KaleidoScope_DrawWorldMap(PlayState* play, GraphicsContext* gfxCtx) {
         gDPSetPrimColor(POLY_OPA_DISP++, 0, 0, 150, 255, 255, pauseCtx->alpha);
         gDPSetEnvColor(POLY_OPA_DISP++, 0, 0, 0, 0);
 
-        POLY_OPA_DISP = KaleidoScope_QuadTextureIA8(POLY_OPA_DISP, pauseCtx->nameSegment + 0x400, 80, 32, 4);
+        POLY_OPA_DISP = KaleidoScope_QuadTextureIA8(POLY_OPA_DISP, pauseCtx->nameSegment + MAP_NAME_TEX1_SIZE,
```
Dragorn421, line 713: all `nameSegment + MAP_NAME_TEX1_SIZE` should be `nameSegment + MAX(MAP_NAME_TEX1_SIZE, ITEM_NAME_TEX_SIZE)`, cf `KaleidoScope_UpdateNamePanel`

src/overlays/misc/ovl_kaleido_scope/z_kaleido_scope_PAL.c
```diff
@@ -1140,7 +1140,8 @@ void KaleidoScope_DrawInfoPanel(PlayState* play) {
                 gDPSetPrimColor(POLY_OPA_DISP++, 0, 0, 255, 255, 255, 255);
             }
 
-            POLY_OPA_DISP = KaleidoScope_QuadTextureIA4(POLY_OPA_DISP, pauseCtx->nameSegment, 128, 16, 0);
+            POLY_OPA_DISP = KaleidoScope_QuadTextureIA4(POLY_OPA_DISP, pauseCtx->nameSegment, MAP_NAME_TEX1_WIDTH,
+                                                        MAP_NAME_TEX1_HEIGHT, 0);
         }
```
Dragorn421, line 1145: at least comment these are also `ITEM_NAME_TEX_WIDTH`/height ?

src/code/z_construct.c
```diff
@@ -179,7 +186,7 @@ void Message_Init(PlayState* play) {
 
     View_Init(&msgCtx->view, play->state.gfxCtx);
 
-    msgCtx->textboxSegment = GameState_Alloc(&play->state, 0x2200, "../z_construct.c", 349);
+    msgCtx->textboxSegment = GameState_Alloc(&play->state, MESSAGE_STATIC_TEX_SIZE + ICON_ITEM_TEX_SIZE + 0x200, "../z_construct.c", 349);
```
Dragorn421, line 189: ->
```
MESSAGE_STATIC_TEX_SIZE + MAX(MAX(ICON_ITEM_TEX_SIZE, ICON_ITEM_24_TEX_SIZE), 2 * MESSAGE_TEXTURE_STATIC_TEX_SIZE)
```
cf `Message_LoadItemIcon`, `Message_Decode`

include/z64dma.h
```diff
@@ -21,4 +22,28 @@ typedef struct {
...
+
+#define DMAMGR_DEFAULT_BUFSIZE ALIGN16(0x2000)
+
+// Standard DMA Requests
+
+s32 DmaMgr_SyncDmaRequest(void* ram, uintptr_t vrom, size_t size);
+s32 DmaMgr_SyncDmaRequestDebug(void* ram, uintptr_t vrom, size_t size, const char* file, s32 line);
+s32 DmaMgr_AsyncDmaRequest(DmaRequest* req, void* ram, uintptr_t vrom, size_t size, u32 unk5, OSMesgQueue* queue,
+                           OSMesg msg, const char* file, s32 line);
+
```
fig02, line 38: my gut feeling is that these names feel a little "backward" to me. 
It fees to me like the name should be structured like DmaMgr_ThingType, so `DmaRequestAsync`/`DmaRequestSync`
And I also kinda feel like "Dma" being there is a bit redundant with the system name already specifying it (`DmaMgr_RequestAsync`/`DmaMgr_RequestSync`/`DmaMgr_RequestSyncDebug`)

not 100% sold on any of these names, but just throwing my thoughts out there

Dragorn421: I did some research on conventions out there for sync vs async vs (nothing), prefix vs suffix, and there is no big common pattern like `Impl` can sometimes be

https://www.reddit.com/r/csharp/comments/s6w5fv/why_do_people_add_async_to_method_names_in_new/ summarizes it best:
> Ehh. Ask 100 devs and you'll probably get 200 different answers.

I personally ... don't have an opinion

-------

As for "Dma" I must say I could see it omitted indeed, `DmaMgr_SyncRequest` looks fine, but I'm also fine if it stays

src/overlays/misc/ovl_kaleido_scope/z_kaleido_scope_PAL.c
```diff
@@ -1140,7 +1140,8 @@ void KaleidoScope_DrawInfoPanel(PlayState* play) {
                 gDPSetPrimColor(POLY_OPA_DISP++, 0, 0, 255, 255, 255, 255);
             }
 
-            POLY_OPA_DISP = KaleidoScope_QuadTextureIA4(POLY_OPA_DISP, pauseCtx->nameSegment, 128, 16, 0);
+            POLY_OPA_DISP = KaleidoScope_QuadTextureIA4(POLY_OPA_DISP, pauseCtx->nameSegment, ITEM_NAME_TEX_WIDTH,
+                                                        ITEM_NAME_TEX_HEIGHT, 0);
```
Dragorn421, line 1144: these dimensions are also `MAP_NAME_TEX1_WIDTH`, `MAP_NAME_TEX1_HEIGHT` (`nameSegment+0` may hold either), worth a comment imo

Thar0: I don't really understand what the conditions for this code to run are. Can it run for both item_name and map_name textures, or is it intended for one or the other? If it's only intended for one I think whichever one it is should be used and it's not worth a comment (unless there's a bug)

Dragorn421: basically, this code runs most of the time, it's the label at the bottom of the pause menu that shows on every page (except when the cursor is on the "special positions" for switching page)

Don't be deceived by `cursorItem`/`namedItem`, it doesn't always hold items, it holds a map/area index when on the pause menu's world map page

Thar0: I see, thanks.
I've added some comments about this assumption the code makes

Dragorn421 (approved):

fig02 (approved):

PR #1416, opened 2022-10-22
Move object table from `z_scene.c` to new file `object_table.c`

Dragorn421:
To stay up-to-date with MM and I wanted to do this anyway since `sNaviQuestHintFiles[2]` was sus as file padding

- https://github.com/zeldaret/mm/pull/1136


EllipticEllipsis (approved):

fig02:
dlftbl is their term for overlay table, which these are not. So I'm not sure it's exactly the best thing to go with. It also makes it sound like the scene table 

This will eventually get a better name anyway, so I won't argue it too much, but I don't think it's the best idea

Roman971:
"dlftbl" is assumed to be short for "DLF table", with DLF being their term for overlay (theorized as Dynamically Loaded File), which is why it's used for the actor, effect and game state overlay tables. But it doesn't make sense for this file since it's not for overlays, so I think we should have a different name for this file. Also we should probably go with something based on objects rather than scenes since that's what  the file is about

AngheloAlf:
Explaining a bit more my reasoning on naming this file like this:
- MM can't put the table on `z_scene.c` because of data paddings which show the file boundary.
- It doesn't make sense to put the table in the next file either.
- I didn't want to name the new file `code_XXXXXXXX` because we know exactly its contents.
- The equivalent actor table lives on `z_actor_dlftbls.c`, with the main difference that file contains a few functions and this one doesn't.
- The name `z_scene_dlftbls.c` follow the "the files are in alphabetical order in the spec" logic.
And that's basically it.

I agree the name is not ideal, but I think it is fine for now unless anybody comes up with a better name.

On the other hand I would argue objects are dynamically loaded files too, because, well, they get loaded dynamically lol . So I think they can be considered overlays, just they lack the "ovl_" prefix and have an `object_` prefix instead

fig02:
The dlftbls scheme is very clearly meant for overlays by their usage. Overlays are pretty clearly defined as files that use the relocation system when loaded, which these are not at all.

We are not arguing against a separate file, it makes sense given the padding shown in both oot and mm. We can hopefully agree on a more suitable name

Dragorn421:
Renamed the file to `object_table.c` following the topic being raised on Discord

@AngheloAlf

fig02 (approved):

krm01 (approved):

Roman971 (approved):

PR #1418, opened 2022-10-25
DoorShutter logic docs

Dragorn421:
Time for another define-declared struct Pepe HYPERS

<details>

<summary>
Graph of the main behavior funcs
</summary>

graph showing which functions call which among the main behavior functions in the DoorShutter actor

</details>


krm01 (approved):
the whole #define for a struct thing is super weird to me, I don't really get why it's necessary at all and it's one of those things I would rather just have a little code duplication than trying to macro it (if that's the reason it exists). I feel like the fix is worse than the problem now with this macro leaking out into z64actor.h

but anyway like I said I don't really understand why this was done so I may be missing something, but all the rest of the changes LGTM

Dragorn421:
The define hack is to accomodate for the fact that player accesses both `DoorGerudo + 0x164` and `DoorShutter + 0x164` with the same line (and 0x164 is outside of the base Actor struct),
so these two actors' structs have a field in common and it must stay that way for the game to not break

So ideally DoorGerudo and DoorShutter would use as a substruct `SlidingDoorActorBase`, but that doesn't work due to struct alignment.
Using the substruct introduces two bytes of padding at 0x166.
To make things even worse, 0x166 is a u8 for DoorGerudo, and a s16 for DoorShutter, so it can't be in `SlidingDoorActorBase` either (or as a union but we ruled out that solution previously with the same hack done for `DoorActorBase`)

The two solutions left to prevent the DoorGerudo and DoorShutter structs to go out of sync from `SlidingDoorActorBase`, is to define the common struct layout as a macro, or just write a comment about it to copypaste any changes made to the other structs

src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
```diff
@@ -798,7 +874,7 @@ void DoorShutter_GohmaBlockFall(DoorShutter* this, PlayState* play) {
         if (!GET_EVENTCHKINF(EVENTCHKINF_70)) {
             BossGoma* parent = (BossGoma*)this->dyna.actor.parent;
 
-            this->unk_164 = 10;
+            this->moveState = 10;
```
engineer124, line 877: So `moveState` is used both as a timer and as a bool to indicate in-use? Does it make sense for player to set the timer to 1 instead of a bool being set to true?

src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
```diff
@@ -773,18 +846,21 @@ void func_809973E8(DoorShutter* this, PlayState* play) {
...
 void DoorShutter_JabuDoorClose(DoorShutter* this, PlayState* play) {
     if (Math_StepToS(&this->jabuDoorClosedAmount, 100, 10)) {
-        func_80997220(this, play);
+        DoorShutter_SetupClosed(this, play);
     }
 }
 
-void func_80997568(DoorShutter* this, PlayState* play) {
-    if (this->unk_16F++ > 30) {
+void DoorShutter_WaitPlayerShock(DoorShutter* this, PlayState* play) {
```
engineer124, line 862: `PlayerShock` made me think this was related to an electric shock in some way. Presumably this is when the player turns around, and surprised about the door being locked? Maybe
`DoorShutter_WaitPlayerSurpriseTurnAroundAnim`? Bit verbose though, might not need all those words

Dragorn421: Probably WaitPlayerSurprised is enough and indeed less ambiguous 👍 ?

src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
```diff
@@ -349,39 +357,43 @@ s32 DoorShutter_SetupDoor(DoorShutter* this, PlayState* play) {
...
                 doorType = SHUTTER_FRONT_CLEAR;
+            } else if (doorType == SHUTTER_BOSS) {
+                doorType = SHUTTER_BACK_LOCKED;
             } else {
-                doorType = (doorType == SHUTTER_BOSS) ? SHUTTER_BACK_LOCKED : SHUTTER;
+                doorType = SHUTTER;
             }
         }
     }
     this->gfxType = (doorType == SHUTTER) ? styleInfo->gfxType1 : styleInfo->gfxType2;
```
hensldm, line 375: Wasn't you, but what does `SHUTTER` by itself mean? Also could possibly give more descriptive names than `type1` and `type2` based on this ternary?

src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
hensldm: Perhaps this would be a good use of a union, since this one variable does seem to have two different usages (1 as a boolean and 1 as a timer)?

Dragorn421: Note `moveState` is already sketchy from being from the `SLIDING_DOOR_ACTOR_BASE` macro

I'm not opposed to making a union but I don't think it's necessary

The name isn't completely inappropriate either, gohma block / phantom ganon bars use it as a movement-related timer, so "move state" isn't that bad (right? COPIUM)

src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
Dragorn421: Afaict it's just "normal" but I'm honestly not sure.

The block of code just above with `frontRoom` I'm not sure I understand either. I'm waiting on documentation for transition actors to be merged to investigate room loading and this transition actor "side" usage

src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
hensldm: I will say I would probably prefer a union, more so the player usage with a boolean named `moveState` seems strange. Or perhaps just name the boolean more boolean like, and just comment the hacky nature for `gohma block / phantom ganon bars'.

(I will say that this isn't a deal breaker, so will approve anyway :) )

hensldm (approved):

src/overlays/actors/ovl_Door_Shutter/z_door_shutter.c
fig02: the gohma case does seem more of an afterthought, would agree with naming it after its main purpose and commenting on its re-use

Dragorn421: Does "moveState" not work for the player interaction?

I was thinking of it as "when the player interacts with the door, get the door to move"
(and it's not far from gohmablock/phantomganonbars using it as a movement timer)

What would be a better name?

hensldm: I'd prefer something like `isUsing` or maybe `isMoving`.

The main thing about `moveState` is thinking of a `true` moveState just seems strange. It seems more of something that needs an enum associated with it, rather than acting as a boolean.

Dragorn421: -> `isActive`

include/z64actor.h
```diff
@@ -527,6 +527,16 @@ typedef struct DoorActorBase {
     /* 0x0000 */ DOOR_ACTOR_BASE;
 } DoorActorBase;
 
+// DoorShutter and DoorGerudo share isActive
+// Due to alignment, a substruct cannot be used in the structs of these actors.
+#define SLIDING_DOOR_ACTOR_BASE      \
+    /* 0x0000 */ DynaPolyActor dyna; \
+    /* 0x0164 */ s16 isActive // Set to true by player when using the door
```
fig02, line 534: `// Set to true by player when using the door. Also as a timer for other niche cases in DoorShutter`
its probably worth a mention, but dont think extensive details are needed. if one is curious they can just search it up

fig02 (approved):

fig02:
we've met merge requirements, but dunno if you want to approve your suggestions @engineer124

engineer124 (approved):
lgtm

PR #1419, opened 2022-10-26
`EnHoll` docs (door-less transition between rooms)

Dragorn421:
Document EnHoll, also known as "transition planes".
However I moved away from calling it "plane" anywhere, since it's more of a cuboid or cylinder shaped volume, and just used "holl"

------------

`EnHollType` usage stats:

| `EnHollType` | Total uses count | Amount of scenes it's used in|
|-|-|-|
|`ENHOLL_H_VISIBLE_NARROW`   | 49  | 9                     |
|`ENHOLL_V_DOWN_FADE_LARGE`  | 3   | 1 (Fire temple)       |
|`ENHOLL_V_INVISIBLE`        | 33  | 8                     |
|`ENHOLL_H_FADE_SWITCHFLAG`  | 4   | 1 (Dampe's grave)     |
|`ENHOLL_H_INVISIBLE`        | 101 | 13                    |
|`ENHOLL_V_FADE`             | 1   | 1 (Haunted wasteland) |
|`ENHOLL_H_INVISIBLE_NARROW` | 8   | 1 (Water temple)      |


src/code/z_play.c
```diff
@@ -1190,7 +1190,7 @@ void Play_Draw(PlayState* this) {
                 }
 
                 if ((HREG(80) != 10) || (HREG(84) != 0)) {
-                    Environment_FillScreen(gfxCtx, 0, 0, 0, this->unk_11E18, FILL_SCREEN_OPA);
+                    Environment_FillScreen(gfxCtx, 0, 0, 0, this->enHollFillAlpha, FILL_SCREEN_OPA);
```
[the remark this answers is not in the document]
Dragorn421: I wouldn't mind, though I think it makes sense to name it after the actor, like, "this is how (some) holls darken the area when traversing them"

not "bg" though, that's overused
`roomShapeDarkenAlpha` ? :shrug:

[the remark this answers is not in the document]
fig02: its definitely generic enough of a thing that it doesnt need to be named after holl, imo
I think background cover describes its purpose extremely well. its a case where im fine with bg despite its collision ties. its pretty clear its not bgcheck related imo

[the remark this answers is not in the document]
Dragorn421: Changed to `bgCoverAlpha`

src/overlays/actors/ovl_En_Holl/z_en_holl.c
```diff
@@ -152,162 +213,186 @@ void func_80A58DD4(EnHoll* this, PlayState* play) {
...
+    Vec3f relSubjectPos;
+    s32 isKokiriLayer8;
+    f32 hollHalfWidth;
+    f32 orthogonalDistToSubject;
+
+    func_8002DBD0(&this->actor, &relSubjectPos, useViewEye ? &play->view.eye : &player->actor.world.pos);
+    hollHalfWidth =
+        (ENHOLL_GET_TYPE(&this->actor) == ENHOLL_H_INVISIBLE_NARROW) ? ENHOLL_H_HALFWIDTH_NARROW : ENHOLL_H_HALFWIDTH;
+
+    isKokiriLayer8 = EnHoll_IsKokiriLayer8();
```
engineer124, line 228: While you're here, I'm looking at the function:
```c
s32 EnHoll_IsKokiriLayer8(void) {
    return gSaveContext.entranceIndex == ENTR_SPOT04_0 && gSaveContext.sceneLayer == 8;
}
```
And Fig's Cutscene PR:
```c
case CS_DEST_KOKIRI_FOREST_FROM_BLUE_WARP:
    play->nextEntranceIndex = ENTR_SPOT04_0;
    gSaveContext.cutsceneIndex = 0xFFF8;
    play->transitionTrigger = TRANS_TRIGGER_START;
    play->transitionType = TRANS_TYPE_FADE_WHITE;
    gSaveContext.nextTransitionType = TRANS_TYPE_FADE_WHITE;
    break;
```
Methinks `EnHoll_IsKokiriLayer8` can be named to something like `EnHoll_IsKokiriForestEntranceFromBlueWarp` or something similar.

src/overlays/actors/ovl_En_Holl/z_en_holl.c
```diff
@@ -324,20 +409,20 @@ void EnHoll_Update(Actor* thisx, PlayState* play) {
...
             gfxP = POLY_OPA_DISP;
-            setupDlIdx = SETUPDL_37;
+            setupDlIndex = SETUPDL_37;
         } else {
             gfxP = POLY_XLU_DISP;
-            setupDlIdx = SETUPDL_0;
+            setupDlIndex = SETUPDL_0;
         }
-        gfxP = Gfx_SetupDL(gfxP, setupDlIdx);
+        gfxP = Gfx_SetupDL(gfxP, setupDlIndex);
```
engineer124, line 425:
```suggestion
        gfxP = Gfx_SetupDL(gfxP, setupDLIndex);
```
So it's like `Gfx_SetupDL_XXX`?

src/overlays/actors/ovl_En_Holl/z_en_holl.c
engineer124: Oh wait, did I mess this up, it might need `0xFFF4` for the cutsceneIndex, brb

fig02: this is cutscene 8 which is layer 12:

0 child day
1 child night
2 adult day
3 adult night
4 cutscene 0
...
12 cutscene 8

engineer124: Thanks, brain not functioning tonight

src/overlays/actors/ovl_En_Holl/z_en_holl.c
Dragorn421: sure, I often miss Dl vs DL

src/overlays/actors/ovl_En_Holl/z_en_holl.c
Dragorn421: Nice try though :sunglasses:

src/overlays/actors/ovl_En_Holl/z_en_holl.c
```diff
@@ -80,11 +109,10 @@ s32 EnHoll_IsKokiriLayer8(void) {
 }
 
 void EnHoll_ChooseAction(EnHoll* this) {
-    s32 action;
+    s32 action = ENHOLL_GET_TYPE(&this->actor);
```
hensldm, line 112:
```suggestion
    s32 type = ENHOLL_GET_TYPE(&this->actor);
```
type seems to make more sense now.

hensldm (approved):
1 small nit

src/overlays/actors/ovl_En_Holl/z_en_holl.c
```diff
@@ -2,26 +2,75 @@
...
+// actor, in the actor's local z direction.
+
+// Defines the height of horizontal holls (all kinds) with a Y range relative to the actor position.
+#define ENHOLL_H_Y_MIN -50.0f
+#define ENHOLL_H_Y_MAX 200.0f
+
+// Defines the width of horizontal holls (all kinds),
+// with a half-width extending on both lateral sides (towards local +x and -x).
+#define ENHOLL_H_HALFWIDTH_NARROW 100.0f
+#define ENHOLL_H_HALFWIDTH 200.0f
```
fig02, line 16: wonder if itd make sense to put these defines near where the action functions for the different types are (if it even works out that way order wise in the file)

the massive wall of 60 lines of defines at the top is kinda unfortunate, but not the end of the world lol

src/overlays/actors/ovl_En_Holl/z_en_holl.c
```diff
@@ -116,34 +144,67 @@ void EnHoll_SwapRooms(PlayState* play) {
     play->roomCtx.unk_30 ^= 1;
 }
 
-// Horizontal Planes
-void func_80A58DD4(EnHoll* this, PlayState* play) {
+/**
+ * When traversing a holl of this kind, it attempts to:
+ *   Load Current Room (fails as it is already loaded)
```
fig02, line 149: not saying it needs to change now, but hopefully we come up with a better word than "holl" in the future lol

src/overlays/actors/ovl_En_Holl/z_en_holl.c
```diff
@@ -116,34 +144,67 @@ void EnHoll_SwapRooms(PlayState* play) {
...
+     * [3] : End of fade region region, where the plane is fully transparent
+     *
+     * Within the fade region, the plane goes:
+     *   opaque -> transparent if approaching,
+     *   transparent -> opaque if receding
+     */
+    static f32 sTriggerDists[2][4] = {
+        { 200.0f, 150.0f, 100.0f, 50.0f }, // default
+        { 100.0f, 75.0f, 50.0f, 25.0f },   // SCENE_JYASINZOU
+    };
```
fig02, line 177: does this need to be in-static to match? if not id be ok with moving out just for the sake of not cluttering the function itself with the large explanation of the data

src/overlays/actors/ovl_En_Holl/z_en_holl.h
```diff
@@ -4,6 +4,33 @@
...
+ *     For the player to move horizontally between rooms.
+ *     Cuboid-shaped, 200 units high plus 50 units into the ground, 400 units wide or, if narrow, 200 units wide.
+ *   Vertical (V):
+ *     For the player to move vertically between rooms.
+ *     Cylinder-shaped, radius of 120 units.
+ *
+ * Three kinds of visual effects:
+ *   Visible: A plane is drawn at the holl location, and fades away as the player approaches.
+ *   Fade: The whole screen fades black as the player approaches.
+ *   Invisible: Approaching the holl does not cause any particular visual effect.
```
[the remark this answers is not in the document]
fig02: also, may be good to overall refer to it as BG_FADE?
when reading most of the code in this review, i read "fade" as the types that themselves fade, which you have called "visible" currently

src/overlays/actors/ovl_En_Holl/z_en_holl.c
```diff
@@ -116,34 +144,67 @@ void EnHoll_SwapRooms(PlayState* play) {
...
+ *   Load Next Room
+ *   Load Previous Room
+ *   Load Next Room
+ *
+ *  @bug The striped nature of holls can cause some actors to unload due to
+ *      conflicting Object Lists between the two rooms
+ *
+ *  @bug If you can get around to the other side of the holl without triggering it,
+ *      you can load the room on the other side multiple times
+ */
```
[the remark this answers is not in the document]
Dragorn421: Most of the comments on `EnHoll_HorizontalVisibleNarrow` come from master, it just appears as new because I moved it

in master: https://github.com/zeldaret/oot/blob/aa48c66e10fc067e4d41c3a1d097bd83dea53acf/src/overlays/actors/ovl_En_Holl/z_en_holl.c#L48-L68

I'm okay with rewording or removing some of it, but I mean I can't defend it as I didn't write them

src/overlays/actors/ovl_En_Holl/z_en_holl.c
Dragorn421: "room transition region" or the familiar "loading zone" may work, but it would possibly encompass more than EnHoll :thinking:

fig02: yea. loading zone, at least in my experience, more pertains to scene changing loading zones. its tough for sure.
but yeah, will probably be a relevant discussion when we come up with new actor names in 2032

src/overlays/actors/ovl_En_Holl/z_en_holl.c
Dragorn421: It can be outside of the function, it is in master

I thought it was better inside it as only the `EnHoll_HorizontalVisibleNarrow`/`ENHOLL_H_VISIBLE_NARROW` type of holl makes use of this data, and it would be better to not think it's used by more types

I can put it outside and name it `sHorizontalVisibleNarrowTriggerDists` to address that too?

Also if I do move it outside, should it be near the top of the file, or is just above the function fine as well

src/overlays/actors/ovl_En_Holl/z_en_holl.c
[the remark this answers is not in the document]
fig02: alright. yeah I kinda feel like its not worth mentioning at all, but if others want to keep it, it should prob just be without a bug tag in whatever place is relevant 

it shuold be relevant to more than just this type anyway, so it doesnt make sense on just this function as-is

src/overlays/actors/ovl_En_Holl/z_en_holl.c
fig02: I would personally do it just above the function. but yeah this is a huge preference thing that we have no official guidelines for

I just personally dont like it when data+comments is taking up huge chunks of a function, i want the function to just get to the code asap

src/overlays/actors/ovl_En_Holl/z_en_holl.c
Dragorn421: We can also have no define at all Sadge
and just use the floats directly everywhere

Of course that's not what I'd prefer, I think it's easier to read (let alone modify) this actor with the defines

Some (most?) defines are used by more than one type, those can't really be properly placed near a function, so should stay at the top?
Then it would be weird to have defines a bit scattered around, at least to me it would

And tbh yeah it's not the end of the world, on my screen it's a screen full which is tolerable?

src/overlays/actors/ovl_En_Holl/z_en_holl.c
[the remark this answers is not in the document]
Dragorn421: Actually I need to check but it may be just this type that has the "duplicate room load" bug. I will know more once room loading is documented, which is also why I left this as-is

src/overlays/actors/ovl_En_Holl/z_en_holl.h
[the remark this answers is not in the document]
Dragorn421: "bg_fade" would be unfortunate because it'd break my fantastic naming scheme of "one word = one feature",
for example `ENHOLL_V_DOWN_FADE_LARGE` = vertical + down + fade + large
and `ENHOLL_V_DOWN_BG_FADE_LARGE` = vertical + down + bg + fade + large ? peepoCryTeddy

Could be alleviated with "bgfade" instead (`ENHOLL_V_DOWN_BGFADE_LARGE`)
(though not in function named `EnHoll_VerticalDownBgFadeLarge` but that's less important)

With that one play state unk being `bgCoverAlpha` maybe `BGCOVER` would be better? I don't think you could mistake that for the "plane fading" (visible) kind of holl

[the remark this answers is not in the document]
Dragorn421: I pushed fade -> bgcover, lmk how it looks

src/overlays/actors/ovl_En_Holl/z_en_holl.h
```diff
@@ -4,6 +4,33 @@
...
 #include "global.h"
 
+#define ENHOLL_GET_TYPE(thisx) (((thisx)->params >> 6) & 7)
+#define ENHOLL_GET_SWITCH_FLAG(thisx) ((thisx)->params & 0x3F)
+
+/**
+ * Two kinds of holls:
+ *   Horizontal (H):
+ *     For the player to move horizontally between rooms.
+ *     Cuboid-shaped, 200 units high plus 50 units into the ground, 400 units wide or, if narrow, 200 units wide.
```
fig02, line 14: theres alot going on in this sentence. maybe
```
Cuboid-shaped: 
- 200 units high, plus 50 unit into the ground
- 400 units wide, or 200 units if narrow
```
or something similar to break the information up

fig02 (approved):

src/overlays/actors/ovl_En_Holl/z_en_holl.h
```diff
@@ -4,6 +4,35 @@
...
+#define ENHOLL_GET_TYPE(thisx) (((thisx)->params >> 6) & 7)
+#define ENHOLL_GET_SWITCH_FLAG(thisx) ((thisx)->params & 0x3F)
+
+/**
+ * Two kinds of holls:
+ *   Horizontal (H):
+ *     For the player to move horizontally between rooms.
+ *     Cuboid-shaped:
+ *      - height: 200 units high plus 50 units into the ground
+ *      - width: 400 units wide or, if narrow, 200 units wide
```
engineer124, line 16: I'm sure it'd be quick to lookup, but a cuboid has 3 dimensions. Which dimension is missing? also, is width (left <-> right) or depth? (refence is player walking into cuboid)

src/overlays/actors/ovl_En_Holl/z_en_holl.c
```diff
@@ -2,26 +2,75 @@
...
+
+// Defines the width of horizontal holls (all kinds),
+// with a half-width extending on both lateral sides (towards local +x and -x).
+#define ENHOLL_H_HALFWIDTH_NARROW 100.0f
+#define ENHOLL_H_HALFWIDTH 200.0f
+
+// Defines the depth range from horizontal invisible holls (`ENHOLL_H_INVISIBLE`, `ENHOLL_H_INVISIBLE_NARROW`),
+// at which rooms get loaded.
+// i.e. when the player's distance from the actor (along local z) is within this range, the corresponding room is loaded
+// Note: This means there is a depth range in the middle of `2*ENHOLL_H_INVISIBLE_LOAD_DEPTH_MIN` where nothing happens.
```
engineer124, line 21:
```suggestion
// Note: This means there is a depth range in the middle of `2 * ENHOLL_H_INVISIBLE_LOAD_DEPTH_MIN` where nothing happens.
```

src/overlays/actors/ovl_En_Holl/z_en_holl.c
```diff
@@ -2,26 +2,75 @@
...
 #define FLAGS ACTOR_FLAG_4
 
-// Horizontal Plane parameters
-
-#define PLANE_Y_MIN -50.0f
-#define PLANE_Y_MAX 200.0f
-
-#define PLANE_HALFWIDTH 100.0f
-#define PLANE_HALFWIDTH_2 200.0f
+// Horizontal holls parameters (`ENHOLL_H_*`)
```
engineer124, line 5:
```suggestion
/* Horizontal holls parameters (`ENHOLL_H_*`) */
```
Just a styling suggestion, don't have to take it. But with this big block of comments/defines, this might help pop the sections a bit. Same suggestion to
```c
/* Vertical holls parameters (`ENHOLL_V_*`) */
```
or maybe some equal signs in there idk:
```c
/* === Vertical holls parameters (`ENHOLL_V_*`) === */
```

src/overlays/actors/ovl_En_Holl/z_en_holl.h
Dragorn421: Depth is missing, (depth = along the dimension the player traverses the cuboid (maybe I'm not using height/width/depth properly here though))
as well as height for the vertical (cylinder) holls

The thing is there isn't a single depth/height, the distances vary by type, and some types have several "depth distance triggers"
For example one type has a depth range in which the room on the other side loads, and another inner range in which the visible plane fades away, what would be its depth

I'm not sure how to document all of it in a few words

src/overlays/actors/ovl_En_Holl/z_en_holl.c
Dragorn421: I made a multiline block comment out of the h/v "headers", lmk how it looks

(I avoid things like `======` breaks because it's ~~hard~~ impossible to keep consistent across the repo)

src/overlays/actors/ovl_En_Holl/z_en_holl.c
```diff
@@ -116,34 +148,69 @@ void EnHoll_SwapRooms(PlayState* play) {
...
+ * [3] : End of fade region region, where the plane is fully transparent
+ *
+ * Within the fade region, the plane goes:
+ *   opaque -> transparent if approaching,
+ *   transparent -> opaque if receding
+ */
+static f32 sHorizontalVisibleNarrowTriggerDists[2][4] = {
+    { 200.0f, 150.0f, 100.0f, 50.0f }, // default
+    { 100.0f, 75.0f, 50.0f, 25.0f },   // SCENE_SPIRIT_TEMPLE
+};
```
mzxrules, line 167: Considering how much documentation is being done on this one structure, it might make sense to just define a type for it right here, and then in-line the data within the function.

fig02: by define a type do you mean use a struct? dont think the codegen would be the same

mzxrules: Yes, a struct, and I've already tested that it OKs perfectly fine.

Dragorn421: Since this type is jank I don't know what to name the struct members though
Will be clearer when I understand the room loading system

src/overlays/actors/ovl_En_Holl/z_en_holl.h
```diff
@@ -4,6 +4,35 @@
 #include "ultra64.h"
 #include "global.h"
 
+#define ENHOLL_GET_TYPE(thisx) (((thisx)->params >> 6) & 7)
+#define ENHOLL_GET_SWITCH_FLAG(thisx) ((thisx)->params & 0x3F)
+
+/**
+ * Two kinds of holls:
```
mzxrules, line 11: A personal thing, but I'm still stuck on the mentality of thinking of holls as being 2d plane/portal you cross with extra things being attached to make the game work nice, instead of this newer idea of en_holls being a three dimensional "hallway" that enable travel either vertically or horizontally. Because of it every single instance of horizontal and vertical is backwards to me because e.g. `EnHoll_HorizontalVisibleNarrow` computations deal with vertical planes. 

So I would like it if this line was more direct in explaining the dichotomy of en_holl instances being split by direction of travel, something like this but... better

```suggestion
 * There are two kinds of holls, which describe the direction in which the player can travel between rooms:
```

And then I also would like it if the first line of each function that has horizontal/vertical in the name would just spell out that you're referring to the direction of travel in some way, just to hammer the concept in.

Dragorn421: The direction of the player movement is shown below in this comment, I think it would just duplicate it to also mention it here

And idk just forget about holls being 2d, I know it doesn't help there is one or two types that draw a 2d plane but it's also only one or two types, not all of them

mzxrules: I didn't mean to suggest duplicating the information in this file, I was suggesting to hoist it up closer into the first line since it's important and it would make the text read better.

And I can't just "forget" about it because of how I visualize the computations, and like you said, `EnHoll_HorizontalVisibleNarrow` is actually visually represented as a vertical plane. If there's going to be a disconnect between visual representation and implementation names I think it's fair to repeat the information by the functions at least.

Dragorn421: What's the takeaway?

I think I wrote this block comment in a way that I think can be parsed quickly, "tree-like", since typically it should (?) be a quick glance to look at
Adding details at the top instead of at the leaves would go against that
Also style-wise if the top line for v/h was amended I'd like if the one for visual effects was similarly amended

fig02: I dont think it needs any more special treatment in the comment. if you just read the next few lines youll see the information you are looking for.

If there is a way we can fit that information in a nice way near the functions as well, we can possibly come to a compromise to move this forward

engineer124 (approved):

PR #1421, opened 2022-10-29
Doc/cleanup pass on effects 1

Dragorn421:
The things you do to get fig's attention FeelsAmazingMan


src/overlays/effects/ovl_Effect_Ss_Bomb/z_eff_ss_bomb.c
```diff
@@ -7,8 +7,10 @@
 #include "z_eff_ss_bomb.h"
 #include "assets/objects/gameplay_keep/gameplay_keep.h"
 
+#define LIFESPAN 20
+
```
fig02, line 11: can probably prefix this, though just "BOMB_ isnt great lmao
maybe EXPLOSION, idk

krm01: the way it's used makes me think it was meant to be 1 second and this dev just hardcoded it to 20 frames

fig02: that is the case, but the define allows you to change the duration in one place and have all of the relevant places update at the same time so that it still works correctly

krm01 (approved):

src/overlays/effects/ovl_Effect_Ss_Bomb/z_eff_ss_bomb.c
krm01: oh yeah I meant instead of them using the R_UPDATE_RATE reg or something to get 20

fig02: im lost on what youre trying to point out then 😅 my comment is just saying to add a prefix to the define name so that it more clearly belongs to this file/system. hardcoded timer durations is nothing special at all

src/overlays/effects/ovl_Effect_Ss_Blast/z_eff_ss_blast.c
```diff
@@ -7,15 +7,15 @@
...
-#define rAlphaTarget regs[8]
+#define rInnerColorR regs[0]
+#define rInnerColorG regs[1]
+#define rInnerColorB regs[2]
+#define rInnerColorA regs[3]
+#define rOuterColorR regs[4]
+#define rOuterColorG regs[5]
+#define rOuterColorB regs[6]
+#define rOuterColorA regs[7]
+#define rAlphaStep regs[8]
```
EllipticEllipsis, line 18: Why change only some of these? Isn't the point that these are all used the same way between effects?

fig02: these regs are effect specific. he only changed the ones where he is sure that prim/env maps to inner/outer. Not all effects will use prim and env for that purpose

EllipticEllipsis: What does inner/outer mean here?

fig02: red inner color, blue outer color
image

src/overlays/effects/ovl_Effect_Ss_Bomb/z_eff_ss_bomb.c
krm01: Sorry I was really thinking out loud and I'm realizing this comment thread isn't the place for it. I was lamenting at another instance of devs using hardcoded framerate instead of referencing a global framerate value. I agree completely with what you're saying about the define name.

src/overlays/effects/ovl_Effect_Ss_Blast/z_eff_ss_blast.c
EllipticEllipsis: But they're still prim and env? I don't see what's inner and outer there, I think that's a much more subjective description than sticking to what they're actually used in.

fig02: prim and env are just the names of the color channels. The asset will use those channels for different purposes depending on how its mapped. The prim color is in the center of the ring, the inner color. and env is the outside.
Naming it after how the asset uses prim and env is much more helpul information to have

Dragorn421: If you look at the I8 texture for the shockwave `gEffShockwaveTex`, it's whiteish on the inside of the ring and black on the outside
The combiner set in the dlist `gEffShockwaveDL` lerps from env to prim using that texture as weight (`PRIMITIVE, ENVIRONMENT, TEXEL0, ENVIRONMENT`, `(prim-env)*tex0+env`) so when the texture is white, prim comes out, that is, the color showing on the inside

the inner/outer is "explained" in the main `EffectSs_` function for spawning this effect

fig02: to take another example from the skulltula actor:
`gDPSetEnvColor(POLY_OPA_DISP++, this->teethR, this->teethG, this->teethB, 0);`
naming this after the fact that its the color of its teeth is much more helpful than saying its the env color. Its about what the prim/env color represent in this context

EllipticEllipsis: I thought it just LERP'd between them using the texture in the usual way? Much as it does in most of the other effects, I don't see why they all warrant different names.

fig02: this is going to depend entirely on how the specific effect asset is using the colors. Im not sure what you mean about "the usual way". The colors have context dependent meaning

Dragorn421: Elliptic are you arguing these regs should be named env/prim ?
If yes, only these regs or also the args in `EffectSs_` functions?
If no, what are we debating

EllipticEllipsis: I think the point here is that I don't see these as context-dependent because when you look at how they're used, they're pretty much used in the same way as they are in every other effect, as prim and env? As far as I can see, all effect use them for is various different sorts of `gsDPSetCombineLERP` with their textures. If anything, the context is in the combiner mode used rather than what the individual effects look like? I also don't think it will be possible to apply this in general, even if it's apparently the case that here one looks like an "inner" colour and the other an "outer" one?

fig02: prim and env tell you nothing about how it will visually end up looking/what its purpose is. It does depend on the combiner yes, but the point is to name it in the code after what it will actually end up representing. Pointing back to my previous example, naming that variable as "envColor" tells you nothing, while teeth color does. 

In cases where prim/env clearly map to certain visual aspects of the end result, like an inner and outer color, it makes perfect sense to name it after those things

src/overlays/effects/ovl_Effect_Ss_Bomb/z_eff_ss_bomb.c
Dragorn421: -> `EFFSSBOMB_LIFESPAN` :shrug:

src/overlays/effects/ovl_Effect_Ss_Blast/z_eff_ss_blast.c
Roman971: In general I agree with Elliptic that they should just be named prim/env, since how they are used is down to the color combiner and texture. Tho I also agree it can be good to be more explicit in situations where the colors have a very clear purpose. But I don't think this effect in particular falls in the category where it's clearly inner/outer (if anything most of the texture is a mix of both colors afaict), so I would personally keep prim/env here.

Dragorn421: To bump on this, I think it's nonsense to call things we have names for, "env" and "prim"

There are plenty examples of named colors that show it's much more useful to refer to the use it gets rather than the color register (that's how prim/env/blend/fog colors are called, they're color registers) in use
For example Link's tunic, gauntlets, the actors' colorFilterParams, Volvagia's head/body alpha, Phantom Ganon's eyes brightness, Gohma's eye color, dog colors...

The C equivalent of restricting names to "env" and "prim" would be naming all function arguments after the ABI, for example a0 for every first non-float argument

Roman971: > Can you elaborate on what this would mean for EffectSsBlast_Spawn's args and docs

I would name the args and docs with prim/env, following the names used inside the effect

> To bump on this, I think it's nonsense to call things we have names for, "env" and "prim"

I think it makes sense to use proper names when the purpose of the colors is clear, but I don't really agree that this is the case here. At least looking at the image above I don't see the red and blue being clear inner/outer colors. In fact most of the texture ends up being purple which I assume is because a lot of it is a mix of both prim and env colors. If it was clearly inner/outer, I would expect a part of the ring to be full red, and another part to be full blue, with no mix in between (in terms of IA texture with this combiner, it would mean max alpha on one part, and 0 alpha on the other). At least this is how I interpret what's going on with the texture here.

Dragorn421: Do you have a better description than inner/outer then?

To be clear, using env/prim is as bad as naming them color1/color2

hensldm: I've been thinking on this for a while now. I'd probably argue that inner/outer while not perfect, are fine as is. The inner color seems to affect the inner portion more (it seems more redish in my eyes), and the same for the outer (at least I see a blueish outline).

src/overlays/effects/ovl_Effect_Ss_Fhg_Flash/z_eff_ss_fhg_flash.c
```diff
@@ -25,15 +25,15 @@ EffectSsInit Effect_Ss_Fhg_Flash_InitVars = {
     EffectSsFhgFlash_Init,
 };
 
-static UNK_TYPE D_809A5178[258];
-static Gfx D_809A5100[15];
+// Should eventually come from assets/overlays/ovl_Effect_Ss_Fhg_Flash/ovl_Effect_Ss_Fhg_Flash.h
```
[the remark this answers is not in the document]
Dragorn421: Because ZAPD doesn't write anything to the .h for whatever reason, this is `assets/overlays/ovl_Effect_Ss_Fhg_Flash/ovl_Effect_Ss_Fhg_Flash.h` :
```c
#ifndef OVL_EFFECT_SS_FHG_FLASH_H
#define OVL_EFFECT_SS_FHG_FLASH_H 1


#endif
```

src/overlays/effects/ovl_Effect_Ss_Fhg_Flash/z_eff_ss_fhg_flash.c
```diff
@@ -172,10 +173,10 @@ void EffectSsFhgFlash_UpdateShock(PlayState* play, u32 index, EffectSs* this) {
     s16 randBodyPart;
     Player* player;
     BossGanondrof* phantomGanon;
-    s16 rand;
+    s16 rotStep;
 
-    rand = (Rand_ZeroOne() * 20000.0f);
-    this->rXZRot = (this->rXZRot + rand) + 0x4000;
+    rotStep = (Rand_ZeroOne() * 20000.0f);
```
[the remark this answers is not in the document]
Dragorn421: OoT's stance on parentheses is to not have a stance
That said, I think most agree these are useless (even the `(a == b) && (c == d)` team)

src/overlays/effects/ovl_Effect_Ss_Fhg_Flash/z_eff_ss_fhg_flash.c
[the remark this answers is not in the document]
hensldm: I think it is because everything is extracted as `static` so it doesn't get put in `.h` file since they are meant to be internal only.

src/overlays/effects/ovl_Effect_Ss_Fhg_Flash/z_eff_ss_fhg_flash.c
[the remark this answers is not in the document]
fig02: the "stance" was referring to the second part of his comment, defining it where its declared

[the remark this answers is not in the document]
Dragorn421: Oh `int a; a = 421;` vs `int a = 421;` ?

Yeah typically the latter is preferred but it doesn't make a huge amount of sense to me in this case

src/overlays/effects/ovl_Effect_Ss_Fhg_Flash/z_eff_ss_fhg_flash.c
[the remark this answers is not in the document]
Dragorn421: I would argue the .h itself is internal only,
but I guess it'd be weird to have either externs for static symbols, or forward static declarations which would amount to bss if there's no actual declaration later on 🤷

Should I change this comment somehow if we shouldn't assume ZAPD will ever change this in some way?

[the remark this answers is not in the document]
fig02: eventually tm tm these kinda of files wont be extracted anyway (though the assets will), so id say its a temporary situation regardless. but the comment is fine imo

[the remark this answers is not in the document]
hensldm: Perhaps just add a todo, with a comment to investigate ZAPD forward declaring static variables. Also note you can't just extern them in the `.h` file, the compiler yells at you.

hensldm (approved):

src/overlays/effects/ovl_Effect_Ss_Fhg_Flash/z_eff_ss_fhg_flash.c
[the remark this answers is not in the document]
Dragorn421: I added a TODO 403201da867853c889480944eeeca640fb631fdb

src/overlays/effects/ovl_Effect_Ss_Blast/z_eff_ss_blast.c
fig02: Of the people who have voiced their opinion, we are 3 to 2 in favor of inner/outer
Id love to come to a better compromise/consensus, but this has been in limbo for a while without much progress. So Id say to go with the majority, even if slight 

ofc anyone open to add their opinion if they havent already

src/code/z_effect_soft_sprite_old_init.c
```diff
@@ -745,12 +771,12 @@ void EffectSsFhgFlash_SpawnLightBall(PlayState* play, Vec3f* pos, Vec3f* velocit
 }
 
 /**
- * Spawn a shock effect
+ * Spawn a purple shock/lighting effect.
```
fig02, line 774: think its probably best to just leave it at shock? dont think it resembles lightning at all really. Its just a bunch of "electricity" that doesnt really resemble much of a lightning strike

I know I put lightning in the description a couple years ago, but I think it was my mistake to do that lol

src/code/z_effect_soft_sprite_old_init.c
```diff
@@ -745,12 +771,12 @@ void EffectSsFhgFlash_SpawnLightBall(PlayState* play, Vec3f* pos, Vec3f* velocit
...
  *
- * param determines where the ligntning should go
- * 0: don't attach to any actor. spawns at the position specified by pos
- * 1: spawn at one of Player's body parts, chosen at random
- * 2: spawn at one of Phantom Ganon's body parts, chosen at random
+ * @param actor If param is `FHGFLASH_SHOCK_PG`, the Phantom Ganon actor. Unused otherwise.
+ * @param pos If param is `FHGFLASH_SHOCK_NO_ACTOR`, the position of the effect. Unused otherwise.
+ * @param scale The effect will be `20*rand(1..2)*scale/100` units wide.
```
fig02, line 778: its probably fine to omit the specifics on how the scale is calculated here, it can be found in the code.
can maybe say "the base scale to which some randomness is applied", or something like that

src/code/z_effect_soft_sprite_old_init.c
Dragorn421: Actually I wrote "lighting" here but I don't remember if that's a typo for "lightning" Thonk

src/code/z_effect_soft_sprite_old_init.c
Dragorn421: yeah I guess the rand() part can be omitted,
but wouldn't it be useful to have a basic idea of how many units a scale value means?
I don't know how much it varies across effects, but if there are effects for which the scale should be 0.1 and others 10 to have a 10 units wide particle, browsing the code would get annoying fast imo

(and trying random numbers is cool but we can do better skawoUHHUH)

Dragorn421: (and the final size is not just about the code, it's the assets too)

fig02: if it can be conveyed in a more digestible format sure, but id rather not be writing code in comments.
The effects do go for a consistent scaling with / 100 being very common

while I do agree with the sentiment, even with the information provided in the comment, I still dont think itd be easy to formulate what the scale should be. looking at existing usages and trying numbers is probably always going to be the best way lol

Dragorn421: Maybe something like "A scale of 50 makes the effect around 10 units wide" ? Or some other wording that suggests a "typical" scale

Very simply: `@param scale e.g. pass 50 for 10 units` ?

fig02: I guess id be fine with "20 * scale / 100 units wide, with some randomness applied", or something

src/code/z_effect_soft_sprite_old_init.c
Dragorn421: Probably a typo for "lightning", I guess I took that from `FhgFlashLightningParam` maybe or elsewhere

Dragorn421: lightning -> "a ball of electrical arcs" but I can also remove

fig02: it doesnt add anything for me personally, but am fine with it staying

fig02 (approved):
this pr has technically met review requirements and it ready to merge, but its been open for a long time, so I will let it sit for a day or so incase anyone has any closing thoughts

PR #1423, opened 2022-11-02
Fix misc 18

Dragorn421:


src/code/z_collision_check.c
```diff
@@ -2666,8 +2666,8 @@ void CollisionCheck_SetOCvsOC(Collider* left, ColliderInfo* leftInfo, Vec3f* lef
     if (leftActor == NULL || rightActor == NULL || left->ocFlags1 & OC1_NO_PUSH || right->ocFlags1 & OC1_NO_PUSH) {
         return;
     }
-    rightMassType = CollisionCheck_GetMassType(leftActor->colChkInfo.mass);
-    leftMassType = CollisionCheck_GetMassType(rightActor->colChkInfo.mass);
+    leftMassType = CollisionCheck_GetMassType(leftActor->colChkInfo.mass);
+    rightMassType = CollisionCheck_GetMassType(rightActor->colChkInfo.mass);
```
fig02, line 2670: unrelated to the actual fix, ive always found the usage of "left" and "right" here kinda weird. I guess something like 1 and 2 isnt that much better, but implying some kind of direction with them has always sounded weird to me

dont have to do anything about it in this pr, but just putting that thought out there

fig02 (approved):

Roman971 (approved):

src/overlays/actors/ovl_Obj_Makekinsuta/z_obj_makekinsuta.h
```diff
@@ -11,8 +11,8 @@ typedef void (*ObjMakekinsutaActionFunc)(struct ObjMakekinsuta*, PlayState*);
 typedef struct ObjMakekinsuta {
     /* 0x0000 */ Actor actor;
     /* 0x014C */ ObjMakekinsutaActionFunc actionFunc;
-    /* 0x150  */ s16 timer;
-    /* 0x152  */ s16 unk_152;
+    /* 0x0150 */ s16 timer;
+    /* 0x0152 */ s16 unk_152;
 } ObjMakekinsuta; // size = 0x0154
```
EllipticEllipsis, line 16: This should use 3 digits instead of 4 anyway.

Dragorn421: idk about "should" but sure I'll change it now
This isn't exactly consistent in the repo

EllipticEllipsis: Usually we do it with the required number of digits' padding and no more.

Dragorn421: looking for `/* 0x000 ` I find 61 results,
while with (regex) `/\* 0x0.*\n\}` (matches ends of structs with a leading `/* 0x0...` comment on the last member) I find 800

looks sus, at least definitely not usual

AngheloAlf: What Elliptic is probably trying to say is there's no need to have the extra zeroes since the struct's size is just `3 digits (without the `0x`)
At least in MM land we have been trying to avoid those extra zeroes

Dragorn421: I do understand the concept of minimal zero padding, just not convinced OoT even made an attempt at it

fig02: the bootstrap script that generated our initial actor scripts used 4 0's no matter what, which is probably why most dont currently do it. we only really standardized omitting leading zeros in asset xmls afaik. but yeah, its something we can look to do in further cleanup prs

src/overlays/actors/ovl_Demo_6K/z_demo_6k.c
```diff
@@ -812,7 +812,7 @@ void func_809691BC(Demo6K* this, PlayState* play, s32 params) {
 
     temp = Environment_LerpWeight(csAction->endFrame, csAction->startFrame, play->csCtx.frames);
 
-    this->actor.world.pos.x = (((endPos.x - startPos.x) * temp) + startPos.x);
-    this->actor.world.pos.y = (((endPos.y - startPos.y) * temp) + startPos.y);
-    this->actor.world.pos.z = (((endPos.z - startPos.z) * temp) + startPos.z);
+    this->actor.world.pos.x = LERP(startPos.x, endPos.x, temp);
+    this->actor.world.pos.y = LERP(startPos.y, endPos.y, temp);
+    this->actor.world.pos.z = LERP(startPos.z, endPos.z, temp);
```
hensldm, line 817: Looks like `VEC3F_LERPIMPDST(&this->actor.world.pos, &startPos, &endPos, temp);` also works here. Not sure if that would be preferred.

src/overlays/actors/ovl_Demo_Du/z_demo_du.c
```diff
@@ -454,9 +454,9 @@ void DemoDu_CsGoronsRuby_DaruniaFalling(DemoDu* this, PlayState* play) {
             endPos.y = npcAction->endPos.y;
             endPos.z = npcAction->endPos.z;
 
-            pos->x = ((endPos.x - startPos.x) * traveledPercent) + startPos.x;
-            pos->y = ((endPos.y - startPos.y) * traveledPercent) + startPos.y;
-            pos->z = ((endPos.z - startPos.z) * traveledPercent) + startPos.z;
+            pos->x = LERP(startPos.x, endPos.x, traveledPercent);
+            pos->y = LERP(startPos.y, endPos.y, traveledPercent);
+            pos->z = LERP(startPos.z, endPos.z, traveledPercent);
```
hensldm, line 459: Same comment looks like the `VEC3F` macro would work as well.

hensldm (approved):
Seems, but did notice a couple places the `VEC3F` `LERP` could be used, but not sure if that would be preferred.

src/overlays/actors/ovl_Demo_6K/z_demo_6k.c
Dragorn421: Do you like the vec macros yourself?

I like the `SQXZ` family, maybe `VEC_SET`, but more involved code in macros idk

I've been cleaning up some math macros since they were freely introduced during decomp, but at 100% it is revealed some of them are nearly useless, or only local to some files
The z64math.h macros are eligible for cleanup and there has been mixed opinions expressed about vec macros in the past

hensldm: TBH, I don't really have a preference either way (though for the `SQ` and `DOT` ones, I'd probably prefer them taking pointers, but iirc that doesn't work all the time). I do like the small bit of documentation the macros provide, but they are also all simple enough that it isn't really needed.

Dragorn421: Well I'm adding this macro to my list of things to consider (along with `TRANS_TYPE_FILL_WHITE2` :sunglasses:)

Can we "overlook" this for this PR and go ahead with merging?

hensldm: Yeah that is fine :)

Dragorn421: In the future I'll just assume an approval with comments means we can merge without waiting for the comments, to make communication less clunky (so if you have comments that should be addressed don't approve right away)

PR #1427, opened 2022-11-10
Cleanup `z_collision_check` 1

Dragorn421:


include/z64collision_check.h
```diff
@@ -9,90 +9,140 @@
...
     /* 0x16 */ u8 bumperFlags; // Information flags for AC collisions
     /* 0x17 */ u8 ocElemFlags; // Information flags for OC collisions
     /* 0x18 */ Collider* atHit;                // object touching this element's AT collider
     /* 0x1C */ Collider* acHit;                // object touching this element's AC collider
-    /* 0x20 */ struct ColliderInfo* atHitInfo; // element that hit the AT collider
-    /* 0x24 */ struct ColliderInfo* acHitInfo; // element that hit the AC collider
-} ColliderInfo; // size = 0x28
+    /* 0x20 */ struct ColliderElement* atHitInfo; // element that hit the AT collider
+    /* 0x24 */ struct ColliderElement* acHitInfo; // element that hit the AC collider
+} ColliderElement; // size = 0x28
```
EllipticEllipsis, line 126: This seems like an excellent way to mix up the general collider properties with those of the sub-collider things that JntSphs have.

src/overlays/actors/ovl_Boss_Dodongo/z_boss_dodongo.c
```diff
@@ -1226,10 +1226,10 @@ void BossDodongo_SpawnFire(BossDodongo* this, PlayState* play, s16 params) {
 
 void BossDodongo_UpdateDamage(BossDodongo* this, PlayState* play) {
     s32 pad;
-    ColliderInfo* item1;
+    ColliderElement* item1;
     u8 swordDamage;
     s32 damage;
-    ColliderInfo* item2;
+    ColliderElement* item2;
```
EllipticEllipsis, line 1232: Especially in a function like this, where you get both!

include/z64collision_check.h
```diff
@@ -9,90 +9,140 @@
...
-    /* 0x10 */ u8 atFlags; // Information flags for AT collisions.
-    /* 0x11 */ u8 acFlags; // Information flags for AC collisions.
-    /* 0x12 */ u8 ocFlags1; // Information flags for OC collisions.
-    /* 0x13 */ u8 ocFlags2;  // Flags related to which colliders it can OC collide with.
-    /* 0x14 */ u8 colType; // Determines hitmarks and sound effects during AC collisions.
-    /* 0x15 */ u8 shape; // JntSph, Cylinder, Tris, or Quad
+    /* 0x00 */ struct Actor* self; // Actor owning this collider.
+    /* 0x04 */ struct Actor* otherAC; // Other actor owning the AC collider that collided with this AT collider.
+    /* 0x08 */ struct Actor* otherAT; // Other actor owning the AT collider that collided with this AC collider.
+    /* 0x0C */ struct Actor* otherOC; // Other actor owning the OC collider that collided with this OC collider.
```
EllipticEllipsis, line 45: This does seem like a good idea, but I think `self` would confuse people coming from languages (Python, Rust) that use `self` to refer to the object itself. Maybe, as your comment suggests, `owner` would work?

Dragorn421: Sure I'm open to changing it, `owner` sounds fine

Fwiw "self" is already used in some parts of zcolchk iirc, but it can be changed there too

It's not super straightforward to change though so I'd like more opinions to consolidate the name before commiting to another rename

include/z64collision_check.h
Dragorn421: Can you elaborate? `ColliderElement` is exactly the base struct for `ColliderJntSphElement`

EllipticEllipsis: Oh, I thought it lived in the collider itself like it does for the other types.

Dragorn421: I can't fix illiteracy

EllipticEllipsis: There is the issue that the base struct for JntSph just behaves completely differently, it feels like there ought to be some sort of distinction.

Dragorn421: wdym it behaves completely differently

For example if you look at an AT Cylinder vs an AC JntSph

`CollisionCheck_AT` checks the cylinder's `Collider.atFlags`
then `CollisionCheck_AC` checks the jntSph's `Collider.acFlags`, (and the `AC_TYPE_ALL` bits between cylinder's `Collider.atFlags` and jntSph's `Collider.acFlags`)
then `CollisionCheck_ATCylVsACJntSph` makes use of `ColliderElement atCyl->elem` and `ColliderElement acJntSph->elements[...].base`

outside VS funcs: use `Collider` base
inside VS funcs: use `ColliderElement` base (or elem for things like cylinder with a single element)

Dragorn421: Also note the symmetry in `CollisionCheck_SetATvsAC` with how it handles all shapes the same

include/z64collision_check.h
mzxrules: I guess I don't understand what's wrong with just using `actor` here.

What's far worse in this part of code is that you've inverted the meaning of the at, ac, oc variables, which makes the code harder to follow and use correctly.  

The at, ac, oc variables are meant to store the _RESULT_ of the collision check computation requested when calling SetAT/SetAC/SetOC, relative to the reference frame of the collider itself. Did the collider's attack land? `at` stores the actor that it hit. Did the attack collider get hit? `ac` stores the actor that touched it. Did it bump into an object? `oc` stores the actor it touched.

Similarly, `atFlags`, and `acFlags` are named to be associated with the `at` and `ac` variables, since they effect the result of the collision check. This is why e.g. `Collider_ResetATBase` nulls `at` and masks `atFlags` to a default value in the same function.

Lastly, inverting the name creates some misgivings about how the system works. Using your renames, `otherAT`, `otherAC`, `otherOC`, one might assume that reading e.g. `otherAT` would return an actor currently subscribed to AT collisions. This is entirely incorrect, since the lists are purged after performing collision check (which happens after actors are processed and before the next frame when the results are available to actors), and actors are free to redefine what lists to subscribe a collider to.

include/z64.h
```diff
@@ -892,6 +892,8 @@ typedef struct {
     /* 0x74 */ s16 unk_74[2]; // context-specific data used by the current scene draw config
 } RoomContext; // size = 0x78
 
+#define SAC_ENABLE (1 << 0)
```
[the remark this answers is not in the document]
Dragorn421: right below this line is `CollisionCheckContext`

include/z64collision_check.h
```diff
@@ -102,12 +152,12 @@ typedef struct {
 } ColliderJntSphElementDimInit; // size = 0x0C
 
 typedef struct {
-    /* 0x00 */ ColliderInfo info;
+    /* 0x00 */ ColliderElement base;
```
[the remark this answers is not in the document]
Dragorn421: The parent of `ColliderElement` for JntSph and other multi-elements colliders, is already an element, so this would mean `jntSphElem->elem` instead of the more natural (?) `jntSphElem->base`

and cylinders and other single-element colliders ofc already have `base` for the collider, and they don't have a specific element substruct, so their (basic) element is just "elem"

I could change the single-element colliders' elements to be specific structs, but it would just be bloat for no reason imo

include/z64collision_check.h
Dragorn421: What you explain matches what I know of how zcolchk works, and I also guessed the same about what the logic behind the previous names was

If I changed them, it's because I believe the names I suggest are better.
For example when you're inside an actor's code, it's not obvious, it's confusing even, what `collider.at` means. `collider.otherAC` seemed clearer: it's another (not me), and it's an AC (I attacked it)

mzxrules: `collider.at` is the result of the SetAT call you made last frame; the core advantage is that you don't have to do a mental inversion use the correct variable name since all `AT` things are named `AT`, and you're not mixing AT with AC.

and again, `otherAC` doesn't return an AC; it returns an actor or NULL. ACs no longer exist by the time you read the variable.

Dragorn421: > and again, otherAC doesn't return an AC; it returns an actor or NULL. ACs no longer exist by the time you read the variable.

sure? does that not apply to `at`

mzxrules: It does so yes. I think the difference between us is our perspective on what the variable name is actually naming. 

The idea behind the `at` name is that it's shorthand for "This is the result Actor of this collider's AT Collision Check". The thing being called `at` is the result pointer. The actor being referenced (or NULL) isn't the thing with an `AT` property or an accessible `AT` collider. Another way of putting it is that `at` is really a handle name used to get the collider's AT Collision Check result... if things were implemented in python, `ColliderCollisionResult["AT"]` better expresses the intent behind the variable name. Given this, I find it weird to change the handle to get AT collision results to one that has AC in the name.

With `otherAC` my understanding is that this name was chosen to describe a property of the actor being stored here, e.g. dereferencing `otherAC` gives you an actor with an AC enabled collider or an actor with some sort of AC, which isn't a correct interpretation. AT/AC/OC aren't properties attached to actors, they are properties attached to colliders.

Dragorn421: Well, for an actor to end up in `otherAC`, it needs to use colliders and have had a collider set as AC
I didn't make it `otherACOwner` because that looks bad and seems redundant

I feel like naming it like a result works less well, `atResult` sounds less natural

How I came to pick `otherAC` was like:
1) `at`/`ac`/`oc` on its own isn't descriptive enough, it's confusing whether that refers to "self" or "other", and also reading code you can't tell it's not a part of the collider struct -> use `other`
2) `atOther` can be confused between "other actor using AT" and "other actor relevant to self being AT" -> `otherAC`

```c
    if (this->colliderBody.base.atFlags & AT_HIT) {
        this->colliderBody.base.atFlags &= ~AT_HIT;
        if (this->colliderBody.base.at == &player->actor) {
```

```c
    if (this->colliderBody.base.atFlags & AT_HIT) {
        this->colliderBody.base.atFlags &= ~AT_HIT;
        if (this->colliderBody.base.otherAC == &player->actor) {
```

I really think the second one reads better

include/z64.h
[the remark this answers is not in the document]
mzxrules: I guess I'm more asking why are we moving defines out of the headers they belong in instead of moving the structs out of the massive header blob into the headers they belong in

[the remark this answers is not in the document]
Dragorn421: I agree `CollisionCheckContext` should move to the adequate header
I didn't target header cleanup with this PR though, more the collider system itself

include/z64collision_check.h
mzxrules: > Well, for an actor to end up in `otherAC`, it needs to use colliders and have had a collider set as AC I didn't make it `otherACOwner` because that looks bad and seems redundant

for NULL to end up there it has to do neither

> I feel like naming it like a result works less well, `atResult` sounds less natural

Disagree; I think it's pretty natural to think of the system working as such:

* Set AT Flags
* Enable AT collision on collider
* Wait
* Check AT collision result on collider

> at/ac/oc on its own isn't descriptive enough, it's confusing whether that refers to "self" or "other"

Why would a collider collide with itself? It should be implicit that a Collider would only collide with something else.

> I really think the second one reads better

It's slightly more readable, but it breaks the consistency that existed, making it harder to use. You could easily achieve even greater clarity with code while maintaining naming consistency like this:

    if (this->colliderBody.base.atFlags & AT_HIT) {
        this->colliderBody.base.atFlags &= ~AT_HIT;
        if (this->colliderBody.base.atHitActor == &player->actor) {

Not suggesting that `at` -> `atHitActor`, but rather that it's easier to understand why `atFlags` are being checked here in this snippet, whereas with `otherAC` it's less obvious.

Dragorn421: So like are you open to renaming these to anything or do you want all of them to stay as they are in master actor/at/ac/oc

mzxrules: I'm personally fine with the names actor/at/ac/oc staying as they are, and I would be fine with at/ac/oc changing to be more expressive to something like atResult/acResult/ocResult. The key thing I care about is that whatever name is used, at/ac/oc all line up like before so that we avoid mixing at and ac flags.

include/z64collision_check.h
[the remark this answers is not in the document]
mzxrules: reason for calling Collider `base` in e.g. JntSph struct is because Colliders utilize a bit of polymorphism (e.g. SetAT accepts Collider* and downcasts to ColliderJntSph*). I don't think this is the case for ColliderElement. Additionally, I'd say it's a bit weird to only have some ColliderElement consumers to express polymorphism but not others.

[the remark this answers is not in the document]
Dragorn421: `ColliderElement` is at the start of every `Collider*Element`, and that naming pattern seems to justify the inheritance pattern (note the naming comes with the PR, `ColliderElement` was `ColliderInfo`-something)

See also https://github.com/zeldaret/oot/pull/1427#discussion_r1020958637 for why I think it makes sense to consider cylinders and such as single-elements colliders

and again
> I could change the single-element colliders' elements to be specific structs, but it would just be bloat for no reason imo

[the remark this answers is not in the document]
mzxrules: but then you also continued saying 

> it would just be bloat for no reason imo

My position is you either implement ColliderShapeElements for all shapes, or you do it for none and change `base`->`elem`. There's no reason to go half and half here. Personally, I don't think the inheritance pattern is strictly necessary, which is why I didn't use it when decomping this code.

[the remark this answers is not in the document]
Dragorn421: I think it's a bad idea to make a change that would turn
```c
static ColliderCylinderInit sCylinderInit = {
    {
        COLTYPE_NONE,
        AT_NONE,
        AC_ON | AC_TYPE_ALL,
        OC1_ON | OC1_TYPE_ALL,
        OC2_TYPE_2,
        COLSHAPE_CYLINDER,
    },
    {
        ELEMTYPE_UNK2,
        { 0x00000000, 0x00, 0x00 },
        { 0xFFCFFFFF, 0x00, 0x00 },
        TOUCH_NONE,
        BUMP_ON,
        OCELEM_ON,
    },
    { 25, 60, 0, { 0, 0, 0 } },
};
```
into
```c
static ColliderCylinderInit sCylinderInit = {
    {
        COLTYPE_NONE,
        AT_NONE,
        AC_ON | AC_TYPE_ALL,
        OC1_ON | OC1_TYPE_ALL,
        OC2_TYPE_2,
        COLSHAPE_CYLINDER,
    },
    {
        {
            ELEMTYPE_UNK2,
            { 0x00000000, 0x00, 0x00 },
            { 0xFFCFFFFF, 0x00, 0x00 },
            TOUCH_NONE,
            BUMP_ON,
            OCELEM_ON,
        },
        { 25, 60, 0, { 0, 0, 0 } },
    },
};
```
or (if one comma is omitted):
```c
static ColliderCylinderInit sCylinderInit = {
    {
        COLTYPE_NONE,
        AT_NONE,
        AC_ON | AC_TYPE_ALL,
        OC1_ON | OC1_TYPE_ALL,
        OC2_TYPE_2,
        COLSHAPE_CYLINDER,
    },
    { {
          ELEMTYPE_UNK2,
          { 0x00000000, 0x00, 0x00 },
          { 0xFFCFFFFF, 0x00, 0x00 },
          TOUCH_NONE,
          BUMP_ON,
          OCELEM_ON,
      },
      { 25, 60, 0, { 0, 0, 0 } } },
};
```

(note these are the Init structs but I assume we want to keep them mirroring the live structs)

even though it would fit the pattern of elements being `ColliderElement base;` followed by element dimensions

Or do you really think it's worth it

include/z64collision_check.h
```diff
@@ -9,90 +9,140 @@
...
     /* 0x15 */ u8 toucherFlags; // Information flags for AT collisions
     /* 0x16 */ u8 bumperFlags; // Information flags for AC collisions
     /* 0x17 */ u8 ocElemFlags; // Information flags for OC collisions
     /* 0x18 */ Collider* atHit;                // object touching this element's AT collider
     /* 0x1C */ Collider* acHit;                // object touching this element's AC collider
-    /* 0x20 */ struct ColliderInfo* atHitInfo; // element that hit the AT collider
-    /* 0x24 */ struct ColliderInfo* acHitInfo; // element that hit the AC collider
-} ColliderInfo; // size = 0x28
+    /* 0x20 */ struct ColliderElement* atHitInfo; // element that hit the AT collider
+    /* 0x24 */ struct ColliderElement* acHitInfo; // element that hit the AC collider
```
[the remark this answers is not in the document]
Dragorn421: -> #1436

include/z64collision_check.h
Dragorn421: -> #1436

include/z64collision_check.h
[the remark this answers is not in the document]
mzxrules: well... that's the price you pay for consistency. I do hate it makes var access a little longer, but you end up with this:

    cylinder->element.base
    cylinder->element.dim
    jntSphr->elements[0].base
    jntSphr->elements[0].dim

Dragorn421:
I would like to have feedback on this change:
https://github.com/zeldaret/oot/pull/1427/commits/cfc8c32ace2970869a2c93be0666786f87ca5e1a
(introducing a new struct `ColliderCylinderElement`)

before I update all the actors. That commit only changes three actors, as a demonstration

AngheloAlf:
Personally I like it, I think it looks better and makes more sense this way

include/z64collision_check.h
Dragorn421: Can this be considered resolve?
I'll consider so in a week or so

include/z64collision_check.h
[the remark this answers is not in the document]
Dragorn421: Sticking to `ColliderJntSph.elements[].base` and `ColliderCylinder.elem`, at least for now.

include/z64collision_check.h
```diff
@@ -9,90 +9,140 @@
...
     /* 0x17 */ u8 ocElemFlags; // Information flags for OC collisions
-    /* 0x18 */ Collider* atHit;                // object touching this element's AT collider
-    /* 0x1C */ Collider* acHit;                // object touching this element's AC collider
-    /* 0x20 */ struct ColliderInfo* atHitInfo; // element that hit the AT collider
-    /* 0x24 */ struct ColliderInfo* acHitInfo; // element that hit the AC collider
-} ColliderInfo; // size = 0x28
+    /* 0x18 */ Collider* atHit; // object touching this element's AT collider
+    /* 0x1C */ Collider* acHit; // object touching this element's AC collider
+    /* 0x20 */ struct ColliderElement* atHitInfo; // element that hit the AT collider
+    /* 0x24 */ struct ColliderElement* acHitInfo; // element that hit the AC collider
```
mzxrules, line 125:
```suggestion
    /* 0x20 */ struct ColliderElement* atHitElem; // element that hit the AT collider
    /* 0x24 */ struct ColliderElement* acHitElem; // element that hit the AC collider
```

hensldm (approved):
Changes seem good to me :) .

fig02 (approved):

src/overlays/actors/ovl_Boss_Fd/z_boss_fd.c
```diff
@@ -1279,11 +1279,11 @@ void BossFd_Effects(BossFd* this, PlayState* play) {
 
 void BossFd_CollisionCheck(BossFd* this, PlayState* play) {
     ColliderJntSphElement* headCollider = &this->collider.elements[0];
-    ColliderInfo* hurtbox;
+    ColliderElement* hurtbox;
```
mzxrules, line 1282:
```suggestion
    ColliderElement* acHitElem;
```

src/overlays/actors/ovl_Boss_Fd2/z_boss_fd2.c
```diff
@@ -806,31 +806,31 @@ void BossFd2_Wait(BossFd2* this, PlayState* play) {
 
 void BossFd2_CollisionCheck(BossFd2* this, PlayState* play) {
     s16 i;
-    ColliderInfo* hurtbox;
+    ColliderElement* hurtbox;
```
mzxrules, line 809:
```suggestion
    ColliderElement* acHitElem;
```

src/overlays/actors/ovl_En_Bubble/z_en_bubble.c
```diff
@@ -97,7 +97,7 @@ void EnBubble_SetDimensions(EnBubble* this, f32 dim) {
 }
 
 u32 func_809CBCBC(EnBubble* this) {
-    ColliderInfo* info = &this->colliderSphere.elements[0].info;
+    ColliderElement* info = &this->colliderSphere.elements[0].base;
```
mzxrules, line 100:
```suggestion
    ColliderElement* elem = &this->colliderSphere.elements[0].base;
```

src/overlays/actors/ovl_En_Fw/z_en_fw.c
```diff
@@ -135,11 +135,11 @@ Vec3f* EnFw_GetPosAdjAroundCircle(Vec3f* dst, EnFw* this, f32 radius, s16 dir) {
 }
 
 s32 EnFw_CheckCollider(EnFw* this, PlayState* play) {
-    ColliderInfo* info;
+    ColliderElement* info;
```
mzxrules, line 138:
```suggestion
    ColliderElement* elem;
```

src/overlays/actors/ovl_En_Fd/z_en_fd.c
```diff
@@ -275,15 +275,15 @@ s32 EnFd_CheckHammer(EnFd* this, PlayState* play) {
 
 s32 EnFd_ColliderCheck(EnFd* this, PlayState* play) {
     Player* player = GET_PLAYER(play);
-    ColliderInfo* info;
+    ColliderElement* info;
```
mzxrules, line 278:
```suggestion
    ColliderElement* elem;
```

src/overlays/actors/ovl_En_Ssh/z_en_ssh.c
```diff
@@ -434,17 +434,17 @@ void EnSsh_Sway(EnSsh* this) {
 }
 
 void EnSsh_CheckBodyStickHit(EnSsh* this, PlayState* play) {
-    ColliderInfo* info = &this->colCylinder[0].info;
+    ColliderElement* info = &this->colCylinder[0].elem;
```
mzxrules, line 437:
```suggestion
    ColliderElement* elem = &this->colCylinder[0].elem;
```

fig02:
Ive made a pull request to this branch with some organizational changes id like to see in the colcheck header: https://github.com/Dragorn421/oot/pull/3

Mainly:
- I think it makes sense for cylinders to be the first "shape" in the header, instead of JntSphere, simply because of how common cylinders are in comparison to anything else
- Within each collider "shape", I separated the "live" collider structs from the init data. It helps me personally to see all of the nested structs next to each other, and have them be separated by their purpose.
- I added a brief description of each shape, mainly so that there is a reference of what "jntsph" even means. And it helps quickly explain which are a "single" entity and which are collections

fig02:
I reverted the change to make Cyl come first, mzx brought up a good point in the PR that the current order respects the enum order. I didnt consider that.

include/z64collision_check.h
```diff
@@ -8,114 +8,168 @@
...
-    /* 0x11 */ u8 acFlags; // Information flags for AC collisions.
-    /* 0x12 */ u8 ocFlags1; // Information flags for OC collisions.
-    /* 0x13 */ u8 ocFlags2;  // Flags related to which colliders it can OC collide with.
-    /* 0x14 */ u8 colType; // Determines hitmarks and sound effects during AC collisions.
-    /* 0x15 */ u8 shape; // JntSph, Cylinder, Tris, or Quad
+    /* 0x10 */ u8 atFlags;
+    /* 0x11 */ u8 acFlags;
+    /* 0x12 */ u8 ocFlags1;
+    /* 0x13 */ u8 ocFlags2; // Flags related to which colliders it can OC collide with.
+    /* 0x14 */ u8 colType; // Determines hitmarks and sound effects during AC collisions. See `ColliderType` enum
```
hensldm, line 49: Probably not in this PR, but thoughts on shortening `colType` to just `type`. Seems weird type gets the qualifier, but `shape` doesn't

hensldm (approved):
Changes still seem good to me :)

include/z64collision_check.h
mzxrules: strong disagree. If any property in Collider were to be called type, `shape` would be it, since the shape of a collider is it's most distinguishing feature; calling `colType` simply `type` would arguably be confusing and misleading. 

The way I see it, `colType`/ColliderType is just a placeholder name until we have a more complete understanding of the ColliderType enum's purpose, and so the name is fine as is.

Dragorn421: yeah this field isn't as important as "type" may make it sound

an appropriate name may or may not be `material` or `texture`

fig02:
Merging sometime during the day on wednesday, last call for reviews

PR #1432, opened 2022-11-15
Small `Actor_UpdateAll` doc

Dragorn421:


src/code/z_actor.c
```diff
@@ -2046,18 +2058,18 @@ void Actor_UpdateAll(PlayState* play, ActorContext* actorCtx) {
...
     }
 
     if ((player->stateFlags1 & PLAYER_STATE1_6) && ((player->actor.textId & 0xFF00) != 0x600)) {
         sp74 = player->targetActor;
     }
 
-    for (i = 0; i < ARRAY_COUNT(actorCtx->actorLists); i++, sp80++) {
-        unkCondition = (*sp80 & player->stateFlags1);
+    for (i = 0; i < ARRAY_COUNT(actorCtx->actorLists); i++, canIgnoreCategoryPlayerStateFlags1++) {
+        canIgnoreCategory = (*canIgnoreCategoryPlayerStateFlags1 & player->stateFlags1);
```
AngheloAlf, line 2072: So if player has any of those or'ed flags of the current actor category then update is skipped for every actor of this category?

src/code/z_actor.c
```diff
@@ -1991,28 +1991,40 @@ void Actor_InitContext(PlayState* play, ActorContext* actorCtx, ActorEntry* play
     func_8002FA60(play);
 }
 
-u32 D_80116068[ACTORCAT_MAX] = {
+u32 sCanIgnoreCategoryPlayerStateFlags1Masks[ACTORCAT_MAX] = {
```
AngheloAlf, line 1994: The name is kinda too verbose but I can't think of a better name either :sweat:

src/code/z_actor.c
Dragorn421: > So if player has any of those or'ed flags of the current actor category

yes

> then update is skipped for every actor of this category?

not exactly,
if `canIgnoreCategory` is true(ish) then actors of the category may not be updated if the condition ```(sp74 != actor) && (actor != player->naviActor) &&
                        (actor != player->heldActor) && (&player->actor != actor->parent)``` passes

the condition can be rewritten as ```!((sp74 == actor) || (actor == player->naviActor) ||
                        (actor == player->heldActor) || (&player->actor == actor->parent))```
i.e. that condition is exceptions to ignoring all the category's actors (basically, keep updating actors interacting with player it looks like)

src/code/z_actor.c
fig02: Overall I think it'd be beneficial to use the term "freeze" over "ignore". Low level it does ignore the update, but it achieves the effect of actors freezing during various cutscene-like states or playing ocarina or something

Can try to stay more high level like sCategoryFreezeFlags and let a comment describe the intricacies instead of fitting it all in the name

AngheloAlf: I agree "freeze" is better than "ignore", but it makes me think on actual freezing tho

EllipticEllipsis: Pause?

Dragorn421: Note "freeze" is already used by `actor->freezeTimer` (so it would make sense to reuse it)

AngheloAlf: Ohh, fair point, in that case using "freeze" here sounds good to me

fig02: pause is an interesting idea, and could be discussed to be used instead of freeze. but yeah, would be good to keep everything in sync.
theres also this, which in my z_demo pr i named `haltAllActors`:
```c
                    if (!this->unk_11DE9) {
                        Actor_UpdateAll(this, &this->actorCtx);
                    }
```
so, that one should prob stay in sync with this terminology as well. whether halt is good or not for all of its uses, idk yet. need to put more thought (and it can prob happen outside of this pr)

src/code/z_actor.c
```diff
@@ -1991,28 +1991,40 @@ void Actor_InitContext(PlayState* play, ActorContext* actorCtx, ActorEntry* play
     func_8002FA60(play);
 }
 
-u32 D_80116068[ACTORCAT_MAX] = {
+u32 sFreezeCategoryPlayerFlags1Masks[ACTORCAT_MAX] = {
```
EllipticEllipsis, line 1994: I think `PlayerState` rather than `PlayerFlags`, since firstly the latter is probably going to end up being ambiguous (I'm reasonably sure Player has several other sets of flags apart from the state ones), and secondly it fits better with the enum.

src/code/z_actor.c
```diff
@@ -2046,18 +2058,18 @@ void Actor_UpdateAll(PlayState* play, ActorContext* actorCtx) {
                     refActor->world.pos.z, 0, 0, 0, 1);
     }
 
-    sp80 = &D_80116068[0];
+    freezeCategoryPlayerFlags1 = &sFreezeCategoryPlayerFlags1Masks[0];
```
EllipticEllipsis, line 2061: I noticed while reading this that it feels a bit ambiguous as to whether this is flags-per-category or categories for each flag when not read with complete context. I know the array is right next to it but I think we can still be clearer here. Maybe `Mask` can still be included? `categoryFreezeMask`?

Dragorn421: you mean
`freezeCategoryPlayerFlags1`
-> `categoryFreezeMask`
?

why not
it bothers me the order isn't the same as `sFreezeCategoryPlayerFlags1Masks` though

fig02: fwiw I think `sCategoryFreezeMasks` is also fine for the data itself. the fact that its player state flags 1 is pretty easily gleaned from usage and the fact that all of the data is already state 1 (And it can also be clarified in a commend above)

Dragorn421: I'll change it to `freezeCategoryMask`, tell me what you think

AngheloAlf (approved):

fig02 (approved):
im with P, but would probably slightly prefer Ptr. we have both in the codebase atm so its not a big deal, but we can prob standardize it more at some point later

EllipticEllipsis (approved):
Naming sp74 needs that player member renaming anywa, looks good now. I have some thoughts about the structure of the loop, but they can wait.

PR #1434, opened 2022-11-16
Skelanime/Animation header cleanup

Dragorn421:
- Move "legacy" stuff to `z64animation_legacy.h`
- Move functions out of functions.h
- Reorganize `z64animation.h` (this is probably very subjective)

Compared to the PR this is inspired by from MM, https://github.com/zeldaret/mm/pull/1147 , I didn't change LinkAnimation to PlayerAnimation as I'm not comfortable with that change yet and it's a bit involved (would massively increase the diff)


fig02:
do wonder if its time to start considering skelanime -> skelanim (sorry anime enjoyers)
at least, we abbreviate animation to anim everywhere in the system already, all else that really needs to change is some function prefixes and the skelanime struct itself. 
and this header would be better of as skelanim cause it handles both skeletons and animations

Dragorn421:
On the topic of renaming animation things, we can also discuss (eventually):

`LinkAnimation` -> `PlayerAnimation` (avoid mention "Link")

`AnimationHeader` -> `Animation` ("header" isn't really relevant nor correct)
 (and other `...Header` structs)

`Skeleton` -> `SausageSkeleton` (  https://github.com/buu342/N64-Sausage64 )
`FlexSkeleton` -> `SkinnedSkeleton` ( https://en.wikipedia.org/wiki/Skeletal_animation , sm64 people)
`z_skin` skeleton -> `WeightedSkinnedSkeleton` (trust me bro)

AngheloAlf (approved):

AngheloAlf:
Btw can we not name the Skeleton `SausageSkeleton`?

include/z64animation.h
```diff
@@ -88,41 +69,185 @@ typedef struct {
...
+
+BAD_RETURN(s32)
+SkelAnime_Init(struct PlayState* play, SkelAnime* skelAnime, SkeletonHeader* skeletonHeaderSeg,
+               AnimationHeader* animation, Vec3s* jointTable, Vec3s* morphTable, s32 limbCount);
+BAD_RETURN(s32)
+SkelAnime_InitFlex(struct PlayState* play, SkelAnime* skelAnime, FlexSkeletonHeader* skeletonHeaderSeg,
+                   AnimationHeader* animation, Vec3s* jointTable, Vec3s* morphTable, s32 limbCount);
+BAD_RETURN(s32)
+SkelAnime_InitSkin(struct PlayState* play, SkelAnime* skelAnime, SkeletonHeader* skeletonHeaderSeg,
+                   AnimationHeader* animation);
```
fig02, line 131: not sure if this was on purpose or the formatter, but I think we should bring these back up on the same line as the type

include/z64animation.h
```diff
@@ -88,41 +69,185 @@ typedef struct {
...
+} AnimationMode;
+
+typedef enum {
+    /* -1 */ ANIMTAPER_DECEL = -1,
+    /*  0 */ ANIMTAPER_NONE,
+    /*  1 */ ANIMTAPER_ACCEL
+} AnimationTapers;
+
+#define ANIM_FLAG_UPDATEY (1 << 1)
+#define ANIM_FLAG_NOMOVE (1 << 4)
```
fig02, line 92: not your change, but i think UPDATE_Y and NO_MOVE would be an improvement

Dragorn421: skelanime moveFlags need a cleanup/documentation pass anyway, I can change this now but it'll probably be touched again in the near future anyway

include/z64animation.h
Dragorn421: you mean `BAD_RETURN(s32)` on its own line? yeah that's the formatter
tbh I would prefer we deviate as little as possible from what the formatter does even in header files, it makes them much easier to maintain

Dragorn421: the formatter does the same in the .c btw
https://github.com/zeldaret/oot/blob/1587c5505ad66706ee53326c375c614c5c381e34/src/code/z_skelanime.c#L1438-L1440

fig02: I dont really agree with that sentiment in general, since we exclude .h from formatting specifically so that we can have the freedom to do what we want. and I think it looks bad like this

Dragorn421: I won't say it looks good but at least it's consistent and easier to maintain than formatting by hand (unless we're okay with breaking line length limits on prototypes, then everything is a single line, however long, and we don't ever worry about it again)

fig02: whenever I use the formatter on a header I go through and undo changes that look weird. its an understanding that youre not really supposed to use them there, but it can be convenient for some things

if you want to argue full committing to the formatter on everything, even headers, then its a discussion to have. but as things stand theres no need to be limited by it if we're not using it

Dragorn421: Should this be
```c
BAD_RETURN(s32) SkelAnime_Init(struct PlayState* play, SkelAnime* skelAnime, SkeletonHeader* skeletonHeaderSeg, AnimationHeader* animation, Vec3s* jointTable, Vec3s* morphTable, s32 limbCount);
BAD_RETURN(s32) SkelAnime_InitFlex(struct PlayState* play, SkelAnime* skelAnime, FlexSkeletonHeader* skeletonHeaderSeg, AnimationHeader* animation, Vec3s* jointTable, Vec3s* morphTable, s32 limbCount);
BAD_RETURN(s32) SkelAnime_InitSkin(struct PlayState* play, SkelAnime* skelAnime, SkeletonHeader* skeletonHeaderSeg, AnimationHeader* animation);
```

fig02: id prob do something like this personally
```c
BAD_RETURN(s32) SkelAnime_Init(struct PlayState* play, SkelAnime* skelAnime, SkeletonHeader* skeletonHeaderSeg, 
                               AnimationHeader* animation, Vec3s* jointTable, Vec3s* morphTable, s32 limbCount);
BAD_RETURN(s32) SkelAnime_InitFlex(struct PlayState* play, SkelAnime* skelAnime, FlexSkeletonHeader* skeletonHeaderSeg, 
                                   AnimationHeader* animation, Vec3s* jointTable, Vec3s* morphTable, s32 limbCount);
BAD_RETURN(s32) SkelAnime_InitSkin(struct PlayState* play, SkelAnime* skelAnime, SkeletonHeader* skeletonHeaderSeg, 
                                   AnimationHeader* animation);
```
which will prob look like shit on github, but its just roughly respecting our normal line limit with the macro still being on the same line (what youd expect the formatter to do in the first place if it wasnt dumb)

Dragorn421: alright, done

fig02 (approved):

engineer124 (approved):

PR #1437, opened 2022-11-16
Player doc: anim wrappers

Dragorn421:
Rough documentation pass on wrapper functions in z_player for switching animations

## `Player_AnimPlay...`/`Player_AnimChange...` functions:

the name depends on `LinkAnimation_Play*` vs `LinkAnimation_Change` usage :shrug: 

## `Player_AnimReplace...` functions:

Some kind of system for switching animation properly/instantly visually, in the sense that it resets horizontal speed and the prev frame root limb's rot/pos that is used for actor movement based on the root limb movement.

CDi's player docs branch calls it "AnimMovement"/"AnimWithMovement", which sounds like a better name maybe? but I haven't looked at all the details

### `AnimReplaceNormal`:

uses flags `0x1C` which seems to be a neutral set of flags (those flags need docs ofc, I'm waiting for #1434)

## `Adj` suffix:

animation is played at 2/3 the speed.

~~My theory is they initially planned anims for 30fps and this handles 20fps instead by *adj*usting the speed :shrug: , but I didn't check if the anims used with those are consistent~~

EDIT: this can only be wrong, to play 30 frames of an animation over 20 frames, the speed would need to be 3/2, not 2/3
so idk. maybe it's just to save some data, since it would save 1/3rd of the bytes compared to storing the animations at the rate it's played at :shrug:


src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -1573,14 +1573,14 @@ static LinkAnimationHeader* D_808543D4[] = {
 };
 
 // return type can't be void due to regalloc in func_8084FCAC
-BAD_RETURN(s32) func_80832210(Player* this) {
+BAD_RETURN(s32) Player_ZeroSpeedXZ(Player* this) {
     this->actor.speedXZ = 0.0f;
     this->linearVelocity = 0.0f;
 }
```
engineer124, line 1579: Worth noting that this conflicts with https://github.com/zeldaret/oot/pull/1346, where this function is named `Player_StopHorizontalMovement`

fig02: yea have been pondering on that one for a bit. I think a combination of both could work: `_StopHorizontalSpeed`
ZeroSpeedXZ is literally what it does, but saying that its stopping horizontal speed is a bit less technical i guess, and could read nicer when looking at code.
But I think "movement" is going too high level, cause movement can involve a few different factors, not just the literal speed value

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -13623,9 +13628,25 @@ void func_80850E84(Player* this, PlayState* play) {
...
+    Player_AnimReplaceNormalPlayLoopAdj2,
+    func_80851134,
+    Player_AnimPlayOnce2,
+    Player_AnimPlayLoop2,
+    func_808511D4,
+    func_808511FC,
+    func_80851294,
+    Player_AnimChangeFreezeZeroRootYawSpeed,
+    Player_AnimPlayOnceAdj2,
+    Player_AnimPlayLoopAdj2,
```
engineer124, line 13647: All the wrappers to the base functions that end in 2 eg. `Player_AnimPlayOnce2` all are for player cutscenes, so I wonder if the names should reflect that. Similar to every function in `D_80854AA4`

fig02: wait so are the "2" wrappers also used in the `D_80854AA4` functions? case thats like a layered/priority action system

fig02: i mixed it up with another function array, disregard

engineer124: Sorry, I phrased that poorly, was mainly referring to other non-2 functions in `D_80854AA4`, but they're all part of this group of functions, and every function in `D_80854AA4` is unique to `D_80854AA4`. Cdi's docs on `D_80854AA4` for reference:

```c
static void (*csModePlaybackFuncs[])(PlayState*, Player*, void*) = {
    NULL,
    Player_AnimPlaybackType0,
    Player_AnimPlaybackType1,
    Player_AnimPlaybackType2,
    Player_AnimPlaybackType3,
    Player_AnimPlaybackType4,
    Player_AnimPlaybackType5,
    Player_AnimPlaybackType6,
    Player_AnimPlaybackType7,
    Player_AnimPlaybackType8,
    Player_AnimPlaybackType9,
    Player_AnimPlaybackType10,
    Player_AnimPlaybackType11,
    Player_AnimPlaybackType12,
    Player_AnimPlaybackType13,
    Player_AnimPlaybackType14,
    Player_AnimPlaybackType15,
    Player_AnimPlaybackType16,
    Player_AnimPlaybackType17,
};
```

src/overlays/actors/ovl_player_actor/z_player.c
Dragorn421: Fwiw I went with speedXZ mostly because it's shorter

Also this name gets bundled in some other wrapper functions iirc

I'm okay with HorizontalSpeed or HorizSpeed or smth else

src/overlays/actors/ovl_player_actor/z_player.c
Dragorn421: Yeah I noticed this peculiar set of functions too

The goal of the PR was to name every wrapper somewhat consistently, so I didn't hold off from naming most of the callbacks in this subsystem, but I'd be fine introducing a (probably temporary) subprefix for it

fig02: imo its fine to just wait until someone documents the cs system

Dragorn421: reverted

src/overlays/actors/ovl_player_actor/z_player.c
fig02: yea both work. i see ZeroSpeedXZ as the mechanical approach, literally saying what the code does. while horizontal speed is longer, it might help you contextualize what the call means in the places it gets used, rather than having to think "what does setting speedXZ to zero do"

even though its a very obvious operation so it doesnt matter too much, just explaining my thought process

Dragorn421: So change to `_StopHorizontalSpeed` ?
tbh I'm not a fan of "stop" with "speed", it sounds weird to say "to stop the speed"
and I'd rather have `_StopHorizontalMovement` than `_StopHorizontalSpeed`

my tier list from best to worst (imo, ofc):
- `_ZeroSpeedXZ`
- `_StopHorizontalMovement`
- `_ZeroHorizontalSpeed`
- `_StopMovementXZ`
- `_StopSpeedXZ`
- `_StopHorizontalSpeed`

Also what would this mean for the fantastic anim wrapper names like
`Player_AnimChangeOnceMorphZeroRootYawSpeed` (yes I know it's horrible :( but it basically has to be this, there are a lot of wrappers)
(note "zero" and "speed", with "zero" applying to both "root yaw" and "speed")
`Player_AnimChangeOnceMorphZeroRootYawStopSpeed` ?

fig02: maybe we unname this for now, since we now need to have the discussion "refer to this as speed or speedXZ"
and naming this at the same time as properly naming `player->linearVelocity` would also be good.

Though I do know some of the anim wrappers do depend on this name, so its a tricky situation lol

fig02: tbh i am likely going to just open a pr to get this conversation started very soon (today unless I run into issues with naming something)

Dragorn421: He did it pog
- #1480
so waiting for that to be merged

fig02: resolving now that 1480 has been merged

fig02 (approved):
in general I think that some of these names are very literal (describing exactly what they do) when in practice, with more docs, a wrapper functions utility may become more clear and can be adjusted to reflect that.

But for now these are fine, and can be revisited if deemed necessary

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -1913,11 +1913,11 @@ void func_80832E48(Player* this, s32 flags) {
     func_808322FC(this);
 }
 
-void func_80832F54(PlayState* play, Player* this, s32 flags) {
+void Player_AnimReplaceApplyFlags(PlayState* play, Player* this, s32 flags) {
```
engineer124, line 1916: Is this suppose to mean "replace the `applyFlags`", or "replace and apply these `flags`"?

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -1590,19 +1590,19 @@ s32 func_8083224C(PlayState* play) {
...
 
-void func_80832284(PlayState* play, Player* this, LinkAnimationHeader* anim) {
+void Player_AnimPlayLoop(PlayState* play, Player* this, LinkAnimationHeader* anim) {
     LinkAnimation_PlayLoop(play, &this->skelAnime, anim);
 }
 
-void func_808322A4(PlayState* play, Player* this, LinkAnimationHeader* anim) {
+void Player_AnimPlayLoopAdj(PlayState* play, Player* this, LinkAnimationHeader* anim) {
     LinkAnimation_PlayLoopSetSpeed(play, &this->skelAnime, anim, 2.0f / 3.0f);
 }
```
engineer124, line 1603: What about using `Slower` instead of `Adj`, so `Player_AnimPlayLoopSlower`? Slightly more descriptive

fig02: did have this thought as well
I was thinking the name would probably be revisited as it may become clear with docs which anims are formatted this way/if theres any pattern

Do think it may imply that the anim itself will be really slow, when in reality the anims played through these functions look "normal", they were just "encoded" in a way that needs this playback speed

Dragorn421: "Dilute" ? :sunglasses: :droplet: 
as in, spread the same amount of butter over a larger slice of bread Pepega :fork_and_knife:  :butter:

src/overlays/actors/ovl_player_actor/z_player.c
Dragorn421: it means ' in the "AnimReplace" subsystem, "Apply" (an anim replacement) with "Flags" '

But I can see how it's a weird name.
Note this does not change the animation, it's instead typically called after an animation is changed

See also the "Player_AnimReplace... functions" bit in my opening comment above
Basically "AnimReplace" as in it does a bit more than changing the animation (the functions wrapping this one, at least)

As fig said the names are very literal :sweat: 
The flags here are mostly SkelAnime.moveFlags (see #1489 ) so that could be worth to keep in mind naming this. (but also I'm not sure `SkelAnime.moveFlags` is a perfect name so...)

src/overlays/actors/ovl_player_actor/z_player.c
fig02: adj is fine and probably accurate to the situation ("adjusted" to look normal despite the weird format)
If we can afford the whole word "Adjusted" it may be better, but I know there are wrapper functions that can cause these to get quite long

Dragorn421: Adj -> Adjusted seems fine, 4 letters is either okay because it doesn't make the name much longer or the name is already stupid long anyway (hi `Player_AnimChangeOnceMorphAdjZeroRootYawSpeed`)

can also come up with a stupid acronym like ATTS "Adjusted Two Third Speed" or A23S "Adjusted 2/3 Speed" :shrug:

Dragorn421: Packed, compacted, shrunken, slumped, crushed, squashed, mashed, pressed

fig02: I think adjusted is fine until we learn more about this. some more docs may reveal some other reason other than "compress them to a smaller size"

Dragorn421: I've changed Adj -> Adjusted
And added a `PLAYER_ANIM_ADJUSTED_SPEED` define, currently used by Adjusted anim play functions. It looks like it may be used elsewhere in player eventually if the purpose is the same

src/overlays/actors/ovl_player_actor/z_player.c
Dragorn421: I added a comment, let me know if it helps (not the greatest comment)
https://github.com/zeldaret/oot/pull/1437/commits/772f5556b02e658295e6f066feefae56e7155ae1

fig02 (approved):

hensldm (approved):

fig02:
Note: `FLAG_FUNC_80832F54_8` and `FLAG_FUNC_80832F54_9` have been renamed to `ANIM_REPLACE_APPLY_FLAG_X`, as `func_80832F54` was renamed to `Player_AnimReplaceApplyFlags` in this PR.

The name of the flags (and things in this system in general) will probably be reworked in general anyway as player docs progress, but this will work for now.

fig02:
I talked with @engineer124 about merging this despite unresolved conversations from him. As always, further discussion and follow up PRs following the merge are welcome.

PR #1438, opened 2022-11-17
Use intptr in most remaining code files

Roman971:
Another decent chunk of intptr type updates. This should cover most remaining code files, with essentially the only ones left being audio files, malloc/arena related files, and z_DLF.c which is already handled in another PR.


fig02 (commented):
as usual, i dont know shit about the uintptr topic. so im waiving my approval spot to another contributor who is more knowledgeable on the subject

it looks good to me in all aspects other than "where its appropriate to use the thing"

AngheloAlf (approved):

src/code/z_room.c
```diff
@@ -295,7 +295,7 @@ void Room_DrawBackground2D(Gfx** gfxP, void* tex, void* tlut, u16 width, u16 hei
     Room_DecodeJpeg(SEGMENTED_TO_VIRTUAL(tex));
 
     bg = (uObjBg*)(gfx + 1);
-    gSPBranchList(gfx, (u32)bg + sizeof(uObjBg));
+    gSPBranchList(gfx, (uintptr_t)bg + sizeof(uObjBg));
```
Thar0, line 298:
```suggestion
    gSPBranchList(gfx, bg + 1);
```
These casts on `bg` are actually not needed at all thanks to pointer arithmetic, although it may make it slightly less clear at a glance.

Thar0 (changes requested):
Once again I think we're being too eager to use (u)intptr types where regular pointer types are sufficient. I agree with the uses when bitwise operations are involved but not for many others, I didn't comment on them all since there are so many.

src/code/z_room.c
Roman971: Imo the code is much clearer with the sizeof, and this code is already pretty low level so the casts seem fine to me, but it's probably just personal preference

Dragorn421: I think `gSPBranchList(gfx, (Gfx*)(bg + 1));` would be better but I don't care too much

src/code/loadfragment2.c
```diff
@@ -1,7 +1,8 @@
 #include "global.h"
 
 void* Overlay_AllocateAndLoad(uintptr_t vRomStart, uintptr_t vRomEnd, void* vRamStart, void* vRamEnd) {
-    void* allocatedVRamAddr = SystemArena_MallocRDebug((s32)vRamEnd - (s32)vRamStart, "../loadfragment2.c", 31);
+    void* allocatedVRamAddr =
+        SystemArena_MallocRDebug((intptr_t)vRamEnd - (intptr_t)vRamStart, "../loadfragment2.c", 31);
```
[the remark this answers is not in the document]
Dragorn421: on the whole pointer arithmetic vs (u)intptr_t thing, I think I prefer pointer arithmetic where possible

Why have a size_t cast here though?

src/code/game.c
```diff
@@ -481,7 +481,7 @@ void* GameState_Alloc(GameState* gameState, size_t size, char* file, s32 line) {
     }
     if (ret != NULL) {
         osSyncPrintf(VT_FGCOL(GREEN));
-        osSyncPrintf("game_alloc(%08x) %08x-%08x [%s:%d]\n", size, ret, (u32)ret + size, file, line);
+        osSyncPrintf("game_alloc(%08x) %08x-%08x [%s:%d]\n", size, ret, (uintptr_t)ret + size, file, line);
```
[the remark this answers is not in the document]
Dragorn421: (for example, `ret` in this same line)

src/code/loadfragment2.c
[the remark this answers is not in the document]
Dragorn421: > Why have a size_t cast here though?

@Thar0

Dragorn421 (commented):
Any definitive ruling on

- `(uintptr_t)base + offset` vs `(u8*)base + offset`
- `(uintptr_t)a - (uintptr_t)b` vs `(u8*)a - (u8*)b`
- `printf("%x", (uintptr_t)ptr)` vs `printf("%x", ptr)`

?

src/code/z_room.c
hensldm: I also prefer just the pointer arithmetic. Also just for the sake of a slightly different perspective `&bg[1]`?

hensldm: Actually on second thought Dragorn's suggestion seems good to me. Clear that it is going past the `bg` to grab the next thing as `Gfx`.

Roman971: I changed it to `bg + 1` since it seems to be the popular option (fun fact it's basically how I wrote it when I decompiled it, until it was changed to using sizeof at some point 😅)

Roman971:
> Any definitive ruling on
> 
> * `(uintptr_t)base + offset` vs `(u8*)base + offset`
> * `(uintptr_t)a - (uintptr_t)b` vs `(u8*)a - (u8*)b`
> * `printf("%x", (uintptr_t)ptr)` vs `printf("%x", ptr)`
> 
> ?

I don't think we should try to make a ruling on these in this PR since it's relevant to a lot more than just the changes here, and it would need its own PR(s) imo. (Also it often affects codegen so it's not a simple choice we have to make)

To be clear the purpose of this PR was really just to remove uses of s32/u32 and use intptr types instead, but it doesn't mean we can't change some of them to use pointer types afterwards if we want to. I did change a few cases to use pointer types already as suggested, tho only the ones I think everyone would agree with like I mentioned above. For the rest I'd say we should leave it for a more dedicated conversation/PR

hensldm (approved):
LGTM, agreed we can probably have a more dedicated PR for the rest of the discussion.

Thar0 (approved):

fig02 (approved):

src/code/z_actor.c
```diff
@@ -2807,17 +2807,19 @@ Actor* Actor_Spawn(ActorContext* actorCtx, PlayState* play, s16 actorId, f32 pos
...
         }
 
-        actorInit = (void*)(u32)((overlayEntry->initInfo != NULL)
-                                     ? (void*)((u32)overlayEntry->initInfo -
-                                               (s32)((u32)overlayEntry->vramStart - (u32)overlayEntry->loadedRamAddr))
-                                     : NULL);
+        actorInit = (void*)(uintptr_t)(
+            (overlayEntry->initInfo != NULL)
+                ? (void*)((uintptr_t)overlayEntry->initInfo -
+                          (intptr_t)((uintptr_t)overlayEntry->vramStart - (uintptr_t)overlayEntry->loadedRamAddr))
```
Dragorn421, line 2821: Casting a uintptr_t-uintptr_t to intptr_t here, shouldn't that be the inexistent "signed size_t" or ptrdiff_t instead?

src/code/z_map_mark.c
```diff
@@ -55,18 +55,19 @@ static MapMarkData** sLoadedMarkDataTable;
...
 
     sLoadedMarkDataTable = gMapMarkDataTable;
-    sLoadedMarkDataTable = (void*)(u32)(
-        (overlay->vramTable != NULL)
-            ? (void*)((u32)overlay->vramTable - (s32)((u32)overlay->vramStart - (u32)overlay->loadedRamAddr))
-            : NULL);
+    sLoadedMarkDataTable =
+        (void*)(uintptr_t)((overlay->vramTable != NULL)
+                               ? (void*)((uintptr_t)overlay->vramTable -
+                                         (intptr_t)((uintptr_t)overlay->vramStart - (uintptr_t)overlay->loadedRamAddr))
```
Dragorn421, line 69: Same

src/code/z_play.c
```diff
@@ -396,12 +396,11 @@ void Play_Init(GameState* thisx) {
...
     zAllocSize = THA_GetRemaining(&this->state.tha);
-    zAlloc = (u32)GameState_Alloc(&this->state, zAllocSize, "../z_play.c", 2918);
+    zAlloc = (uintptr_t)GameState_Alloc(&this->state, zAllocSize, "../z_play.c", 2918);
     zAllocAligned = (zAlloc + 8) & ~0xF;
-    ZeldaArena_Init((void*)zAllocAligned, zAllocSize - zAllocAligned + zAlloc);
+    ZeldaArena_Init((void*)zAllocAligned, zAllocSize - (zAllocAligned - zAlloc));
     // "Zelda Heap"
-    osSyncPrintf("ゼルダヒープ %08x-%08x\n", zAllocAligned,
-                 (s32)(zAllocAligned + zAllocSize) - (s32)(zAllocAligned - zAlloc));
+    osSyncPrintf("ゼルダヒープ %08x-%08x\n", zAllocAligned, zAllocAligned + zAllocSize - (s32)(zAllocAligned - zAlloc));
```
Dragorn421, line 403: Not same but same pattern, the uintptr_t subtraction is cast to s32 here

Dragorn421:
Besides the other comments I just made, sure let's go

src/code/z_actor.c
Roman971: Maybe but that would mean figuring out which type to use and I'd rather not get into that debate here😅(we don't have either of these types in the repo rn btw)

At least for now intptr_t works fine imo and should make it 64 bit compatible

src/code/z_play.c
Roman971: I think this is fine to keep as s32 since it's the alignment difference and should never be more than a few bytes, but sure if we had a signed size_t it might make sense to use it here

src/code/z_actor.c
Dragorn421: compared to a s32 cast it would be harder to catch that for fixing up later

src/code/z_play.c
Dragorn421: the standard types are not just about bit width but defined as "can hold a value representing <x> and <list of uses> will work", though I agree this can stay s32 since the code deeply assumes the n64 architecture anyway

I mostly pointed this out in contrast to the other 2 occurrences

src/code/z_actor.c
Roman971: A s32 cast wouldn't be 64 bit compatible here so I would rather not downgrade it. I think it would be better to keep a note that this pattern could be revisited if we decide to use some form of signed size_t

Dragorn421: This is where I'm not sure what the goal behind using the proper types is, I would assume proper C but using `(intptr_t)` here is not proper C (it's not specified as able to hold a difference between `uintptr_t`s), nor is `(s32)` of course.
I hear the argument that it makes it 64bit compatible but that's just an implementation detail of the 64bit implementations we know of, honestly as a temporary fix I would rather see a new type introduced now, perhaps only temporarily, than introducing a temporary mistake
We could even do something stupid like
```suggestion
#ifndef NON_MATCHING
                          (s32)
#endif
                          ((uintptr_t)overlayEntry->vramStart - (uintptr_t)overlayEntry->loadedRamAddr))
```

Roman971: As far as I'm concerned, the main goal behind using intptr types is to to make the code 64 bit compatible (otherwise I think it's not worth it if the types are just meant as an alias to s32/u32). I agree `intptr_t` isn't the most ideal type here, but afaik `intptr_t` is really just defined as an integer type capable of holding a pointer which is all we need here, so I think it's "fine" to use it in this context.

Also fwiw `intptr_t` is what MM has been using in this pattern, and it was part of what we wanted to merge in [1334](https://github.com/zeldaret/oot/pull/1334) (without the macros)

Roman971: I don't really mind reverting it but I think it would be unfortunate and it's definitely going to come up again in the future

Dragorn421 (approved):
ok

PR #1441, opened 2022-11-18
`z_en_toryo` doc

blackgamma7:
- named funcs and fields
- clarified some values with enums
- removed align bytes from struct

### DIFF
--- a/src/overlays/actors/ovl_En_Toryo/z_en_toryo.c
+++ b/src/overlays/actors/ovl_En_Toryo/z_en_toryo.c
@@ -14,7 +14,7 @@ void EnToryo_Destroy(Actor* thisx, PlayState* play);
 void EnToryo_Update(Actor* thisx, PlayState* play);
 void EnToryo_Draw(Actor* thisx, PlayState* play);
 
-void func_80B20914(EnToryo* this, PlayState* play);
+void EnToryo_TalkAction(EnToryo* this, PlayState* play);
 s32 EnToryo_OverrideLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Vec3f* pos, Vec3s* rot, void* thisx);
 void EnToryo_PostLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Vec3s* rot, void* thisx);
 
@@ -128,7 +128,7 @@ void EnToryo_Init(Actor* thisx, PlayState* play) {
                      sEnToryoAnimation.morphFrames);
     this->stateFlags |= 8;
     this->actor.targetMode = 6;
-    this->actionFunc = func_80B20914;
+    this->actionFunc = EnToryo_TalkAction;
 }
 
 void EnToryo_Destroy(Actor* thisx, PlayState* play) {
@@ -137,7 +137,7 @@ void EnToryo_Destroy(Actor* thisx, PlayState* play) {
     Collider_DestroyCylinder(play, &this->collider);
 }
 
-s32 func_80B203D8(EnToryo* this, PlayState* play) {
+s32 EnToryo_TalkRespond(EnToryo* this, PlayState* play) {
     s32 pad;
     Player* player = GET_PLAYER(play);
     s32 ret = 1;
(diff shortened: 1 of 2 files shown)


src/overlays/actors/ovl_En_Toryo/z_en_toryo.c
```diff
@@ -281,64 +281,64 @@ s32 func_80B206A0(EnToryo* this, PlayState* play) {
...
+        if ((posX >= 0) && (posX <= SCREEN_WIDTH) && (posY >= 0) && (posY <= SCREEN_HEIGHT)) {
+            this->actor.textId = EnToryo_GetTextId(this, play);
             func_8002F298(&this->actor, play, 100.0f, EXCH_ITEM_POACHERS_SAW);
         }
     }
 }
 
-void func_80B20914(EnToryo* this, PlayState* play) {
+void EnToryo_TalkAction(EnToryo* this, PlayState* play) {
     SkelAnime_Update(&this->skelAnime);
```
fig02, line 339: usually we like to name the action functions as just the verb itself (Tall/Walk/Run/Whatever)
The function that you called _Talk id say is more of a lower level routine like `UpdateMessageState`/`HandleDialogue` or something similar (not set on those names, just brainstorming)

This action runs, really, during the duration of this actors lifetime. So its more than just "Talk", which makes it actually pretty hard to name. Its handles both idle and talking. So I dont have a sufficient suggestion yet

Dragorn421: I can understand why calling this "Action" since it's the only function `actionFunc` is set to 8)

I don't have outstanding ideas either
`EnToryo_Main`
`EnToryo_UpdateTalk`
`EnToryo_Action`
`EnToryo_Be`
`EnToryo_BeThere`
`EnToryo_Live`
`EnToryo_Act`
... :shrug:

Dragorn421 (approved):
Naming functions for NPCs that only talk is hard...
This looks OK to me, though I'm not brimming with enthusiasm I don't see how to improve it

src/overlays/actors/ovl_En_Toryo/z_en_toryo.c
hensldm: Personally, I think `EnToryo_Idle` would be good enough.

hensldm (approved):
Did leave one comment in the resolved conversation.

src/overlays/actors/ovl_En_Toryo/z_en_toryo.c
fig02: think i agree, mainly because if you look at it from the perspective of what "action" it handles most, the actor is going to be idle 99% of the time.

if the actor was implemented "properly" this function would have then passed execution on to a proper talking action, instead of just calling a function directly on its own.
But yeah idk, annoying situation all around

Dragorn421: So is anyone expecting `_Main` to be changed to `_Idle` ? (I'd be fine with it personally)

fig02: yes, id like it changed

src/overlays/actors/ovl_En_Toryo/z_en_toryo.c
```diff
@@ -233,11 +233,11 @@ s32 func_80B205CC(EnToryo* this, PlayState* play) {
     return ret;
 }
 
-u32 func_80B20634(EnToryo* this, PlayState* play) {
+u32 EnToryo_OfferItem(EnToryo* this, PlayState* play) {
     u32 ret;
```
fig02, line 237: we recently started naming the "get item" functions, and have potentially settled on using "offer" for offering the get item exchange. considering this doesnt deal with that interaction at all, and is instead more about getting text ids, we should maybe reconsider

src/overlays/actors/ovl_En_Toryo/z_en_toryo.c
```diff
@@ -281,64 +281,64 @@ s32 func_80B206A0(EnToryo* this, PlayState* play) {
     return ret;
 }
 
-void func_80B20768(EnToryo* this, PlayState* play) {
+void EnToryo_Talk(EnToryo* this, PlayState* play) {
     Player* player = GET_PLAYER(play);
```
fig02, line 285: I expressed this in my previous comment, but may have not been clear, but I dont think this one should be named a simple verb like an action function when it isnt really one. Id be fine with `HandleTalking` or something

Dragorn421 (approved):
:+1:

fig02 (approved):

PR #1431, opened 2022-11-12
`z_en_bird` doc

blackgamma7:
Documented `z_en_bird`

PR #1442, opened 2022-11-18
`z_en_reeba` doc

blackgamma7:
Leever fields and functions now ID'd.


src/overlays/actors/ovl_En_Reeba/z_en_reeba.c
```diff
@@ -203,17 +203,17 @@ void func_80AE5054(EnReeba* this, PlayState* play) {
...
+            Math_ApproachF(&this->offsetYStep, 300.0f, 1.0f, 5.0f);
         } else {
-            this->unk_288 = 0.0f;
+            this->offsetYStep = 0.0f;
             this->actor.shape.yOffset = 0.0f;
             playerLinearVel = player->linearVelocity;
 
-            switch (this->unk_280) {
+            switch (this->rotYSwitch) {
                 case 0:
```
fig02, line 217: Id definitely avoid calling this "switch" just because its used in a switch statement. it has more meaning than that.

higher level it seems to be more about how it aims itself before attacking link. some modes here are trying to aim a bit in front of link to predict where hes walking, some just aim where hes standing at the time of surfacing. some more thought will need to be put into naming these modes, and naming the variable sufficiently after that point

if you dont want to work out those different types now, thats fair, but id leave it unnamed until theyre figured out

Dragorn421 (commented):
good job

src/overlays/actors/ovl_En_Reeba/z_en_reeba.h
```diff
@@ -15,17 +15,17 @@ typedef struct EnReeba {
...
-    /* 0x0270 */ s16 unk_270;
-    /* 0x0272 */ s16 unk_272;
-    /* 0x0274 */ s16 unk_274;
-    /* 0x0276 */ s16 unk_276;
-    /* 0x0278 */ s16 unk_278;
+    /* 0x0270 */ s16 bigLeeverTimer; // big Leever uses to close and make distance from Link.
+    /* 0x0272 */ s16 surfaceTimer;
+    /* 0x0274 */ s16 sfxTimer; // delay between move sounds. 2x if big Leever.
+    /* 0x0276 */ s16 painTimer;
+    /* 0x0278 */ s16 stunTimer;
```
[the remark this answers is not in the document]
blackgamma7: It seems to also be used when surfacing as a timer before moving. I have renamed it to reflect such.

src/overlays/actors/ovl_En_Reeba/z_en_reeba.c
```diff
@@ -499,76 +510,76 @@ void func_80AE5C38(EnReeba* this, PlayState* play) {
...
-            this->unk_27E = 0;
+            this->stunType = LEEVER_STUN_NONE;
 
             switch (this->actor.colChkInfo.damageEffect) {
-                case 11: // none
-                case 12: // boomerang
-                    if ((this->actor.colChkInfo.health > 1) && (this->unk_27E != 4)) {
-                        this->unk_27E = 4;
+                case LEEVER_DMGEFF_UNUSED: 
+                case LEEVER_DMGEFF_BOOMERANG:
```
engineer124, line 538: The extra space on these switch cases indicates this needs a formatter run

src/overlays/actors/ovl_En_Reeba/z_en_reeba.c
```diff
@@ -499,76 +510,76 @@ void func_80AE5C38(EnReeba* this, PlayState* play) {
...
-                    if (this->unk_27E != 4) {
-                        this->unk_27E = 4;
+                case LEEVER_DMGEFF_UNK:
+                    if (this->stunType != LEEVER_STUN_OTHER) {
+                        this->stunType = LEEVER_STUN_OTHER;
                         Actor_SetColorFilter(&this->actor, 0, 0xFF, 0, 80);
-                        this->actionfunc = func_80AE58EC;
+                        this->actionfunc = EnReeba_SetupStunned;
                     }
                     break;
```
engineer124, line 584: Can this switch case fit a
```c
default:
    break;
```
at the end?

src/overlays/actors/ovl_En_Reeba/z_en_reeba.h
```diff
@@ -15,17 +15,17 @@ typedef struct EnReeba {
...
     /* 0x0268 */ char unk_268[0x4];
     /* 0x026C */ EnReebaActionFunc actionfunc;
-    /* 0x0270 */ s16 unk_270;
-    /* 0x0272 */ s16 unk_272;
-    /* 0x0274 */ s16 unk_274;
-    /* 0x0276 */ s16 unk_276;
-    /* 0x0278 */ s16 unk_278;
+    /* 0x0270 */ s16 bigLeeverTimer; // big Leever uses to close and make distance from Link.
+    /* 0x0272 */ s16 surfaceTimer;
+    /* 0x0274 */ s16 sfxTimer; // delay between move sounds. 2x if big Leever.
```
engineer124, line 20:
```suggestion
    /* 0x0274 */ s16 sfxTimer; // delay between moving sfx. 2x if big Leever.
```

src/overlays/actors/ovl_En_Reeba/z_en_reeba.h
```diff
@@ -15,17 +15,17 @@ typedef struct EnReeba {
...
+    /* 0x0278 */ s16 waitTimer;
     /* 0x027A */ s16 isBig;
-    /* 0x027C */ s16 unk_27C;
-    /* 0x027E */ s16 unk_27E;
-    /* 0x0280 */ s16 unk_280;
-    /* 0x0284 */ f32 unk_284;
-    /* 0x0288 */ f32 unk_288;
+    /* 0x027C */ s16 unkDamageField; // set when hit with ice and most other weapons. Never read.
+    /* 0x027E */ s16 stunType;
+    /* 0x0280 */ s16 aimType; // inc'd by spawner. dictates rotateY modification
```
engineer124, line 26:
```suggestion
    /* 0x0280 */ s16 aimType; // incremented by the spawner. Dictates rotateY modification
```
I wasn't sure what this abbreviation was, had to look up the code. I think we can afford "incremented" here.

src/overlays/actors/ovl_En_Reeba/z_en_reeba.c
Dragorn421: Right, we made a somewhat recent style decision for switches, add a default case if it matches, and put line breaks in between cases
Like this one here for example
https://github.com/zeldaret/oot/blob/b57f2162ee71b23643fd4e4133f7455905c2969a/src/code/z_kankyo.c#L351-L383

src/overlays/actors/ovl_En_Reeba/z_en_reeba.h
fig02: Im not against this change, but I also think we dont need to be this strict about referring to sfx as sfx everywhere. high level in this context "sound" is fine too imo

src/overlays/actors/ovl_En_Reeba/z_en_reeba.c
```diff
@@ -232,17 +243,17 @@ void func_80AE5054(EnReeba* this, PlayState* play) {
             }
 
             if (this->isBig) {
```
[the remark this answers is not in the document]
Dragorn421: Bump on this suggested change (isBig vs type != small), was it not done on purpose?

src/overlays/actors/ovl_En_Reeba/z_en_reeba.h
```diff
@@ -15,24 +15,28 @@ typedef struct EnReeba {
...
     /* 0x028C */ f32 scale;
     /* 0x0290 */ ColliderCylinder collider;
 } EnReeba; // size = 0x02DC
 
 typedef enum {
-    /* 0 */ LEEVER_SMALL,
-    /* 1 */ LEEVER_BIG
+    /* 0 */ LEEVER_TYPE_SMALL,
+    /* 1 */ LEEVER_TYPE_BIG
 } LeeverParam;
```
engineer124, line 36:
```suggestion
    /* 0 */ LEEVER_TYPE_SMALL,
    /* 1 */ LEEVER_TYPE_BIG
} LeeverType;
```

src/overlays/actors/ovl_En_Reeba/z_en_reeba.c
engineer124: bump on adding a space after the `break;` here

src/overlays/actors/ovl_En_Reeba/z_en_reeba.c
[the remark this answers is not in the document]
blackgamma7: I'm concerned that by no longer treating it as a bool, the code may be more cluttered. It seems only to check for a 0 or non-zero value. The "Reeba Boss Appears" log also suggests this field was likely solely used as a bool determining if it was the big purple Leever or small green one.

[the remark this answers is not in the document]
fig02: imo its not cluttering anything. actors having a type variable and comparing aginst a type enum is a common construction. it only works out here cause the leever happens to have 2 types (0 and 1). You can imagine a modding scenario where more types are added (and it would work just fine with this variable)

that print message doesnt say anything about the variable and its type/how it was treated. its just to let the devs know that a leever that is big has spawned :D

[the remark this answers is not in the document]
blackgamma7: another point of conflict comes from a `[note from another PR](https://github.com/zeldaret/oot/pull/1431#discussion_r1020726133)` - wherein it's recommended to treat the int as a bool if it still matches. Thus far, this seems to be the case, and, IMO, `if (this->isBig)` seems to have as much legibility and conveys as much information as `if (this->leeverType == LEEVER_TYPE_BIG)` without being as verbose.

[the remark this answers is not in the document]
fig02: that comment isnt really relevant to this situation. that one is pretty clearly a boolean.
This one is not as cut and dry. It *can* be treated as a boolean yes, but it also can just as well be a type variable (which, in the case of this actor, only has two types, 0 and 1)

Given that "type" is such a common pattern among all actors for this kind of thing, treating this value as a boolean is not really preferred.

src/overlays/actors/ovl_En_Reeba/z_en_reeba.c
```diff
@@ -499,81 +511,88 @@ void func_80AE5C38(EnReeba* this, PlayState* play) {
...
                         Audio_PlayActorSfx2(&this->actor, NA_SE_EN_GOMA_JR_FREEZE);
-                        this->actionfunc = func_80AE58EC;
+                        this->actionfunc = EnReeba_SetupStunned;
                         break;
                     }
                     FALLTHROUGH;
-                case 14:
-                    this->unk_27C = 6;
+
+                case LEEVER_DMGEFF_OTHER:
```
Dragorn421, line 561: (we've been putting line breaks after `break;`s but not `FALLTHROUGH;`s)

src/overlays/actors/ovl_En_Reeba/z_en_reeba.c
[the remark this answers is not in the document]
fig02: to be clear, something matching as `if (var)` doesnt mean it was a boolean originally. pretty much any `if (var != 0)` can match as `if (var)`. That comment just specifies that when we DO treat it as a boolean, it should be changed to that format.

Our point here is that we dont want to treat this as a boolean

Dragorn421:
I think the only thing holding the PR besides formatting style ( https://github.com/zeldaret/oot/pull/1442#discussion_r1050944624 ) is:

`isBig` should be changed from a boolean to a `LeeverType`-enum-using `type`

https://github.com/zeldaret/oot/pull/1442#discussion_r1032733874

Dragorn421 (approved):

fig02 (approved):

PR #1443, opened 2022-11-19
Player doc: some rotation logic/data

Dragorn421:
I tried to document TargetContext but there were too many player unks in the way, including these
Take that, player unks zfgSmug


include/z64player.h
```diff
@@ -557,15 +567,11 @@ typedef struct Player {
     /* 0x06A8 */ Actor*     unk_6A8;
     /* 0x06AC */ s8         unk_6AC;
     /* 0x06AD */ u8         unk_6AD;
-    /* 0x06AE */ u16        unk_6AE;
-    /* 0x06B0 */ s16        unk_6B0;
+    /* 0x06AE */ u16        rotsNoRestFlags; // See `NOREST_ROT_` macros. If its flag isn't set, a rot steps to a rest value.
```
AngheloAlf, line 570: Why "No Rest"?

Dragorn421: when the flag is set, the rotation does *not* step to a rest value

AngheloAlf: I see. Next question, what is a rest value? Something like the idle value?

Dragorn421: That's a good question and I'm open to suggestions on changing that

I got the term from Blender (a 3d modeling piece of software https://www.blender.org/ ), let me define a few terms

- armature = skeleton, made of bones
- pose = rotations and locations of the bones in an armature

For example when animated the armature has a different pose every frame

The "rest pose" is the pose the armature has when all its bones aren't transformed (all rotations/translations/... are 0/identity)

-------

Back to OoT, all rotations involved with `NOREST_ROT_` are stepped to 0 when their flag isn't set
For all these rotations, 0 corresponds visually to a neutral pose of Link, with only the animation affecting his bones, so I refered to those as "rest" values

Also the use case for these flags is when setting these rots to specific values, typically to look at something, which also can be seen as deviating from the "rest" animation-provided pose

engineer124: Could you maybe think of this as a "hold and release" system? So the `rot` is held to a target when the flag is set, and released back to rest when the flag is unset? So something like `HOLD_ROT_`? Or would that conflate with the "HOLD" action of "holding" something above Link's Head?

Dragorn421: For example in the else block in `func_80836AB8`, the rotations aren't simply set but stepped to different values, so "hold" may be confusing in the sense the values can be still moving, just not to 0/rest, but it may not be such a big deal?

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -3335,48 +3335,68 @@ void func_808368EC(Player* this, PlayState* play) {
...
+    s16 clampedOffset;
+    s16 valueBeforeClamp;
 
-    temp1 = temp2 = arg4 - *pValue;
-    temp2 = CLAMP(temp2, -arg5, arg5);
-    *pValue += (s16)(temp1 - temp2);
+    // Clamp value to [constraintMid - constraintRange , constraintMid + constraintRange]
+    // This is more involved than a simple `CLAMP`, to account for binang wrapping
+    offset = clampedOffset = constraintMid - *pValue;
+    clampedOffset = CLAMP(clampedOffset, -constraintRange, constraintRange);
```
engineer124, line 3355: offset vs. diff? I see diff used more often with angles, but idk

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -3335,48 +3335,68 @@ void func_808368EC(Player* this, PlayState* play) {
...
 
-    Math_ScaledStepToS(pValue, arg1, arg2);
+    Math_ScaledStepToS(pValue, target, step);
 
-    temp3 = *pValue;
-    if (*pValue < -arg3) {
-        *pValue = -arg3;
-    } else if (*pValue > arg3) {
-        *pValue = arg3;
+    valueBeforeClamp = *pValue;
```
engineer124, line 3360: Since there's also a clamp above, maybe `valueBeforeOverflow`? or `valueBeforeOverflowClamp`

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -3335,48 +3335,68 @@ void func_808368EC(Player* this, PlayState* play) {
...
-        this->unk_6AE |= 0xD9;
+        // Step the head pitch to the focus pitch.
+        // If the head cannot be pitched enough, pitch the upper body.
+        Player_ScaledStepBinangClamped(&this->upperLimbRot.x,
+                                       Player_ScaledStepBinangClamped(&this->headLimbRot.x, this->actor.focus.rot.x,
+                                                                      600, 10000, this->actor.focus.rot.x, 0),
+                                       200, 4000, this->headLimbRot.x, 10000);
+
+        // Step the upper body and head yaw to the focus yaw.
+        // Eventually prefers turning the upper body rather than the head.
```
engineer124, line 3387: What does `Eventually` mean here?

include/z64player.h
```diff
@@ -560,15 +570,11 @@ typedef struct Player {
...
+    /* 0x06B0 */ s16        upperLimbYawSecondary;
     /* 0x06B2 */ char       unk_6B4[0x004];
-    /* 0x06B6 */ s16        unk_6B6;
-    /* 0x06B8 */ s16        unk_6B8;
-    /* 0x06BA */ s16        unk_6BA;
-    /* 0x06BC */ s16        unk_6BC;
-    /* 0x06BE */ s16        unk_6BE;
-    /* 0x06C0 */ s16        unk_6C0;
+    /* 0x06B6 */ Vec3s      headLimbRot;
+    /* 0x06BC */ Vec3s      upperLimbRot;
```
engineer124, line 577: Is `Limb` necessary here? Which limbs are encompassed in `upperLimbRot`?

include/z64player.h
```diff
@@ -560,15 +570,11 @@ typedef struct Player {
     /* 0x06A8 */ Actor*     unk_6A8;
     /* 0x06AC */ s8         unk_6AC;
     /* 0x06AD */ u8         unk_6AD;
-    /* 0x06AE */ u16        unk_6AE;
-    /* 0x06B0 */ s16        unk_6B0;
+    /* 0x06AE */ u16        rotsNoRestFlags; // See `NOREST_ROT_` macros. If its flag isn't set, a rot steps to a rest value.
+    /* 0x06B0 */ s16        upperLimbYawSecondary;
```
engineer124, line 574:
```suggestion
    /* 0x06B0 */ s16        upperLimbYawOffset;
```
Maybe?

src/overlays/actors/ovl_player_actor/z_player.c
Dragorn421: When turning towards something, both the head and upper body rotate towards the target

Since the head is attached to the upper body (*citation needed*), it turns faster overall and faces the target before the upper body does

The upper body then continues to turn towards the target, with the head turning backwards to accommodate for the upper body rotating and to keep looking the right way

So *eventually* the upper body is turned towards the target, with the head actually having no rotation of its own (relative to the upper body)

Does that make sense?

(at least this is what I understood of this when I first looked at it and wrote this. maybe it's wrong :cold_sweat: )

include/z64player.h
Dragorn421: It corresponds to `PLAYER_LIMB_UPPER`

include/z64player.h
Dragorn421: why not

"secondary" is because
```c
if (!(this->rotsNoRestFlags & NOREST_ROT_UPPER_Y)) {
    if (this->upperLimbYawSecondary != 0) {
        Player_ApproachZeroBinang(&this->upperLimbYawSecondary);
    } else {
        Player_ApproachZeroBinang(&this->upperLimbRot.y);
    }
}
```
, it steps this one to 0 before the main one

include/z64player.h
engineer124: It's a subtle difference, but what about `nonRestFlags`? For some reason, sounds better to me but can't articulate why. Might just be me

engineer124: Maybe it's because `no rest` is two words, so it feels more separate, while `non-rest` is one word, so it's easier to group together as a single concept?

fig02: I need to review the pr in full again, but I think in general itd be good to try a positive name instead of a negative one, like "modified" instead of "no rest" (which i interpret to mean "non default, offset from the rotation the animation is applying)

Dragorn421: It is related to three rotations / rotation offsets: focus, head, upper (body)

I'm not sure what's up with the focus one, but head and upper are I guess used to make Link face something

The flags control if these rotations are stepped toward "0" (not actually always 0 but something that is "normal"/"default" hence "rest")
- flag unset = rotations are stepped toward "0"
- flag set = rotations are not stepped toward "0"

Phrased this way it makes it arduous to make a positive name but I'm not sure how to phrase it another way 

maybe like "flag set = rot is active" but idk I don't have many ideas

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -3335,48 +3335,68 @@ void func_808368EC(Player* this, PlayState* play) {
...
-    s16 temp3;
+/**
+ * Step a value by `step` to a `target` value.
+ * Constrains the value to be no further than `constraintRange` from `constraintMid` (accounting for wrapping).
+ * Constrains the value to be no further than `overflowRange` from 0.
+ * If this second constraint is enforced, return how much the value was past by the range, or return 0.
+ *
+ * @return The amount by which the value overflowed the absolute range defined by `overflowRange`
+ */
+s32 Player_ScaledStepBinangClamped(s16* pValue, s16 target, s16 step, s16 overflowRange, s16 constraintMid,
```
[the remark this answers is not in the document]
hensldm: Since the args are named with `constraint`, I think this would make sense.

include/z64player.h
hensldm: From what I can tell down below, it seems like setting this flag just stops rotation back to the rest position (0). You might have been slightly joking, but `activeRot` doesn't seem too bad. Could possibly also do something like `stayRot/remainRot`, or maybe `keepRot`?

engineer124: Just brainstorming: what about something a little more general like `overrideRot` or `overrideRotFlags`, and leave it to the comments to describe that the default behaviour is to restore the rot to resting position, and that by setting this flag, it gives you manual control to control the rotations?

fig02 (approved):
I think the name of "upper limb" needs to change at some point, but obv not in scope here

fig02:
Would like to revive this and get it merged so I can continue to build on this documentation.

Dragorn421:
Is there something I need to do?

fig02:
no just another review would be good

include/z64player.h
```diff
@@ -832,15 +842,11 @@ typedef struct Player {
     /* 0x06A8 */ Actor* unk_6A8;
     /* 0x06AC */ s8 unk_6AC;
     /* 0x06AD */ u8 unk_6AD;
-    /* 0x06AE */ u16 unk_6AE;
-    /* 0x06B0 */ s16 unk_6B0;
+    /* 0x06AE */ u16 unk_6AE_rotFlags; // See `UNK6AE_ROT_` macros. If its flag isn't set, a rot steps to 0.
+    /* 0x06B0 */ s16 upperLimbYawSecondary;
```
engineer124, line 846: What is the secondary yaw? Is `upperLimbRot.y` the primary yaw? Could you add a small comment on how they differ?

fig02: It looks like its only meaningfully used when riding a horse. Id be fine with this temporary name (and no comment) for now until we figure it out more, i imagine the name will change

engineer124 (approved):

PR #1445, opened 2022-11-21
`z_actor` documentation pass

AngheloAlf:
Porting some documentation from zeldaret/mm#1154

- Name the `Actor_OfferGetItem` family of functions
- Name `Actor_SetClosestSecretDistance` and `Player.closestSecretDistSq`
- `COLORFILTER_` macros for the `Actor_SetColorFilter` function and `Actor.colorFilterParams` member
  - Something weird happens with the parameters of `Actor_SetColorFilter` because they are flags but for some reason they are signed which leads to some weird `-0x8000` shenanigans which does not happen on MM. I'm not really sure how we'll want to handle this
  - The weirdest example of this happens in EnRd: 
  ```c
  Actor_SetColorFilter(&this->actor, -COLORFILTER_COLORFLAG_RGB, -0x8000 | 200, COLORFILTER_XLUFLAG_OPA, 255);
  ```

I didn't port the rest of that MM PR mainly because those changes are handled in other PRs or are MM-specific


src/overlays/actors/ovl_En_Box/z_en_box.c
```diff
@@ -267,7 +267,7 @@ void EnBox_FallOnSwitchFlag(EnBox* this, PlayState* play) {
     s32 treasureFlag = this->dyna.actor.params & 0x1F;
 
     if (treasureFlag >= ENBOX_TREASURE_FLAG_UNK_MIN && treasureFlag < ENBOX_TREASURE_FLAG_UNK_MAX) {
-        func_8002F5F0(&this->dyna.actor, play);
+        Actor_SetClosestSecretDistance(&this->dyna.actor, play);
```
Dragorn421, line 270: :eyes:

src/code/z_actor.c
```diff
@@ -2215,25 +2222,25 @@ void Actor_Draw(PlayState* play, Actor* actor) {
     if (actor->colorFilterTimer != 0) {
         Color_RGBA8 color = { 0, 0, 0, 255 };
 
-        if (actor->colorFilterParams & 0x8000) {
-            color.r = color.g = color.b = ((actor->colorFilterParams & 0x1F00) >> 5) | 7;
-        } else if (actor->colorFilterParams & 0x4000) {
-            color.r = ((actor->colorFilterParams & 0x1F00) >> 5) | 7;
+        if (actor->colorFilterParams & COLORFILTER_COLORFLAG_RGB) {
+            color.r = color.g = color.b = COLORFILTER_GET_COLORINTENSITY(actor->colorFilterParams) | 7;
```
Dragorn421, line 2226: Actually this could be better as "gray" ?
since r=g=b=intensity means go from black (intensity=0) to white (intensity=255)

src/code/z_actor.c
```diff
@@ -2497,7 +2504,10 @@ void func_800315AC(PlayState* play, ActorContext* actorCtx) {
     CLOSE_DISPS(play->state.gfxCtx, "../z_actor.c", 6563);
 }
 
-void func_80031A28(PlayState* play, ActorContext* actorCtx) {
+/**
+ * Kills every actor which its object is not loaded
```
Dragorn421, line 2508:
```suggestion
 * Kill every actor which depends on an object that is not loaded.
```

src/code/z_actor.c
```diff
@@ -3631,7 +3644,7 @@ void func_8003424C(PlayState* play, Vec3f* arg1) {
 }
 
 void Actor_SetColorFilter(Actor* actor, s16 colorFlag, s16 colorIntensityMax, s16 xluFlag, s16 duration) {
```
Dragorn421, line 3646: xluFlag could be renamed to bufFlag maybe (as in, opa buffer, xlu buffer)
mostly suggesting it because `xluflag_opa` looks weird

include/z64actor.h
```diff
@@ -169,6 +169,17 @@ typedef struct {
 #define ACTOR_FLAG_27 (1 << 27)
 #define ACTOR_FLAG_28 (1 << 28)
 
+#define COLORFILTER_GET_XLUFLAG(colorFilterParams) ((colorFilterParams) & 0x2000)
+#define COLORFILTER_GET_COLORINTENSITY(colorFilterParams) (((colorFilterParams) & 0x1F00) >> 5)
+#define COLORFILTER_GET_DURATION(colorFilterParams) ((colorFilterParams) & 0xFF)
+
+#define COLORFILTER_COLORFLAG_RGB  0x8000
+#define COLORFILTER_COLORFLAG_R    0x4000
+#define COLORFILTER_COLORFLAG_B    0x0000
```
Dragorn421, line 178: tbh spelling out "red" and "blue" in full could be worth it?

Dragorn421 (commented):
:medal_sports:

src/code/z_actor.c
AngheloAlf: Ohh, good point. I'll rename it

src/overlays/actors/ovl_En_Bb/z_en_bb.c
```diff
@@ -1083,14 +1083,14 @@ void EnBb_SetupStunned(EnBb* this) {
     }
     switch (this->dmgEffect) {
         case 8:
-            Actor_SetColorFilter(&this->actor, -0x8000, 0xC8, 0, 0x50);
+            Actor_SetColorFilter(&this->actor, -COLORFILTER_COLORFLAG_GRAY, 200, COLORFILTER_BUFFLAG_OPA, 80);
```
Dragorn421, line 1086: about `-COLORFILTER_COLORFLAG_GRAY`, I'm on board with just using `COLORFILTER_COLORFLAG_GRAY` instead :shrug:

AngheloAlf: I swear I tried this before and it didn't match, but now it does, I guess I did something wrong before

src/overlays/actors/ovl_En_Ge2/z_en_ge2.c
```diff
@@ -556,7 +556,7 @@ void EnGe2_Update(Actor* thisx, PlayState* play) {
     } else if (this->collider.base.acFlags & AC_HIT) {
         if ((this->collider.info.acHitInfo != NULL) &&
             (this->collider.info.acHitInfo->toucher.dmgFlags & DMG_HOOKSHOT)) {
-            Actor_SetColorFilter(&this->actor, 0, 120, 0, 400);
+            Actor_SetColorFilter(&this->actor, COLORFILTER_COLORFLAG_BLUE, 120, COLORFILTER_BUFFLAG_OPA, 400);
```
Dragorn421, line 559: I took a quick look if every `Actor_SetColorFilter` usage was mindful of the `duration` being 8bits...
Turns out no, at least this one isn't (haven't found others but it was a really quick glance)

`400 = 0b1_1001_0000` so the maxDuration in the bitfield becomes `0b1001_0000 = 144` and there's a bit leaking into the maxIntensity

However the timer is still initialized to 400, I wonder if that causes a noticeable something in-game

Anyway, is this a `@bug` ?

AngheloAlf: I'm not sure if I would count it as a bug, maybe a limitation/bug of the system? It technically allows to set a duration over 255 since that parameter is a `s16` but as you said it is leaking one extra bit, with the weird bitpacking it ends up being an extra 8 (out of 255) of the color intensity. It probably isn't that noticeable but it may be worth checking

Not sure if we should tag it as a `@bug`, or if it should be tagged in the actor or in the `Actor_SetColorFilter` function itself

Dragorn421: I tested in-game and it doesn't do anything special...

Because `colorFilterTimer` is a u8 so the 400 passed to it also becomes 144 PepeLaugh

Definitely a bug then imo. The system just doesn't handle durations over 255

hensldm (approved):
LGTM

src/code/z_actor.c
```diff
@@ -1570,7 +1570,14 @@ u32 Actor_HasParent(Actor* actor, PlayState* play) {
     }
 }
 
-s32 func_8002F434(Actor* actor, PlayState* play, s32 getItemId, f32 xzRange, f32 yRange) {
+/**
+ * Allows to pick up an item (GetItem or GI), lift an actor or catch various actors in bottles
```
engineer124, line 1574:
```suggestion
 * Allows to pick up an item (GetItem), lift an actor or catch various actors in bottles
```
GI and GetItem are the same right? Or is this just here to help with the abbreviation

fig02: I think we need to workshop the wording of this comment in general more. I tried writing a new one, but the exact words arent coming to me at the moment. The current wording is still as if its called "Actor_PickUp", so we need to get into a bit more of the intricacies of an "offer" and what that means for the process of getting an item

Dragorn421: my attempt:
(current comment)
```c
 * Allows to pick up an item (GetItem or GI), lift an actor or catch various actors in bottles
 * within the specified range.
 *
 * GI_NONE is usually used as a special case to lift an actor
 * GI_MAX is usually used to catch an actor in a bottle
```
->
```c
 * This function covers various interactions with the player actor, using Get Item IDs (see `GetItemID` enum).
 * It is typically used to give items to the player, but also has other purposes.
 *
 * This function does not perform the requested action itself, it only carries the request onto the player actor if context allows it (e.g. the player is in range and not busy with certain things).
 * Hence its name using "Offer", as it is up to the player actor to honor the request.
 *
 * This function handles most requests the same, besides a few checks.
 * The following description of what the `getItemId` values can do is provided here for completeness, but these behaviors are entirely out of the scope of this function. All behavior is defined by the player actor.
 *
 * Positive values (`GI_NONE < getItemId < GI_MAX`): give an item to the player. The player may not get it immediately (for example if underwater), but is expected to in a near future.
 * Negative values (`-GI_MAX < getItemId < GI_NONE`): used by treasure chests to indicate the chest can be opened (by pressing A)
 * `GI_NONE`: allows the player to pick up the actor (by pressing A), to carry it around.
 * `GI_MAX`: allows the player to catch the actor in a bottle.
```

fig02: yea this is a good start

- I find the `GI_NONE < getItemId < GI_MAX` notation kinda confusing and feel it can be omitted, but that may just be me
- in a near future -> in the near future
- for chests, we can try to fit in that its the negative version of an otherwise positive get item id. idk how to word it well
- "Allows the player to catch *specific actors* in a bottle", just want to stress that not any actor can be caught in a bottle, player has to have implemented it for specific actors on his end
- `@return true if Player has passed all checks and is currently able to accept an offer, false if not` or something like that

Dragorn421: Not sure about `GI_NONE < getItemId < GI_MAX`, I feel like it's needed to exclude these specific `GI_` values, since they're otherwise part of Get Item IDs
Also shows `-GI_MAX` does nothing (wasted opportunity for Morita there to map a fishing thing to it)

2nd draft taking into account other remarks
```
 * This function covers various interactions with the player actor, using Get Item IDs (see `GetItemID` enum).
 * It is typically used to give items to the player, but also has other purposes.
 *
 * This function does not perform the requested action itself, it only carries the request onto the player actor if
 * context allows it (e.g. the player is in range and not busy with certain things).
 * Hence its name using "Offer", as it is up to the player actor to honor the request.
 *
 * This function handles most requests the same, besides a few checks.
 * The following description of what the `getItemId` values can do is provided here for completeness, but these
 * behaviors are entirely out of the scope of this function.
 * All behavior is defined by the player actor.
 *
 * Positive values (`GI_NONE < getItemId < GI_MAX`):
 *   Give an item to the player. The player may not get it immediately (for example if underwater), but is expected to
 *   in the near future.
 * Negative values (`-GI_MAX < getItemId < GI_NONE`):
 *   Used by treasure chests to indicate the chest can be opened (by pressing A).
 *   The item gotten corresponds to the positive Get Item ID `abs(getItemId)`.
 * `GI_NONE`:
 *   Allows the player to pick up the actor (by pressing A), to carry it around.
 * `GI_MAX`:
 *   Allows the player to catch specific actors in a bottle.
 *
 * @return true If the player actor is in range, seems in capacity of handling the offer, and the offer has been passed on to it. 
```

src/code/z_actor.c
```diff
@@ -2527,7 +2537,10 @@ void Actor_FreezeAllEnemies(PlayState* play, ActorContext* actorCtx, s32 duratio
     }
 }
 
-void func_80031B14(PlayState* play, ActorContext* actorCtx) {
+/**
+ * Kill actors on room change and update flags accordingly
+ */
+void Actor_KillAllFromUnloadedRooms(PlayState* play, ActorContext* actorCtx) {
```
fig02, line 2543: i think the other stuff its doing beyond just killing actors makes it a bit more broader scope. colliders are cleared, temp flags are reset, and some message context variable is zerod (even tho its unused)

Id say something more like `Actor_ResetForRoomChange` or similar wording can be used. not settled on that exact wording, but on that track

Dragorn421: yeah when I suggested this name I missed that it did other things
I don't have name ideas rn

Dragorn421: I think `CollisionCheck_ClearContext` is only there to prevent issues with possibly holding references to actors that were just *deleted* (note this function actually doesn't just kill, and it's very similar to `Actor_UpdateAll`'s actor garbage collection
https://github.com/zeldaret/oot/blob/b57f2162ee71b23643fd4e4133f7455905c2969a/src/code/z_actor.c#L2099-L2104
)

So in hindsight this really isn't the best name, maybe "Delete" would be better than "Kill" for starters

Other than that I think the only thing to worry about is
```c
    actorCtx->flags.tempClear = 0;
    actorCtx->flags.tempSwch &= 0xFFFFFF;
```
(let's ignore the unused message context thing)

include/z64actor.h
```diff
@@ -169,6 +169,19 @@ typedef struct {
 #define ACTOR_FLAG_27 (1 << 27)
 #define ACTOR_FLAG_28 (1 << 28)
 
+#define COLORFILTER_GET_XLUFLAG(colorFilterParams) ((colorFilterParams) & 0x2000)
```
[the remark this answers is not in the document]
Dragorn421: `#define IS_BUFFLAG_XLU (params & COLORFILTER_BUFFLAG_XLU)`
may make more sense

but is this macro that necessary? why not just use `actor->colorFilterParams & COLORFILTER_BUFFLAG_XLU` directly in the two spots it's used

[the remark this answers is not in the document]
fig02: agree on macro not being needed

src/code/z_actor.c
mzxrules: It's still a bit more wordy than it needs to be, and the first line really needs to cut straight to the point quicker... right now the wording of the first line is too nebulous.

I'd say the core purpose is to send a message to the player that an "item" can be obtained, with the return value being true if the message was properly sent. We should lead with that concept in the first sentence, and bang out the details later. Similarly, the return description should be simplified to just `@return true if the message was passed to the player`, because the checks performed are complex and are better expressed by the code itself.

Paragraphs 2 and 3 are filled with needless fluff, especially if you restructure with the core purpose above in mind. I was able to cut it down to just these lines:

    This function passes the request onto the player actor only if context allows it (e.g. the player is in range and not busy).
    The following description of what the `getItemId` values can do is provided here for completeness

src/overlays/actors/ovl_En_Ge2/z_en_ge2.c
```diff
@@ -556,6 +556,7 @@ void EnGe2_Update(Actor* thisx, PlayState* play) {
     } else if (this->collider.base.acFlags & AC_HIT) {
         if ((this->collider.info.acHitInfo != NULL) &&
             (this->collider.info.acHitInfo->toucher.dmgFlags & DMG_HOOKSHOT)) {
+            // !@bug duration parameter is larger than 255 which messes with the internal bitpacking of the colorfilter
```
Dragorn421, line 559: It's probably note-worthy too that the timer (400) ends up being only 144
To be clear this has an impact on gameplay, this timer is the one used as the stun timer when hookshotting gerudos

And just a format thing but `//! @bug`, right?

src/code/z_actor.c
AngheloAlf: `_GarbageCollectFromUnloadedRooms`, `_DestroyFromUnloadedRooms`, `_DeleteFromUnloadedRooms` ?

Dragorn421: I can't think of this function as any less than "notify the actor system a room has been unloaded"

The flags clearing
```c
    actorCtx->flags.tempClear = 0;
    actorCtx->flags.tempSwch &= 0xFFFFFF;
```
is pretty important,
clearing `tempClear` means enemies in a cleared room respawn unless something (like the chest actor) set the permanent clear flag,
and clearing switch flags 0x30..0x3F is useful for some puzzles resetting when reentering the room

So I agree with Fig on naming this along the lines of "process stuff when room changes/unloads"

... `Actor_ProcessRoomUnload` ? 4Head

fig02: id prob be ok with ProcessRoomChange, though I think including clear or reset in there somehow is appropriate (like my original suggestion, `Actor_ResetForRoomChange`)

i dont think we need to say unload, since the actor system isnt really aware of any loading or unloading thats going on. just that the room is changing (hence the check on cur and prev room)

Dragorn421: I feel like "room changing" may happen at the start of a room change though, while "room unload" sounds more like it happens at the end

fig02: its the actor system handling its own end of the entire room change process

Dragorn421: I mean "room change" is a few parts, including "load next room" "both rooms are loaded, wait one of them unload" "unload other room"

fig02: right, and my point is that the actor system isnt concerned with those details about whats loaded when. it is simply handling its own things for the room change process as a whole. whether the room is loaded in memory or not isnt really relevant

Dragorn421: Yeah the room being in memory isn't super relevant, but I think being at the end of the room change is, it would be weird to kill prev rooms' actors at the *start* of the transition
For example with EnHoll, loading another room would trigger a "room change", but you wouldn't want that to instantly delete all actors, in particular if the player went back into the room it would just mean the room got depopulated for no reason

fig02: well, I dont think unload is sufficent to describe that aspect that youre looking for then. Its hinging on things being loaded or unloaded which is a lower level operation, which is not really relevant to this at all

I still think its fine to describe this as "the function that handles the actor systems tasks for the overall room change process", but I dont think we will agree on that front, and should get some other opinions

Dragorn421: Well we can always name this however and revisit later once room loading is documented (I want to do that once all transition actors PRs are in)
My impression is `func_80097534` which calls the function we're talking about is something like "commit to loading that next room", maybe having a better view of how the whole system works will bring better terminology

fig02: agree. id even say its best not to name this right now until rooms have better docs

src/code/z_actor.c
fig02: I do agree it can be trimmed down a bit. But some of the details youre cutting out are still important. it doenst need to be that succinct. This is a complicated interaction, so we can spare a few words

when I can, I will make my own edits to our current version of the comment

src/code/z_actor.c
Dragorn421: Waiting for @AngheloAlf 's input on this
(would you be fine to not name this right now?)

Dragorn421:
Good for me besides possibly having a more fleshed out comment on `Actor_OfferGetItem` (see discussion above https://github.com/zeldaret/oot/pull/1445#discussion_r1032730099 )

Big fan of the PR, it really helps with reading actors

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -10428,7 +10428,7 @@ void func_80848C74(PlayState* play, Player* this) {
 
 void func_80848EF8(Player* this) {
     if (CHECK_QUEST_ITEM(QUEST_STONE_OF_AGONY)) {
-        f32 temp = 200000.0f - (this->unk_6A4 * 5.0f);
+        f32 temp = 200000.0f - (this->closestSecretDistSq * 5.0f);
```
engineer124, line 10431:
```suggestion
        f32 temp = (SQ(200.0f) * 5.0f) - (this->closestSecretDistSq * 5.0f);
```
Perhaps? And the `4000000.0f` in this function could be `SQ(2000.0f)`
Might be worth naming `unk_6A0` and `func_80848EF8` since both are only used here with `closestSecretDistSq`, so might be good to document this as a group.

Maybe `func_80848EF8` -> `Player_RumbleNearSecret`
`unk_6A0` -> `secretRumbleTrigger`
Just brainstorming?

Not essential that these are named now though

fig02 (approved):
needs a conflict fixed

Dragorn421 (approved):
I could go even further, a conflict needs to be fixed

Dragorn421 (approved):
:pause_button: :champagne: :question:

engineer124 (approved):

PR #1446, opened 2022-11-23
Player doc: Temporary `PLAYER_ANIMGROUP` names from original animation names

Dragorn421:
Based on the names of animations in `D_80853914` for each `PLAYER_ANIMGROUP_`, I updated the names of `PLAYER_ANIMGROUP_` values.
Of course this is meant to be temporary, as noted in a comment above the enum definition

I think keeping the lowercase is good because it avoids interpretation of the original names, and it makes it look temporary


fig02 (approved):

engineer124 (approved):

Dragorn421:
Merging soon tm

hensldm (approved):

PR #1450, opened 2022-11-29
Add ZAPD configs for entrance lists

louist103:
A draft until zeldaret/ZAPD#271 is merged


fig02:
is it possible to parse entrance_table.h so we dont need to maintain two lists
Id imagine zap would even be able to use DEFINE_ENTRANCE macros on its end and stringify each entry, but idk

louist103:
> is it possible to parse entrance_table.h so we dont need to maintain two lists Id imagine zap would even be able to use DEFINE_ENTRANCE macros on its end and stringify each entry, but idk

I suppose its possible. I would rather do it all at once with the actor and object lists though.

Dragorn421:
> is it possible to parse entrance_table.h so we dont need to maintain two lists
> Id imagine zap would even be able to use DEFINE_ENTRANCE macros on its end and stringify each entry, but idk 

Idk if this really is a good idea, as entrance_table.h may be shifted and not match vanilla

fig02:
i mean by the time youre shifting it, youre already in modding territory which this isnt really relevant for.
these files are going to be committed anyway so its only a short term problem anyway i guess

Dragorn421:
> i mean by the time youre shifting it, youre already in modding territory which this isnt really relevant for.

Well it'd break shifting vanilla scenes but if that's fine then whatever. It can be worked around anyway (checkout master before extracting scenes)

> these files are going to be committed anyway so its only a short term problem anyway i guess

Sorry I didn't get that part?

fig02:
scene files will be committed at some point, with only the "sensitive" assets being extracted. so things like actor/object/entrance lists will be committed to the repo, and there will be no need to maintain external lists to extract them

Dragorn421:
Ah right, I forgot about that

This is a very distant future though so imo we should still consider implications of these temporary changes
(personally I don't have further remarks and I'm ok with it)

It's very nice to see this done in ZAPD!

fig02 (approved):

tools/ZAPDConfigs/MqDbg/SpecialEntranceList_OoTMqDbg.txt
```diff
@@ -0,0 +1,7 @@
+ENTR_RETURN_YOUSEI_IZUMI_YOKO
+ENTR_RETURN_SYATEKIJYOU
+ENTR_RETURN_2
+ENTR_RETURN_SHOP1
+ENTR_RETURN_4
+ENTR_RETURN_DAIYOUSEI_IZUMI
+ENTR_RETURN_GROTTO
```
Dragorn421, line 7: at least these have changed since the PR was opened

I forgot if the other entrances also did

Dragorn421:
Also is there a reason this is separate from #1457 ?

Dragorn421:
> Also is there a reason this is separate from #1457 ?

bump

PR #1468, opened 2022-12-18
Move libultra functions and variables to libultra headers

Thar0:
Not totally happy with this, since `macros.h` and the new `stack.h` are not libultra headers but are still used in some libultra files for now, however this is the bulk of the changes and further improvements will inevitably be found.


Dragorn421:
Not sure where to start reviewing but I noticed `FLT_MAX` -> `MAXFLOAT`, the former being C-standard (I think), the latter coming from where?

Thar0:
In the three MIPS cross compilers I have installed (`mips-linux-gnu-`, well-known, `mips64-ultra-elf-`, glank's toolchain, `mips64-elf-`, libdragon toolchain) I cannot find the definition `FLT_MAX`, only `MAXFLOAT` in `math.h`.

Supposedly `FLT_MAX` is in `float.h`, an ISO C99 header, however I can't find a physical copy of such a header (I'd love to know where it is)
It seems `MAXFLOAT` is not strictly ISO C, it's an "extension", but it seems sufficiently widely adopted?

src/libultra/gu/coss.c
```diff
@@ -1,4 +1,4 @@
-#include "global.h"
+#include "guint.h"
```
Dragorn421, line 1:
```suggestion
#include "gu.h"
```
Note I don't know what you're going for with includes

I assume they should be minimal?
But for example would a
```c
#include "guint.h"
```
->
```c
#include "ultratypes.h"
#include "gu.h"
```
be better if something didn't need what guint.h itself contains?

src/libultra/io/dpsetstat.c
```diff
@@ -1,4 +1,5 @@
-#include "global.h"
+#include "rcp.h"
+#include "rdp.h"
```
Dragorn421, line 2:
```suggestion
```
?
and dpgetstat

Thar0: `rdp.h` contains the prototype for `osDpSetStatus`, the header containing the prototype needs to be included so the compiler can warn/error if the prototype does not match the function itself

Makefile
```diff
@@ -98,7 +98,8 @@ OBJDUMP    := $(MIPS_BINUTILS_PREFIX)objdump
 EMULATOR = mupen64plus
 EMU_FLAGS = --noosd
 
-INC        := -Iinclude -Isrc -Ibuild -I.
+INC        := -Iinclude -Iinclude/libc -Isrc -Ibuild -I.
+build/src/libultra/%.o: INC := -Iinclude/libc -Iinclude/ultra64 -Isrc/libultra
```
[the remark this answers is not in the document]
Thar0: I think it's just more correct than using `ultra64/...` paths in libultra files, and libultra files ideally shouldn't see any headers from the base include dir (although unfortunately as I mentioned I still couldn't get rid of two headers yet)
It may be better if we moved `src/libultra` to say `lib/libultra` with it's own makefile or something, but that's a bit too far of a departure from what I wanted to achieve for now (but I think doing this later may be good)

src/libultra/gu/coss.c
Thar0: I've mostly been using the `*int` headers for each libultra module, since I believe that's how it was originally based on mdebug includes, but maybe a more minimal approach is just better idk. Though note that as I mention below the minimal set of includes must also contain the prototype for the functions in the file, not just the necessary external declarations that the function(s) uses.

include/ultra64/viint.h
```diff
@@ -34,4 +37,12 @@
 #define VINTR(v) v
 #define HSTART START
```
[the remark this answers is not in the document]
Thar0: I'm hesitant to do this since we know what should be in `vi.h` (`os_vi.h`), but we don't have a reference `viint.h` so we can put just about whatever we like in there.

Makefile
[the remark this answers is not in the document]
Dragorn421: Yeah I understand the appeal when looking at libultra specifically/separately and I agree libultra should be completely cut off from the rest, my beef is mostly/only that it's confusing when navigating the repo

It would be nice to document this somewhere other than "self-documenting" in the Makefile. I wouldn't be opposed to a `README` file in src/libultra and include/libultra but idk if that would be a popular suggestion

include/ultra64/viint.h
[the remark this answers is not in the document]
Dragorn421: Forgetting about the original headers, I think it makes more sense for them to be with the `OSViMode` definition,
but I'll leave it up to you and the "follow the original headers" stance

Makefile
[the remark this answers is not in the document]
Dragorn421: Also I'm starting with criticism but I should also congratulate you on a job well done, it's nice to see header cleanup and this looks quite involved FeelsOKMan

Dragorn421:
Very interesting last set of changes!
What is the future you envision for explicit-width "ultratypes" s32 & co ? What should use int & co "natives" instead, or not? (does that make sense)

Thar0:
So far I've only been removing the ultratypes from libc files/functions, since libc shouldn't know anything about libultra. Otherwise I have no current plans to start removing them elsewhere

include/ultra64/cache.h
```diff
@@ -0,0 +1,14 @@
+#ifndef ULTRA64_CACHE_H
+#define ULTRA64_CACHE_H
+
+#include "ultratypes.h"
+
+#define	OS_DCACHE_ROUNDUP_ADDR(x) (void*)(((((u32)(x) + 0xf) / 0x10) * 0x10))
+#define	OS_DCACHE_ROUNDUP_SIZE(x)   (u32)(((((u32)(x) + 0xf) / 0x10) * 0x10))
```
hensldm, line 7:
```suggestion
#define	OS_DCACHE_ROUNDUP_ADDR(x) (void*)(((((u32)(x) + 0xF) / 0x10) * 0x10))
#define	OS_DCACHE_ROUNDUP_SIZE(x)   (u32)(((((u32)(x) + 0xF) / 0x10) * 0x10))
```

include/ultra64/ai.h
```diff
@@ -0,0 +1,10 @@
+#ifndef ULTRA64_AI_H
+#define ULTRA64_AI_H
+
+#include "ultratypes.h"
+
+u32 osAiGetLength(void);
+s32 osAiSetFrequency(u32);
+s32 osAiSetNextBuffer(void*, u32);
```
hensldm, line 8: Curious was it deliberate to leave out the arg names here, and just use types?

include/ultra64/gu.h
```diff
@@ -1,8 +1,51 @@
...
 #define ULTRA64_GU_H
 
+#include "ultratypes.h"
+#include "mbi.h"
+
 #define GU_PI 3.1415926
 
 #define ROUND(x) (s32)(((x) >= 0.0) ? ((x) + 0.5) : ((x) - 0.5))
 
+void guMtxIdent(Mtx*);
```
hensldm, line 11: Relating to the arg question again, some functions here have arg names and some don't. Any particular reason why?

include/libc/assert.h
```diff
@@ -0,0 +1,41 @@
...
+#  define assert(cond) ASSERT(cond, ., ., .)
+# endif
+#else
+# define ASSERT(cond, msg, file, line) ((void)0)
+# define assert(cond) ((void)0)
+#endif
+
+// Static/compile-time assertions
+
+#if defined(__GNUC__) || (__STDC_VERSION__ >= 201112L)
```
Dragorn421, line 28: Wouldn't just
```suggestion
#if (__STDC_VERSION__ >= 201112L)
```
be enough / make more sense? (unless gcc did have the macro way before C11)

include/is_debug.h
```diff
@@ -0,0 +1,12 @@
+#ifndef ISVIEWER_H
+#define ISVIEWER_H
+
+#include "ultra64.h"
```
Dragorn421, line 4: Should this include be in the .c only? It's not needed by the .h, I think

src/libultra/os/exceptasm.s
```diff
@@ -308,7 +308,7 @@ cart:
     // Load cart callback set by __osSetHWIntrRoutine
     lui     $t1, %hi(__osHwIntTable)
     addiu   $t1, %lo(__osHwIntTable)
-    lw      $t2, (OS_INTR_CART*HWINT_SIZE+HWINT_CALLBACK)($t1)
+    lw      $t2, (OS_INTR_CART*HWINT_SIZE+HWINT_HANDLER)($t1)
```
Dragorn421, line 311: since `HWINT_CALLBACK` -> `HWINT_HANDLER` , should references to "callback" in sentences be changed too?

src/libultra/os/afterprenmi.c
```diff
@@ -1,5 +1,4 @@
-#include "ultra64.h"
-#include "global.h"
+#include "internal.h"
 
 s32 osAfterPreNMI(void) {
```
Dragorn421, line 3: So per the comment about includes https://github.com/zeldaret/oot/pull/1468#discussion_r1052701432
should this include `system.h` ? (with the `osAfterPreNMI` prototype)

The same kind of comment seems to apply to other files

include/ultra64/libc.h
```diff
@@ -0,0 +1,12 @@
+#ifndef ULTRA64_LIBC_H
+#define ULTRA64_LIBC_H
+
+#include "stddef.h"
+
+void osSyncPrintf(const char* fmt, ...);
+
+void bzero(void* __s, size_t __n);
+int bcmp(const void* __sl, const void* __s2, size_t __n);
+void bcopy(const void* __src, void* __dest, size_t __n);
```
AngheloAlf, line 10: The original libultra headers use `int` instead of `size_t`. Why deviate from the original headers here?
Also, why the `__` prefix on the arguments?

ariahiro64:
idk what this entails but theres also the "modern n64 sdk" which includes the entirety of libultra. the legality is murky at best.

include/ultra64/libc.h
cadmic: IIRC `int` is required for matching anyway. Currently in HEAD it's

```
#ifdef __GNUC__
void bzero(void* __s, unsigned int __n);
int bcmp(const void* __sl, const void* __s2, unsigned int __n);
void bcopy(const void* __src, void* __dest, unsigned int __n);
#else
void bzero(void* __s, int __n);
int bcmp(const void* __sl, const void* __s2, int __n);
void bcopy(const void* __src, void* __dest, int __n);
#endif
```

Dragorn421:
We'll eventually need at least part of this PR revived for header cleanup.

The goals with libultra should be clarified: align with ultralib? if not, which includes in .c files ([IWYU](https://github.com/include-what-you-use/include-what-you-use/blob/master/README.md) ?) ?

PR #1471, opened 2022-12-19
Fbdemo Docs: `TransitionTile`

hensldm:
Decided to call the system `TransitionTile` as it converts the current screen to a tiled version of it (of course always open to suggestions :) ).

Static Version: https://discord.com/channels/688807550715560050/718842910749229066/1053350561849811015
Dynamic Version: https://discord.com/channels/688807550715560050/718616443834335293/1053171306939568128


include/z64transition.h
```diff
@@ -4,24 +4,31 @@
...
+    /* 3 */ TRANS_TILE_PROCESS
+} TransitionTileStatus;
+
 typedef struct {
-    f32 unk_0;
-    f32 unk_4;
-} TransitionUnkData;
+    /* 0x0 */ f32 x;
+    /* 0x4 */ f32 y;
+} TransitionTileVtxData;
```
engineer124, line 17:
```suggestion
    /* 0x0 */ f32 x;
    /* 0x4 */ f32 y;
} TransitionTileVtxData; // size = 0x8
```

include/functions.h
```diff
@@ -1208,13 +1208,11 @@ void func_800AD920(struct_80166500* this);
...
-void TransitionUnk_InitGraphics(TransitionUnk* this);
-void TransitionUnk_InitData(TransitionUnk* this);
-void TransitionUnk_Destroy(TransitionUnk* this);
-TransitionUnk* TransitionUnk_Init(TransitionUnk* this, s32 row, s32 col);
-void TransitionUnk_SetData(TransitionUnk* this);
-void TransitionUnk_Draw(TransitionUnk* this, Gfx**);
-void func_800B23E8(TransitionUnk* this);
+void TransitionTile_Destroy(TransitionTile* this);
+TransitionTile* TransitionTile_Init(TransitionTile* this, s32 rows, s32 cols);
+void TransitionTile_Draw(TransitionTile* this, Gfx**);
```
engineer124, line 1213:
```suggestion
void TransitionTile_Draw(TransitionTile* this, Gfx** gfxP);
```

src/code/z_fbdemo.c
```diff
@@ -1,9 +1,25 @@
+/**
+ * @file z_fbdemo.c
+ *
+ * This file implements an unused transition system that takes the current screen and converts it to large tiles.
+ *
+ * The static version will just display this tiled screen, while the dynamic version has the visual effect of `sucking`
+ * it tiled screen into a single point.
```
engineer124, line 7:
```suggestion
 * The static version will just display this tiled screen, while the dynamic version has the visual effect of `sucking`
 * the tiled screen into a single point.
```

src/code/z_play.c
```diff
@@ -489,20 +489,21 @@ void Play_Update(PlayState* this) {
...
+                        gTransitionTileState = TRANS_TILE_PROCESS;
                         R_UPDATE_RATE = 1;
                     }
                     break;
-                case 3:
-                    func_800B23E8(&sTrnsnUnk);
+
+                case TRANS_TILE_PROCESS:
+                    TransitionTile_UpdateStatic(&sTransitionTile);
                     break;
```
engineer124, line 507:
```suggestion
                case TRANS_TILE_PROCESS:
                    TransitionTile_UpdateStatic(&sTransitionTile);
                    break;
                
                 default:
                     break;
```
assuming it matches

include/z64transition.h
```diff
@@ -4,24 +4,31 @@
 #include "ultra64.h"
 #include "color.h"
 
+typedef enum {
+    /* 0 */ TRANS_TILE_OFF,
+    /* 1 */ TRANS_TILE_SETUP,
+    /* 2 */ TRANS_TILE_INIT,
+    /* 3 */ TRANS_TILE_PROCESS
+} TransitionTileStatus;
```
fig02, line 12: this heavily mirrors the names given to the pause background state given in https://github.com/zeldaret/oot/pull/1368/files
which notably is still under debate. so ideally we come to an agreement there before we start introducing things like it. we need to just get that conversation started again and pull it back out of limbo tho

Dragorn421: please do despair

src/code/z_fbdemo.c
```diff
@@ -1,9 +1,25 @@
+/**
+ * @file z_fbdemo.c
+ *
+ * This file implements an unused transition system that takes the current screen and converts it to large tiles.
+ *
+ * The static version will just display this tiled screen, while the dynamic version has the visual effect of `sucking`
+ * the tiled screen into a single point.
```
fig02, line 7: "sucked" is in backticks like its referencing a variable or something, but it isnt right?
can maybe just say "visual effect of being sucked into a single point", or something

src/code/z_fbdemo.c
```diff
@@ -60,59 +75,67 @@ void TransitionUnk_InitGraphics(TransitionUnk* this) {
...
         }
+
+        colTex += 0x20;
     }
 
     gDPPipeSync(gfx++);
     gSPEndDisplayList(gfx++);
 
-    LOG_NUM("this->col * (1 + this->row * (1 + 7 + 1)) + 1 + 1", this->col * (1 + this->row * 9) + 2, "../z_fbdemo.c",
+    LOG_NUM("this->col * (1 + this->row * (1 + 7 + 1)) + 1 + 1", this->cols * (1 + this->rows * 9) + 2, "../z_fbdemo.c",
```
Dragorn421, line 114: not sure what the original "col" and "row" meant, but it's not "amount of columns" and "amount of rows"

unless they had vertical rows and horizontal columns

include/z64transition.h
fig02: no, I thought they were the same until you pointed that out just now lol

src/code/z_fbdemo.c
```diff
@@ -45,11 +58,13 @@ void TransitionUnk_InitGraphics(TransitionUnk* this) {
...
         vtx = (this->frame == 0) ? this->vtxFrame1 : this->vtxFrame2;
-        for (colTex = 0, col = 0; col < this->col + 1; colTex += 0x20, col++) {
-            for (rowTex = 0, row = 0; row < this->row + 1; row++) {
-                vtx2 = &vtx->v;
-                vtx++;
+        colTex = 0;
+        for (col = 0; col < this->cols + 1; col++) {
+            rowTex = 0;
+            for (row = 0; row < this->rows + 1; row++) {
+                Vtx_t* vtx2 = &vtx->v;
```
Thar0, line 65: While you're here, this should use `Vtx_tn` since these vertices end up drawn using the settings in `sTransTileSetupDL`. which sets `G_LIGHTING` in the geometrymode

src/code/z_fbdemo.c
hensldm: Yeah, they do seem reversed. Lowering the numbers below 5 and 4 (not 100% sure why that is the limit since the screen seems divided by 10 and 7), I can control the number of rows and cols that get drawn. For instance with `cols=2` and `rows=3`.
image

hensldm:
Small note, I think it was better to rename `UpdateStatic` to just `Update` and `UpdateDynamic` to `Suck` which could presumably just be called from `Update` if it was to be used (inspiration from Dragorn's branch that messes with this system which just calls `UpdateDynamic` from within `UpdateStatic`).

fig02 (approved):

src/code/z_fbdemo.c
Dragorn421: Consider using angrylion, a very accurate low level emulation video plugin, https://github.com/ata4/angrylion-rdp-plus , or the parallel video plugin (which idk how to get on pj64, it's typically bundled with other emulators though like ares or simple64),
instead of high level emulation video plugins, even if GlideN64 is very good for a HLE video plugin it will often struggle or act odd with weird stuff like this

> Lowering the numbers below 5 and 4 (not 100% sure why that is the limit since the screen seems divided by 10 and 7)

limit? what do you mean? (I didn't try lowering these myself)

src/code/z_play.c
```diff
@@ -186,9 +186,9 @@ void Play_Destroy(GameState* thisx) {
     EffectSs_ClearAll(this);
     CollisionCheck_DestroyContext(this, &this->colChkCtx);
 
-    if (gTrnsnUnkState == 3) {
-        TransitionUnk_Destroy(&sTrnsnUnk);
-        gTrnsnUnkState = 0;
+    if (gTransitionTileState == TRANS_TILE_PROCESS) {
+        TransitionTile_Destroy(&sTransitionTile);
+        gTransitionTileState = TRANS_TILE_OFF;
     }
```
engineer124, line 192:
```suggestion
    if (gTransitionTileState == TRANS_TILE_DONE) {
        TransitionTile_Destroy(&sTransitionTile);
        gTransitionTileState = TRANS_TILE_OFF;
    }
```
I think there was a comment to sync this more with:
```c
typedef enum {
    /* 0 */ PAUSE_BG_PRERENDER_OFF, // Inactive, do nothing.
    /* 1 */ PAUSE_BG_PRERENDER_SETUP, // The current frame is only drawn for the purpose of serving as the pause background.
    /* 2 */ PAUSE_BG_PRERENDER_PROCESS, // The previous frame was PAUSE_BG_PRERENDER_SETUP, now apply prerender filters.
    /* 3 */ PAUSE_BG_PRERENDER_DONE, // The pause background is ready to be used.
    /* 4 */ PAUSE_BG_PRERENDER_MAX
} PauseBgPreRenderState;
```
Is there part of this that doesn't line up with this?

fig02: (my comment didnt secify that it *should* be synced with pause bg, I thought it already *was*)
But, if it can be synced, then sure

src/code/z_fbdemo.c
hensldm: Okay, using simple64 it acts completely different (I should have known using PJ64 was going to mess with me someday). It more blends things to a point (if you use the dynamic version), where the cols and rows tell which part of the screen to blend. Static doesn't seem to do anything.

src/code/z_play.c
hensldm: I could probably sync them up, the main hesitation for me is that for the transition state 3, it continually calls update so doesn't seem `done` to me. However, I do see that the same thing happen for `PAUSE_BG_PRERENDER` where 3 continually calls `PreRender_RestoreFramebuffer`.

src/code/z_fbdemo.c
hensldm: With this information, perhaps `TransitionTile` isn't best, but maybe `TransitionBlend`.

hensldm: For reference this is with `cols=3` and `rows=5`
image

src/code/z_play.c
hensldm: Question. Thoughts on changing `Done` to `Ready`? indicating the prerender or transition are ready to be used?

Dragorn421: I like "ready"

src/code/z_fbdemo.c
Dragorn421: > Okay, using simple64 it acts completely different (I should have known using PJ64 was going to mess with me someday).

:sweat_smile: 

> It more blends things to a point (if you use the dynamic version), where the cols and rows tell which part of the screen to blend.

yeah

> Static doesn't seem to do anything.

visually nothing happens yeah, it just draws the tiles "where they belong" so there's nothing to see. 

> With this information, perhaps `TransitionTile` isn't best, but maybe `TransitionBlend`?

TransitionTile works for me since the main feature is splitting the saved framebuffer into "rectangles" (so "tiles") that can then be drawn at arbitrary locations

---------

Try the following patch (bottom of this message), I think it should help a lot clearing up what's happening here

(to apply a patch, run `git apply`, copypaste to console, then hit Ctrl+D, or save the patch to a file and `git apply path/to/patch.txt`)

This adds the following controls on controller 1:
- L to turn on/off TransitionTile
- while TransitionTile is on:
  - Z to clear the framebuffer using `sTransTileBackgroundDL` (as long as Z is held)
  - dpad to reset TransitionTile with different cols/rows values (e.g. dpad left decreases the amount of cols, "shrinking" horizontally)

The effect achieved is drawing the saved framebuffer, scaled down and moving across the screen.

Testing suggestion:

1. Load file 1 (for some reason using this on title screen is jank)
2. Press L
   Screenshot after half a second:
   <details>

   image

   </details>
3. Observe the "trail" left by the moving image. This is because the framebuffer is not cleared
4. Hold Z. This clears the framebuffer every frame with `sTransTileBackgroundDL`
   Screenshot:
   <details>

   image

   </details>
6. Let go of Z, if you want to
7. Press dpad down. you see garbage drawn, this is data past the valid framebuffer data saved in the zbuffer (note dpad right has code to prevent having more than 10 cols, since for a reason I am unaware of that crashes the game)
   Screenshot:
   <details>

   image

   </details>
9. press dpad up/down/left/right to adjust the amount of cols/rows
   Screenshot:
   <details>

   image

   </details>

Note: this isn't crash-free and I didn't bother to debug why

<details>

<summary>

**>>> PATCH <<<**

</summary>

```diff
diff --git a/include/functions.h b/include/functions.h
index e59a2a51a..6cf9521f1 100644
--- a/include/functions.h
+++ b/include/functions.h
@@ -1210,7 +1210,7 @@ void func_800AD958(struct_80166500* this, Gfx** gfxp);
 void PlayerCall_InitFuncPtrs(void);
 void TransitionTile_Destroy(TransitionTile* this);
 TransitionTile* TransitionTile_Init(TransitionTile* this, s32 cols, s32 rows);
-void TransitionTile_Draw(TransitionTile* this, Gfx** gfxP);
+void TransitionTile_Draw(TransitionTile* this, Gfx** gfxP, PlayState* play);
 void TransitionTile_Update(TransitionTile* this);
 void TransitionTriforce_Start(void* thisx);
 void* TransitionTriforce_Init(void* thisx);
diff --git a/src/code/z_fbdemo.c b/src/code/z_fbdemo.c
index cf6f6a988..a1fcb67bc 100644
--- a/src/code/z_fbdemo.c
+++ b/src/code/z_fbdemo.c
@@ -204,10 +204,12 @@ void TransitionTile_SetVtx(TransitionTile* this) {
     }
 }
 
-void TransitionTile_Draw(TransitionTile* this, Gfx** gfxP) {
+void TransitionTile_Draw(TransitionTile* this, Gfx** gfxP, PlayState* play) {
     Gfx* gfx = *gfxP;
 
-    gSPDisplayList(gfx++, sTransTileSetupDL);
+    if (play->state.input[0].cur.button & BTN_Z) {
+        gSPDisplayList(gfx++, sTransTileBackgroundDL);
+    }
     TransitionTile_SetVtx(this);
     gSPMatrix(gfx++, &this->projection, G_MTX_NOPUSH | G_MTX_LOAD | G_MTX_PROJECTION);
     gSPMatrix(gfx++, &this->modelView, G_MTX_NOPUSH | G_MTX_LOAD | G_MTX_MODELVIEW);
@@ -246,7 +248,36 @@ void TransitionTile_Suck(TransitionTile* this) {
     }
 }
 
+f32 fmodf(f32 x, f32 y);
+
+void TransitionTile_Test(TransitionTile* this) {
+    s32 row;
+    s32 col;
+    f32 diffX;
+    f32 diffY;
+    f32 scale;
+    static s32 timer = 0;
+    f32 tileWidth = 20;
+    f32 tileHeight = 20;
+    f32 dx, dy;
+
+    timer++;
+    dx = fmodf(timer, SCREEN_WIDTH - this->cols * tileWidth);
+    dy = fmodf(timer, SCREEN_HEIGHT - this->rows * tileHeight);
+
+    for (row = 0; row < this->rows + 1; row++) {
+        for (col = 0; col < this->cols + 1; col++) {
+            f32 x, y;
+            x = col * tileWidth + dx;
+            y = row * tileHeight + dy;
+            (this->vtxData + col + row * (this->cols + 1))->x = CLAMP(x, 0, SCREEN_WIDTH);
+            (this->vtxData + col + row * (this->cols + 1))->y = CLAMP(y, 0, SCREEN_HEIGHT);
+        }
+    }
+}
+
 void TransitionTile_Update(TransitionTile* this) {
+    TransitionTile_Test(this);
 }
 
 s32 func_800B23F0(TransitionTile* this) {
diff --git a/src/code/z_play.c b/src/code/z_play.c
index f2640886c..51daf445e 100644
--- a/src/code/z_play.c
+++ b/src/code/z_play.c
@@ -455,6 +455,15 @@ void Play_Update(PlayState* this) {
 
     input = this->state.input;
 
+    if (input->press.button & BTN_L) {
+        if (gTransitionTileState == TRANS_TILE_OFF) {
+            gTransitionTileState = TRANS_TILE_SETUP;
+        } else if (gTransitionTileState == TRANS_TILE_DONE) {
+            TransitionTile_Destroy(&sTransitionTile);
+            gTransitionTileState = TRANS_TILE_OFF;
+        }
+    }
+
     if ((SREG(1) < 0) || (DREG(0) != 0)) {
         SREG(1) = 0;
         ZeldaArena_Display();
@@ -493,9 +502,36 @@ void Play_Update(PlayState* this) {
         }
 
         if (gTransitionTileState != TRANS_TILE_OFF) {
+            static s32 cur_cols = 10, cur_rows = 7;
+            s32 cols = cur_cols, rows = cur_rows;
+            s32 refresh = false;
+            u16 pressBtn = this->state.input[0].press.button;
+
+            if (pressBtn & BTN_DLEFT) {
+                cols--;
+            }
+            if (pressBtn & BTN_DRIGHT) {
+                cols++;
+            }
+            if (pressBtn & BTN_DUP) {
+                rows--;
+            }
+            if (pressBtn & BTN_DDOWN) {
+                rows++;
+            }
+            cols = CLAMP_MIN(cols, 0);
+            cols = CLAMP_MAX(cols, SCREEN_WIDTH / 32);
+            rows = CLAMP_MIN(rows, 0);
+
+            refresh = cols != cur_cols || rows != cur_rows;
+            if (refresh) {
+                cur_cols = cols;
+                cur_rows = rows;
+            }
+
             switch (gTransitionTileState) {
                 case TRANS_TILE_PROCESS:
-                    if (TransitionTile_Init(&sTransitionTile, 10, 7) == NULL) {
+                    if (TransitionTile_Init(&sTransitionTile, cur_cols, cur_rows) == NULL) {
                         osSyncPrintf("fbdemo_init呼出し失敗！\n"); // "fbdemo_init call failed!"
                         gTransitionTileState = TRANS_TILE_OFF;
                     } else {
@@ -506,7 +542,12 @@ void Play_Update(PlayState* this) {
                     break;
 
                 case TRANS_TILE_DONE:
-                    TransitionTile_Update(&sTransitionTile);
+                    if (refresh) {
+                        TransitionTile_Destroy(&sTransitionTile);
+                        gTransitionTileState = TRANS_TILE_SETUP;
+                    } else {
+                        TransitionTile_Update(&sTransitionTile);
+                    }
                     break;
 
                 default:
@@ -1116,7 +1157,7 @@ void Play_Draw(PlayState* this) {
         if (gTransitionTileState == TRANS_TILE_DONE) {
             Gfx* sp88 = POLY_OPA_DISP;
 
-            TransitionTile_Draw(&sTransitionTile, &sp88);
+            TransitionTile_Draw(&sTransitionTile, &sp88, this);
             POLY_OPA_DISP = sp88;
             goto Play_Draw_DrawOverlayElements;
         } else {
```

</details>

src/code/z_fbdemo.c
```diff
@@ -159,72 +180,75 @@ TransitionUnk* TransitionUnk_Init(TransitionUnk* this, s32 row, s32 col) {
...
 }
 
-void TransitionUnk_Update(TransitionUnk* this) {
-    f32 temp_f00;
-    f32 temp_f12;
-    s32 col;
-    f32 phi_f14;
+/**
+ * Blurs tiles which has the visual effect of sucking those tiles into a single point
+*/
```
Dragorn421, line 225: (this file needs formatting)

src/code/z_fbdemo.c
hensldm: > TransitionTile works for me since the main feature is splitting the saved framebuffer into "rectangles" (so "tiles") that can then be drawn at arbitrary locations

Makes sense.

src/code/z_fbdemo.c
```diff
@@ -45,74 +57,84 @@ void TransitionUnk_InitGraphics(TransitionUnk* this) {
...
         this->frame = frame;
         vtx = (this->frame == 0) ? this->vtxFrame1 : this->vtxFrame2;
-        for (colTex = 0, col = 0; col < this->col + 1; colTex += 0x20, col++) {
-            for (rowTex = 0, row = 0; row < this->row + 1; row++) {
-                vtx2 = &vtx->v;
+        rowTex = 0;
+        for (row = 0; row < this->rows + 1; row++) {
+            colTex = 0;
+            for (col = 0; col < this->cols + 1; col++) {
+                Vtx_tn* vtxn = &vtx->n;
```
Dragorn421, line 64: ( @Thar0 ) in `sTransTileSetupDL`, `gsSPClearGeometryMode(... | G_SHADE | ... | G_LIGHTING` uses "clear" so vertices are actually not lit, and also not shaded at all I guess (`G_SHADE` cleared too)
so idk if Vtx_t / Vtx_tn is more appropriate here

include/z64transition.h
```diff
@@ -4,24 +4,31 @@
 #include "ultra64.h"
 #include "color.h"
 
+typedef enum {
+    /* 0 */ TRANS_TILE_OFF, // Inactive, do nothing
+    /* 1 */ TRANS_TILE_SETUP, // Save the necessary buffers
+    /* 2 */ TRANS_TILE_PROCESS, // Initalize the transition using the saved buffers
```
Dragorn421, line 10:
```suggestion
    /* 2 */ TRANS_TILE_PROCESS, // Initialize the transition
```
the framebuffer isn't used at this point, it's only referenced in the built dlist, and that dlist is executed under `TRANS_TILE_READY`

src/code/z_fbdemo.c
Thar0: oh yeah I can't read, I was not expecting it to be this weird I guess, in this case it doesn't matter ultimately.
I think I still prefer Vtx_tn though,
```
vtxn->n[0] = 0;
vtxn->n[1] = 0;
vtxn->n[2] = 120;
```
This makes slightly more sense as a normal I think (although notably it isn't normalized 🤷)

Dragorn421 (approved):

Thar0 (approved):

PR #1472, opened 2022-12-19
Document debug text drawing functions in z_debug.c

fig02:
I have some notes on my reasoning for different names, and some stuff to keep in mind, which I will leave in review comments below


src/code/z_debug.c
```diff
@@ -106,60 +108,59 @@ void Regs_Init(void) {
     }
 }
 
-// Called when free movement is active.
-void func_8006375C(s32 arg0, s32 arg1, const char* text) {
+// Function is stubbed, name is assumed by `Debug_SetScreenTextColored`, arguments, and usage
+void Debug_SetScreenText(u8 x, u8 y, const char* text) {
 }
```
fig02, line 113: This function is stubbed, so this name is entirely speculation.
But, I think its a fair assumption considering the function right below it, the args supplied to it which look like screen coordinates, and its one usage in z_player. If you change it so that it actually draws, it looks like this in noclip mod
debugmode
e:

src/code/z_debug.c
```diff
@@ -282,14 +281,14 @@ void func_80063D7C(GraphicsContext* gfxCtx) {
     GfxPrint_Open(&printer, gfx);
 
     if ((OREG(0) == 1) || (OREG(0) == 8)) {
-        func_80063828(&printer);
+        Debug_DrawOnScreenText(&printer);
```
fig02, line 284: Important to note that this text will only draw if these O(gawa) regs are set
This is only used in camera, debug cam, and that one call for noclip in player. Its prob safe to say that the camera dev implemented noclip

That said, I opted to name it more generally and not after camera. Wasnt totally sure what to do, but I think general is slightly preferable when its only tie to camera is the regs that allow it to draw

include/z64.h
```diff
@@ -1774,4 +1774,15 @@ typedef struct {
...
+typedef enum {
+    /* 0 */ DEBUG_TEXT_YELLOW,
+    /* 1 */ DEBUG_TEXT_PEACH,
+    /* 2 */ DEBUG_TEXT_BROWN,
+    /* 3 */ DEBUG_TEXT_ORANGE,
+    /* 4 */ DEBUG_TEXT_GOLD,
+    /* 5 */ DEBUG_TEXT_WHITE,
+    /* 6 */ DEBUG_TEXT_BLUE,
+    /* 7 */ DEBUG_TEXT_GREEN
+} DebugTextColor;
```
fig02, line 1786: I could not think of a better place to put this. If there are any suggestions that dont involve making a new header right now, feel free

also, the colors in game:
colors

src/code/z_debug.c
```diff
@@ -106,60 +108,59 @@ void Regs_Init(void) {
...
-// Called when free movement is active.
-void func_8006375C(s32 arg0, s32 arg1, const char* text) {
+// Function is stubbed. Name is assumed by similarities in signature to `Debug_SetScreenTextColored` and usage.
+void Debug_SetScreenText(u8 x, u8 y, const char* text) {
 }
 
-// Store text during Update, to be drawn later during Draw
-void func_8006376C(u8 x, u8 y, u8 colorIndex, const char* text) {
-    PrintTextBufferEntry* entry;
+void Debug_SetScreenTextColored(u8 x, u8 y, u8 colorIndex, const char* text) {
```
Dragorn421, line 115: I think "Set" can be omitted, the operation is more a "append" anyway, not that the users of this function care that much
`Debug_ScreenText(Colored) (...)` seems fine

src/code/z_debug.c
```diff
@@ -14,17 +14,19 @@ typedef struct {
 
 RegEditor* gRegEditor;
 
-PrintTextBufferEntry sDebugPrintTextBuffer[22];
-s16 sDebugPrintTextBufferNumUsed = 0;
+PrintTextBufferEntry sDebugTextBuffer[22];
+
+s16 sDebugTextEntryCount = 0;
+
 Color_RGBA8 sDebugPrintTextColors[] = {
```
Dragorn421, line 21: "text" vs "print text" ?
"print text" would be -> master,
committing to "text" may be:
```suggestion
DebugTextBufferEntry sDebugTextBuffer[22];

s16 sDebugTextEntryCount = 0;

Color_RGBA8 sDebugTextColors[] = {
```

Dragorn421 (commented):
(new strat: PR reviewbait) I didn't know OoT had Peach

src/code/z_debug.c
fig02: yeah had this idea but it felt weird without a "verb"
but the verb cant be print or draw cause that happens somewhere else lol

but yeah ill just remove set

src/code/z_debug.c
fig02: yeah just missed these while removing masters "print"

fig02:
i think i got owned by someones jenkins agent but idk what the error is trying to say

Dragorn421 (approved):
smh didn't take the """opportunity""" to clean up db_camera.c strings morphon

src/code/z_debug.c
```diff
@@ -267,7 +266,7 @@ void Regs_DrawEditor(GfxPrint* printer) {
     }
 }
 
-void func_80063D7C(GraphicsContext* gfxCtx) {
+void Debug_Draw(GraphicsContext* gfxCtx) {
```
engineer124, line 269: Despite this being in `z_debug.c` and called directly from `GameState_Draw`, `Debug_Draw` feels like a bit too generic of a name? It's primarily for `Debug_DrawOnScreenText` and `Regs_DrawEditor` right? Is `DebugText_Draw` or something else able to capture both of these? I'm not sure if you've had any other ideas. Might be fine as is though

fig02: yea any more specific name I can think of makes it sound like its only for the printed screen text and not the reg editor. `DebugText_Draw` is an ok direction, but introducing a new system name for the one function is probably not ideal.
But at the same time, the two things it handles (camera screen text and reg editor) are kind of separate. so idk 

at the very least debug isnt wrong, it is two pretty seperate debugging features. but can understand feeling its generic.

I think in an ideal world other debug text in the game used this system and it would make more sense, but that is not how it played out in practice

src/code/z_debug.c
engineer124: This system does seem designed to be more generalized with one exception: these `OREG`s are annoying because if an external system changes `OREG(0)` to draw the screen text, then `Camera_Init` on the main camera (i.e. upon `Play_Init`) will overwrite `OREG(0)` based on `sOREGInit`:

```c
for (i = 0; i < sOREGInitCnt; i++) {
    OREG(i) = sOREGInit[i];
}
```

src/code/z_debug.c
engineer124: Yeah, the regs and text are different enough that it's hard to group them. Could make the other `Debug_` prefixes into `DebugText_` prefixes so it's not just one function, but then you'd have `Text` twice in those other functions. So `Debug_Draw` is probably fine, can't really think of a better alternative

src/code/z_debug.c
fig02: yea
my main reasoning for keeping the name general is because the text functions themselves arent really tied in any way. only the call to draw it in the `Debug_Draw` function (which we are still discussing

src/code/z_debug.c
Dragorn421: I think for now "this file is z_debug" is enough to name this as a "Debug" system.
We can change it later if needed, when more "debug only" stuff is named (not that there is a lot of it, is my impression), or if a better idea arises

In the meantime having this named at all is :+1: for me

src/code/z_debug.c
Dragorn421: Kinda brainstorming but what about naming this "screen debug text" system after the O in OREG (or non-canonically O for Ogawa)
like:
- `Debug_ScreenTextColored` -> `Debug_OTextColored`
- `Debug_DrawOnScreenText` -> `Debug_DrawOText`
- `DebugTextBufferEntry sDebugTextBuffer` -> `DebugOTextBufferEntry sDebugOTextBuffer`

:sunglasses:

include/z64.h
mzxrules: Not the greatest color options. Feel like `peach` -> `red`, and although gold, orange are ok names, `gold` looks closer to orange and `orange` looks closer to mud.

Dragorn421: From https://rgbcolorcode.com/ :
laser lemon
light salmon
brown (traditional)
dark goldenrod
saffron
platinum
cornflower blue
chartreuse (web)

fig02: I dont really agree on red, its not red enough 

the way orange/gold are used in debug camera, its often switched between with a ternary, implying that the darker "orange" is like an "off" version of the same color. But this is entirely speculation from just pasting the enum values in. Personally I think its fine like this

src/code/z_debug.c
fig02: yea I still think I want to not tie it to Oreg for now, since nothing about the functions are actually tied to it, just how its called in debug_draw
But as you mention in the comment down below, we may have a better naming scheme for these debug related things overall later. Opinion may change then

src/code/z_debug.c
Thar0: `Debug_DrawAllText` maybe?

fig02: possibly. it does require that you know that this "system" only handles regs and ogawa's text lol
as opposed to the other systems which draw their own debug text with gfxprint

but I guess that context is needed no matter what we name this thing

fig02:
with some time to think, and through our discussions, personally feel like its more appropriate to make this more tied to camera/debug camera than my original names set it out to be.

for now the system name `DbCamera_` is used to sync with db_cam.c, but in general would like to see `DebugCam_` used instead for the debug cam in general

src/code/z_debug.c
```diff
@@ -14,17 +14,19 @@ typedef struct {
...
-Color_RGBA8 sDebugPrintTextColors[] = {
-    { 255, 255, 32, 192 },  // 0
-    { 255, 150, 128, 192 }, // 1
-    { 128, 96, 0, 64 },     // 2
-    { 192, 128, 16, 128 },  // 3
-    { 255, 192, 32, 128 },  // 4
-    { 230, 230, 220, 64 },  // 5
-    { 128, 150, 255, 128 }, // 6
-    { 128, 255, 32, 128 },  // 7
+DbCameraTextBufferEntry sDebugTextBuffer[22];
```
Dragorn421, line 17:
```suggestion
DbCameraTextBufferEntry sDbCameraTextBuffer[22];
```

src/code/z_debug.c
```diff
@@ -106,60 +108,59 @@ void Regs_Init(void) {
...
 
-        color = &sDebugPrintTextColors[entry->colorIndex];
         GfxPrint_SetColor(printer, color->r, color->g, color->b, color->a);
         GfxPrint_SetPos(printer, entry->x, entry->y);
         GfxPrint_Printf(printer, "%s", entry->text);
     }
 }
 
-// Process inputs to control the reg editor
 void Regs_UpdateEditor(Input* input) {
```
Dragorn421, line 156: idk I like that comment sadge
You could make it a `/** * */` doc comment though

Dragorn421 (approved):

include/z64.h
```diff
@@ -1774,4 +1774,15 @@ typedef struct {
...
+typedef enum {
+    /* 0 */ DBCAMERA_TEXT_YELLOW,
+    /* 1 */ DBCAMERA_TEXT_PEACH,
+    /* 2 */ DBCAMERA_TEXT_BROWN,
+    /* 3 */ DBCAMERA_TEXT_ORANGE,
+    /* 4 */ DBCAMERA_TEXT_GOLD,
+    /* 5 */ DBCAMERA_TEXT_WHITE,
+    /* 6 */ DBCAMERA_TEXT_BLUE,
+    /* 7 */ DBCAMERA_TEXT_GREEN
+} DbCameraTextColor;
```
engineer124, line 1786: This could probably go to the bottom of z64camera.h where the other DbCamera stuff currently resides?

src/code/z_debug.c
```diff
@@ -282,14 +288,14 @@ void func_80063D7C(GraphicsContext* gfxCtx) {
     GfxPrint_Open(&printer, gfx);
 
     if ((OREG(0) == 1) || (OREG(0) == 8)) {
-        func_80063828(&printer);
+        DbCamera_DrawScreenText(&printer);
     }
```
engineer124, line 292: Would be nice to name/doc `OREG(0)` here, but don't worry about it if you don't want to. Can always do it later

fig02: it has some meanings as 1 and 8 that I dont fully understand, and is sometimes treated as a flag and sometimes not.. it just kinda a mess and idk what to call it

engineer124: It's only used in two placed: here in `z_debug.c` and in `Camera_PrintSettings` which are used to print the status/settings/mode/func/bgCamIndex. So all it's doing is printing cam info to the screen (except for no-clip but w/e). So maybe just `R_DBG_CAM_PRINT_TEXT`?

Looking at values:
`OREG(0) = 0` -> doesn't print anything
`OREG(0) = 1` -> print both `DbCamera_DrawScreenText` and `Camera_PrintSettings`
`OREG(0) = 8` -> print only `DbCamera_DrawScreenText`
`OREG(0) = 1 | 8` -> print only `Camera_PrintSettings`
(not sure why they didn't just give each bit to each of the two types of printing instead of weird usages of = and bit checks)

fig02: im not sure just "print text" captures that face that it can switch between printing a few different things well
seems like something that can prob be named with another more OREG/debug cam related pass

engineer124: What do you mean by a few different things?

Yeah, we can name this in a different pass, just curious what you mean by that.

fig02: the different meanings for 1 and 8 you describe above. it can print different kinds of information on the screen, and I think the name should probably capture somehow

fig02: phrased another way, it is not a simple switch that you turn on and it will print all of the information. it has deeper meaning and lets you switch between what will get printed to the screen

engineer124 (approved):

src/code/z_debug.c
hensldm: Might I suggest `AddScreenText` or possibly `AddToScreenText`?

hensldm (approved):
Note I did add one comment/idea to the resolved conversation (https://github.com/zeldaret/oot/pull/1472#discussion_r1051979380). Though would be fine as it is.

src/code/z_debug.c
fig02: `AddScreenText` is okay. its accurate to what its actually doing, but I do kinda think that to someone not aware of the underlying workings, "add screen text" may be confusing. I see what "Add to" is going for, but think its probably even more confusing

I think just saying what it is, that its screen text, is probably fine. but more opinions welcome

Dragorn421:
I don't have an opinion on `Debug_ScreenText(Colored)` (current) vs `Debug_AddScreenText(Colored)` (suggested, see https://github.com/zeldaret/oot/pull/1472#discussion_r1057313106 )

If nothing else comes up and that doesn't progress I guess we'll just merge the PR as is eventually (in a few days to ~one week)

PR #1476, opened 2022-12-21
Name movement related functions in z_actor

fig02:
The diff is large because the function is called in many actors, but its all find and replace outside of z_actor.c.

This is mostly taken from MM's current documentation, as I find the function names for the main movement related functions are solid. I reworded comments a bit (mainly taking out information I feel isnt needed).

A difference from MM is `Actor_UpdatePosFromSkelAnime` -> `Actor_UpdatePosByAnimation` (this func is unused)

Also, I'm not sure what to name `func_8002D9A4`. Its currently called `Actor_SetSpeeds` in MM, but I think this is too general.
This function gets used in 3 whole actors: hookshot, arrow, and boomerang. It takes in a general speed value and factors in the actors pitch to set speedXZ and y velocity correctly. 
Its only used by projectiles, so my first instinct was `_SetProjectileSpeed`. But in conversation with @Thar0, we left off feeling like it should probably be more general. But Im not entirely sure and would like more opinions.


src/code/z_actor.c
```diff
@@ -841,54 +841,73 @@ void Actor_Destroy(Actor* actor, PlayState* play) {
...
-void func_8002D908(Actor* actor) {
+/**
+ * Update actor's velocity without gravity.
+ */
+void Actor_UpdateVelocityWithoutGravity(Actor* actor) {
     f32 sp24 = Math_CosS(actor->world.rot.x) * actor->speedXZ;
 
     actor->velocity.x = Math_SinS(actor->world.rot.y) * sp24;
     actor->velocity.y = Math_SinS(actor->world.rot.x) * actor->speedXZ;
     actor->velocity.z = Math_CosS(actor->world.rot.y) * sp24;
```
fig02, line 886: any ideas for this temp name, lmk. `xzSpeedScale`? should it involve pitch as well? idk

engineer124: So `Actor_UpdateVelocityWithoutGravity` accounts for pitch but `Actor_UpdateVelocityWithGravity` does not. I'm just trying to compare that to `func_8002D9A4` which also does account for pitch. `Actor_UpdateSpeedWithoutGravity`?

For `sp24`, camera would name that `pitchAdjSpeedXZ` (pitch adjusted speed xz) as an idea

include/functions.h
```diff
@@ -358,12 +358,12 @@ void Actor_Kill(Actor* actor);
...
-void func_8002D908(Actor* actor);
-void func_8002D97C(Actor* actor);
-void func_8002D9A4(Actor* actor, f32 arg1);
+void Actor_UpdatePos(Actor* actor);
+void Actor_UpdateVelocityWithGravity(Actor* actor);
+void Actor_MoveWithGravity(Actor* actor);
+void Actor_UpdateVelocityWithoutGravity(Actor* actor);
+void Actor_MoveWithoutGravity(Actor* actor);
+void func_8002D9A4(Actor* actor, f32 speed);
 s16 Actor_WorldYawTowardActor(Actor* actorA, Actor* actorB);
```
engineer124, line 367: Should `Actor_UpdatePosByAnimation` be in here? What's the future intent/style for declarations of completely unused functions, but seem to be designed just fine and could be used externally? Would that be something to wait for more split header files?

Dragorn421:
I'm not a fan of WithGravity/WithoutGravity, I would expect the difference between the two to be only `vel.y += gravity`, not a different meaning of `Actor.speedXZ`

master | #1476 | summary | suggestion?
-|-|-|-
`func_8002D7EC`     | `UpdatePos`                    | `pos += velocity + displacement` (xyz)                  | `-`
`func_8002D868`     | `UpdateVelocityWithGravity`    | `velocity xz = speedXZ`, gravity                        | `UpdateVelocityXZGravity`
`Actor_MoveForward` | `MoveWithGravity`              | `UpdateVelocityWithGravity`, `UpdatePos`    | `MoveXZGravity`
`func_8002D908`     | `UpdateVelocityWithoutGravity` | `velocity xyz = speedXZ`                                | `UpdateVelocityXYZ`
`func_8002D97C`     | `MoveWithoutGravity`           | `UpdateVelocityWithoutGravity`, `UpdatePos` | `MoveXYZ`
`func_8002D9A4`     | `-`                                  | `speedXZ, velocity.y = speed`                           | `SetSpeedXYZ`
`func_8002D9F8`     | `UpdatePosByAnimation`         | `pos += animTranslation`                                | `-`

src/code/z_actor.c
fig02: uh oh, does the instance variable need to change? lmao. since its context dependent on what move function is used..
speedXZ/the xz + gravity function is way more common, and is why it has the name it does. but... yeah

include/functions.h
fig02: any function that seems like it is meant to be called externally will be exposed. personally dont really care for functions.h that much though, cause it will just be fixed when proper headers exist. we were very inconsistent about what gets included in functions.h in the decomp phase, so its a problem everywhere

fig02:
I agree that your suggestions are more accurate, but with all the axis letters and stuff they are really clunky names. I personally understand exactly what the MM names meant, and would want a function that gets called in almost every actor to be nice to read/easy to remember

Im not shooting down the suggestions (since again I agree they are more correct), but want more opinions

src/code/z_actor.c
fig02: more I think about it, the instance variable probably should be just `speed`, with a comment clarifying that it can be XZ or XYZ depending on which move function is used

Dragorn421: I thought we were aware speedXZ not always being speedXZ and being fine with the """odd""" uses (at least I was aware :sweat_smile: )

If that's not the case yeah it may be worth debating renaming it to `speed`

fig02: I think i instinctively knew cause Id still set it when using the non-gravity move functions, but seeing it pointed out like this is still surprising to me yea

I think its just called speedXZ due to legacy reasons (rom hackers + speedrunners calling it "linear velocity" wrt link)

fig02: also with this change it would make more sense to include which axis youre moving on in the name of the function call, since it literally does depend on that and the speed value tells you nothing about it. So in that case I would deal with the "clunky" names.

Dragorn421:
Here's a graph showing the functions this PR touches and which actors use them:
[out.pdf](https://github.com/zeldaret/oot/files/10282458/out.pdf)
(keep in mind eg `Actor_MoveWithGravity` calls `Actor_UpdateVelocityWithGravity`)

Made with:
[ootgrep.py](https://github.com/zeldaret/oot/files/10282460/ootgrep.py.txt)
(for the graph `apt install graphviz`, `pip install graphviz`)

I am tempted to suggest calling the XYZ functions "Flying"

fig02:
I do "colloquially" refer to it as the flying variant, but I am not sure its best to impose that much meaning behind how the function is most commonly used in the name

as noted in comments above I think indicating the axes it operates in will make sense if we change `speedXZ` to `speed`, since it doesnt actually hold any information about what axes it operates in on its own

fig02:
Ive implemented the names in dragorns table of suggestions above.
If #1477 is accepted, and we more correctly call this speed variable without any axis information, it makes sense for the move functions to specify which axes the speed value will have meaning for.

fig02:
maybe just clarify that nuance in the function comment

src/code/z_actor.c
```diff
@@ -841,54 +841,73 @@ void Actor_Destroy(Actor* actor, PlayState* play) {
...
+ */
+void Actor_MoveXYZ(Actor* actor) {
+    Actor_UpdateVelocityXYZ(actor);
+    Actor_UpdatePos(actor);
 }
 
-void func_8002D9A4(Actor* actor, f32 arg1) {
-    actor->speedXZ = Math_CosS(actor->world.rot.x) * arg1;
-    actor->velocity.y = -Math_SinS(actor->world.rot.x) * arg1;
+void Actor_SetSpeedXYZ(Actor* actor, f32 speed) {
```
hensldm, line 898: There's been a little discussion, but I think since this sets `speed` for the `XZ` variant, I would say name it `Actor_SetSpeedXZ` and than probably name the argument `speedXYZ` since that is what it takes in.

Dragorn421: Sounds like a good idea but what about velocity.y being set? maybe `Actor_SetSpeedXZVelY` ? Thonk

fig02: my main concern with this is (all of these names) it sounding so general when it seems more specialized.
In conversations with tharo, he said that other actors probably could have made use of this function, even though its only used by projectiles in the final game. But just want to make sure that is the case before naming it so generally

Also he noted this
```
oh this in Actor_GetProjectileActor depends on projectiles using the specific function:

deltaX = Math_SinS(actor->world.rot.y) * (actor->speedXZ * 10.0f);
deltaY = actor->velocity.y + (actor->gravity * 10.0f);
deltaZ = Math_CosS(actor->world.rot.y) * (actor->speedXZ * 10.0f);
```
Im not sure this is enough to make it projectile specific, but something to note

hensldm: Hmm, to me at least the function seems kind of like a general convert a `speedXYZ` into `speedXZ` and `velocity.y` for use with the move with gravity function. I think most of the time, an actor with gravity just doesn't really care about their total speed (XYZ), but really only care about their `XZ` speed so don't end up using this function.

fig02: (conversation continued in discord, and as of posting, still ongoing: https://discord.com/channels/688807550715560050/838852326231769089/1056077160667951154)

src/code/z_actor.c
```diff
@@ -841,54 +841,73 @@ void Actor_Destroy(Actor* actor, PlayState* play) {
...
     f32 speedRate = R_UPDATE_RATE * 0.5f;
 
     actor->world.pos.x += (actor->velocity.x * speedRate) + actor->colChkInfo.displacement.x;
     actor->world.pos.y += (actor->velocity.y * speedRate) + actor->colChkInfo.displacement.y;
     actor->world.pos.z += (actor->velocity.z * speedRate) + actor->colChkInfo.displacement.z;
 }
 
-void func_8002D868(Actor* actor) {
+/**
+ * Update actor's velocity accounting for gravity (without exceeding terminal velocity)
```
Thar0, line 856:
```suggestion
 * Update actor's velocity accounting for gravity (without dropping below minimum y velocity)
```

src/code/z_actor.c
```diff
@@ -841,54 +841,73 @@ void Actor_Destroy(Actor* actor, PlayState* play) {
...
+ * The actor will move in the direction of its world yaw.
+ */
+void Actor_MoveXZGravity(Actor* actor) {
+    Actor_UpdateVelocityXZGravity(actor);
+    Actor_UpdatePos(actor);
 }
 
-void func_8002D908(Actor* actor) {
+/**
+ * Update actor's velocity without gravity.
```
Thar0, line 879: It's probably worth mentioning in comments what the role of `speedXZ` (soon to be`speed`) is in these functions?

Thar0 (changes requested):

src/code/z_actor.c
fig02: can you write a comment, im not sure exactly how to word what you want to convey

src/code/z_actor.c
```diff
@@ -841,54 +841,75 @@ void Actor_Destroy(Actor* actor, PlayState* play) {
...
+ */
+void Actor_MoveXYZ(Actor* actor) {
+    Actor_UpdateVelocityXYZ(actor);
+    Actor_UpdatePos(actor);
 }
 
-void func_8002D9A4(Actor* actor, f32 arg1) {
-    actor->speed = Math_CosS(actor->world.rot.x) * arg1;
-    actor->velocity.y = -Math_SinS(actor->world.rot.x) * arg1;
+void Actor_SetProjectileSpeed(Actor* actor, f32 speedXYZ) {
```
Dragorn421, line 900:
```c
/**
 * From a XYZ speed, set the corresponding XZ speed as `actor.speed`, and Y speed as Y velocity.
 * Only the actor's world pitch is factored in, with positive pitch moving downwards.
 */
```
idk (again this is correct afaik but idk about wording)

Dragorn421 (approved):

src/code/z_actor.c
Dragorn421: This concluded as

(quoting fig)
> I’m changing the function to SetProjectileSpeed in the PR cause it’s my personal opinion of what is best, but I don’t consider this conversation resolved yet. If you guys come to a consensus on something else I’ll gladly change it

and yeah :shrug: let's go

src/code/z_actor.c
Dragorn421: For context, #1477 happened (Actor.speedXZ -> speed) and this function is as of writing this `Actor_UpdateVelocityXYZ` with the temp named `speedXZ`

src/code/z_actor.c
fig02: Yeah I dont consider it concluded, but the conversation is also not really moving anywhere. If you guys (tharo, anon, engineer, anyone else) have more to say about it then feel free

hensldm (approved):
LGTM

Thar0 (approved):

Dragorn421:
Merge soon :tm: ?

Any more comment @engineer124 ? (since you reviewed this previously), or anyone else

src/code/z_actor.c
```diff
@@ -841,54 +841,79 @@ void Actor_Destroy(Actor* actor, PlayState* play) {
...
     f32 speedRate = R_UPDATE_RATE * 0.5f;
 
     actor->world.pos.x += (actor->velocity.x * speedRate) + actor->colChkInfo.displacement.x;
     actor->world.pos.y += (actor->velocity.y * speedRate) + actor->colChkInfo.displacement.y;
     actor->world.pos.z += (actor->velocity.z * speedRate) + actor->colChkInfo.displacement.z;
 }
 
-void func_8002D868(Actor* actor) {
+/**
+ * Update actor's velocity accounting for gravity (without dropping below minimum y velocity)
```
mzxrules, line 856:
```suggestion
 * Update actor's velocity accounting for gravity, without dropping below minimum y velocity
```
I wonder how necessary it is to specify that y velocity is clamped. Are there cases where y velocity is allowed exceed the minimum?

fig02: I think its worth a mention, though yeah its not critical. That is also why its in parenthesis currently, as a side note. So I dont think it really needs to change

engineer124 (approved):

src/code/z_actor.c
Dragorn421: :shrug: 

Solved?

mzxrules: Not quite. I hadn't considered the bigger picture until now, but I had assumed that `actor->minVelocityY` always clamped y velocity but this isn't the case. It only seems to do so when the gravity calculation is being applied, which suggests that the variable needs to be renamed.

Dragorn421: Renamed to what?

fig02: can add a comment to the struct member clarifying that its only relevant when move functions that involve gravity are used. but im not sure we'll be able to convey all that nuance in the name alone

mzxrules: I don't have any strong opinions on names... `minGravityVelocity` perhaps?

Dragorn421: I think `minGravityVelocitY` makes less sense than `minVelocityY`, both with and without knowing what it actually is

fig02: yea its hard to covey "minimum velocity when using gravity" in a concise name. 
So I say 
`// Sets the lower bounds cap on velocity along the Y axis`
->
`// Sets the lower bounds cap on velocity along the Y axis when gravity is applied`

hensldm: Perhaps I'm missing something, but what about `terminalVelocity`.

Dragorn421: `terminalVelocity` is definitely the right idea but it's also very wrong in the physics sense, terminal velocity results from gravity being compensated by the force of drag, and in oot there's no force of drag. What happens is more "gravity disappears beyond a certain speed"

so idk, not a fan of refering to terminal velocity

hensldm: Hmm, perhaps something like `minFallVelocity`? Feel like it conveys the same idea, but wouldn't get conflated with terminal velocity. (I'm also fine leaving it as is with a good comment)

Dragorn421: I like that better than `terminalVelocity`, though it sounds more twisted to think of `minFallVelocity` as a lower bound compared to `minVelocityY`

Maybe "limit" somewhere? `fallVelocityLimit` ?

But tbh all this is just playing on words to have the same meaning as `minVelocityY`. Since the PR is otherwise ready and doesn't currently touch `minVelocityY` I'm down to merging now and change `minVelocityY` in another PR if one holds dear to it

fig02: I clarified this issue in the comment. I dont see a better name being found, but yea ofc open to ideas if they surface down the line

PR #1480, opened 2022-12-27
Cleanup player speed and yaw variable names

fig02:
Name change justifications (sorry for the wall of text, but I think its important to detail my reasoning):

`player->linearVelocity` -> `player->speedXZ`:

At first, I thought this variable should be named something like "speed xz target", because this value can technically be a target value which the actor struct speed variable steps to.
However, when thinking larger picture, its only a "target" value for a relatively niche number of cases: standing on ice and wearing hover boots. Otherwise it is a direct assignment from `speedXZ` to `actor.speed`. 
Additionally, the generalization of `actor.speed` allows player's variable to be more specific as `speedXZ` without being conflated with the actor variable.
Laslty, `speedXZ` is referenced in many places outside of player. These actors dont really care about the underlying implementation detail that it can be a target value sometimes. They read it as a current reading of players speed. This makes me feel more comfortable about this slight abstraction when it comes to naming.

`player->currentYaw` -> `player->yaw`:

This variable has the same mechanic as `speedXZ` where it can act as a target value for ice and hoverboots. Because of this, I think its good to take out the word "current". 
For the same reasons given for `speedXZ`, I think its ok to not specify "target".
This will need more research, but as far as I can tell, this yaw value is used for both world and shape in many places. This is why I chose a very generic "yaw" without specifying what type of yaw. It again seems like a higher level abstraction which gets referenced externally in other actors and engine systems, like `speedXZ`.

`player->targetYaw` -> `player->zTargetYaw`

Just adds z to make it clear its for z targeting. I know our terms regarding z targeting are still under debate. This name will likely change, but this is necessary to clarify for now.

`func_80832210` -> `Player_ZeroSpeedXZ`

Dragorn's original suggestion from #1437. With the changes above, I think this name is best.


Dragorn421 (commented):
What is `zTargetYaw`?

fig02:
I dont have a specific answer for now other than "related to z targeting"

It gets set in `func_808355DC`, which is also where links yaw related variables are set when targeting a wall.
Though its not just wall targeting, `zTargetYaw` will get set to shape y rotation when that function is called regardless of the wall interaction. This is relevant for things like throwing the boomerang (`func_808359FC`)

I dont doubt this name will probably change as we settle our z target/parallel debates, but for now this change makes it more obvious that its z target related and not "a target value"

Dragorn421:
Ah so "z target" as in `(z-parallel) || (z-target something)`, and Link's yaw

I thought it refered to the latter only (`z-target something`) and wasn't sure if it refered to Link's yaw or the z-targeted thing's yaw

I would like to suggest something but no idea (yet?) :shrug:

fig02:
relevant blocks of code from cdi's branch for cross reference: 
https://github.com/CDi-Fails/oot/blob/player_docs/src/overlays/actors/ovl_player_actor/z_player.c#L2991
https://github.com/CDi-Fails/oot/blob/player_docs/src/overlays/actors/ovl_player_actor/z_player.c#L3440

some more usages in input related functions, i imagine to handle how analog stick/moving is handled while holding target. more research/docs necessary

fig02:
notably shape.rot.y is stepped toward `zTargetYaw` when targeting. And regular `yaw` in most other cases.
`yaw` being used for shape here, and `world` in the hoverboots/ice case is why I chose to make `yaw` very generic, it covers both

include/z64player.h
```diff
@@ -686,9 +686,9 @@ typedef struct Player {
     /* 0x0834 */ s16        unk_834;
     /* 0x0836 */ s8         unk_836;
     /* 0x0837 */ u8         unk_837;
-    /* 0x0838 */ f32        linearVelocity;
-    /* 0x083C */ s16        currentYaw;
-    /* 0x083E */ s16        targetYaw;
+    /* 0x0838 */ f32        speedXZ;
+    /* 0x083C */ s16        yaw;
+    /* 0x083E */ s16        zTargetYaw;
```
Dragorn421, line 691:
```suggestion
    /* 0x0838 */ f32        speedXZ; // Main parameter for horizontal speed. Controls `Actor.speed`
    /* 0x083C */ s16        yaw; // Related to yaw (current or target)
    /* 0x083E */ s16        zTargetYaw; // Related to the player yaw when using Z to target an actor or a wall
```
?

Dragorn421: I forgot, for `zTargetYaw` : "(also boomerang related)" ?

fig02: boomerang was just an example of an interaction with target i gave to show you that its related to targeting.
when you throw the boomerang, you enter a z target mode even tho you arent holding Z

fig02: but yea I can add comments, tho I will prob rephrase

fig02: I added some comments. I made the z target one fairly vague, at least until we know all of the details

Dragorn421 (approved):

hensldm (approved):

Roman971 (approved):

PR #1481, opened 2022-12-29
document bug in eff_ss_d_fire

fig02:
thanks to @Thar0 for helping me debug this issue in our rom hack 🙃


Dragorn421 (approved):
This is especially funny with three other effects having a `oldSeg6`-mechanism for temporarily changing the segment 6 base, and two of those three other effects don't need it PepeLaugh

Should there be a comment near `EffectSsDFire_Spawn` as well?

hensldm (approved):

engineer124 (approved):

fig02:
Im not against adding a note on the spawn function, and understand why it makes sense to, since the spawn functions are more public-facing.
But do also think we should probably keep the comment on the bug where the actual bug is. 

maybe in general we add comments to all the effect spawn functions that need a special object present that isnt gameplay keep (in another pr)

PR #1483, opened 2022-12-30
Document pause states (excluding gameover)

Dragorn421:
- Add a `PauseState` enum for `PauseContext.state` values
- Add macros for common state checks: `IS_PAUSE_STATE_GAMEOVER` and `IS_PAUSED`
- (re)name a few related functions and temps

Please refer to #1366 if you need the bigger picture of what's happening in some parts
I'd be fine with a bit of scope creep but it can get out of hand really quick (full docs diff is +4000 -2000 :fearful:)


include/z64pause.h
```diff
@@ -26,6 +26,35 @@ typedef enum {
...
 #define PAUSE_PLAYER_SEGMENT_GAMEPLAY_KEEP_BUFFER_SIZE 0x5000
 
+typedef enum {
+    /*  0 */ PAUSE_STATE_OFF,
+    /*  1 */ PAUSE_STATE_WAIT_LETTERBOX, // Request no letterboxing and wait for it.
+    /*  2 */ PAUSE_STATE_WAIT_BG_PRERENDER, // Wait for the pause background prerender to be done.
+    /*  3 */ PAUSE_STATE_INIT, // Load data and initialize/setup various things.
+    /*  4 */ PAUSE_STATE_OPENING_1, // Animate the pause menu coming together with rotations and other animations.
+    /*  5 */ PAUSE_STATE_OPENING_2, // Finish some animations for opening the menu.
+    /*  6 */ PAUSE_STATE_MAIN, // Pause menu ready for player inputs.
```
fig02, line 36: maybe active instead of main? both work

include/z64pause.h
```diff
@@ -26,6 +26,35 @@ typedef enum {
...
+    /* 10 */ PAUSE_STATE_10,
+    /* 11 */ PAUSE_STATE_11,
+    /* 12 */ PAUSE_STATE_12,
+    /* 13 */ PAUSE_STATE_13,
+    /* 14 */ PAUSE_STATE_14,
+    /* 15 */ PAUSE_STATE_15,
+    /* 16 */ PAUSE_STATE_16,
+    /* 17 */ PAUSE_STATE_17,
+    /* 18 */ PAUSE_STATE_CLOSING, // Animate the pause menu closing
+    /* 19 */ PAUSE_STATE_UNPAUSE
```
fig02, line 49: whats the difference here? is "unapuse" that state the whole time while the game is unpaused and the player is playing the game? in that cause "unpaused" might be better, or "inactive" if main is changed to "active"

include/z64pause.h
engineer124: Something to consider is that there's a substate enum that will be needed for this state. Something like:
`PAUSE_MAIN_STATE_XXX`
or 
`PAUSE_ACTIVE_STATE_XXX`
which will be `pauseCtx->unk_1E4`

include/z64pause.h
Dragorn421: Unpause is the state that deals with unpausing, when the game isn't paused the state is OFF

I didn't realize state_unpause could be understood as state_unpaused, would state_unpausing be better?

include/z64pause.h
Dragorn421: Yeah this is where the scope can widen significantly, the substates of the main state (there also are substates for the save dialog state) involve logic from the equips page, which is a beast of its own. I could introduce dummy enums (named by number) though if it helps provide context

include/z64pause.h
fig02: it would, but at a cursory glance (without diving into to the code) am wondering how that differs from closing. the menu closing is usually associated with unpausing in general, at a very high level

Dragorn421: CLOSING is the animation of the pause menu going (rotating) away,
UNPAUSE is do things to resume normal play state execution like restoring the object buffer contents, segtovirtual the collision headers, and some various other stuff

fig02: I see
maybe a name centered around "resume" or "restore" would be nice, just to try to be more specific about what its doing. havent put much thought into it though, will let you know if I come up with anything

include/z64pause.h
Dragorn421: I could also leave this state named as a number (`PAUSE_STATE_6`) until introducing the sub-state enum

include/z64pause.h
Dragorn421: `UNPAUSE` -> `RESUME_GAMEPLAY` and added a comment in the enum

include/z64pause.h
Dragorn421: Well I did just that. sounds like the better idea

kemenaran: For what is worth: in the LADX disassembly, when a system is finally ready after several stages of initialization, we named this kind of state "interactive". That would give something like `PAUSE_INTERACTIVE_STATE` in that case.

I find it descriptive enough. Of course it is lengthier to write (but I guess that could be abridged for sub-enums). And it doesn't perfectly convey that other states may also be responsive to user input, like the save prompt (but neither does "_ACTIVE").

src/code/z_player_lib.c
```diff
@@ -1637,7 +1637,7 @@ void Player_PostLimbDrawGameplay(PlayState* play, s32 limbIndex, Gfx** dList, Ve
     }
 }
 
-u32 func_80091738(PlayState* play, u8* segment, SkelAnime* skelAnime) {
+u32 Player_InitDrawPause(PlayState* play, u8* segment, SkelAnime* skelAnime) {
     s16 linkObjectId = gLinkObjectIds[(void)0, gSaveContext.linkAge];
```
fig02, line 1641: I totally get what this name is going for, but wondering if we can improve it. I get its related to drawing in that its loading assets and initializing skelanime, but personally wouldnt call this as "init for draw". It feels more logic-side to me.

Especially considering the larger point of this function is the dma-ing thats happening (and the size that gets returned), I wonder if the name should be more centered around that. or maybe just `Player_InitForPause`, removing draw and making it more general "init things for the pause menu"

Dragorn421: To explain why I initially picked "InitDrawPause": If you look at `Player_DrawPause`, that's where the "use" side of the buffer this sets up is, so I just added "Init" to that name 4Head

fwiw `Player_InitForPause` sounds to me like it would change something in the player system, but afaik pause menu does nothing to the player (`Player`)

fig02: yea the intent is "init player related things for drawing". But thats fair either way.
I cant put into words why "init draw" makes my brain hurt. I know it can be taken to understand "init the things that will be used for drawing", but it just doesnt sound right lol

I think a name more centered on the data dma it does is still potentially viable, `Player_InitPauseData` or something, with skelanime being lumped in there as a side thing, idk

It also loads gameplay keep which isnt even player centric. Which makes me kinda doubt the player prefix at all, but i guess a majority of the function is player related

Dragorn421: Afaik gkeep is only (partially) loaded for the purpose of drawing player, for example for `gHylianShieldDesignTex` and other textures (and `gLinkPauseChildJointTable` and family apparently) (actually maybe all the ~0x2000 first bytes of gkeep, though 0x5000 are kept and idk what/if the next 0x3000 are used for)

The whole buffer is named `playerSegment` (`PauseContext.playerSegment`) so if we move away from player at all (but at this time I don't see a reason we should), it's something to rename too

`Player_PauseInitPlayerSegment` ? bleh

Currently:
```
Player_InitDrawPause
Player_DrawPause
Player_DrawPauseImpl
```

Could also do:
```
Player_PauseInitSegment
Player_PauseDraw
Player_PauseDrawImpl
```
or something

hensldm: The second list doesn't seem too bad. Could also `Player_InitPauseData` and `Player_InitDrawPause` and do something like `Player_InitDrawPauseData/Player_InitPauseDrawData`?

hensldm (commented):
Overall LGTM, added a comment to the `Player_InitDrawPause` discussion.

src/code/z_player_lib.c
fig02: i am not opposed to `Player_InitPauseDrawData` tbh, it gets rid of my issue of "init" and "draw" being next to each other. it tells you that its initializing draw data

fig02: `Player_InitPauseDrawingData` maybe to elimate the confusion of "init pause, draw the data", but that also doesnt really make sense anyway and think itd be clear already

Dragorn421: I'm fine with `Player_InitPauseDrawData`

do I keep `Player_DrawPause` and `Player_DrawPauseImpl`,
or change to `Player_PauseDraw` and `Player_PauseDrawImpl`

fig02: Id say DrawPause still, to stay in line with things like `Player_DrawGameplay`. think its fine if the data init func doesnt have the same scheme

fig02 (approved):

hensldm (approved):

PR #1484, opened 2022-12-31
Import fault bss under NON_MATCHING

Thar0:
Importing for NON_MATCHING allows shifting the structs without having to manually change sizes in the asm files. The `FaultMgr` struct contains various substructs such as `OSThread` and `Input`, those structs are not shiftable prior to this change and it's not immediately obvious that it's due to unimported fault bss.
This is intended as a temporary solution until fault bss reordering is worked out for matching.

### DIFF
--- a/data/fault.bss.s
+++ b/data/fault.bss.s
@@ -7,6 +7,8 @@
 
 .section .bss
 
+// Note: This file is only included in the MATCHING build, the data is imported for non-matching
+
 .balign 16
 
 glabel sFaultInstance
--- a/data/fault.bss.s
+++ b/data/fault.bss.s
@@ -7,7 +7,7 @@
 
 .section .bss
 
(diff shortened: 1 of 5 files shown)


Dragorn421 (commented):
Have you considered putting the non-matching in the .s files instead of .o ? eg like this https://github.com/Dragorn421/oot/commit/23ad47e1da28a4ee3b9e2645f59ddb1946c4d069

mostly it would make the comment irrelevant or at least make it more obvious, if we should keep a comment

Thar0:
asm files in data/ aren't preprocessed by cpp, that could change but generally it's not very useful since files in data/ are meant to be temporary

Dragorn421 (approved):
Oh I didn't realize that :+1: yeah nbd

Roman971 (approved):

hensldm (approved):

PR #1485, opened 2022-12-31
Identified and named some player states

TheClave:
Hi there, I could identify some of the PLAYER_STATE... definitions in z64player.h and renamed them and their references in other files. I modded the debug overlay to be able to identify the player flags (not part of the commit). There are still about 20 states left to identify, but hopefully, this will help to understand some of the code better.

Thanks for all your work, it is truly amazing. Have a great start into the new year!

Johann


include/functions.h
```diff
@@ -877,9 +877,9 @@ f32 Math_SmoothStepToDegF(f32* pValue, f32 target, f32 fraction, f32 step, f32 m
 s16 Math_SmoothStepToS(s16* pValue, s16 target, s16 scale, s16 step, s16 minStep);
 void Math_ApproachS(s16* pValue, s16 target, s16 scale, s16 step);
 void Color_RGBA8_Copy(Color_RGBA8* dst, Color_RGBA8* src);
-void func_80078884(u16 sfxId);
-void func_800788CC(u16 sfxId);
-void func_80078914(Vec3f* arg0, u16 sfxId);
+void Audio_PlaySfxGeneralWithDefaultSettings1(u16 sfxId);
+void Audio_PlaySfxGeneralWithDefaultSettings2(u16 sfxId);
+void Audio_PlaySfxGeneralWithDefaultSettingsAtPosition(Vec3f* arg0, u16 sfxId);
```
Dragorn421, line 882: Note this clashes with #1470 , and overall with PlaySfx functions our local resident audio person @engineer124 has a big plan for naming the *many* PlaySfx wrappers (see the comment on that PR)

Dragorn421 (commented):
Again as I said on Discord thanks for your contribution and the kind words :)
Just one comment on something done in another PR

Before reviewing the state names, I want to discuss the approach for documenting it, namely experiment in-game vs looking at code

---------------

In the past I've strongly prefered "100% looking at code", my reasoning being that it is likely to misinterpret an observed behavior or miss an aspect of it, or even coming to a plain wrong conclusion.
But I've also changed my stance since, for two reasons:

1) Player is incredibly complicated. imo it's most of the oot engine that's in there
2) It's hard to tell what was intended from code because code bugs, code hacks and odd choices of implementation are common

Which leads to having a high picture grasp through in-game testing, being really useful to contextualize stuff when looking at the code then.

So I'm conflicted and would like other opinions to make up my mind

I wanted to discuss this directly in Discord but had to go afk shortly after saying hi :sweat_smile: 

----------

Also for reference the three current resources I know of with player documentation are:
- https://docs.google.com/spreadsheets/d/12RpxLB-AhiilGi6aG7MNKia2lhkAKivjly9V8GL-utg/edit#gid=0
- https://docs.google.com/spreadsheets/d/1W7QQFoaVz4cEWm-jfQUfN05QriI1YBLKZXL4qLXwOFk/edit#gid=664420014
- https://github.com/CDi-Fails/oot/blob/player_docs/src/overlays/actors/ovl_player_actor/z_player.c

include/functions.h
TheClave: True, I haven't read this PR in more detail. I am fine with having there a second underscore. @engineer124 Nothing to do here, I will just adjust my code to yours if this is alright for all of you:

```suggestion
void Audio_PlaySfx_GeneralWithDefaultSettings1(u16 sfxId);
void Audio_PlaySfx_GeneralWithDefaultSettings2(u16 sfxId);
void Audio_PlaySfx_GeneralWithDefaultSettingsAtPosition(Vec3f* arg0, u16 sfxId);
```

It's frustrating that both functions "Audio_PlaySfx_GeneralWithDefaultSettings" do the same, look the same, but are not allowed to be optmized...

TheClave:
I have renamed some player states after viewing the docs and doing more testing and research. I will make another commit for the Lib_PlaySfx function names.

fig02:
Hi 
Just dropping my 2 cents wrt the approach we take for player documentation:
This is an extremely complex system, and I believe that the best approach will be one sub system at a time. Player is made of little sub systems like the item handling system, player cutscene system, action function system, etc. And I think documenting these sub systems in depth, as opposed to broad sweeping things like this, will be both more effective/accurate and easier for us to review.

I do appreciate the efforts put in to name these states, and think it will be a good reference to have going forward. As Dragorn linked, we've had similar attempts at naming these states based on a more observational approach in the past before. I think it will be good for people to double check these resources as different sub systems (and associated states) get documented. But I am not sure its best to merge them to master at the moment.

Player documentation is certainly lacking, but I hope to be jump starting that process of documenting player very soon. It may help to see what I mean about "sub-system level documentation" with an example, when Im ready to start that process (hopefully within a few weeks)

TheClave:
Hi fig, thank you for speaking out. I agree that deciphering and documenting needs a bigger plan. But I don't really understand your proposed approach of sub-systems. Can you explain it in more detail or should I wait for you example then?

Thanks,
Johann

fig02:
For example I would consider all of the cutscene cues/cs modes a sub system (`D_80854B18`/`func_80852C50`/`func_80852C50` and affiliated functions)
Handling links items (func_82C, functions in `D_80853EDC`/`D_80853FE8`, etc ) are another
Various groups of actions (func_674) as well

All of these are mostly isolated systems with a specialized purpose within player that can be documented without (alot of) spilling out into other areas. This approach I believe will lead to the most accurate results documentation wise, and also be easier for us to review, as we dont need to "fact check", so to speak over, such a large spectrum of code.

With large changes like this, we either have to do all of the documentation work you did to come to these conclusions, or take your word for it until more of players code is documented. 

Thats not to say I dont trust that you did your best to come up with accurate names. Just that we like to verify everything with the code, and its hard to do that with broad level changes without the underlying code also being documented.

TheClave:
Alright, then I will close this PR for now.

PR #1489, opened 2023-01-03
Placeholder names for skelanime moveflags

Dragorn421:
Partly companion to #1437

- Add "missing" defines for `SkelAnime.moveFlags` flags, all of them only relevant/effect-bearing in player.
- Animation system callbacks: name `diffScaleY`/`moveDiffScaleY`
- Add `FLAG_FUNC_80832F54_8` (which suprisingly is unused unless I missed something), `FLAG_FUNC_80832F54_9`
- Try to name them but it's too FeelsUnkMan, only add "related to" comments

I wish I could have named them, but still PRing this as it's not trivial to travel from/to the flags in hex to/from where they're used, so it should be useful

The same flags in CDi's docs:
https://github.com/CDi-Fails/oot/blob/player_docs/include/z64player.h#L620
https://github.com/CDi-Fails/oot/blob/14fb990358dbf4fb144a58882dc4b316fd5bb54e/include/z64player.h#L620
One outstanding difference is (this PR) `ANIM_FLAG_PLAYER_SETMOVE` vs (CDi) `PLAYER_ANIMMOVEFLAGS_KEEP_ANIM_Y_TRANSLATION`
I think CDi named it based on https://github.com/CDi-Fails/oot/blob/14fb990358dbf4fb144a58882dc4b316fd5bb54e/src/overlays/actors/ovl_player_actor/z_player.c#L1954 (?) but I think the other use is a lot more important: https://github.com/CDi-Fails/oot/blob/14fb990358dbf4fb144a58882dc4b316fd5bb54e/src/overlays/actors/ovl_player_actor/z_player.c#L11137-L11138 :shrug:


src/overlays/actors/ovl_En_Nb/z_en_nb.c
```diff
@@ -745,7 +745,7 @@ void EnNb_InitDemo6KInConfrontation(EnNb* this, PlayState* play) {
 }
 
 void func_80AB2688(EnNb* this, PlayState* play) {
-    this->skelAnime.moveFlags |= 1;
+    this->skelAnime.moveFlags |= ANIM_FLAG_PLAYER_0;
     AnimationContext_SetMoveActor(play, &this->actor, &this->skelAnime, 1.0f);
 }
```
fig02, line 750: (I know nothing about these flags so sorry for the ignorance)
why are these named after player if other actors are using them? its used on their own skelanime, and they pass themselves to "set move actor"

does "set move actor" pertain to player specifically?

Dragorn421: Only player has code reacting to the `ANIM_FLAG_PLAYER_`-named flags, so it only has a purpose when used in player

I'm not sure what to make of the fact this one is also used outside of player (with no effect afaict)

fig02: it does make it seem to me like its not intended to be player specific, even if hes the only one who ended up implementing it 
Id prob lean on the safe side and name it more generally, and later on if we feel its right, can name it after him

fig02 (approved):

hensldm (approved):

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -1905,17 +1905,20 @@ void func_80832E48(Player* this, s32 flags) {
     func_808322FC(this);
 }
 
+#define FLAG_FUNC_80832F54_8 0x100
+#define FLAG_FUNC_80832F54_9 0x200
```
Roman971, line 1909: Is there a reason you defined those with the raw values instead of (1 << x) like typical flags?

Dragorn421: maybe because of the mask below `this->skelAnime.moveFlags = flags & 0xFF` to make it clearer those are function specific flags but it's all ugly anyway, I changed it to shifts

Roman971 (approved):

PR #1491, opened 2023-01-04
Some build system cleanup

Dragorn421:
- Rework how the Makefile lists
  - folders (zapd destination folders under assets, source folders, ...)
  - and files (png, jpg, .s, .c ...)
- Move text extraction from extract_assets.py to running the msgdis script.
  - To extract text without `make setup`: `python3 tools/msgdis.py --text-out assets/text/message_data.h --staff-text-out assets/text/message_data_staff.h` (can easily be found in the makefile too)
- Remove unused tool makeromfs
- Remove elf2rom: replaced with
  - objcopy,
  - padding script `tools/pad_rom.py`
  - and `tools/n64checksum_6105` for the rom header checksum (from https://github.com/Dragorn421/n64checksum )

These are mostly changes I've wanted to do for a long time, but really have a reason to make now with the topics of assets system and licensing


fig02 (approved):
I am not savy enough with makefiles to be able to effectively review those changes, but am ok with everything else

hensldm (approved):
LGTM :)

Makefile
```diff
@@ -152,27 +149,30 @@ ELF := $(ROM:.z64=.elf)
...
-# source files
-C_FILES       := $(filter-out %.inc.c,$(foreach dir,$(SRC_DIRS) $(ASSET_BIN_DIRS),$(wildcard $(dir)/*.c)))
-S_FILES       := $(foreach dir,$(SRC_DIRS) $(UNDECOMPILED_DATA_DIRS),$(wildcard $(dir)/*.s))
-O_FILES       := $(foreach f,$(S_FILES:.s=.o),build/$f) \
-                 $(foreach f,$(C_FILES:.c=.o),build/$f) \
-                 $(foreach f,$(wildcard baserom/*),build/$f.o)
+# Source files
+# List all .c files in assets and source directories, excluding .inc.c files
+C_FILES := $(shell find $(ASSETS_ZAPD_DEST_DIRS) $(SRC_DIRS) -name '*.inc.c' -o -name '*.c' -print)
+S_FILES := $(shell find $(SRC_DIRS) data -name '*.s')
```
Thar0, line 174: Why replace `foreach` with `shell find` here?

Makefile
```diff
@@ -181,13 +181,11 @@ OVL_RELOC_FILES := $(shell $(CPP) $(CPPFLAGS) $(SPEC) | grep -o '[^"]*_reloc.o'
...
-TEXTURE_FILES_JPG := $(foreach dir,$(ASSET_BIN_DIRS),$(wildcard $(dir)/*.jpg))
-TEXTURE_FILES_OUT := $(foreach f,$(TEXTURE_FILES_PNG:.png=.inc.c),build/$f) \
-					 $(foreach f,$(TEXTURE_FILES_JPG:.jpg=.jpg.inc.c),build/$f) \
-
 # create build directories
-$(shell mkdir -p build/baserom build/assets/text $(foreach dir,$(SRC_DIRS) $(UNDECOMPILED_DATA_DIRS) $(ASSET_BIN_DIRS),build/$(dir)))
+stdout := $(shell mkdir -p $(dir $(O_FILES)))
+ifneq ($(stdout),)
+$(error $(stdout))
+endif
```
Thar0, line 188: Could `stdout` get a more descriptive/purposed name?

Makefile
```diff
@@ -152,27 +149,30 @@ ELF := $(ROM:.z64=.elf)
 # description of ROM segments
 SPEC := spec
 
-ifeq ($(COMPILER),ido)
-SRC_DIRS := $(shell find src -type d -not -path src/gcc_fix)
-else
-SRC_DIRS := $(shell find src -type d)
+# Folders ZAPD extracts assets to
+ASSETS_ZAPD_DEST_DIRS := assets/code assets/misc assets/objects assets/overlays assets/scenes assets/textures
```
Thar0, line 153: Can this simply be `ASSET_DIRS`?

Makefile
```diff
@@ -276,9 +275,11 @@ test: $(ROM)
 #### Various Recipes ####
 
 $(ROM): $(ELF)
-	$(ELF2ROM) -cic 6105 $< $@
+	$(OBJCOPY) -O binary $< $@
+	python3 tools/pad_rom.py $@
```
Thar0, line 279: I'm not thrilled with using python for any build step, even if this is probably already faster overall than the previous elf2rom solution. I don't think it's a disaster if the checksum tool also did the padding, but if that's no good surely this is possible with a bash script?

tools/Makefile
```diff
@@ -30,11 +30,10 @@ distclean: clean
...
 .PHONY: all clean distclean
 
-elf2rom_SOURCES      := elf2rom.c elf32.c n64chksum.c util.c
-makeromfs_SOURCES    := makeromfs.c n64chksum.c util.c
 mkdmadata_SOURCES    := mkdmadata.c spec.c util.c
 mkldscript_SOURCES   := mkldscript.c spec.c util.c
 reloc_prereq_SOURCES := reloc_prereq.c spec.c util.c
+n64checksum_6105_SOURCES := n64checksum_6105.c
 vtxdis_SOURCES       := vtxdis.c
 yaz0_SOURCES         := yaz0tool.c yaz0.c util.c
```
Thar0, line 38:
```suggestion
mkdmadata_SOURCES        := mkdmadata.c spec.c util.c
mkldscript_SOURCES       := mkldscript.c spec.c util.c
n64checksum_6105_SOURCES := n64checksum_6105.c
reloc_prereq_SOURCES     := reloc_prereq.c spec.c util.c
vtxdis_SOURCES           := vtxdis.c
yaz0_SOURCES             := yaz0tool.c yaz0.c util.c
```

Thar0 (commented):
Some thoughts

Makefile
Dragorn421: I changed `SRC_DIRS` to only list the "root" folders

Makefile
Dragorn421: Not really, it's specifically the ones ZAPD touches (or we can find a better name to refer to this group of assets but idk)
For example it leaves out text, and in the future audio stuff too

Makefile
Dragorn421: so the purpose of these is if somehow something gets printed on stdout
(afaik on any normal system there shouldn't be anything in any case, even assuming the weirdest errors),
it doesn't get parsed as part of the makefile like e.g. an `$(shell echo myvar := myvalue)` line would

and assert the stdout is indeed nothing

I'm open to comments on this because as much as I think it's a proper way to handle things, I'm not a huge fan

Makefile
Dragorn421: btw sure we can not use python, would a C utility be fine

AngheloAlf: Is there really any difference? Does it matter the value used to pad the rom? Do the rom need to be padded?

tbh I wouldn't be against to just pad the rom with zeroes instead of using two different values.
For diffing against the baserom we could just modify/strip the padding. We already mess with the baserom in `fixbaserom.py`, so I think we could consider it

AngheloAlf: Actually, taking a look at `fixbaserom.py` seems like _we_ are the ones that are padding the ROM with `0xFF`.
https://github.com/zeldaret/oot/blob/0ce03133425867723dd81fc66f57bbb8d40aaf7a/fixbaserom.py#L76-L77
Changing that to pad using `0` instead could allow us to simply use objcopy to pad the rom:
https://github.com/Dragorn421/oot/compare/meta_buildsystem_cleanup...AngheloAlf:oot:meta_buildsystem_cleanup_padding
(I had to change the hash because the changes to the rom makes it change obviously )

Dragorn421: The explicit 0xFF padding is somewhat recent: https://github.com/zeldaret/oot/pull/987

Dragorn421: taken to discord https://discord.com/channels/688807550715560050/688851317593997489/1127217091397812327
will try to reach experts

Makefile
Dragorn421: we have to maintain it manually anyway, because of conditional "inclusion" of src/gcc_fix, and it contains `assets/text` which isn't in src/

Thar0: We don't currently have to manually maintain every inclusion of every source dir on account of `src/gcc_fix`, since we use `find` to create the list, so I don't see what this change achieves at least with respect to the source dirs?

Dragorn421: basically I think rather than (master) "the ido path knows about gcc_fix and removes it"
```mk
ifeq ($(COMPILER),ido)
SRC_DIRS := $(shell find src -type d -not -path src/gcc_fix)
else
SRC_DIRS := $(shell find src -type d)
endif
```
it's cleaner to do (this PR) "add gcc_fix if relevant"
```mk
SRC_DIRS := assets/text src/audio src/boot src/buffers src/code src/dmadata src/elf_message src/libultra src/makerom src/overlays
ifeq ($(COMPILER),gcc)
SRC_DIRS += src/gcc_fix
endif
```
and it allows moving in assets/text there too as a bonus

Dragorn421: Note I would agree there is an underlying issue here of poor structure of the repo

Thar0: Is this possible?
```make
SRC_DIRS := assets/text $(shell find src -type d -not -path src/gcc_fix)
ifeq ($(COMPILER),gcc)
SRC_DIRS += src/gcc_fix
endif
```

AngheloAlf: Why do we conditionally compile the files from `gcc_fix` instead of just building them and ignoring them on IDO builds?
If it is because of the `__asm__` statement then we could either move that asm into its own `.s` file or wrap it into a macro that expands to nothing on IDO.
If the C functions are problematic then we could just ifdef the whole file, so building with IDO just produces an empty file.

Thar0: Wrapping the file in `#ifndef __sgi` seems like a good option

tools/n64checksum_6105.c
Dragorn421, line 1: after https://github.com/zeldaret/oot/pull/1614 this can be yeeted in favor of `python3 -m ipl3checksum sum --cic 6105 --update`

this whole PR could also take inspiration from https://github.com/zeldaret/mm/pull/1542

AngheloAlf: Oh, I completely forgot about the existence of this PR lmao. I should have checked it before working on the other MM PR.
From a quick glance seems like you could replace `n64checksum_6105.c` and the `pad_rom.py` script with the `checksummer.py` script from MM#1542. I think having both functionalities on one script would be better because it avoids the overhead of running a second Python interpreter

fig02:
author agreed to close-- some of these things have already been implemented in other PRs. Some other things have not been decided yet, and can be handled separately

PR #1492, opened 2023-01-06
En_Cow Doc

philtyl:


src/overlays/actors/ovl_En_Cow/z_en_cow.c
```diff
@@ -114,39 +115,39 @@ void EnCow_Init(Actor* thisx, PlayState* play) {
...
+            R_PLAYED_EPONAS_SONG = 0;
             break;
         case 1:
             SkelAnime_InitFlex(play, &this->skelAnime, &gCowTailSkel, NULL, this->jointTable, this->morphTable, 6);
             Animation_PlayLoop(&this->skelAnime, &gCowTailIdleAnim);
-            this->actor.update = func_809DFE98;
-            this->actor.draw = func_809E0070;
-            this->actionFunc = func_809DFA84;
-            func_809DEF94(this);
+            this->actor.update = EnCow_AnimateTail;
```
Dragorn421, line 140:
```suggestion
            this->actor.update = EnCow_UpdateTail;
```

Dragorn421 (commented):
Note MM is ahead in terms of documenting EnCow, and it looks similar if not identical to OoT's

https://github.com/zeldaret/mm/blob/master/src/overlays/actors/ovl_En_Cow/z_en_cow.c
https://github.com/zeldaret/mm/blob/master/src/overlays/actors/ovl_En_Cow/z_en_cow.h

It is an implicit goal to keep the two repos consistent, so if there are names you like there you could use them, if not you can just leave it to someone more invested in the consistency.
(I wouldn't take the MM names as perfect so please feel free to suggest your own in this PR anyway!)

--------

Also hijacking this comment to bump https://github.com/Dragorn421/oot-licensing , did you see it (in particular the issue you were tagged in)

src/overlays/actors/ovl_En_Cow/z_en_cow.c
```diff
@@ -57,9 +58,9 @@ static ColliderCylinderInit sCylinderInit = {
     { 30, 40, 0, { 0, 0, 0 } },
 };
 
-static Vec3f D_809E010C = { 0.0f, -1300.0f, 1100.0f };
+static Vec3f headFocusOffset = { 0.0f, -1300.0f, 1100.0f };
```
Dragorn421, line 61:
```suggestion
static Vec3f sHeadFocusOffset = { 0.0f, -1300.0f, 1100.0f };
```
we s-prefix symbol names that are used in a single file
and g-prefix the others
(not directly related to using `static`, many s-prefixed symbols are not static just so they appear in the map file build/z64.map)

src/overlays/actors/ovl_En_Cow/z_en_cow.c
```diff
@@ -70,32 +71,32 @@ void func_809DEE00(Vec3f* vec, s16 rotY) {
     vec->x = xCalc;
 }
 
-void func_809DEE9C(EnCow* this) {
+void EnCow_PositionCow(EnCow* this) {
```
Dragorn421, line 74: I think MM's `EnCow_SetColliderPos` is better since this function indeed deals with positioning the colliders

src/overlays/actors/ovl_En_Cow/z_en_cow.c
```diff
@@ -106,188 +107,195 @@ void EnCow_Init(Actor* thisx, PlayState* play) {
...
+        this->animationTimer = Rand_ZeroFloat(500.0f) + 40.0f;
         Animation_Change(&this->skelAnime, &gCowBodyChewAnim, 1.0f, this->skelAnime.curFrame,
                          Animation_GetLastFrame(&gCowBodyChewAnim), ANIMMODE_ONCE, 1.0f);
     }
 
-    if ((this->actor.xzDistToPlayer < 150.0f) && !(this->unk_276 & 2)) {
-        this->unk_276 |= 2;
-        if (this->skelAnime.animation == &gCowBodyChewAnim) {
-            this->unk_278 = 0;
+    if ((this->actor.xzDistToPlayer < 150.0f)) {
```
Dragorn421, line 174:
```suggestion
    if (this->actor.xzDistToPlayer < 150.0f) {
```

src/overlays/actors/ovl_En_Cow/z_en_cow.h
```diff
@@ -3,21 +3,37 @@
...
-    /* 0x0228 */ Vec3s jointTable[6];
-    /* 0x024C */ Vec3s morphTable[6];
-    /* 0x0270 */ Vec3s someRot;
-    /* 0x0276 */ u16 unk_276;
-    /* 0x0278 */ u16 unk_278;
-    /* 0x027A */ u16 unk_27A;
+    /* 0x0228 */ Vec3s jointTable[COW_LIMB_MAX];
+    /* 0x024C */ Vec3s morphTable[COW_LIMB_MAX];
+    /* 0x0270 */ Vec3s headTilt;
+    /* 0x0276 */ u16 flags;
```
Dragorn421, line 34: We avoid naming actor structs fields `flags` because there's already such a field in the base `Actor` struct, and it can easily be confused (`this->flags` vs `this->actor.flags`)
You can use `cowFlags` instead
(smh MM's EnCow)

src/overlays/actors/ovl_En_Cow/z_en_cow.h
```diff
@@ -3,21 +3,37 @@
...
+typedef enum {
+    /*  0 */ EN_COW_TYPE_DEFAULT,
+    /*  1 */ EN_COW_TYPE_TAIL
+} EnCowType;
+
+typedef enum {
+    /*  0 */ EN_COW_COLL_FRONT,
+    /*  1 */ EN_COW_COLL_REAR,
+    /*  2 */ EN_COW_COLL_MAX
+} EnCowCollider;
```
Dragorn421, line 22: I would spell `COLLIDER` in full instead of `COLL`
there are issues with disambiguation between colliders (z_collision_check) and surface collision (z_bgcheck) so avoiding confusion where possible

src/overlays/actors/ovl_En_Cow/z_en_cow.h
```diff
@@ -3,21 +3,37 @@
...
 #include "ultra64.h"
 #include "global.h"
+#include "assets/objects/object_cow/object_cow.h"
+
+#define EN_COW_FLAG_PLAYER_HAS_APPROACHED (1 << 1)
+#define EN_COW_FLAG_WONT_GIVE_MILK (1 << 2)
+
+#define EN_COW_TYPE(thisx) ((thisx)->actor.params)
+typedef enum {
+    /*  0 */ EN_COW_TYPE_DEFAULT,
```
Dragorn421, line 13: Ah I was about to suggest "en_cow_type_cow" because "default" looks weird, but MM has `EN_COW_TYPE_ABDUCTED` so it makes sense there, hm

this kind of thing is hard to rule, since it's anachronistic to name things in OoT based on MM, but also the consistency is nice

I guess I'd say `EN_COW_TYPE_COW` would be better since "default" doesn't make sense in OoT, but I'm open to more opinions

src/overlays/actors/ovl_En_Cow/z_en_cow.c
```diff
@@ -106,188 +107,195 @@ void EnCow_Init(Actor* thisx, PlayState* play) {
...
+    this->animationCycle++;
+    if (this->animationCycle > 48) {
+        this->animationCycle = 0;
     }
 
     // (1.0f / 100.0f) instead of 0.01f below is necessary so 0.01f doesn't get reused mistakenly
-    if (this->unk_27A < 0x20) {
-        this->actor.scale.x = ((Math_SinS(this->unk_27A << 0xA) * (1.0f / 100.0f)) + 1.0f) * 0.01f;
+    if (this->animationCycle < 32) {
+        this->actor.scale.x = ((Math_SinS(this->animationCycle * 1024) * (1.0f / 100.0f)) + 1.0f) * 0.01f;
```
Dragorn421, line 190: Didn't you have `breathPhase` or `breathTimer` instead of MM's `animationCycle` at one point? I think it'd be a better name

src/overlays/actors/ovl_En_Cow/z_en_cow.c
```diff
@@ -358,11 +368,11 @@ void func_809DFE98(Actor* thisx, PlayState* play) {
 s32 EnCow_OverrideLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Vec3f* pos, Vec3s* rot, void* thisx) {
     EnCow* this = (EnCow*)thisx;
 
-    if (limbIndex == 2) {
-        rot->y += this->someRot.y;
-        rot->x += this->someRot.x;
+    if (limbIndex == COW_LIMB_HEAD) {
+        rot->y += this->headTilt.y;
```
Dragorn421, line 372: I know it's from MM but
```suggestion
        rot->y += this->headRot.y;
```
is more consistent with how we name these rotation fields

src/overlays/actors/ovl_En_Cow/z_en_cow.c
```diff
@@ -106,188 +107,195 @@ void EnCow_Init(Actor* thisx, PlayState* play) {
...
-                    this->actionFunc = func_809DF8FC;
+                    (ABS((s16)(this->actor.yawTowardsPlayer - this->actor.shape.rot.y)) < 25000)) {
+                    R_PLAYED_EPONAS_SONG = false;
+                    this->actionFunc = EnCow_Talk;
                     this->actor.flags |= ACTOR_FLAG_16;
                     func_8002F2CC(&this->actor, play, 170.0f);
                     this->actor.textId = 0x2006;
                 } else {
-                    this->unk_276 |= 4;
+                    this->cowFlags |= EN_COW_FLAG_WONT_GIVE_MILK;
```
Dragorn421, line 273: I'm confused by the `EN_COW_FLAG_WONT_GIVE_MILK` flag, what effect/purpose does it have in game?

(`msgCtx.ocarinaMode` docs courtesy of @engineer124 : https://github.com/zeldaret/oot/compare/master...engineer124:oot:ocarina )
`OCARINA_MODE_00` is `OCARINA_MODE_NONE` 

so most of the time `this->cowFlags &= ~EN_COW_FLAG_WONT_GIVE_MILK;` will run and the flag will be cleared :thinking:

Dragorn421: >  If this flag was enabled on the next tick along with R_PLAYED_EPONAS_SONG, it would mean that there wasnt another actor who could handle the epona's song reg flag, so it turns both of them off.

Aaaah! :brain: thanks that makes a lot of sense
I'm not sure how to name the flag then, could you write a brief comment somewhere (near `cowFlags & EN_COW_FLAG_WONT_GIVE_MILK` maybe) about what it achieves?

src/overlays/actors/ovl_En_Cow/z_en_cow.h
Dragorn421: then I think COW is better, typically these enums are important to be understood simply when used externally, to avoid having to browse through the code again to figure them out

Dragorn421 (approved):

Dragorn421 (approved):

src/overlays/actors/ovl_En_Cow/z_en_cow.c
```diff
@@ -106,188 +107,196 @@ void EnCow_Init(Actor* thisx, PlayState* play) {
...
-            this->actor.draw = func_809E0070;
-            this->actionFunc = func_809DFA84;
-            func_809DEF94(this);
+            this->actor.update = EnCow_UpdateTail;
+            this->actor.draw = EnCow_DrawTail;
+            this->actionFunc = EnCow_IdleTail;
+            EnCow_SetTailPos(this);
             this->actor.flags &= ~ACTOR_FLAG_0;
-            this->unk_278 = ((u32)(Rand_ZeroFloat(1000.0f)) & 0xFFFF) + 40.0f;
+            this->animationTimer = ((u32)(Rand_ZeroFloat(1000.0f)) & 0xFFFF) + 40.0f;
```
hensldm, line 148:
```suggestion
            this->animationTimer = (u16)Rand_ZeroFloat(1000.0f) + 40.0f
```
Also works and seems cleaner.

src/overlays/actors/ovl_En_Cow/z_en_cow.c
```diff
@@ -106,188 +107,196 @@ void EnCow_Init(Actor* thisx, PlayState* play) {
...
-            this->unk_27A = 0;
+            Actor_SpawnAsChild(&play->actorCtx, &this->actor, play, ACTOR_EN_COW,
+                               this->actor.world.pos.x, this->actor.world.pos.y, this->actor.world.pos.z,
+                               0, this->actor.shape.rot.y, 0, EN_COW_TYPE_TAIL);
+            this->animationTimer = Rand_ZeroFloat(1000.0f) + 40.0f;
+            this->breathTimer = 0;
             this->actor.targetMode = 6;
-            DREG(53) = 0;
+            R_PLAYED_EPONAS_SONG = false;
             break;
```
hensldm, line 138: Can you put a newline between the break and case? Makes it easier to see different cases.

hensldm (approved):

src/overlays/actors/ovl_En_Cow/z_en_cow.c
```diff
@@ -106,188 +112,209 @@ void EnCow_Init(Actor* thisx, PlayState* play) {
...
+void EnCow_Idle(EnCow* this, PlayState* play) {
     if ((play->msgCtx.ocarinaMode == OCARINA_MODE_00) || (play->msgCtx.ocarinaMode == OCARINA_MODE_04)) {
-        if (DREG(53) != 0) {
-            if (this->unk_276 & 4) {
-                this->unk_276 &= ~0x4;
-                DREG(53) = 0;
+        if (R_EPONA_CALLED) {
+            if (this->cowFlags & COW_FLAG_DONT_GIVE_MILK) {
+                this->cowFlags &= ~COW_FLAG_DONT_GIVE_MILK;
+                R_EPONA_CALLED = false;
```
Dragorn421, line 280: I think the comment there was nice to have because imo it's not trivial what this achieves (working properly with other things reacting to epona song)

Also not convinced by naming the reg after calling epona when it's really "has played epona song" isn't it?

Dragorn421: Iirc this also handles having a single cow give milk if there are several in range, but it's been a long time I kinda forgot

And about the reg I mean, cows react to playing the song, and the code listens on that reg does it not? (Again a long time has passed)

fig02: ah, yeah I did miss that the cow is listening for that reg to continue its own process. Thought it was just unsetting it instantly to prevent epona from spawning.

Ill change the name back, but still not sure on the comment, maybe it can be reworded. It still seems to me like its always the actor's responsibility to set the flag to false, since the message code (which sets it to true) doesnt handle that.

What is the "next tick" in the original comment referring to? It just gets unset immediately if a cow is giving milk, not sure what is happening on the next tick/frame

Dragorn421: For the tick thing I don't remember exactly but I do think there is a one frame delay before actions are taken, maybe the previous discussion on this will help (I don't remember well enough to write an explanation right now) https://github.com/zeldaret/oot/pull/1492#discussion_r1070426005

fig02: alright i understand now and have pushed a change.

I changed `COW_WONT_GIVE_MILK` -> `COW_FLAG_FAILED_TO_GIVE_MILK`. I think this better expresses the intent of why the flag exists/how it behaves in this interaction.
I also wrote a pretty lengthy comment that expands alot on the original. I think it is warranted for how complex the interaction is (especially cause I could not even notice the complexity of the interaction with the old comment)

src/overlays/actors/ovl_En_Cow/z_en_cow.c
```diff
@@ -106,188 +112,215 @@ void EnCow_Init(Actor* thisx, PlayState* play) {
...
-        if (DREG(53) != 0) {
-            if (this->unk_276 & 4) {
-                this->unk_276 &= ~0x4;
-                DREG(53) = 0;
+        // There is a complex interaction between `R_EPONAS_SONG_PLAYED` and `COW_FLAG_FAILED_TO_GIVE_MILK` to allow
+        // multiple cows to try and give milk on the same frame.
+        // `COW_FLAG_FAILED_TO_GIVE_MILK` gets set if a cow is not in range with the player to interact.
+        // In the case of a failure, `R_EPONAS_SONG_PLAYED` is not set to false incase another cow can succeed.
+        // On the following frame, if both `R_EPONAS_SONG_PLAYED` and `COW_FLAG_FAILED_TO_GIVE_MILK` are set, the
+        // first cow that updates can assume all other cows also failed and can safely unset `R_EPONAS_SONG_PLAYED`.
```
Dragorn421, line 282: this is a much better comment, I don't think I fully realized what was going on either (specifically ensuring `R_EPONAS_SONG_PLAYED` doesn't stay set)

```suggestion
        // multiple cows to try and give milk on the same frame.
        // `COW_FLAG_FAILED_TO_GIVE_MILK` gets set if this cow is not in range with the player to interact.
        // In the case of a failure, `R_EPONAS_SONG_PLAYED` is not set to false in case another cow can succeed.
        // On the following frame, if both `R_EPONAS_SONG_PLAYED` and `COW_FLAG_FAILED_TO_GIVE_MILK` are set, the
        // first cow that updates can assume all other cows also failed and can safely unset `R_EPONAS_SONG_PLAYED`. All cows also unset `COW_FLAG_FAILED_TO_GIVE_MILK`
```
the "this" bit as a clarification it's not checking other cows to set the flag

I suggest to add the last remark because it's not obvious `COW_FLAG_FAILED_TO_GIVE_MILK` always also ends up unset, it's in two different code paths, and if it wasn't unset properly it would clear `R_EPONAS_SONG_PLAYED` early the next time

(formatting to be fixed)

fig02: makes sense, pushed change

Dragorn421 (approved):

PR #1496, opened 2023-01-27
Fix deprecated warnings becoming errors for clang 15

philtyl:
https://www9.open-std.org/jtc1/sc22/wg14/www/docs/n2841.htm

Fixes all the following errors:
```
error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
```

### DIFF
--- a/tools/fado/lib/vc_vector/vc_vector.c
+++ b/tools/fado/lib/vc_vector/vc_vector.c
@@ -112,15 +112,15 @@ bool vc_vector_is_equals(vc_vector* vector1, vc_vector* vector2) {
   return memcmp(vector1->data, vector2->data, size_vector1) == 0;
 }
 
-float vc_vector_get_growth_factor() {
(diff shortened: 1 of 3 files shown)


AngheloAlf:
`fado` is a subrepo, you should first PR a fix there and then update the subrepo here
https://github.com/EllipticEllipsis/fado

AngheloAlf:
And most of your changes seems to be in the vc_vector subrepo, so you would need to PR to them first, then update fado then update here
<https://github.com/skogorev/vc_vector>

hensldm:
Looks like Fado has been updated, so this should be a subrepo pull, not just applying the fix manually.

Dragorn421:
Thanks for PRing fixes to vc_vector / fado, I pulled fado in https://github.com/zeldaret/oot/pull/1501 so that obsoletes this PR

PR #1498, opened 2023-02-16
Further documentation for overlay relocation

Thar0:


src/code/relocation.c
```diff
@@ -25,71 +68,74 @@ void Overlay_Relocate(void* allocatedVRamAddress, OverlayRelocationSection* over
...
-                regValP = &luiVals[(*relocDataP >> 0x15) & 0x1F];
+                luiInstRef = luiRefs[MIPS_REG_RS(*relocDataP)];
+                regValP = &luiVals[MIPS_REG_RS(*relocDataP)];
 
                 // Check address is valid for relocation
                 if ((((*regValP << 0x10) + (s16)*relocDataP) & 0x0F000000) == 0) {
-                    relocOffset = ((*regValP << 0x10) + (s16)*relocDataP) - (uintptr_t)vRamStart;
-                    isLoNeg = (((relocOffset + allocu32) & 0x8000) ? 1 : 0);
+                    relocOffset = ((*regValP << 0x10) + (s16)*relocDataP) - (uintptr_t)vramStart;
+                    isLoNeg = (((relocOffset + allocu32) & 0x8000) ? 1 : 0); // adjust for signed immediate
```
hensldm, line 138:
```suggestion
                    isLoNeg = ((relocOffset + allocu32) & 0x8000) ? 1 : 0; // adjust for signed immediate
```
Perhaps just a personal preference.

src/code/relocation.c
```diff
@@ -1,20 +1,63 @@
...
     u32 relocData;
     uintptr_t unrelocatedAddress;
     u32 i;
     u32* relocDataP;
+    //! MIPS ELF relocation does not generally require tracking register values, so at first glance it appears this
+    //! register tracking was an unnecessary complication. However there is a bug in the IDO compiler that can cause
+    //! relocations to be emitted in the wrong order under rare circumstances when the compiler attempts to reuse a
+    //! previous HI16 relocation for a different LO16 relocation as an optimization. This register tracking is likely
+    //! a workaround to prevent improper matching of unrelated HI16 and LO16 relocations that would otherwise arise
+    //! due to the incorrect ordering.
```
hensldm, line 54: Perhaps this comment would be better where the actual tracking takes place in `R_MIPS_HI16`?

src/code/load.c
```diff
@@ -17,14 +17,17 @@ s32 Overlay_Load(uintptr_t vRomStart, uintptr_t vRomEnd, void* vRamStart, void*
...
 
-    DmaMgr_RequestSync(allocatedVRamAddr, vRomStart, size);
+    // DMA the overlay, wait until transfer completes
+    DmaMgr_RequestSync(allocatedRamAddr, vromStart, size);
 
+    // The overlay file is expected to contain a 32-bit offset from the end of the file to the start of the
+    // relocation section.
     ovlOffset = ((s32*)end)[-1];
-
     ovl = (OverlayRelocationSection*)(end - ovlOffset);
```
hensldm, line 29: Perhaps rename these `relocsOffset` and `ovlRelocs` respectively? Seems a little clearer, and also mirrors how you rename `ovlInfo` to `ovlRelocs` in `Overlay_Relocate`.

include/z64.h
```diff
@@ -1165,24 +1165,35 @@ typedef struct ArenaNode {
...
 #define R_MIPS_HI16 5
 #define R_MIPS_LO16 6
 
+/* Reloc section id, must fit in 2 bits otherwise the relocation format must be modified */
+typedef enum {
+    REL_SECTION_NULL,
+    REL_SECTION_TEXT,
+    REL_SECTION_DATA,
+    REL_SECTION_RODATA,
+
```
hensldm, line 1186: Any reason for the newline?

hensldm (commented):
Nice work. A few thoughts.

include/z64.h
EllipticEllipsis: Needs value comments too.

src/code/relocation.c
Thar0: This is also my preference, I just left it as it was before. Changed

src/code/load.c
Thar0: Changed to `relocSectionOffset` and `ovlRelocs`

include/z64.h
Thar0: Fixed

src/code/relocation.c
Thar0: I don't think it's much of an improvement, I think commenting the variables is the most universal place for the comment since it applies to all usages of these variables. If there was no need for hi/lo register matching the arrays wouldn't exist at all.

Dragorn421 (commented):
Is `REL_` a standard prefix? I liked `RELOC_` more, "rel" can be a lot of things (mainly relative, release)

thanks for clearing this up, and pinning down what "Virtual RAM" refers to :)

Thar0:
I changed it to `REL_` for consistency with a fado enum, however I looked through binutils (the source of all knowledge) and it tends to use reloc more, so I've switched back to that

Dragorn421 (approved):

src/code/relocation.c
fig02: will say from an aesthetic pov i am not a fan of a large comment in the middle of variable declarations like this. But I understand why it was done, and will get over my personal gripes with it if others are ok with it

fig02 (approved):

hensldm (approved):

PR #1500, opened 2023-02-18
changed names of functions in z_item_ocarina

Hyena-Senpai:
and the fado makefile thing is there, idk how to do anything about that. also tried to comment each function to state what they were for as well.

### DIFF
--- a/src/overlays/actors/ovl_Item_Ocarina/z_item_ocarina.c
+++ b/src/overlays/actors/ovl_Item_Ocarina/z_item_ocarina.c
@@ -7,6 +7,8 @@
 #include "z_item_ocarina.h"
 #include "assets/scenes/overworld/spot00/spot00_scene.h"
 
+//ocarina object for the zelda escaping with impa cutscene and flashback in light arrow cutscene
+
 #define FLAGS ACTOR_FLAG_4
 
 void ItemOcarina_Init(Actor* thisx, PlayState* play);
@@ -14,12 +16,12 @@ void ItemOcarina_Destroy(Actor* thisx, PlayState* play);
 void ItemOcarina_Update(Actor* thisx, PlayState* play);
 void ItemOcarina_Draw(Actor* thisx, PlayState* play);
 
-void ItemOcarina_GetThrown(ItemOcarina* this, PlayState* play);
-void ItemOcarina_Fly(ItemOcarina* this, PlayState* play);
+void ItemOcarina_Thrown_Escape(ItemOcarina* this, PlayState* play);
+void ItemOcarina_Falling_Escape(ItemOcarina* this, PlayState* play);
 void ItemOcarina_WaitInWater(ItemOcarina* this, PlayState* play);
 void ItemOcarina_StartSoTCutscene(ItemOcarina* this, PlayState* play);
-void func_80B864EC(ItemOcarina* this, PlayState* play);
-void func_80B865E0(ItemOcarina* this, PlayState* play);
+void ItemOcarina_Falling_LACS(ItemOcarina* this, PlayState* play);
+void ItemOcarina_Thrown_LACS(ItemOcarina* this, PlayState* play);
 void ItemOcarina_DoNothing(ItemOcarina* this, PlayState* play);
 
 ActorInit Item_Ocarina_InitVars = {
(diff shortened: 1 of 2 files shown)


src/overlays/actors/ovl_Item_Ocarina/z_item_ocarina.c
```diff
@@ -7,19 +7,21 @@
 #include "z_item_ocarina.h"
 #include "assets/scenes/overworld/spot00/spot00_scene.h"
 
+//ocarina object for the zelda escaping with impa cutscene and flashback in light arrow cutscene
```
Dragorn421, line 10: You can move this above, where `Description: Ocarina of Time` is
(possibly polishing up the punctuation :blush: )

src/overlays/actors/ovl_Item_Ocarina/z_item_ocarina.c
```diff
@@ -7,19 +7,21 @@
...
-void ItemOcarina_GetThrown(ItemOcarina* this, PlayState* play);
-void ItemOcarina_Fly(ItemOcarina* this, PlayState* play);
+void ItemOcarina_Thrown_Escape(ItemOcarina* this, PlayState* play);
+void ItemOcarina_Falling_Escape(ItemOcarina* this, PlayState* play);
 void ItemOcarina_WaitInWater(ItemOcarina* this, PlayState* play);
 void ItemOcarina_StartSoTCutscene(ItemOcarina* this, PlayState* play);
-void func_80B864EC(ItemOcarina* this, PlayState* play);
-void func_80B865E0(ItemOcarina* this, PlayState* play);
+void ItemOcarina_Falling_LACS(ItemOcarina* this, PlayState* play);
+void ItemOcarina_Thrown_LACS(ItemOcarina* this, PlayState* play);
```
Dragorn421, line 24: typically we avoid having several underscores `_` in names

typically we put the "subsystem"/"group" name first, so here it would be `EscapeThrown`/`EscapeFalling`/...

also I'm fine with the LACS abbreviation, though I'd rather we not use abbreviations, at the very least it would be nice to explain what it means somewhere. Possibly as part of the description at the top (like "in the LACS (Light Arrow CutScene)")

src/overlays/actors/ovl_Item_Ocarina/z_item_ocarina.c
```diff
@@ -76,7 +78,8 @@ void ItemOcarina_Init(Actor* thisx, PlayState* play) {
 void ItemOcarina_Destroy(Actor* thisx, PlayState* play) {
 }
 
-void ItemOcarina_Fly(ItemOcarina* this, PlayState* play) {
+//ocarina after being thrown in the escape cutscene and beginning to fall
```
Dragorn421, line 81: Please polish the capitalization/punctuation
Also have a space after the `//` (or run the formatter: `./format.py` at the repo's root)

Dragorn421 (commented):
Hey, thanks for your contribution

src/overlays/actors/ovl_Item_Ocarina/z_item_ocarina.c
mzxrules: I've stated it earlier in discord, but I disagree with referring to the entire set of cutscenes as the "Light Arrow Cutscene". The part where you actually obtain the light arrows is a very small part of the full sequence, and that specific moment takes place in a different scene entirely. Another way to put it, deleting the Temple of Time deletes the true Light Arrow cutscene, but has no effect on the function of this cutscene (until you hit the bad scene swap and crash obv, but that's not important). 

So for all cutscenes I think we should try to use names that relate to only what the one `Cutscene` is doing first, instead of using a common naming the entire sequence of `Cutscene`s. And in this case, I think either ZeldaEscape/ZeldaFlashback, or possibly the simpler Escape/Flashback is the way to go (I think I favor the former though).

I also disagree with you on the LACS abbreviation, Light Arrow isn't common enough to be abbr'd and you end up having to add a comment to explain the abbreviation anyway. Also, specifically with LACS there's an inconsistent naming pattern being used for the two cutscenes (We use Light Arrow Cutscene, but it's fine to leave it as just Escape?).

Dragorn421: what would be a better choice for naming the four functions?

fig02:
Hi
This PR has been waiting for a while with unaddressed feedback. We will be closing this PR in about a week due to inactivity, unless you express intent to work on it in the near future. Thanks!

fig02:
See above. 
Thank you for your interest in contributing, and feel free to re-open this at a later time.

PR #1511, opened 2023-04-09
introduce pausectx enum

playerskel:
Fairly straightforward. Was looking at some code and thought we could use this.

### DIFF
--- a/include/z64pause.h
+++ b/include/z64pause.h
@@ -126,4 +126,27 @@ typedef struct {
 
 typedef PauseMapMarkData PauseMapMarksData[3];
 
+typedef enum {
+    PAUSECTX_0,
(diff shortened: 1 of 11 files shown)


playerskel:
Let me know if it should be renamed (e.g., `PAUSECTXSTATE_` or `PAUSECTX_STATE_`) or if it's not desired.

Dragorn421:
😅 This is already in PR with some docs
https://github.com/zeldaret/oot/pull/1483
I also have a draft PR open that is a nearly 100% doc of the pause menu (merging it bit by bit)

playerskel:
Good stuff, thanks! Will do a pass through PRs someday so I won't duplicate your work in the future, hopefully. Closing :-)

PR #1513, opened 2023-04-30
Fixup assets extraction

Dragorn421:
These came up while getting someone to extract assets on windows

### DIFF
--- a/extract_assets.py
+++ b/extract_assets.py
@@ -6,6 +6,7 @@ import os
 import signal
 import time
 import multiprocessing
+from pathlib import Path
 
 
 EXTRACTED_ASSETS_NAMEFILE = ".extracted-assets.json"
@@ -21,7 +22,13 @@ def ExtractFile(xmlPath, outputPath, outputSourcePath):
         # Don't extract if another file wasn't extracted properly.
         return
 
-    execStr = f"tools/ZAPD/ZAPD.out e -eh -i {xmlPath} -b baserom/ -o {outputPath} -osf {outputSourcePath} -gsf 1 -rconf tools/ZAPDConfigs/MqDbg/Config.xml {ZAPDArgs}"
+    zapdPath = Path("tools") / "ZAPD" / "ZAPD.out"
+    configPath = Path("tools") / "ZAPDConfigs" / "MqDbg" / "Config.xml"
+
+    Path(outputPath).mkdir(parents=True, exist_ok=True)
+    Path(outputSourcePath).mkdir(parents=True, exist_ok=True)
+
+    execStr = f"{zapdPath} e -eh -i {xmlPath} -b baserom -o {outputPath} -osf {outputSourcePath} -gsf 1 -rconf {configPath} {ZAPDArgs}"
 
     if "overlays" in xmlPath:
         execStr += " --static"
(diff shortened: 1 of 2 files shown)


extract_assets.py
```diff
@@ -21,7 +22,13 @@ def ExtractFile(xmlPath, outputPath, outputSourcePath):
         # Don't extract if another file wasn't extracted properly.
         return
 
-    execStr = f"tools/ZAPD/ZAPD.out e -eh -i {xmlPath} -b baserom/ -o {outputPath} -osf {outputSourcePath} -gsf 1 -rconf tools/ZAPDConfigs/MqDbg/Config.xml {ZAPDArgs}"
+    zapdPath = Path("tools") / "ZAPD" / "ZAPD.out"
```
Dragorn421, line 25: This fixes:
image

extract_assets.py
```diff
@@ -21,7 +22,13 @@ def ExtractFile(xmlPath, outputPath, outputSourcePath):
...
+    configPath = Path("tools") / "ZAPDConfigs" / "MqDbg" / "Config.xml"
+
+    Path(outputPath).mkdir(parents=True, exist_ok=True)
+    Path(outputSourcePath).mkdir(parents=True, exist_ok=True)
```
Dragorn421, line 29: This fixes:
image

apparently when creating e.g. assets/code/something without assets/code existing first

IMO this is on ZAPD for not creating the directories recursively but apparently it's hard to do safely in C++ ?

In normal circumstances the folder creation is in the Makefile iirc

tools/msgdis.py
```diff
@@ -391,7 +391,7 @@ def extract_all_text(text_out, staff_text_out):
                 out += "\n#endif"
             out += "\n\n"
 
-        with open(text_out, "w") as outfile:
+        with open(text_out, "w", encoding="utf8") as outfile:
```
Dragorn421, line 394: this addresses:
image

though I thought python defaulted to utf8 regardless of system/environment/... :shrug:

extract_assets.py
```diff
@@ -143,14 +150,21 @@ def main():
...
             print("Extracting assets with " + str(numCores) + " CPU core" + ("s" if numCores > 1 else "") + ".")
-            with multiprocessing.get_context("fork").Pool(numCores,  initializer=initializeWorker, initargs=(mainAbort, args.unaccounted, extractedAssetsTracker, manager)) as p:
+            try:
+                mp_context = multiprocessing.get_context("fork")
+            except ValueError as e:
+                raise CannotMultiprocessError() from e
+            with mp_context.Pool(numCores,  initializer=initializeWorker, initargs=(mainAbort, args.unaccounted, extractedAssetsTracker, manager)) as p:
                 p.map(ExtractFunc, xmlFiles)
-        except (multiprocessing.ProcessError, TypeError):
+        except (multiprocessing.ProcessError, TypeError, CannotMultiprocessError):
```
Dragorn421, line 167: windows thing idk, this fixes:
image

AngheloAlf (approved):
Since you are introducing pathlib I wonder if it would be better to replace the other path stuff to pathlib too, like all the `os.path` stuff and `os.walk`, but is fine either way

fig02 (approved):

Dragorn421:
> Since you are introducing pathlib I wonder if it would be better to replace the other path stuff to pathlib too, like all the `os.path` stuff and `os.walk`, but is fine either way

I started and it would end up rewriting most of the script, paths are passed around a lot
Since extract_assets.py is gone in the assets system rewrite I'm planning I don't want to put "that much" effort into it 8)

hensldm (approved):
LGTM

Roman971 (approved):

PR #1517, opened 2023-07-08
Prefer clang-format-14

Dragorn421:
We were staying with clang-format-11 because there were formatting behavior changes from some version and it was annoying to deal with everyone having different versions
The same situation still applies but clang-format-11 is pretty old at this point
 
Availability of `clang-format-*` in the Ubuntu packages:
image
"x" is "in  ubuntu's apt package lists"
https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=clang-format-14&searchon=names
-> clang-format-14 sounds reasonable

I haven't checked if clang-format-14 can be installed on ubuntu 20.04 using https://apt.llvm.org/ , I'd guess yes (also this exists https://apt.llvm.org/focal/dists/llvm-toolchain-focal-14/ so it makes me think yes but idk). Any testing would be nice

------

MM, Fado and ZAPD are also aligned on clang-format-11 so it would be nice to coordinate

Guess I'm opening this to spearhead the transition

### DIFF
--- a/format.py
+++ b/format.py
@@ -14,8 +14,8 @@ from typing import List
 
 
 # clang-format, clang-tidy and clang-apply-replacements default version
-# Version 11 is used when available for more consistency between contributors
-CLANG_VER = 11
+# This specific version is used when available, for more consistency between contributors
+CLANG_VER = 14
 
 # Clang-Format options (see .clang-format for rules applied)
 FORMAT_OPTS = "-i -style=file"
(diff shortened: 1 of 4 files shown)


fig02 (approved):
Just approving to say im fine with upgrading to 14. Idk if you want to wait for everything else to agree to an update before merging

Dragorn421:
As Anghelo put it in the MM companion PR I would prefer we wait for consensus/readiness to merge yeah

hensldm (approved):
Just want to officially express my approval as well (even though it was kind of implied by opening the MM companion PR :) )

Dragorn421:
Similar changes have been merged in zapd, fado and mm

We can green the press button

PR #1518, opened 2023-07-28
Mapfile Parser

hensldm:
Same as https://github.com/zeldaret/mm/pull/1342

the python package mapfile_parser handles both older and newer ld (>=2.40 is newer) so integrated that into the wrappers for first_diff.py and sym_info.py so they work.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone &
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
-RUN python3 -m pip install --user colorama ansiwrap attrs watchdog python-Levenshtein mapfile-parser
+RUN python3 -m pip install --user colorama ansiwrap attrs watchdog python-Levenshtein "mapfile-parser>=1.2.1,<2.0.0" "rabbitizer>=1.0.0,<2.0.0"
 RUN python3 -m pip install --upgrade attrs pycparser
 
 ENV LANG C.UTF-8
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone &
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
-RUN python3 -m pip install --user colorama ansiwrap attrs watchdog python-Levenshtein
+RUN python3 -m pip install --user colorama ansiwrap attrs watchdog python-Levenshtein mapfile-parser
 RUN python3 -m pip install --upgrade attrs pycparser
 
 ENV LANG C.UTF-8
--- a/first_diff.py
+++ b/first_diff.py
@@ -6,13 +6,13 @@ from pathlib import Path
 try:
     import rabbitizer
 except ImportError:
-    print("Missing dependency rabbitizer, install it with `python3 -m pip install rabbitizer`")
+    print("Missing dependency rabbitizer, install it with `python3 -m pip install rabbitizer>=1.0.0,<2.0.0`")
     exit(1)
 
 try:
     import mapfile_parser
 except ImportError:
-    print("Missing dependency mapfile_parser, install it with `python3 -m pip install mapfile-parser`")
+    print("Missing dependency mapfile_parser, install it with `python3 -m pip install mapfile-parser>=1.2.1,<2.0.0`")
     exit(1)
 
 
--- a/first_diff.py
+++ b/first_diff.py
@@ -6,13 +6,13 @@ from pathlib import Path
 try:
     import rabbitizer
 except ImportError:
-    print("Missing dependency rabbitizer, install it with `python3 -m pip install rabbitizer>=1.0.0,<2.0.0`")
+    print("Missing dependency rabbitizer, install it with `python3 -m pip install 'rabbitizer>=1.0.0,<2.0.0'`")
     exit(1)
 
 try:
     import mapfile_parser
 except ImportError:
-    print("Missing dependency mapfile_parser, install it with `python3 -m pip install mapfile-parser>=1.2.1,<2.0.0`")
+    print("Missing dependency mapfile_parser, install it with `python3 -m pip install 'mapfile-parser>=1.2.1,<2.0.0'`")
     exit(1)
 
 
--- a/first_diff.py
+++ b/first_diff.py
@@ -1,10 +1,13 @@
 #!/usr/bin/env python3
 
-from __future__ import annotations
-
 import argparse
 from pathlib import Path
-import rabbitizer
+
+try:
+    import rabbitizer
+except ImportError:
+    print("Missing dependency rabbitizer, install it with `python3 -m pip install rabbitizer`")
+    exit(1)
 
 try:
     import mapfile_parser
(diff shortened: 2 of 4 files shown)


sym_info.py
```diff
@@ -1,139 +1,31 @@
...
+    parser.add_argument("-e", "--expected", dest="use_expected", action="store_true", help="use the map file in expected/build/ instead of build/")
 
-            prev_ram = ram
-            prev_rom = rom
-            prev_sym = sym
-            prev_file = cur_file
+    args = parser.parse_args()
 
-    return "at end of rom?"
+    BUILTMAP = Path(f"build") / f"mm.map"
```
AngheloAlf, line 22: Found a copy-paste error

tools/assist.py
```diff
@@ -162,7 +160,7 @@ def do_query(query):
 args = parser.parse_args()
 
 rom_bytes = read_rom()
-map_syms = parse_map(build_dir + "z64.map")
+map_syms = parse_map(build_dir + "mm.map")
```
AngheloAlf, line 163: Another one lol

tools/assist.py
```diff
@@ -1,19 +1,17 @@
...
 
 script_dir = os.path.dirname(os.path.realpath(__file__))
 root_dir = script_dir + "/../"
-asm_dir = root_dir + "asm/non_matchings/overlays/actors"
+asm_dir = root_dir + "asm/non_matchings/overlays"
 build_dir = root_dir + "build/"
 
 def read_rom():
-    with open("baserom.z64", "rb") as f:
+    with open("baserom_uncompressed.z64", "rb") as f:
```
Dragorn421, line 14: A third one

tools/assist.py
```diff
@@ -1,19 +1,17 @@
...
-import re
-import pickle
-import sys
+
+gAddressWidth = 18 # if your ld >= 2.40 change this to 10
 
 script_dir = os.path.dirname(os.path.realpath(__file__))
 root_dir = script_dir + "/../"
-asm_dir = root_dir + "asm/non_matchings/overlays/actors"
+asm_dir = root_dir + "asm/non_matchings/overlays"
```
Dragorn421, line 10: Why?

(btw I'm not even sure why this script is being updated?)

tools/assist.py
```diff
@@ -1,19 +1,17 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 
-import argparse
+import argparse, os, sys
```
Dragorn421, line 3: It is better style to import different modules on different lines

Why? Because it was said so
https://peps.python.org/pep-0008/#imports
(there are various reasons but style consistency and sticking to a standard beats everything)

first_diff.py
```diff
@@ -1,255 +1,55 @@
 #!/usr/bin/env python3
 
-import os.path
+from __future__ import annotations
```
Dragorn421, line 3: I'm not sure why this is here

It changes the behavior of annotations but I don't think you care here
(And if I missed something and you do I'd suggest just using string type hints instead)
https://stackoverflow.com/questions/61544854/from-future-import-annotations

first_diff.py
```diff
@@ -1,255 +1,55 @@
...
-myimg = f"zelda_ocarina_mq_dbg.z64"
-mymap = f"build/z64.map"
-
-if not os.path.isfile(baseimg):
-    print(f"{baseimg} must exist.")
-    exit(1)
-if not os.path.isfile(myimg) or not os.path.isfile(mymap):
-    print(f"{myimg} and {mymap} must exist.")
+from pathlib import Path
+import rabbitizer
```
Dragorn421, line 7: This would be nice also wrapped in a try except ImportError, rabbitizer isn't a dependency for building oot

first_diff.py
```diff
@@ -1,255 +1,55 @@
...
+
+    parser.add_argument("-c", "--count", type=int, default=5, help="find up to this many instruction difference(s)")
+    parser.add_argument("-v", "--version", help="Which version should be processed", default="mq_dbg")
+    parser.add_argument("-a", "--add-colons", action='store_true', help="Add colon between bytes" )
+
+    args = parser.parse_args()
+
+    buildFolder = Path("build")
+
+    BUILTROM = Path(f"zelda_ocarina_{args.version}.z64")
```
Dragorn421, line 46: This is GoingToChange% but ok

sym_info.py
```diff
@@ -1,139 +1,31 @@
...
-            prev_sym = sym
-            prev_file = cur_file
+    args = parser.parse_args()
 
-    return "at end of rom?"
+    BUILTMAP = Path(f"build") / f"z64.map"
 
+    mapPath = BUILTMAP
+    if args.use_expected:
+        mapPath = "expected" / BUILTMAP
```
Dragorn421, line 26: This does work? (str / Path())

first_diff.py
```diff
@@ -1,255 +1,55 @@
...
+    BUILTMAP = buildFolder / f"z64.map"
+
+    EXPECTEDROM = Path("baserom.z64")
+    EXPECTEDMAP = "expected" / BUILTMAP
+
+    exit(mapfile_parser.frontends.first_diff.doFirstDiff(BUILTMAP, EXPECTEDMAP, BUILTROM, EXPECTEDROM, args.count, mismatchSize=True, addColons=args.add_colons, bytesConverterCallback=decodeInstruction))
```
Dragorn421, line 52: Is calling exit() necessary
(I mostly ask because it's not done in sym_info)

tools/assist.py
hensldm: It is because it also parses the mapfile and was broken (though not many people even use this script which is why I put pretty low effort in "fixing" it).

sym_info.py
hensldm: Yeah, it is the same in first_diff.py

first_diff.py
hensldm: No, I'll remove it.

Dragorn421 (commented):
One tiny thing I didn't think of earlier

I'll test locally then approve

Dockerfile
```diff
@@ -18,7 +18,7 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone &
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
-RUN python3 -m pip install --user colorama ansiwrap attrs watchdog python-Levenshtein
+RUN python3 -m pip install --user colorama ansiwrap attrs watchdog python-Levenshtein mapfile-parser
```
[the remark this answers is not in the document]
Dragorn421: Btw I think in MM's requirements.txt these are versioned (rabbitizer and mapfile-parser), can/should they be versioned here as well

first_diff.py
```diff
@@ -1,255 +1,58 @@
...
-mymap = f"build/z64.map"
-
-if not os.path.isfile(baseimg):
-    print(f"{baseimg} must exist.")
+from pathlib import Path
+
+try:
+    import rabbitizer
+except ImportError:
+    print("Missing dependency rabbitizer, install it with `python3 -m pip install rabbitizer>=1.0.0,<2.0.0`")
```
Dragorn421, line 9:
```suggestion
    print("Missing dependency rabbitizer, install it with `python3 -m pip install 'rabbitizer>=1.0.0,<2.0.0'`")
```
and similar with the others (bash sees the <> as special shell characters)

Dragorn421 (approved):

fig02 (approved):

AngheloAlf (approved):

PR #1520, opened 2023-07-31
Mention why PrimDepth is used in `PreRender_FetchFbufCoverage`

Thar0:
Crazy undocumented blender hardware behavior

### DIFF
--- a/src/code/PreRender.c
+++ b/src/code/PreRender.c
@@ -364,7 +364,7 @@ void PreRender_FetchFbufCoverage(PreRender* this, Gfx** gfxp) {
     // Set the blend color to full white and set maximum depth.
     // It is important that at least dz is set to full here as the blender will shift memory alpha values based on the
     // value of dz even if depth compare is disabled. If per-pixel depth was enabled, fill rectangle always uses 0 z/dz,
-    // resulting in a maximal shift of 4 to be applied to the coverage value. Full dz results in no shift, which is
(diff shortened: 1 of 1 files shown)


Dragorn421 (commented):
angrylion somehow knows about this but only for depth compare enabled
https://github.com/ata4/angrylion-rdp-plus/blob/c24268bddd44bcbb7f7d7f91b51dd30a6a92566b/src/core/n64video/rdp/zbuffer.c#L240-L252

it would be clearer to mention the coverage shift applies on the 8-bits value blender input, since with coverage you typically think of the 3-bits value

you could actually write a whole blog article on this, which makes me wonder if it's that useful to point out here without external references or additional details?

Thar0:
I don't think additional details are especially needed, I only want to explain this particular usage of SetPrimDepth and I think the comment does this adequately. The only sources I can link are implementations in accurate graphics plugins since nobody has written about this otherwise. I'm happy to do so if it would be useful.

Dragorn421:
To be sure, your comment goes against angrylion's code about "doesn't matter if depth compare", so angrylion has this wrong?

Thar0:
The angrylion source implements this here
https://github.com/ata4/angrylion-rdp-plus/blob/c24268bddd44bcbb7f7d7f91b51dd30a6a92566b/src/core/n64video/rdp/zbuffer.c#L349-L365

fig02 (approved):

Dragorn421 (approved):

engineer124 (approved):

PR #1522, opened 2023-08-17
attempt to restructure banks

owlisnotacat1:


owlisnotacat1:
did not mean to pr this to main

PR #1523, opened 2023-08-19
Start Player Item Docs

fig02:
This lays the groundwork for starting to document player items.

Following the process for changing/putting away items is probably one of the most confusing things ive tried to figure out in this game. I wrote a pastebin essay walking through the whole process using the names in this pr: https://pastebin.com/hRB3PjYb

I've further documented the rest of the ItemActionFuncs and related things, but broke this off into a new PR for the sake of easier reviewing.


fig02:
Using cool new gdb stuff, can show a call stack for how all the named functions in this pr interact. This shows the process of initially taking out sword:
image

Then later, after the item change process has completed, the sword is ready to be swung:
image

(note `Player_HandleSubAction` and `Player_Action_Idle` are tentative names for the sake of demonstration and are not in this PR)

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -2081,31 +2107,31 @@ LinkAnimationHeader* func_808335F4(Player* this) {
...
+void Player_InitItemActionWithAnim(PlayState* play, Player* this, s8 itemAction) {
     LinkAnimationHeader* current = this->skelAnime.animation;
     LinkAnimationHeader** iter = D_80853914 + this->modelAnimType;
-    u32 i;
+    u32 animGroup;
 
     this->stateFlags1 &= ~(PLAYER_STATE1_3 | PLAYER_STATE1_24);
 
-    for (i = 0; i < PLAYER_ANIMGROUP_MAX; i++) {
+    for (animGroup = PLAYER_ANIMGROUP_wait; animGroup < PLAYER_ANIMGROUP_MAX; animGroup++) {
```
AngheloAlf, line 2131: I think it is better to keep this as 0 instead of the first element of the enum. With 0 it is clear this loop is iterating from the first entry to the last one

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -2307,114 +2333,131 @@ s32 func_80833CDC(PlayState* play, s32 index) {
...
     s32 i;
 
     if (this->currentMask != PLAYER_MASK_NONE) {
         maskItemAction = this->currentMask - 1 + PLAYER_IA_MASK_KEATON;
-        if (!func_80833C98(C_BTN_ITEM(0), maskItemAction) && !func_80833C98(C_BTN_ITEM(1), maskItemAction) &&
-            !func_80833C98(C_BTN_ITEM(2), maskItemAction)) {
+
+        if (!Player_ItemIsItemAction(C_BTN_ITEM(0), maskItemAction) &&
+            !Player_ItemIsItemAction(C_BTN_ITEM(1), maskItemAction) &&
+            !Player_ItemIsItemAction(C_BTN_ITEM(2), maskItemAction)) {
```
AngheloAlf, line 2364: So having a mask on B (without having a mask on a C button) would not allow you to use it? Interesting

fig02: I had no idea, but yes, just tested this and its true
(ofc, masks on B is only achievable through glitches)

src/overlays/actors/ovl_player_actor/z_player.c
fig02: I suppose. I prefer the enum, but as long as we keep the name as `animGroup` instead of `i` im happy. The value gets used later outside the loop, and naming it `i` makes that less obvious

AngheloAlf: `animGroup` is fine, yeah

AngheloAlf (approved):
Always nice to see progress on player :rocket:

src/overlays/actors/ovl_player_actor/z_player.c
fig02: dont know whats happening here. I tested in 1.0 as child (since masks as adult crashes emu). it shouldnt be a version difference cause decomp is based on mq which is much after 1.2

fig02: I believe in that video he is pressing mask and then talking to gossip stone before the game has a chance to take the mask off. you have a frame or two before it does so.

Yanis42: ahh I see, so this is why it's removing the mask instantly, I thought it was preventing the mask from being used at all, like no frames of equipped then removing it (if that makes sense)

I tried on 1.2 and it's working like it should be 👍

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -353,7 +353,30 @@ static Input* sControlInput;
 
 // .data
 
-static u8 D_80853410[] = { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 };
+static u8 sUpperBodyLimbCopyMap[] = {
```
engineer124, line 356:
```suggestion
static u8 sUpperBodyLimbCopyMap[PLAYER_LIMB_MAX] = {
```
Might as well?

include/z64player.h
```diff
@@ -688,11 +688,11 @@ typedef struct Player {
...
     /* 0x06C2 */ s16        unk_6C2;
     /* 0x06C4 */ f32        unk_6C4;
-    /* 0x06C8 */ SkelAnime  skelAnime2;
-    /* 0x070C */ Vec3s      jointTable2[PLAYER_LIMB_BUF_COUNT];
-    /* 0x079C */ Vec3s      morphTable2[PLAYER_LIMB_BUF_COUNT];
-    /* 0x082C */ PlayerFunc82C func_82C;
-    /* 0x0830 */ f32        unk_830;
+    /* 0x06C8 */ SkelAnime  skelAnimeUpper;
+    /* 0x070C */ Vec3s      jointTableUpper[PLAYER_LIMB_BUF_COUNT];
+    /* 0x079C */ Vec3s      morphTableUpper[PLAYER_LIMB_BUF_COUNT];
```
engineer124, line 693: Stalfos (test.c) has 
```c
/* 0x4A8 */ SkelAnime upperSkelanime;
/* 0x4EC */ Vec3s upperJointTable[STALFOS_LIMB_MAX];
/* 0x65A */ Vec3s upperMorphTable[STALFOS_LIMB_MAX];
```
Which puts the upper before the names instead of after, it should probably be consistent (also noting Skelanime -> SkelAnime in test.c but can be done in a later cleanup)

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -479,8 +502,8 @@ static s32 sPrevFloorProperty = FLOOR_PROPERTY_0; // floor property from the pre
 static s32 sShapeYawToTouchedWall = 0;
 static s32 sWorldYawToTouchedWall = 0;
 static s16 sFloorShapePitch = 0;
-static s32 D_80853614 = 0;
-static s32 D_80853618 = 0;
+static s32 sUseHeldItem = 0;          // When true, the current held item is used. Is reset to false every frame.
+static s32 sHeldItemButtonIsHeld = 0; // Indicates if the controller button for the current held item is held down.
```
engineer124, line 506:
```suggestion
static s32 sUseHeldItem = false;          // When true, the current held item is used. Is reset to false every frame.
static s32 sHeldItemButtonIsHeld = false; // Indicates if the controller button for the current held item is held down.
```

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -9634,7 +9691,7 @@ static EffectBlureInit2 D_8085470C = {
     0, 2, 0, { 0, 0, 0, 0 },         { 0, 0, 0, 0 },
 };
 
-static Vec3s D_80854730 = { -57, 3377, 0 };
+static Vec3s sSekeltonBaseTransl = { -57, 3377, 0 };
```
engineer124, line 9694:
```suggestion
static Vec3s sSkeletonBaseTransl = { -57, 3377, 0 };
```

include/z64player.h
fig02: I went with this mainly because I want to see `skelAnime` right away in the name when its passed around to various animation functions. Do agree stalfos should be consistent but would rather that be done in a separate stalfos pass

src/overlays/actors/ovl_player_actor/z_player.c
fig02: I also changed `sHeldItemButtonIsHeld` -> `sHeldItemButtonIsHeldDown`
mainly because the double meaning of held here broke my brain for a second lol. hopefully this makes it more clear that the first held is for held-item and the second one means the button is physically held down

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -2316,114 +2341,131 @@ s32 func_80833CDC(PlayState* play, s32 index) {
...
+ *
+ * Tasks include:
+ *    - Put away a mask if it is not present on any C button
+ *    - Put away an item if it is not present on the B button or any C button
+ *    - Use an item on the B button or any C button if the corresponding button is pressed
+ *    - Keep track of the curent item button being held down
+ *
+ * Note: The item put-away/take-out process is not handled here.
+ */
+void Player_ItemHandler(Player* this, PlayState* play) {
```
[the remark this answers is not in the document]
fig02: I suppose. I think I was going for something less direct and more administrative, if that makes sense. since this doesnt really do any work for the item changing process itself. but it facilitates the whole button interaction thing and calls `UseItem` when it makes sense to (and puts items away if they leave c buttons). 

Maybe theres a better name for this in general, but I cant think of it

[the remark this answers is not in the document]
fig02: My original ideas for the name were oriented around "updating ItemButtons", but I moved away from it because this function is directly responsible for actually using the item as well. I think `UpdateItemButtons` caters more to the "remove item if its not on a button" aspect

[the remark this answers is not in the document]
fig02: my only other idea is `ProcessItemButtons` - its vague enough to cover both the case of using an item and putting it away if its not present on a button anymore, and even the `sHeldItemButtonIsHeldDown` variable

I think this is prob best tbh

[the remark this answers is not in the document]
fig02: Okay yeah i changed it and really like how this looks

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -2284,23 +2309,23 @@ void func_80833C3C(Player* this) {
     this->unk_870 = this->unk_874 = 0.0f;
 }
 
-s32 func_80833C50(Player* this, s32 item) {
+s32 Player_ItemIsInUse(Player* this, s32 item) {
     if ((item < ITEM_NONE_FE) && (Player_ItemToItemAction(item) == this->itemAction)) {
         return 1;
     } else {
         return 0;
```
engineer124, line 2316: Could make `Player_ItemIsInUse` and `Player_ItemIsItemAction` return `true`/`false`

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -3263,14 +3316,15 @@ void func_80836448(PlayState* play, Player* this, LinkAnimationHeader* anim) {
...
 }
 
-s32 func_808365C8(Player* this) {
+s32 Player_CanUpdateItems(Player* this) {
     return (!(Player_Action_808458D0 == this->actionFunc) ||
-            ((this->stateFlags1 & PLAYER_STATE1_8) &&
+            ((this->stateFlags1 & PLAYER_STATE1_START_CHANGING_HELD_ITEM) &&
              ((this->heldItemId == ITEM_LAST_USED) || (this->heldItemId == ITEM_NONE)))) &&
-           (!(func_80834A2C == this->func_82C) || (Player_ItemToItemAction(this->heldItemId) == this->heldItemAction));
+           (!(Player_IA_ChangeHeldItem == this->itemActionFunc) ||
```
engineer124, line 3323:
```suggestion
           ((Player_IA_ChangeHeldItem != this->itemActionFunc) ||
```
Does this match?

engineer124 (approved):

src/overlays/actors/ovl_player_actor/z_player.c
fig02: doesnt seem to 
image

fig02:
There is a chance that after more docs `ItemActionFunc` will make more sense as something like `UpperActionFunc`, pertaining more to the upper body. Since that is what the return value mainly orients around. And there are a couple special cases of `ItemActionFunc`s that arent directly tied to the `heldItemAction` value. 

Engineer also went with upper action in his wip mm docs: https://github.com/engineer124/oot/blob/71fec02129570c6200cc9291ac4826cf78d82aa5/include/z64player.h#L797

I think we can merge this now and make that call later when the functions themselves are documented

PR #1525, opened 2023-08-21
Document Camera Special9 (door) angle bug

fig02:
this also changes the name of `adjustedPlayerPosRot`, since it can be either the door's PosRot or player's PosRot
example video of the bug: https://clips.twitch.tv/ZanySillyRatWutFace-hNOvBsBb5Bg8R807

### DIFF
--- a/src/code/z_camera.c
+++ b/src/code/z_camera.c
@@ -6920,7 +6920,7 @@ s32 Camera_Special9(Camera* camera) {
             camera->stateFlags &= ~(CAM_STATE_1 | CAM_STATE_2);
             camera->animState++;
 
-            //! @bug There should be a cast to s16 in the condition of the ternary below.
+            //! @bug The angle passed to ABS should be cast to s16.
             //! The lack of a cast means that a door which has an angle of around 0x8000 will calculate an incorrect
             //! angle and use an incorrect `targetYaw`.
             //! In-game, this means if the player opens a susceptible door at a slanted angle, the camera will not
--- a/src/code/z_camera.c
+++ b/src/code/z_camera.c
@@ -6876,7 +6876,7 @@ s32 Camera_Special9(Camera* camera) {
     f32 playerYOffset;
     s32 pad3;
     PosRot* playerPosRot = &camera->playerPosRot;
-    PosRot adjustedPlayerPosRot;
+    PosRot referencePosRot;
     f32 yNormal;
     DoorParams* doorParams = &camera->paramData.doorParams;
     Special9ReadOnlyData* roData = &camera->paramData.spec9.roData;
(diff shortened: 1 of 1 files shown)


src/code/z_camera.c
```diff
@@ -6919,9 +6919,15 @@ s32 Camera_Special9(Camera* camera) {
         case 0:
             camera->stateFlags &= ~(CAM_STATE_1 | CAM_STATE_2);
             camera->animState++;
-            rwData->targetYaw = ABS(playerPosRot->rot.y - adjustedPlayerPosRot.rot.y) >= 0x4000
-                                    ? adjustedPlayerPosRot.rot.y - 0x7FFF
-                                    : adjustedPlayerPosRot.rot.y;
+
+            //! @bug There should be a cast to s16 in the condition of the ternary below.
```
mzxrules, line 6923: so you're saying the result of `ABS()` should be cast to s16?

fig02: the angle subtraction inside of the ABS. like usual subtraction on binary angles always needs to be casted

AngheloAlf: I agree the wording doesn't make clear where the cast is missing.

I originally thought on something like this (casting the result of the operation)
```c
            rwData->targetYaw = (s16)(ABS(playerPosRot->rot.y - referencePosRot.rot.y) >= 0x4000
                                    ? referencePosRot.rot.y - 0x7FFF
                                    : referencePosRot.rot.y);
```

But the cast should be inside the ABS?
```c
            rwData->targetYaw = ABS((s16)(playerPosRot->rot.y - referencePosRot.rot.y)) >= 0x4000
                                    ? referencePosRot.rot.y - 0x7FFF
                                    : referencePosRot.rot.y;
```

fig02: Yes, on the subtraction itself. This is just a fact of subtracting binary angles in general that its needed, not specific to this particular case. Which is why I didnt go into very much detail here

AngheloAlf: The other subtraction (`referencePosRot.rot.y - 0x7FFF`) made me think the whole result needed to be casted. Idk, it may be just be me. Either way the pr seems fine

fig02: that one gets casted implicitly on assignment to `targetYaw`, the one in the condition does not.
I can change the wording, but just making it clear that this is a general requirement and not something weird to this case

Dragorn421:
```suggestion
            //! @bug The angle passed to ABS should be cast to s16.
```
?

fig02: sure

Thar0 (approved):

hensldm (approved):

PR #1526, opened 2023-08-24
Document niche magic spell bug

fig02:
This bug is extremely complex and very niche. To manifest, it requires another glitch to be active as a prerequisite. Even still, the code that manages the magic spell interaction could have prevented this, but it chose to ignore a return value that is intended for error handling.

Long write up of the bug: https://pastebin.com/6QTqzQEi
Example video: 

https://github.com/zeldaret/oot/assets/22045047/5f3938f7-9ef8-4858-b1e7-2df17a3cc2d0


src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -5153,6 +5153,12 @@ void func_8083AF44(PlayState* play, Player* this, s32 magicSpell) {
     func_80835DE4(play, this, func_808507F4, 0);
 
     this->unk_84F = magicSpell - 3;
+
+    //! @bug `MAGIC_CONSUME_WAIT_PREVIEW` is not guaranteed to succeed. Ideally, the return value of
+    //! `Magic_RequestChange` should be checked before allowing the process of using a spell to continue.
+    //! If the magic state change request fails, `magicTarget` will never be set correctly.
```
Thar0, line 5159:
```suggestion
    //! If the magic state change request fails, `gSaveContext.magicTarget` will never be set correctly.
```
possibly?

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -5153,6 +5153,12 @@ void func_8083AF44(PlayState* play, Player* this, s32 magicSpell) {
...
+    //! When `MAGIC_STATE_CONSUME_SETUP` is set in `func_808507F4`, magic will eventually be consumed to a stale target
+    //! value. If the stale target value is higher than the current magic value, it will be drained to zero.
```
Thar0, line 5161: and if it's lower it will just drain to the stale target? (or what really happens in that case?)

Thar0 (commented):
Two things

src/overlays/actors/ovl_player_actor/z_player.c
fig02: I suppose, I feel like its just code stuff that doesnt add anything to the explanation in this case, when im more interested in conveying the meaning here

src/overlays/actors/ovl_player_actor/z_player.c
fig02: It will just drain to target yeah. I only mentioned the higher than case cause its less intuitive to know what it will do

src/overlays/actors/ovl_player_actor/z_player.c
Thar0: It just feels easier to read through this way, rather than wondering if there's some local variable named `magicTarget` that you should be in the know about

src/overlays/actors/ovl_player_actor/z_player.c
Thar0: I think it's nice to add it for completeness so the curious reader won't have to dredge through this stuff again to find the alternative outcome

fig02: `magic will eventually be consumed to a stale target value` before that sentence is meant to convey this. but if you have a suggestion for rewording, lmk

Thar0: I think my confusion stemmed from one using "consumed" and the other "drained" maybe, I think making the wording consistent between both should clear it up

fig02: sure, changed

Thar0 (approved):

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -5184,6 +5184,12 @@ void func_8083AF44(PlayState* play, Player* this, s32 magicSpell) {
     func_80835DE4(play, this, Player_Action_808507F4, 0);
 
     this->actionVar1 = magicSpell - 3;
+
+    //! @bug `MAGIC_CONSUME_WAIT_PREVIEW` is not guaranteed to succeed. Ideally, the return value of
+    //! `Magic_RequestChange` should be checked before allowing the process of using a spell to continue.
+    //! If the magic state change request fails, `gSaveContext.magicTarget` will never be set correctly.
+    //! When `MAGIC_STATE_CONSUME_SETUP` is set in `func_808507F4`, magic will eventually be consumed to a stale target
```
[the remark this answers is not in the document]
fig02: done

Dragorn421 (approved):

PR #1528, opened 2023-08-28
Emulator related changes in the makefile

fig02:
- allow the user to specify their own EMULATOR and EMU_FLAGS environment variables. We can keep mupen as the default I guess, but these days I dont imagine many people are using it

- change "test" to "run". This just makes more sense to me, it runs the ROM that was just built. Not everything is necessarily a "test". And small bonus its 1 character shorter lol


Makefile
```diff
@@ -95,8 +95,8 @@ AS         := $(MIPS_BINUTILS_PREFIX)as
 LD         := $(MIPS_BINUTILS_PREFIX)ld
 OBJCOPY    := $(MIPS_BINUTILS_PREFIX)objcopy
 OBJDUMP    := $(MIPS_BINUTILS_PREFIX)objdump
-EMULATOR = mupen64plus
-EMU_FLAGS = --noosd
+EMULATOR ?= 
+EMU_FLAGS ?=
```
AngheloAlf, line 99: Could you add spaces so this lines up with the other variables?

fig02: done

Makefile
```diff
@@ -269,11 +269,11 @@ setup:
 	python3 extract_baserom.py
 	python3 extract_assets.py -j$(N_THREADS)
 
-test: $(ROM)
+run: $(ROM)
```
Thar0, line 272:
```suggestion
run: $(ROM)
ifeq ($(EMULATOR),)
	$(error Emulator executable not set. Either change EMULATOR in the Makefile or define it as an environment variable to point to an emulator executable.)
endif
```
It would be nice to give a clear error if `EMULATOR` was not set to anything. I'm not happy about how the suggestion reads but it shows how to do it

fig02: "Emulator not found. Set EMULATOR in the Makefile or define it as an environment variable."
short and sweet, gets the same info across?

Thar0: "Emulator path not set. Set EMULATOR in the Makefile or define it as an environment variable."

since this error doesn't cover the case where a path is set but is invalid

fig02: done. i also tested to make sure it works. had to remove the period at the end since apparently "error" adds one on its own
image

Thar0 (approved):

PR #1529, opened 2023-08-28
Fix and tidy up missing_gcc_functions.c

Thar0:
- fixed broken 64-bit integer <-> float conversion routines
- reworked `__powisf2` to avoid `while (1)`
- used assembler macros for 64-bit div and mod routines (e.g. `ddivu $v0, $t6, $t7` instead of `ddivu $zero, $t6, $t7` + manual breaks + mflo)
- tidied up asm formatting
- expanded on comments

### DIFF
--- a/src/gcc_fix/missing_gcc_functions.c
+++ b/src/gcc_fix/missing_gcc_functions.c
@@ -7,7 +7,9 @@
 
 #include "global.h"
 
-// Self-hosted memcmp.
+// Self-hosted libc memory functions, gcc assumes these exist even in a freestanding
+// environment and there is no way to tell it otherwise.
+
 int memcmp(void* s1, const void* s2, size_t n) {
     u8* m1 = (u8*)s1;
     u8* m2 = (u8*)s2;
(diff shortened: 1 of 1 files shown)


fig02 (approved):

src/gcc_fix/missing_gcc_functions.c
```diff
@@ -35,177 +37,212 @@ void* memset(void* str, s32 c, size_t n) {
...
+    sw      $a0, 0x0($sp)                   \n\
+    sw      $a1, 0x4($sp)                   \n\
+    sw      $a2, 0x8($sp)                   \n\
+    sw      $a3, 0xC($sp)                   \n\
+    ld      $t6, 0($sp)                     \n\
+    ld      $t7, 8($sp)                     \n\
+    dremu   $v0, $t6, $t7                   \n\
+    dsll32  $v1, $v0, 0                     \n\
+    dsra32  $v1, $v1, 0                     \n\
+    jr      $ra                             \n\
```
AngheloAlf, line 188: Since you are using inline asm for the other functions, wouldn't it be better to rewrite this functions as C functions with inline asm too? or it isn't worth it?

Thar0: Explained in discord but for visibility: these functions aren't worth rewriting in inline asm due to the o32 abi passing values in 2 registers. The inputs/outputs for inline asm aren't really possible to use in this case and it becomes about as much manual work as these pure-asm functions.

AngheloAlf (approved):

hensldm (approved):

PR #1531, opened 2023-09-06
Player Docs: Name actionFunc

fig02:
While the diff for this PR is large, this pr accomplishes only 3 things:
- rename `PlayerFunc674 func_674` to `PlayerActionFunc actionFunc`
- rename all action functions to be of the form `Player_Action_<addr>`, real names to be given as documentation progresses
- group all the action function declarations together at the top of the file

I am confident enough in my knowledge of players function callbacks to say that this is the `actionFunc`. Player additionally has `itemActionFunc` which is renamed in https://github.com/zeldaret/oot/pull/1523 , and a sub-action function system handled by `D_80854448`, which essentially allows other actions to run within "main" actions. The names in this system are tbd in future docs, and are not touched by this PR.

Prefixing action functions with `_Action_` is not a thing we normally do in actors, but with how complex player is I feel it is warranted. It is helpful to know at a glance whether a function is an action function. In some rare cases these functions do unfortunately get called directly without the callback system. So it is helpful in those cases as well. 
I do plan to prefix `itemActionFunc`s with `_ItemAction_` when those get named in a future PR.


AngheloAlf (approved):
This aligns nicely with MM, since we have named this exact member as player's `actionFunc`

fig02:
Forgot to mention this PR also names `func_80835C58` as `Player_SetupAction`.

Theres `func_80835DAC` and `func_80835DE4`, which wrap `Player_SetupAction`. These will be named something like `SetupActionPreserveAnimMoveFlags` and `SetupActionForItem`, but these are just ideas at a cursory glance. Wont be named until more docs are complete

hensldm (approved):

PR #1533, opened 2023-09-09
Give ObjectContext unk_09 field name numKeepObjects

SlyryD:
Rename unk_09 in ObjectContext to numKeepObjects. The numKeepObjects field of ObjectContext is incremented by Object_Spawn, which is only used to spawn link and objects in gameplay_keep, gameplay_field_keep, and gameplay_dangeon_keep. Thus, numKeepObjects counts the number of "keep" objects (including link) pushed to the ObjectContext status array.

On room transition, Scene_CommandObjectList iterates through objects in the ObjectContext status array from index numKeepObjects to index num - 1, keeping objects that are in the same order between the ObjectContext status array and the next room's object list and removing the rest of the objects from the ObjectContext status array (before adding new objects to the ObjectContext status array). Thus, Scene_CommandObjectList ignores "keep" objects (including link).

### DIFF
--- a/include/z64object.h
+++ b/include/z64object.h
@@ -18,7 +18,7 @@ typedef struct {
     /* 0x0000 */ void*  spaceStart;
     /* 0x0004 */ void*  spaceEnd; // original name: "endSegment"
     /* 0x0008 */ u8     num; // number of objects in bank
-    /* 0x0009 */ u8     unk_09;
+    /* 0x0009 */ u8     numKeepObjects; // number of "keep" objects in bank including link
     /* 0x000A */ u8     mainKeepIndex; // "gameplay_keep" index in bank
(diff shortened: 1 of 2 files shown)


fig02:
Hi 
this is already named `numPersistentEntries` in https://github.com/zeldaret/oot/pull/1227/, which matches the rest of the proposed naming scheme there. Though, the PR has been stuck for a while due to other disagreements.

Feel free to review that PR and give your thoughts if you like.

SlyryD:
Okay that seems fine. I messed around with this in rando and realized my name for it there was bad and I wanted it to match something here for clarity. I don't want to interrupt that massive PR.

Dragorn421:
Thanks for writing a summary of how the system works around this btw, it also matches what I figured out which is always nice to check

Feel free to open such PRs again, it's uneasy to know what's already in PRs :p

PR #1536, opened 2023-09-15
Jenkinsfile master -> main

hensldm:
I believe this is the only change we would need to make in the repo itself if we want to switch over to main. Please let me know if I missed something. Opening as draft until we decide we 100% want to change.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,7 +26,7 @@ pipeline {
         stage('Build') {
             when {
                 not {
-                    branch 'master'
+                    branch 'main'
                 }
             }
             steps {
@@ -35,7 +35,7 @@ pipeline {
         }
         stage('Report Progress') {
             when {
-                branch 'master'
+                branch 'main'
             }
             steps {
                 sh 'mkdir reports'
@@ -47,7 +47,7 @@ pipeline {
         }
         stage('Update Progress') {
             when {
-                branch 'master'
+                branch 'main'
             }
             agent {
                 label 'zeldaret_website'


Dragorn421:
Using github's branch renaming feature it sounds like there is 0 issue to expect. it will even redirect urls using master to main apparently
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/renaming-a-branch

discussion in discord has shown support (or at least lack of anti-support) for changing to "main" so we can merge and change the name at some point

Dragorn421 (commented):
why is the "Build (qemu-irix)" stage still using master
(note: I have no idea how jenkins works)

Dragorn421 (commented):
the readme may be referencing the master branch here:
```
[jenkins]: https://jenkins.deco.mp/job/OOT/job/master
[jenkins-badge]: https://img.shields.io/jenkins/build?jobUrl=https%3A%2F%2Fjenkins.deco.mp%2Fjob%2FOOT%2Fjob%2Fmaster
```
(idk)

hensldm:
Ahh yeah, both of those needed to be changed. You also reminded me to update stuff in the docs folder, so thank you for that :) .

Dragorn421 (approved):
I think the order of operations should go like:
1) rename master to main
2) merge this
whereas the reverse has potential to confuse jenkins I'd think

fig02 (approved):
Look like we will be going ahead with this, but I'll wait until sometime on Friday (2 days from now)

PR #1538, opened 2023-09-17
Name player's generic `actionVar`s

fig02:
Player unfortunately makes extensive use of generic context-dependent variables. There are two that get reset everytime a new action function is set up and constantly change meaning. One is a byte and one is a short.

We will eventually need to decide how we will move forward with giving it meaningful names. The `#define vName` approach exists, but would get very unruly with how often this variable changes meaning. 

Cdi uses unions in his personal docs, shown here: https://github.com/CDi-Fails/oot/blob/player_docs/include/z64player.h#L875
Though this is not nearly a complete list of all the names it will have. Something to consider

### DIFF
--- a/include/z64player.h
+++ b/include/z64player.h
@@ -707,8 +707,8 @@ typedef struct Player {
     /* 0x0846 */ u8 unk_846;
     /* 0x0847 */ s8 unk_847[4];
     /* 0x084B */ s8 unk_84B[4];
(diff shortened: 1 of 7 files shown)


Thar0 (approved):

hensldm (approved):

fig02:
going to merge and we will continue the discussion on how to handle the variable as more actions get documented

PR #1546, opened 2023-09-21
Use `long` over `int` for s32/u32 types

Thar0:
libultra originally used long for the 32-bit integer types, this PR switches to that.

Some code changes were required to maintain matching since IDO of course has different codegen under certain conditions even though both types are the same width. The changes are mostly return types for predicate functions that would return a boolean value (evidence for a `typedef int bool` or similar type being used originally?)


fig02:
real C types in our codebase :monka:
Even if it is evidence of a possible bool type existing originally, im not sure its something we should do. We still wouldnt be able to use it consistently cause they used all types of widths for bools.

Im mostly fine with these just remaining `int`, will just take a sec to get used to

fig02:
Something worth noting also I suppose, the Gamecube zeldas also had their own version of the ultratypes. And since map files show the exact types that functions took, its clear that they mix and matched them all the time (prob dev preference)

Ofc our codebase can continue to use them whenever possible for consistencies sake, but just a fun tidbit for people that may not have been aware lol

include/libc/stdint.h
```diff
@@ -1,8 +1,8 @@
 #ifndef STDINT_H
 #define STDINT_H
 
-typedef signed int intptr_t;
-typedef unsigned int uintptr_t;
+typedef signed long intptr_t;
+typedef unsigned long uintptr_t;
```
hensldm, line 5: Out of curiosity do we have a source for this change?

Thar0: I dug through standard headers provided by GCC to see what the default for `uintptr_t` is, and it uses `unsigned long` usually. It's also required for matching in some spots

AngheloAlf:
> Something worth noting also I suppose, the Gamecube zeldas also had their own version of the ultratypes. And since map files show the exact types that functions took, its clear that they mix and matched them all the time (prob dev preference)

I may be misremembering, but I don't think the map files show the type a function took :thinking:

fig02:
It does, it has mangled function names that tell you types of args (but not names). The gamecube zeldas have this, but AF does not

AngheloAlf:
Ohh, you mean the actual gamecube zeldas, not the OoT/MM gamecube ones lmao

Dragorn421:
Did you change the least s32 to int to keep it matching? (or also more)

Could more `s32`s possibly be changed to int and still match?

Thar0:
This is the minimal changes needed for matching. I'm sure more could be changed to int if there's good reason (probably various standard c functions should use int where possible)

fig02 (approved):

Roman971:
Having to use `int` in the codebase is very unfortunate, I'm wondering if this change is really worth doing because of that

louist103:
> Having to use `int` in the codebase is very unfortunate, I'm wondering if this change is really worth doing because of that

It looks like almost all the cases of functions returning `int` is when it returns something that is meant to be treated as a bool. Likely indicating a typedef existed for it. I don't think its worth removing this change for that.

Dragorn421:
Why is using `int` unfortunate?

fig02:
The only unfortunate thing is using a base C type when we have managed to not do it anywhere else using ultratypes (its out of place)
But if this is confirmed to be the type for s32 from libultra itself then I dont think we really have much say in the matter

AngheloAlf (approved):

include/libc/stdint.h
Dragorn421: ftr my system has
```c
/* Types for `void *' pointers.  */
#if __WORDSIZE == 64
# ifndef __intptr_t_defined
typedef long int		intptr_t;
#  define __intptr_t_defined
# endif
typedef unsigned long int	uintptr_t;
#else
# ifndef __intptr_t_defined
typedef int			intptr_t;
#  define __intptr_t_defined
# endif
typedef unsigned int		uintptr_t;
#endif
```
(aka int for 32 bits, long for 64 bits)

but this really doesn't matter, we can choose whatever we want as long as it works for compiling in accordance to the C standard

Dragorn421 (approved):

PR #1550, opened 2023-09-21
Document pause page switching

Dragorn421:


Dragorn421:
I will say this must be one of the worst bits of the pause menu code
I even forgot to copypaste some bits from the whole docs draft PR

Here are some explanations on what's going on:

`KaleidoScope_DrawPages` draws the item page in the -z direction, equip in -x, quest in +z and map in +x
this turns out as:
schema of pause pages in world space

"scrolling" works by (approximately) the camera being in the center and rotating in increments of 90° to face a different page.

per `KaleidoScope_SetView`, the camera's `at` (where the camera points at) is always 0,0,0.
the eye coordinates (where the camera "is", "looks from") is what's animated to produce the "rotation" effect

when looking at a page and not scrolling, the eye coordinates are e.g. 64,0,0 (64 along +/- x/z) (hence this PR's `#define PAUSE_EYE_DIST (64.0f)`)

when scrolling (and this is where this code really is cursed), `KaleidoScope_UpdateSwitchPage` (named so in this PR) updates the eye coordinates in increments read from the `sPageSwitchEyeDx`/z arrays
These arrays have entries for each combination of each page and the switch direction left/right, aka all `nextPageMode; // (2 * prev pageIndex) + (scroll left ? 1 : 0)` values

For example if currently on the map page, then the eye is on the opposite side of that page from 0,0,0, aka (since map page is towards +x) towards -x, eye=(-64,0,0)

Now if scrolling left, nextPageMode is set (see `KaleidoScope_SwitchPage`) to `PAUSE_MAP * 2 + 1 == 3`
`sPageSwitchEyeDx[3]` is 4, `sPageSwitchEyeDz[3]` is 4
the scrolling takes 16 frames (cf `pauseCtx->switchPageTimer`), so this makes the eye move (linearly) from (-64,0,0) to (-64 + 4*16,0,0  + 4*16)=(0,0,64), making the camera face towards -z, towards the item page, which is indeed left from the map page

This PR puts calculations in `sPageSwitchEyeDx`/Dz instead of the raw values.
For example `sPageSwitchEyeDx[0 == 2*PAUSE_ITEM + 0]` (scroll right from the item page) has:
`-PAUSE_EYE_DIST * (PAUSE_MAP_X - PAUSE_ITEM_X) / 16,  // PAUSE_ITEM  right`
scrolling right from the item page brings the view to the map page hence the usage of `PAUSE_MAP_`
the initial eye.x (when sitting on the item page) is `-PAUSE_EYE_DIST * PAUSE_ITEM_X` (be opposite from the page across 0,0,0)
we want a dx value such that (after the 16 frames of switching):
`eye_x_map = eye_x_item + 16 * dx`
aka `-PAUSE_EYE_DIST * PAUSE_MAP_X = -PAUSE_EYE_DIST * PAUSE_ITEM_X + 16 * dx`
aka `dx = -PAUSE_EYE_DIST * (PAUSE_MAP_X - PAUSE_ITEM_X) / 16`
and this generalizes for any two side by side pages and a scroll direction

-----

there's also `gPageSwitchNextButtonStatus` which for some reason has different indexing logic and is used to set which buttons are dimmed on each page, it could just be indexed by target page but apparently they liked this weird indexing better

src/overlays/misc/ovl_kaleido_scope/z_kaleido_scope_PAL.c
```diff
@@ -2470,28 +2512,32 @@ void KaleidoScope_GrayOutTextureRGBA32(u32* texture, u16 pixelCount) {
...
+
+        pauseCtx->pageIndex = sPageSwitchNextPageIndex[pauseCtx->nextPageMode];
+
         pauseCtx->mainState = PAUSE_MAIN_STATE_IDLE;
         pauseCtx->state++; // PAUSE_STATE_MAIN
+
         pauseCtx->alpha = 255;
         Interface_LoadActionLabelB(play, DO_ACTION_SAVE);
-    } else if (pauseCtx->unk_1EA == 64) {
+    } else if (pauseCtx->switchPageTimer == (4 * 16 * 1)) {
```
engineer124, line 2537: What do these 3 numbers mean for the timer? The `(4 * 16)` appears a few times

src/overlays/misc/ovl_kaleido_scope/z_kaleido_scope_PAL.c
```diff
@@ -490,26 +529,29 @@ void KaleidoScope_SetDefaultCursor(PlayState* play) {
     }
 }
 
+#define SWITCH_PAGE_LEFT_PT 0
+#define SWITCH_PAGE_RIGHT_PT 2
```
engineer124, line 533: What's the `PT` here?

src/overlays/misc/ovl_kaleido_scope/z_kaleido_scope_PAL.c
Dragorn421: The timer increases in increments of 4
`pauseCtx->switchPageTimer += 4 * ZREG(46);`
(`ZREG(46)` is always 1)
-> hence `4*16`

The if before this else has `(4 * 16 * ZREG(47))`
and `ZREG(47)` is always 1
-> so the `4*16*1` is for symmetry

I assume at some point the dev tried `ZREG(47)` == 2 or 3 (or whatever) then this else if would pass at (resp.) 1/2 or 1/3rd of the full animation but idk

(note inside this else if is dead code as the existing "
```
`ZREG(47)` is always 1 so this normally never happens
```
" comment tries to say, is it unclear?)

src/overlays/misc/ovl_kaleido_scope/z_kaleido_scope_PAL.c
Dragorn421: It comes from this debug string

https://github.com/zeldaret/oot/blob/b3486b57ef41a284a366eb40ca07bf82bf4750d6/src/overlays/misc/ovl_kaleido_scope/z_kaleido_scope_PAL.c#L514

PT values are for the pt variable

I guess I could rename it to direction or something maybe?

-----

Pure guess with no proof: maybe "pt" stands for "page turn" 8)

include/z64pause.h
```diff
@@ -98,9 +124,9 @@ typedef struct {
     /* 0x01D6 */ u16 debugState;
     /* 0x01D8 */ Vec3f eye;
     /* 0x01E4 */ u16 mainState;
-    /* 0x01E6 */ u16 mode;
+    /* 0x01E6 */ u16 nextPageMode; // (2 * prev pageIndex) + (scroll left ? 1 : 0)
     /* 0x01E8 */ u16 pageIndex; // "kscp_pos"
```
fig02, line 128: coming form the perspective of someone who doesnt understand this mechanism all too well, the random string of code next to the member variable isnt really telling me anything. id either use words to describe it or just move that explanation somewhere else maybe?

include/z64pause.h
```diff
@@ -20,6 +20,32 @@ typedef enum {
...
+/*
+ *         [item]
+ *   [
+ *   e          x     [
+ *   q       o-->     m
+ *   u       |        a
+ *   i     z v        p
+ *   p                ]
+ *   ]
+ *         [quest]
```
fig02, line 34: im no ascii artist, but this may look better if you spell "equip" and "map" horizontally? at the expense of the whole diagram shifting to the right a bit with the word "equip" pushing it out

src/code/z_kaleido_setup.c
```diff
@@ -1,12 +1,42 @@
...
+    PAUSE_EYE_DIST * -PAUSE_EQUIP_X, // PAUSE_MAP
+    PAUSE_EYE_DIST * -PAUSE_ITEM_X,  // PAUSE_QUEST
+    PAUSE_EYE_DIST * -PAUSE_MAP_X,   // PAUSE_EQUIP
+};
+f32 sKaleidoSetupUnusedEyeZ[] = {
+    PAUSE_EYE_DIST * -PAUSE_QUEST_Z, // PAUSE_ITEM
+    PAUSE_EYE_DIST * -PAUSE_EQUIP_Z, // PAUSE_MAP
+    PAUSE_EYE_DIST * -PAUSE_ITEM_Z,  // PAUSE_QUEST
+    PAUSE_EYE_DIST * -PAUSE_MAP_Z,   // PAUSE_EQUIP
+};
```
fig02, line 20: newline police, I think these were fine squished together when they were a single line but now they may look nicer spaced out

src/code/z_kaleido_setup.c
```diff
@@ -30,24 +60,27 @@ void KaleidoSetup_Update(PlayState* play) {
             WREG(16) = -175;
             WREG(17) = 155;
 
-            pauseCtx->unk_1EA = 0;
-            pauseCtx->mainState = PAUSE_MAIN_STATE_1;
+            pauseCtx->switchPageTimer = 0;
+            pauseCtx->mainState = PAUSE_MAIN_STATE_SWITCHING_PAGE; // irrelevant
```
fig02, line 65: may want to expand a bit on whats irrelevant and why?

src/code/z_kaleido_setup.c
```diff
@@ -30,24 +60,27 @@ void KaleidoSetup_Update(PlayState* play) {
...
 
-            if (ZREG(48) == 0) {
-                pauseCtx->eye.x = sKaleidoSetupEyeX0[pauseCtx->pageIndex];
-                pauseCtx->eye.z = sKaleidoSetupEyeZ0[pauseCtx->pageIndex];
-                pauseCtx->pageIndex = sKaleidoSetupKscpPos0[pauseCtx->pageIndex];
+            if (R_START_LABEL_DD(0) == 0) {
+                // Never reached, unused, and the data would be wrong anyway
+                pauseCtx->eye.x = sKaleidoSetupUnusedEyeX[pauseCtx->pageIndex];
```
fig02, line 68: id be interested in expanding more why its never reached/unused and why the data would be wrong

src/overlays/misc/ovl_kaleido_scope/z_kaleido_scope_PAL.c
```diff
@@ -135,12 +135,30 @@ s16 D_8082AB2C[] = {
...
-static u8 D_8082AB6C[][5] = {
+static u8 gPageSwitchNextButtonStatus[][5] = {
+    // PAUSE_ITEM  + SWITCH_PAGE_LEFT_PT
+    //
+    //  -> PAUSE_EQUIP
     { BTN_ENABLED, BTN_DISABLED, BTN_DISABLED, BTN_DISABLED, BTN_ENABLED },
+    // PAUSE_MAP   + SWITCH_PAGE_LEFT_PT
+    //
+    //  -> PAUSE_ITEM
     { BTN_ENABLED, BTN_ENABLED, BTN_ENABLED, BTN_ENABLED, BTN_DISABLED },
```
fig02, line 146: this is definitely just my lack of understanding of how this is working in general, but these comments are confusing to me. I guess there are entries for each page when coming from another page? may be good to explain that above the array but idk

src/overlays/misc/ovl_kaleido_scope/z_kaleido_scope_PAL.c
```diff
@@ -155,16 +173,37 @@ static s16 D_8082ABA4 = 0;
...
-    PAUSE_MAP, PAUSE_EQUIP, PAUSE_QUEST, PAUSE_ITEM, PAUSE_EQUIP, PAUSE_MAP, PAUSE_ITEM, PAUSE_QUEST,
+static f32 sPageSwitchEyeDx[] = {
+    -PAUSE_EYE_DIST * (PAUSE_MAP_X - PAUSE_ITEM_X) / 16,  // PAUSE_ITEM  right
+    -PAUSE_EYE_DIST*(PAUSE_EQUIP_X - PAUSE_ITEM_X) / 16,  // PAUSE_ITEM  left
+    -PAUSE_EYE_DIST*(PAUSE_QUEST_X - PAUSE_MAP_X) / 16,   // PAUSE_MAP   right
+    -PAUSE_EYE_DIST*(PAUSE_ITEM_X - PAUSE_MAP_X) / 16,    // PAUSE_MAP   left
+    -PAUSE_EYE_DIST*(PAUSE_EQUIP_X - PAUSE_QUEST_X) / 16, // PAUSE_QUEST right
+    -PAUSE_EYE_DIST*(PAUSE_MAP_X - PAUSE_QUEST_X) / 16,   // PAUSE_QUEST left
+    -PAUSE_EYE_DIST*(PAUSE_ITEM_X - PAUSE_EQUIP_X) / 16,  // PAUSE_EQUIP right
+    -PAUSE_EYE_DIST*(PAUSE_QUEST_X - PAUSE_EQUIP_X) / 16, // PAUSE_EQUIP left
```
fig02, line 184: i guess this is a similar thing? how does `// PAUSE_ITEM (coming from the right)`
and such sound? or is my understanding of this wrong?

fig02: also does the formatter let you space out the multiplication? the first line has it but the rest dont

src/overlays/misc/ovl_kaleido_scope/z_kaleido_scope_PAL.c
```diff
@@ -1531,33 +1573,33 @@ void KaleidoScope_UpdateNamePanel(PlayState* play) {
...
+void KaleidoScope_UpdateSwitchPage(PlayState* play, Input* input) {
     PauseContext* pauseCtx = &play->pauseCtx;
-    s32 cond = false;
-    s32 mode;
+    s32 frameAdvanceFreeze = false;
+    s32 nextPageMode;
 
-    if (ZREG(13) && !CHECK_BTN_ALL(input->press.button, BTN_L)) {
-        cond = true;
+    if (R_PAUSE_SWITCH_PAGE_FRAME_ADVANCE_ON && !CHECK_BTN_ALL(input->press.button, BTN_L)) {
```
fig02, line 1581: fwiw PageSwitch sounds better to me than SwitchPage in general, and it seems youve used that phrasing for other things such as `sPageSwitchEyeDx`
I guess this would apply to alot of things like `switchPageTimer`

fig02 (commented):
Sorry for review delay. Distracted by retail, and also this stuff is pretty intimidating.

My comments may come from a place of ignorance, as I have not put as much time into learning how this system works as you have. But I guess its a good test for trying to make the docs understandable to someone new.

src/overlays/misc/ovl_kaleido_scope/z_kaleido_scope_PAL.c
Dragorn421: 16 is now a macro: `PAGE_SWITCH_NSTEPS`

src/overlays/misc/ovl_kaleido_scope/z_kaleido_scope_PAL.c
Dragorn421: the formatter is doing that with the whitespace unfortunately :/

Dragorn421:
I added a bunch more docs/comments, lmk if it helps

include/z64pause.h
```diff
@@ -20,6 +20,33 @@ typedef enum {
...
+ *      [   item   ]
+ *
+ *   [                [
+ *              x
+ * equip     o-->    map
+ *           |
+ *   ]     z v        ]
+ *
+ *      [   quest  ]
+ */
```
fig02, line 35: I was thinking maybe something like this 
```c
/*
 *         [   item   ]
 *
 *                    
 *              x
 * [ equip ]    o-->    [ map ]
 *              |
 *            z v        
 *
 *         [   quest  ]
 */
 ```
 Im not sure if that formatting will look right so ill include an image as well
image

**Review by fig02 — APPROVED** [review]
All of these explanations are what I needed, thanks.
Small suggestion with ascii art below

File: include/z64pause.h (hunk 11 above)
Comment (Dragorn421) [reply include/z64pause.h:35]: programmers art moment
I was trying to mimic the direction in which each page stretches by keeping the [] above/below

maybe `<>v^` would be clearer for that purpose?

```
 *      <   item   >
 *
 *   ^                ^
 *              x
 * equip     o-->    map
 *           |
 *   v     z v        v
 *
 *      <   quest  >
 */
```

Or I'm ok `[linear]`ing it

fig02: Sure that looks nice

jpburnett (approved):

PR #1551, opened 2023-09-21
Name the main pause state and add placeholder enum for main (sub)states

Dragorn421:
Not sold on naming it "main" `PAUSE_STATE_MAIN` but I don't have other ideas

The substate enum `PAUSE_MAIN_STATE_` is placeholders besides the IDLE (0) one, which is almost half of the references and is straightforward to name now imo (the other substates are more involved to name)

"interactive" was suggested but other states are "interactive" too (save prompt, gameover), and some substates are not legitimately interactive (i.e. the substate for song playback with the player listening/watching)

For the documented `PAUSE_MAIN_STATE_` enum instead of the placeholder `_1`, `_2`, ... this PR has: https://github.com/Dragorn421/oot/blob/d43613694caf4574f2375c22529de9918caee054/include/z64pause.h#L73-L82

there is another such substate (not documented in main yet): `PauseContext.unk_1EC`, aka `savePromptState`


fig02 (approved):
im fine with main state personally and think its gets the point across fine

engineer124 (approved):

fig02:
ready to merge after resolved conflict

PR #1554, opened 2023-09-24
Speedmeter Small cleanup

hensldm:
Noticed a couple things will doing MM's version.
The rename to `x` is more just my interpretation of what it probably was, though don't really have proof. If preferred I can revert it back to unk.

### DIFF
--- a/src/code/speed_meter.c
+++ b/src/code/speed_meter.c
@@ -74,6 +74,17 @@ SpeedMeterTimeEntry sSpeedMeterTimeEntryArray[] = {
     { &gGraphUpdatePeriod, 0, 10, GPACK_RGBA5551(255, 0, 255, 1) },
 };
 
+typedef struct {
+    /* 0x00 */ s32 maxval;
+    /* 0x04 */ s32 val;
+    /* 0x08 */ u16 backColor;
+    /* 0x0A */ u16 foreColor;
+    /* 0x0C */ s32 ulx;
+    /* 0x10 */ s32 lrx;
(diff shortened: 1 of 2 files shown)


fig02 (approved):

Roman971 (approved):

fig02:
has met merge requirements but will merge tomorrow incase anyone else wants to take a look

Dragorn421 (approved):

src/code/speed_meter.c
```diff
@@ -237,10 +248,9 @@ void SpeedMeter_DrawAllocEntries(SpeedMeter* meter, GraphicsContext* gfxCtx, Gam
         y++;
     }
 
-    thga = (TwoHeadGfxArena*)&state->tha;
-    //! @bug THA_GetRemaining call should be THGA_GetRemaining like the others below, harmless as-is
```
Thar0, line 241: Why is the bug comment gone?

Thar0 (changes requested):

src/code/speed_meter.c
fig02: he changed the temp to be a tha instead of a thga, so its the right function call

Thar0 (approved):

src/code/speed_meter.c
Dragorn421: I was very confused as well, I think we missed the tha to thga cast back when you (tharo) overhauled th(g)a docs

PR #1560, opened 2023-10-04
Player Docs: Propose Named Unions For Shared Variables

fig02:
The Player Actor makes extensive usage of shared variables in an attempt to save space. 
Currently, the project tends to prefer a `#define` approach to give varying names to the same variable. With how often some of these variables change meaning, I feel very confident saying that this approach is not going to work for these cases. It will get too unruly.

I've proposed here a concept that isn't really used too often in the repo at the moment: a named union for 3 cases (so far) of shared variables in player.

I feel *very* strongly against using unnamed unions. When reading code, and especially when debugging, I need to know right away if something is involved with a shared variable. There are too many opportunities for bugs to arise because of stale or uninitialized values being used. Some wip player docs out there use unnamed unions and I have already gotten very confused trying to read some code.

This is where a named union solves that problem. While yes, it does increase the length of names a bit, I think the information it provides is much more valuable than the downside of a slightly longer name. 
Additionally, the scheme I use currently gives the unions very short and abbreviated names ("av1", "av2", and "cv"). There is a comment on the union explaining what the abbreviations stand for. And it doesn't need to be too descriptive, given its only job is to let you know its a part of a shared variable.

Since this pr has a good amount of changes, I will detail out an example using what I've called the "cutscene variable" with its two different uses:

Union is defined like so:
```c
    /* 0x046A */ union { 
        s16 haltActorsDuringCsAction; // If true, freeze actors belonging to certain categories during a `csAction`
        s16 slidingDoorBgCamIndex; // `BgCamIndex` used during a sliding door cutscene
    } cv; // "Cutscene Variable": context dependent variable that has different meanings depending on what function is called
```
and has uses like so:
```c
// in func_8002DF54
player->cv.haltActorsDuringCsAction = true; 

// in func_8083ADD4
        Player_SetupAction(play, this, Player_Action_80852E14, 0);
        if (this->cv.haltActorsDuringCsAction) {
            this->stateFlags1 |= PLAYER_STATE1_29;
        }
```
```c
// in Player_ActionChange_1
this->cv.slidingDoorBgCamIndex = play->transiActorCtx.list[GET_TRANSITION_ACTOR_INDEX(&slidingDoor->dyna.actor)].sides[(doorDirection > 0) ? 0 : 1].bgCamIndex;

// in DoorShutter_InitOpeningDoorCam
Camera_ChangeDoorCam(play->cameraPtrs[CAM_ID_MAIN], &this->dyna.actor, player->cv.slidingDoorBgCamIndex, 0.0f, 12, doorCamTimer2, 10);
```

We've had many discussions about unions over the years. I think this solution is the best compromise for all of the gripes people have brought up with our various solutions over time.

(Since this PR introduces a big change I will wait for more than 2 reviews before committing)


AngheloAlf (approved):
Named unions are way nice, keeps the code better organized imo

Dragorn421 (approved):

Thar0 (approved):

PR #1562, opened 2023-10-10
Documentation for audio_thread_manager.c

Thar0:
Maybe we want to change the `AudioMgr_` prefix? Not sure what works well.


src/code/audio_thread_manager.c
```diff
@@ -1,91 +1,150 @@
...
+        // Wait for the audio rsp task scheduled on the previous retrace to complete. This looks like it should wait
+        // for the task scheduled on the current retrace, but since the queue is initially filled in AudioMgr_Init
+        // it doesn't wait for the task scheduler to post a message for the most recent task as there is already a
+        // message waiting.
+        osRecvMesg(&audioMgr->taskDoneQueue, NULL, OS_MESG_BLOCK);
+        // Report task done
+        //! @bug As a result of above, the task done notification is sent to the next task instead of the task it was
+        //! intended for. In practice, task done notifications are not used by the audio driver so this is
+        //! inconsequential.
+        AudioMgr_NotifyTaskDone(audioMgr);
```
Dragorn421, line 71: Is it? audioMgr->rspTask is still the previous task at this point isn't it?

src/code/audio_thread_manager.c
```diff
@@ -1,91 +1,150 @@
...
+    // Spin waiting for events
     while (true) {
         osRecvMesg(&audioMgr->interruptQueue, (OSMesg*)&msg, OS_MESG_BLOCK);
+
         switch (*msg) {
             case OS_SC_RETRACE_MSG:
                 AudioMgr_HandleRetrace(audioMgr);
+
+                // Empty the interrupt queue
                 while (!MQ_IS_EMPTY(&audioMgr->interruptQueue)) {
```
Dragorn421, line 116: Is there a reason to have another inner while loop instead of just having the one outer loop?

src/code/audio_thread_manager.c
```diff
@@ -95,11 +154,12 @@ void AudioMgr_Init(AudioMgr* audioMgr, void* stack, OSPri pri, OSId id, Schedule
...
     audioMgr->rspTask = NULL;
 
-    osCreateMesgQueue(&audioMgr->taskQueue, audioMgr->taskMsgBuf, ARRAY_COUNT(audioMgr->taskMsgBuf));
+    osCreateMesgQueue(&audioMgr->taskDoneQueue, &audioMgr->taskDoneMsg, 1);
     osCreateMesgQueue(&audioMgr->interruptQueue, audioMgr->interruptMsgBuf, ARRAY_COUNT(audioMgr->interruptMsgBuf));
-    osCreateMesgQueue(&audioMgr->lockQueue, audioMgr->lockMsgBuf, ARRAY_COUNT(audioMgr->lockMsgBuf));
+    osCreateMesgQueue(&audioMgr->initQueue, &audioMgr->initMsg, 1);
 
-    osSendMesg(&audioMgr->taskQueue, NULL, OS_MESG_BLOCK);
+    // Send a message to the task queue so it is initially full
```
Dragorn421, line 161:
```suggestion
    // Send a message to the task done queue so it is initially full
```
?

Also I know it's the same because these queues are only holding 1 message at most, but I find using "full" confusing instead of saying like "has 1 message available", but idk may just be me

src/code/main.c
```diff
@@ -90,7 +90,7 @@ void Main(void* arg) {
     StackCheck_Init(&sPadMgrStackInfo, sPadMgrStack, STACK_TOP(sPadMgrStack), 0, 0x100, "padmgr");
     PadMgr_Init(&gPadMgr, &sSerialEventQueue, &gIrqMgr, THREAD_ID_PADMGR, THREAD_PRI_PADMGR, STACK_TOP(sPadMgrStack));
 
-    AudioMgr_Unlock(&gAudioMgr);
+    AudioMgr_WaitForInit(&gAudioMgr);
```
Dragorn421, line 93: Any idea why this is, iirc the main thread has lowest priority so the audiomgr thread will have already run up to the point where it's blocking on something basically before AudioMgr_Init returns

Dragorn421 (commented):
What exactly is the "audio driver"? The thing that (allegedly) forms a library that's also (in various versions) in sm64 and mm?

Thar0:
"audio driver" is meant to refer to the contents of src/audio, which is believed to be one library yes

src/code/audio_thread_manager.c
Thar0: Here's what I think happens, in all the gory details

- The task done queue initially holds a message
- First retrace runs
  - `audioMgr->rspTask` is initially null, no task is scheduled
  - Audio driver produces an rsp task (task 1)
  - `audioMgr->rspTask` is still null, don't try to pull a message from the task done queue
  - `audioMgr->rspTask` is updated to the generated rsp task (task 1)
- Second retrace runs
  - `audioMgr->rspTask` is valid, the task is scheduled (task 1)
  - Audio driver produces an rsp task (task 2)
  - `audioMgr->rspTask` is valid, try to read a message. This always succeeds immediately regardless of whether the task actually completed due to the initially full queue. Further, since the queue is full the scheduler blocks waiting for the queue to empty, so the message that task 1 gets is guaranteed to be the "extra" message. The scheduler will then refill the queue with the task done message for task 1 once it actually completes.
  - `audioMgr->rspTask` is updated to the generated rsp task (task 2)
- Third retrace runs
  - `audioMgr->rspTask` is valid, the task is scheduled (task 2)
  - Audio driver produces an rsp task (task 3)
  - `audioMgr->rspTask` is valid, try to read a message. This will read the task done message for task 1 and post it to task 2 since they use the same message queue and the scheduler will have posted the task done message for task 1 before task 2 can start, as the scheduler blocks if the task done queue is full.
  - `audioMgr->rspTask` is updated to the generated rsp task (task 3)
- Repeat

I think my wording can definitely be improved here. If you see a good way to make this more clear let me know.

src/code/audio_thread_manager.c
Thar0: The idea is to only handle a retrace event once until the queue is emptied, I guess if the retrace handling takes so long that another retrace event is already waiting they didn't want to run it again so quickly

src/code/audio_thread_manager.c
Thar0: I don't find "has a message available" any more clear than just using "full" here. I'm open to describing this differently but I can't think of a satisfying way to word it currently.

Dragorn421:
> "audio driver" is meant to refer to the contents of src/audio, which is believed to be one library yes

You mean src/audio/lib then, right?

src/code/main.c
Dragorn421: ah I got confused between the "main" thread and the "graph" thread, thanks

even with main boosting its priority to 15=audiomgr+3, in the mean time the bulk of execution happens in the graph thread which is audiomgr-1 in priority, so it's another reason this is useless

but oh well :shrug: explicit synchronization doesn't hurt

src/code/audio_thread_manager.c
Dragorn421: If that's ever the case you are having big problems anyway, but I guess this is still better than monopolizing the cpu on audio then :shrug:

src/code/audio_thread_manager.c
Dragorn421: ah I see I missed that the generated task is kept for scheduling on the next retrace instead of being scheduled immediately (why?)

I don't think your wording can be improved as much as the code in the first place 8)

maybe instead of

> As a result of above, the task done notification is sent to the next task instead of the task it was intended for

, ? :

> As a result of above, the initial task done notification is sent before the first task is actually done, and then every task done notification is early by one task.

I don't like "is early by one task" but idk how to put it concisely

Dragorn421: :thinking: 

> As a result of above, the first task done notification is sent before the first task is actually done, then the first task being done causes sending a second notification before the second task is actually done, and so on. So every task done notification is early by one task.

Thar0: > ah I see I missed that the generated task is kept for scheduling on the next retrace instead of being scheduled immediately (why?)

I guess it could make the timing more consistent, since this way the rsp task start time is not sensitive to variation in how long the audio driver takes to update?

I tried to adjust the comment, not sure if it's much better though

Dragorn421 (approved):
:musical_note: :thread: :woman_office_worker::chart_with_upwards_trend:

include/audiomgr.h
```diff
@@ -0,0 +1,33 @@
...
+#define AUDIOMGR_H
+
+#include "sched.h"
+#include "z64audio.h"
+
+typedef enum {
+    AUDIOMGR_DEBUG_LEVEL_NONE = 0,
+    AUDIOMGR_DEBUG_LEVEL_NO_RSP = 1,
+    AUDIOMGR_DEBUG_LEVEL_NO_UPDATE = 2
+} AudioMgrDebugLevel;
```
fig02, line 11: we can drop the value assignments right?
theoretically would want offset comments to stay true to the repos convention, but I personally do not care much for them in this case, but meh

fig02 (approved):

PR #1563, opened 2023-10-11
Document two bugs in Camera_Normal1

cadmic:
Side note: I think `sp94` in the first bug should be called `accel` based on the way it's used to call `Camera_CalcDefaultYaw` but it has an unrelated use above and I don't think I'm good enough at decomp yet to fix that.

### DIFF
--- a/src/code/z_camera.c
+++ b/src/code/z_camera.c
@@ -1525,9 +1525,10 @@ s32 Camera_Normal1(Camera* camera) {
     if (sp94 > 1.0f) {
         sp94 = 1.0f;
     }
-    //! @bug This is probably supposed to be "sp94 < -1.0f" to clamp the value between -1 and 1.
-    //! As written, the previous if statement will have no effect.
     if (sp94 > -1.0f) {
+        // The condition here is probably supposed to be "sp94 < -1.0f" to clamp the value between
+        // -1 and 1. As written, the previous if statement has no effect. However, trying to fix
+        // this causes the camera to yaw too quickly when the player is moving.
         sp94 = -1.0f;
     }
 
@@ -1549,9 +1550,9 @@ s32 Camera_Normal1(Camera* camera) {
             Camera_LERPCeilF(R_CAM_PITCH_UPDATE_RATE_INV, camera->pitchUpdateRateInv, sp9C, rate);
     }
 
-    //! @bug pitchUpdateRateInv is updated twice since both branches of the previous if statement
-    //! also update it. This is probably unintentional because all other camera functions update it
-    //! at most once.
+    // pitchUpdateRateInv is updated twice since both branches of the previous if statement also
+    // update it. This might be unintentional because all other camera functions update it at most
+    // once.
     camera->pitchUpdateRateInv = Camera_LERPCeilF(R_CAM_PITCH_UPDATE_RATE_INV, camera->pitchUpdateRateInv, sp9C, rate);
     camera->xzOffsetUpdateRate =
         Camera_LERPCeilF(CAM_DATA_SCALED(R_CAM_XZ_OFFSET_UPDATE_RATE), camera->xzOffsetUpdateRate, spA0, rate);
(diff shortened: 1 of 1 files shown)


fig02:
Fixing both of these bugs makes the camera really fast an snappy. To the point of it being worse, in my opinion:

https://github.com/zeldaret/oot/assets/22045047/676f6d3e-819d-4872-b9a4-d053d5ca3539

This is mainly due to the `pitchUpdateRateInv` change.
Makes me wary about calling it a bug. It seems to me intentional to have it be set to the second value. Obviously not ideal that dead code exists above it. 
Definitely worth a comment, but not `@bug` in my opinion. And possibly move the comments to the two if blocks, since it seems to me like the last one is intended

cadmic:
Thanks for testing! Is that behavior actually due to the `sp94 > -1.0f` change though? `sp94` is later passed to `Camera_CalcDefaultYaw` (as `accel`) and it looks like the yaw is changing a lot in your example. And it doesn't look like the pitch is changing much, maybe try walking towards a cliff to get the camera to pitch down?

cadmic:
Hmm these definitely look like coding mistakes to me, and I think they probably tweaked camera parameters until it looked right without knowing about the mistakes. So a proper "bugfix" would also include camera parameter changes. Should `@bug` be used only for things that are trivially fixable?

cadmic:
(Sorry for the spam) As an example for my last question, I would consider acute angle clips and block clips to be bugs, but they're probably not fixable without a major redesign of the collision system. So probably those do not deserve a `@bug` comment?

fig02:
There isnt really strict criteria for what makes something a bug or not. Its about as subjective as trying to argue what is a glitch or not in speedruns lol

In my view, the end result is what they wanted and its not bugged, even if theres some jank along the way to get there (which like I said, definitely deserves to be commented on). But do see the other perspective as well. Maybe some more opinions needed

cadmic:
Alright, I've removed the `@bug` tags, let me know what you think

src/code/z_camera.c
```diff
@@ -1547,6 +1550,9 @@ s32 Camera_Normal1(Camera* camera) {
             Camera_LERPCeilF(R_CAM_PITCH_UPDATE_RATE_INV, camera->pitchUpdateRateInv, sp9C, rate);
     }
 
+    // pitchUpdateRateInv is updated twice since both branches of the previous if statement also
+    // update it. This might be unintentional because all other camera functions update it at most
+    // once.
```
Dragorn421, line 1555:
```suggestion
    // pitchUpdateRateInv is updated twice, also in the branches just above
```

src/code/z_camera.c
```diff
@@ -1526,6 +1526,9 @@ s32 Camera_Normal1(Camera* camera) {
         sp94 = 1.0f;
     }
     if (sp94 > -1.0f) {
+        // The condition here is probably supposed to be "sp94 < -1.0f" to clamp the value between
+        // -1 and 1. As written, the previous if statement has no effect. However, trying to fix
+        // this causes the camera to yaw too quickly when the player is moving.
```
Dragorn421, line 1531:
```suggestion
    // note: > -1, not < -1
    if (sp94 > -1.0f) {
```

Dragorn421 (commented):
Unfortunately the camera system is full of oddities such as what you pointed out, so I don't think we can afford to point out everything, especially not in many details
Otherwise the code (also in other files than z_camera) would be littered with comments

However maybe we can put one-line comments still as "gotchas" (see attached comments), though I'm not convinced either and wouldn't vehemently argue for them either

cadmic:
Well I very carefully went through the camera code for Normal1 and Parallel1 (and all the functions they call) while trying to reimplement it, and these are the only two blatant coding mistakes I found (rather than merely questionable design decisions). But I don't want to argue about a couple comments so I'll go ahead and close this PR. Thanks everyone for your time

PR #1568, opened 2023-10-23
libc cleanup

Thar0:
Mostly contains the libc changes from #1468, to try and cut that one down to be more manageable.


include/functions.h
```diff
@@ -5,15 +5,21 @@
...
+#ifdef __sgi
 #pragma intrinsic(fabsf)
+#else
+#define fabsf(f) __builtin_fabsf((f32)(f))
 #endif
+
 f32 sqrtf(f32 f);
+#ifdef __sgi
 #pragma intrinsic(sqrtf)
+#endif
```
AngheloAlf, line 17: Why the ifdefs for the pragmas? It shouldn't have any effect on other compilers

include/libc/assert.h
```diff
@@ -0,0 +1,45 @@
+#ifndef ASSERT_H
+#define ASSERT_H
+
+#if !defined(__GNUC__) && !defined(__attribute__)
+#define __attribute__(x)
+#endif
+
+// Runtime assertions
+
+__attribute__((noreturn)) void __assert(const char* assertion, const char* file, int line);
```
AngheloAlf, line 10:
```suggestion
NORETURN void __assert(const char* assertion, const char* file, int line);
```

include/libc/assert.h
```diff
@@ -0,0 +1,45 @@
+#ifndef ASSERT_H
+#define ASSERT_H
+
+#if !defined(__GNUC__) && !defined(__attribute__)
+#define __attribute__(x)
+#endif
```
AngheloAlf, line 6: Wouldn't be better to include `attributes.h` instead of duplicating this define all over the place?

include/functions.h
Thar0: Some compilers (gcc) will warn about unknown pragmas

include/libc/assert.h
Thar0: I'd rather not include non-libc headers in libc headers. Same reason for why I use `__attribute__((noreturn))` above

include/libc/math.h
```diff
@@ -5,24 +5,37 @@
 
 #define M_PI 3.14159265358979323846f
 #define M_SQRT2 1.41421356237309504880f
-#define FLT_MAX 340282346638528859811704183484516925440.0f
+#define MAXFLOAT 3.40282347e+38f
 #define SHT_MAX 32767.0f
```
fig02, line 9: is this a known original name or a standard name for this? 
am fine with it if so, but it just goes against our convention for similar things

Thar0: This is the standard C name for the largest representable float define in `math.h`.
`FLT_MAX` is another standard C name, but is part of `float.h` instead, which we don't currently have. It might be nice to add additional standard headers in the future, but there's some questions about exactly how much to add so I opted not to add too many new headers for now.

fig02 (approved):

include/libc/math.h
Dragorn421: ISO C has `FLT_MAX` but not `MAXFLOAT`
https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3096.pdf

`MAXFLOAT` seems to be POSIX
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/math.h.html

include/ultra64/libc.h
Dragorn421, line 1: What is the meaning of this file name?

I see you put bzero/bcmp/bcopy in here which apparently were posix and today are deprecated/removed

src/libultra/libc/xprintf.c
```diff
@@ -131,85 +150,88 @@ void _Putfld(_Pft* px, va_list* pap, u8 code, u8* ac) {
...
             } else {
-                *va_arg(*pap, u32*) = px->nchar;
+                *va_arg(*pap, unsigned int*) = px->nchar;
             }
             break;
 
         case 'p':
-            px->v.ll = (s32)va_arg(*pap, void*);
-            px->s = (char*)&ac[px->n0];
+            px->v.ll = (long)va_arg(*pap, void*);
```
Dragorn421, line 218:
```suggestion
            px->v.ll = (intptr_t)va_arg(*pap, void*);
```
?

src/libultra/libc/xprintf.c
```diff
@@ -131,85 +150,88 @@ void _Putfld(_Pft* px, va_list* pap, u8 code, u8* ac) {
...
         case 'e':
         case 'f':
         case 'g':
         case 'E':
         case 'G':
-            px->v.ld = px->qual == 'L' ? va_arg(*pap, f64) : va_arg(*pap, f64);
+            px->v.ld = px->qual == 'L' ? va_arg(*pap, LONG_DOUBLE_TYPE) : va_arg(*pap, double);
 
-            if (*(u16*)&px->v.ll & 0x8000) {
+            if (LDSIGN(px->v.ld)) {
```
Dragorn421, line 193: why make a macro for this? and is there a reason to not do `ld < 0` (besides, for us, matching)

this with the macro `isdigit` (not `IS_DIGIT`) makes me wonder if you yoinked this code from somewhere or are looking at a reference?

src/libultra/libc/xlitob.c
```diff
@@ -0,0 +1,57 @@
...
+
+    if (num != 0 || args->prec != 0) {
+        buff[--i] = digs[num % base];
+    }
+
+    args->v.ll = num / base;
+
+    while (args->v.ll > 0 && i > 0) {
+        lldiv_t quotrem = lldiv(args->v.ll, base);
+        args->v.ll = quotrem.quot;
```
Dragorn421, line 39:
```suggestion
        lldiv_t quotrem = lldiv(args->v.ll, base);

        args->v.ll = quotrem.quot;
```

src/libultra/libc/sprintf.c
```diff
@@ -0,0 +1,31 @@
...
+#include "string.h"
+#include "ultra64/xstdio.h"
+
+void* proutSprintf(void* dst, const char* fmt, size_t size) {
+    return (char*)memcpy(dst, fmt, size) + size;
+}
+
+int vsprintf(char* dst, const char* fmt, va_list args) {
+    int ret = _Printf(proutSprintf, dst, fmt, args);
+    if (ret > -1) {
```
Dragorn421, line 12:
```suggestion
    int ret = _Printf(proutSprintf, dst, fmt, args);

    if (ret > -1) {
```
while yer at it

src/libultra/libc/ll.c
```diff
@@ -1,47 +1,48 @@
...
     return l % r;
 }
 
-u64 __ull_div(u64 l, u64 r) {
+unsigned long long __ull_div(unsigned long long l, unsigned long long r) {
     return l / r;
 }
 
-s64 __ll_lshift(s64 l, s64 r) {
+unsigned long long __ll_lshift(unsigned long long l, unsigned long long r) {
```
Dragorn421, line 15: you changed the types from signed to unsigned here and in other functions below, was that intended?

include/ultra64/xstdio.h
Dragorn421, line 1: I think you forgot to `git rm include/ultra64/printf.h`

https://github.com/Thar0/oot/blob/libc-cleanup/include/ultra64/printf.h

src/code/load.c
```diff
@@ -46,7 +46,7 @@ s32 Overlay_Load(uintptr_t vromStart, uintptr_t vromEnd, void* vramStart, void*
             // "Clear BSS area (% 08x-% 08x)"
             osSyncPrintf("BSS領域をクリアします(%08x-%08x)\n", end, end + ovlRelocs->bssSize);
         }
-        bzero((void*)end, ovlRelocs->bssSize);
+        bzero((void*)end, (s32)ovlRelocs->bssSize);
```
Dragorn421, line 49: bruh

src/libultra/libc/xprintf.c
Thar0: I guess they wanted to avoid float operations, possibly for NaN/infinity handling. I tried to sync a bit more with ultralib's xprintf hence the macros

include/ultra64/libc.h
Thar0: idk, it's a libultra header so I could only speculate

src/libultra/libc/ll.c
Thar0: I wanted to better match ultralib, but looking at the libultra debug info the current types in ultralib seem a bit wrong

include/libc/math.h
Thar0: Specifically ISO C has `FLT_MAX` in `float.h`, which is not a header we currently have. I think it may be nice to add `float.h` and `limits.h` at some point, or I can attempt to add them now if it's much desired?

Dragorn421: I'm fine with the way it is now personally

Dragorn421 (approved):
Need to yeet the printf.h include directive

PR #1570, opened 2023-10-27
Player: Rename ItemActionFunc to UpperActionFunc

fig02:
I thought this was a good name at the time, however there is a distinction between an item action and the action function used for links upper body. 
Good examples that show why `ItemActionFunc` doesnt work are holding an actor (`func_808356E8`) and shielding (`func_80834B5C`). These are not tied to `ItemAction` values at all and exist as their own action functions for links upper body.

The related upper body variables in links struct were renamed to have "upper" as a prefix to keep them neat and consistent:
```c
    /* 0x06C8 */ SkelAnime upperSkelAnime;
    /* 0x070C */ Vec3s upperJointTable[PLAYER_LIMB_BUF_COUNT];
    /* 0x079C */ Vec3s upperMorphTable[PLAYER_LIMB_BUF_COUNT];
    /* 0x082C */ UpperActionFunc upperActionFunc;
    /* 0x0830 */ f32 upperAnimBlendWeight;
```

(also a side note, MM is already using "upper" and named these variables the same way)

### DIFF
--- a/include/z64player.h
+++ b/include/z64player.h
@@ -669,7 +669,7 @@ typedef struct {
 #define PLAYER_STATE3_7 (1 << 7)
 
 typedef void (*PlayerActionFunc)(struct Player*, struct PlayState*);
-typedef s32 (*ItemActionFunc)(struct Player*, struct PlayState*);
+typedef s32 (*UpperActionFunc)(struct Player*, struct PlayState*);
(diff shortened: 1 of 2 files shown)


AngheloAlf (approved):

krm01 (approved):
good change, i like the `upper` being a prefix on the struct fields.

Dragorn421 (approved):

PR #1572, opened 2023-11-04
Rename "ITEM_LAST_USED" 

fig02:
I dont know how we ended up calling this "last used". But I guess thats what the documentation process is all about :)

This item value serves the sole purpose of letting Link hold the sword in his hand without the shield too. It gets used when Link places the Master Sword into the pedestal when going back to child link.
It exists as its own item so that Link can use the item system to take it out, like any other item. The item is not hooked up to any sword swinging code though, so it cant be used as a regular sword.

Im not totally sold on calling it "_SWORD_CS". But the fact that its unusable as an actual sword item means its only useful in cutscenes. And the one use case it has in the game is a cutscene.

(There are some unrelated changes due to the formatter)

https://github.com/zeldaret/oot/assets/22045047/02e2bda8-c087-4029-b4d4-bc7453853059


AngheloAlf (approved):
The video is a bit cursed lol

Dragorn421 (approved):

Dragorn421 (approved):

PR #1573, opened 2023-11-04
Fix misc 21

Dragorn421:
Some of these inspired by gcc warnings


src/code/jpegdecoder.c
```diff
@@ -154,7 +154,7 @@ s32 JpegDecoder_ParseNextSymbol(JpegHuffmanTable* hTable, s16* outCoeff, s8* out
     if (sym) {
         *outCoeff = JpegDecoder_ReadBits(sym);
         if (*outCoeff < (1 << (sym - 1))) {
-            *outCoeff += (-1 << sym) + 1;
+            *outCoeff += (-1U << sym) + 1;
```
AngheloAlf, line 157: So no `0xFFFFFFFF`?

src/code/z_player_lib.c
```diff
@@ -1061,7 +1061,7 @@ s32 Player_OverrideLimbDrawGameplayCommon(PlayState* play, s32 limbIndex, Gfx**
         // Note: The increment would not be done for the root limb, even if it had a non-NULL `dList`.
         //       So if the root limb had a non-NULL `dList` (which is not the case in vanilla),
         //       an out-of-bounds write to `bodyPartsPos` would occur.
-        sCurBodyPartPos = &this->bodyPartsPos[-1];
+        sCurBodyPartPos = &this->bodyPartsPos[0] - 1;
```
AngheloAlf, line 1064: I can't believe gcc is fooled by this one lmao

src/code/jpegdecoder.c
Dragorn421: It's debatable but cf the commit message f4c7bf59777e5d08d41238908877d8033bb85c30

src/code/z_player_lib.c
Dragorn421: I guess it makes more sense like this since the later operation is `sCurBodyPartPos++` but yeah it's weird only one of the two is a warning when afaik both writings are exactly the same per the standard

src/code/jpegdecoder.c
AngheloAlf: I think that kind of information shouldn't get lots in a commit's message, would be nice to actually have it on the repo

Dragorn421: I could add a `// See ISO/IEC 10918-1, this part is "F.2.2.1 Huffman decoding of DC coefficients"` somewhere but not sure exactly
And like to be clear I'm only preferring "-1" here because I expect that's what the original code had (based on the standard having it), 0xFFFFFFFF would be fine but possibly misleading in recognizing the algorithm? 

The jpeg handling in general would benefit from a pass by someone who knows what they're doing / is willing to study the standard, across z_jpeg.c, jpegdecoder.c and the jpeg ucode, but that's probably not going to be me

AngheloAlf: I think that comment sounds fine.
Alternatively you could do something like `// -1U is preferred over 0xFFFFFFFF because that's what the jpeg standard uses` or something

src/code/z_player_lib.c
fig02: I prefer the way it was before, but if its for the sake of the warning i guess its fine. I just recognize way quicker what happening with a -1 index instead of having to think about what a -1 after the fact does

src/code/z_play.c
```diff
@@ -1463,7 +1464,9 @@ void Play_SpawnScene(PlayState* this, s32 sceneId, s32 spawn) {
 
     Play_InitScene(this, spawn);
 
-    osSyncPrintf("ROOM SIZE=%fK\n", func_80096FE8(this, &this->roomCtx) / 1024.0f);
+    size = func_80096FE8(this, &this->roomCtx);
+
+    osSyncPrintf("ROOM SIZE=%fK\n", size / 1024.0f);
```
fig02, line 1469: whats the motivation here? think it looked fine as is but not sure if theres a reason im missing

Dragorn421: That func_ call is doing something important about room loading (I forgot the details) and it's better not hidden in a printf, for readability

Also it's good practice not putting code with side effects in (debug) printfs, again for readability and possibly if one was to macro out printf calls in a non-debug build
(I don't know if that's the case for retail)

fig02: fair enough. Yeah it does do the actual allocation of the room buffers, not just get their size

fig02 (approved):

src/overlays/gamestates/ovl_file_choose/z_file_nameset_PAL.c
```diff
@@ -941,6 +943,9 @@ void FileSelect_DrawOptionsImpl(GameState* thisx) {
             gDPSetEnvColor(POLY_OPA_DISP++, 0, 0, 0, 255);
         }
 
+        //! @bug Mistakenly using sOptionsMenuHeaders instead of sOptionsMenuSettings for the height.
+        //! This is also an OOB read that happens to access the height of the first two elements in
+        //! sOptionsMenuSettings, and since all heights are 16, it works out anyway.
```
Roman971, line 948: This should probably become an AVOID_UB (at least eventually) since it relies on the variables being in the same order in memory, which isn't guaranteed

Roman971 (approved):

src/overlays/gamestates/ovl_file_choose/z_file_nameset_PAL.c
Dragorn421: I agree on the surface but

1) not sure if the `#if avoid_ub` guards should be around the whole command (lengthy copypaste) or only the argument (weird)
2) it feels weird to offer a fix for this under avoid_ub but not for the other similar occurrence above (which isn't UB because not OOB, just a copypaste mistake)
3) I'm not even sure what the `avoid_ub` behavior here should be (the semantics for `avoid_ub` is write alternative non-ub-invoking code that behaves the same as the ido-compiled code, not to fix things)

and I'm not a huge fan of `AVOID_UB` anyway so would rather let someone else deal with it if at all

fwiw this causes a warning when compiling with gcc (that's how I found this) so it's not entirely hidden

Roman971: I understand if you don't want to add it yourself, but I'll answer to your points anyway so we can figure out what we'll do in the future:

Since it's always 16 with IDO (as long as memory manipulation isn't involved at least), I'd say the `AVOID_UB` version would just use the right variable so the behaviour is equivalent and correct.
Also, if we feel it would be weird otherwise, I think it's okay to have a fix for the other one under `AVOID_UB`. Imo the goal of `AVOID_UB` should be to cover at least all cases of UB, but it's fine if it makes small improvements at the same time, as long as they make sense.
As for how to format it, I think it would be fine either way, tho I would probably duplicate the command

PR #1575, opened 2023-11-07
Player Docs: Some Item Action function names

fig02:
These are the low hanging fruit upper action/item action related functions.

`func_8083485C` is pretty simple in a vacuum. It lets the player interrupt the item action with shielding. But shielding itself is complicated and needs to be documented separately. So it is not named in this pass.

Bow/Sling/Boomerang/Hookshot are also a bit more complicated with upper actions that flow into each other. So those will be done its own pass as well.


krm01 (approved):
lgtm. only thing that's a little weird is `CarryActor` and `heldActor` differences. but thinking about it, `Player_UpperAction_HoldActor` is sort of ambiguous, and not immediately clear that it's talking about pots/boxes/etc and not weapons. so I think it's fine.

fig02:
I would have liked to change `heldActor` to simply `carryActor` or something of the sort, but link uses `heldActor` for things hes not necessarily carrying. For example, the hookshot actor that hes holding, or an arrow actor hes about to shoot from his bow.

`heldActor` is more generally an actor that is interacting with links right hand (see relevant code in `Player_PostLimbDrawGameplay` for `PLAYER_LIMB_R_HAND`)

While carrying is a subset of that where he is lifting an actor over his head (and it is still positioned relative to his right hand, with an offset to make it seem a bit more centered)

fig02:
(this is a good reminder to myself to change the name of `sGetItemRefPos` soon, as it is much more than just the get item position lol)

fig02:
tbh, would be down `rightHandActor`, being more literal in this case seems like it could be more clear about its purpose (and specific about what hand its associated with). While "held" is more about how right hand actors happen to be used. 

And like, I guess you can say the arrow actor is "held", but its more like its held by the bow? lol

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -2979,12 +2986,12 @@ s32 func_80835644(PlayState* play, Player* this, Actor* arg2) {
...
 void func_80835688(Player* this, PlayState* play) {
     if (!func_80835644(play, this, this->heldActor)) {
-        Player_SetUpperActionFunc(this, func_808356E8);
+        Player_SetUpperActionFunc(this, Player_UpperAction_CarryActor);
         LinkAnimation_PlayLoop(play, &this->upperSkelAnime, &gPlayerAnim_link_normal_carryB_wait);
     }
 }
 
-s32 func_808356E8(Player* this, PlayState* play) {
+s32 Player_UpperAction_CarryActor(Player* this, PlayState* play) {
```
Dragorn421, line 2994: it seems (looking at `func_8083A0F4` which does `Player_SetupAction(Player_Action_80846050) > func_80835688 > Player_SetUpperActionFunc(Player_UpperAction_CarryActor)`) this is indeed the main/default path for carrying stuff,
but `func_8083A0F4` also seems to have a few hardcoded things where the player "carries" but doesn't use `Player_UpperAction_CarryActor` (notably silver rocks? `ACTOR_EN_ISHI` large)

I guess we're sweeping those exceptions under the rug for now

Dragorn421 (approved):

src/overlays/actors/ovl_player_actor/z_player.c
fig02: the special cases are all full action functions, as in his whole body is involved. The upper action is only for his upper body and you can still, for example, walk around

fig02: to be more clear, something like the silver rock or the gold gauntlets block are "too heavy" and use their own action function where you are stuck in place while holding the actor, while the stuff that use `Player_UpperAction_CarryActor` let you still move around and such

PR #1577, opened 2023-11-13
[Doc] Actor Bg_Bdan_Objects (Jabu Jabu misc dungeon actors)

krm01:
I don't have clang-format-14 and can't seem to install it from packages, so hopefully its ok with just clang-format version 10? if not i can figure out how to get 14.

Anyway this actor is 4 unrelated things:
- The elevator above the pool of water with an octo that simply oscillates up and down
- The waterbox level controller in the room with the timed switch puzzle that changes water level
- The odd angled platform that descends into the 2nd room from the entrance once stepped on
- The platform used by Ruto and the BigOcto miniboss fight

I struggled a bit with finding concise names, especially for the BigOcto variant, so any improvements welcome... All the other types only have a couple states they can be in, but the BigOcto version has like 6 so I added a prefix to better separate them.

Also there's this bizarre `BgBdanObjects_GetProperty` / `SetProperty` thing going on here, I tried my best to make it understandable. Extremely weird pattern I haven't come across before...


src/overlays/actors/ovl_Bg_Bdan_Objects/z_bg_bdan_objects.c
```diff
@@ -10,27 +10,38 @@
...
+typedef enum {
+    GET_PROP_CAMERA_SETTING_NORMAL0 = 0,
+    GET_PROP_CAMERA_SETTING_DUNGEON1 = 3,
+    GET_PROP_WATCHED_BIGOCTO_INTRO_CUTSCENE = 4
+} BgBdanObjectPropertyGetter;
+typedef enum {
+    SET_PROP_CAMERA_SETTING_NORMAL0 = 1,
+    SET_PROP_CAMERA_SETTING_DUNGEON1 = 2,
+    SET_PROP_WATCHED_BIGOCTO_INTRO_CUTSCENE = 4
+} BgBdanObjectPropertySetter;
```
fig02, line 22: Yeah im still not sure what to make of this weird system lmao

At the very least, we like prefix all enums with a namespace prefix to prevent possible name clashes. Even though obviously these names are hyper specific and would never clash, we just do it everywhere

You can use a "friendly" name for this prefix, so for example `JABU_OBJECTS_` would be fine.
You can take other shortening measures like changing "CAMERA" to "CAM" if you want

krm01: Right? they could have at least used the same value for the getter/setter... maybe the fact that they didn't is some clue that I'm not understanding. Regardless, added the prefix 👍

src/overlays/actors/ovl_Bg_Bdan_Objects/z_bg_bdan_objects.c
```diff
@@ -75,29 +86,29 @@ static Gfx* sDLists[] = {
...
             return -1;
     }
 }
 
-void BgBdanObjects_SetContactRu1(BgBdanObjects* this, s32 arg1) {
+void BgBdanObjects_SetProperty(BgBdanObjects* this, s32 arg1) {
     switch (arg1) {
-        case 1:
+        case JABU_OBJECTS_SET_PROP_CAM_SETTING_NORMAL0:
             this->cameraSetting = CAM_SET_NORMAL1;
```
Dragorn421, line 106: this sets the camera setting to normal1 not normal0

src/overlays/actors/ovl_Bg_Bdan_Objects/z_bg_bdan_objects.c
```diff
@@ -10,27 +10,38 @@
...
+typedef enum {
+    JABU_OBJECTS_GET_PROP_CAM_SETTING_NORMAL0 = 0,
+    JABU_OBJECTS_GET_PROP_CAM_SETTING_DUNGEON1 = 3,
+    JABU_OBJECTS_GET_PROP_WATCHED_BIGOCTO_INTRO_CUTSCENE = 4
+} BgBdanObjectPropertyGetter;
+typedef enum {
+    JABU_OBJECTS_SET_PROP_CAM_SETTING_NORMAL0 = 1,
+    JABU_OBJECTS_SET_PROP_CAM_SETTING_DUNGEON1 = 2,
+    JABU_OBJECTS_SET_PROP_WATCHED_BIGOCTO_INTRO_CUTSCENE = 4
+} BgBdanObjectPropertySetter;
```
Dragorn421, line 22:
```suggestion
} BgBdanObjectsPropertyGetter;
typedef enum {
    JABU_OBJECTS_SET_PROP_CAM_SETTING_NORMAL0 = 1,
    JABU_OBJECTS_SET_PROP_CAM_SETTING_DUNGEON1 = 2,
    JABU_OBJECTS_SET_PROP_WATCHED_BIGOCTO_INTRO_CUTSCENE = 4
} BgBdanObjectsPropertySetter;
```

src/overlays/actors/ovl_Bg_Bdan_Objects/z_bg_bdan_objects.c
```diff
@@ -114,49 +125,49 @@ void BgBdanObjects_Init(Actor* thisx, PlayState* play) {
...
-        if (thisx->params == 1) {
+        if (thisx->params == BDAN_OBJECT_TYPE_SMALL_AUTO_ELEVATOR) {
             CollisionHeader_GetVirtual(&gJabuElevatorCol, &colHeader);
             this->timer = 512;
-            this->switchFlag = 0;
-            this->actionFunc = func_8086C874;
-        } else {
+            this->camChangeTimer = 0;
+            this->actionFunc = BgBdanObjects_ElevatorOscillate;
+        } else { // thisx->params == BDAN_OBJECT_TYPE_FALLING_PLATFORM
```
Dragorn421, line 164:
```suggestion
        } else { // BDAN_OBJECT_TYPE_FALLING_PLATFORM
```
is enough

src/overlays/actors/ovl_Bg_Bdan_Objects/z_bg_bdan_objects.h
```diff
@@ -6,12 +6,22 @@
 
 struct BgBdanObjects;
 
+typedef enum {
+    BDAN_OBJECT_TYPE_BIG_OCTO_PLATFORM = 0,
+    BDAN_OBJECT_TYPE_SMALL_AUTO_ELEVATOR = 1,
+    BDAN_OBJECT_TYPE_WATERBOX_HEIGHT_CHANGER = 2,
+    BDAN_OBJECT_TYPE_FALLING_PLATFORM = 3
+} BgBdanObjectType;
```
Dragorn421, line 14:
```suggestion
} BgBdanObjectsType;
```

src/overlays/actors/ovl_Bg_Bdan_Objects/z_bg_bdan_objects.h
```diff
@@ -6,12 +6,22 @@
 
 struct BgBdanObjects;
 
+typedef enum {
+    BDAN_OBJECT_TYPE_BIG_OCTO_PLATFORM = 0,
+    BDAN_OBJECT_TYPE_SMALL_AUTO_ELEVATOR = 1,
+    BDAN_OBJECT_TYPE_WATERBOX_HEIGHT_CHANGER = 2,
+    BDAN_OBJECT_TYPE_FALLING_PLATFORM = 3
```
Dragorn421, line 13: you used `JABU_OBJECTS_` as a prefix for the get/set property enums, it'd probably be a good idea to use the same as here

so `JABU_OBJECTS` or `BDAN_OBJECTS` (mind the `S`, following the actor's debug name, though that may be up to discussion)

src/overlays/actors/ovl_Bg_Bdan_Objects/z_bg_bdan_objects.h
```diff
@@ -6,12 +6,22 @@
...
+
 typedef void (*BgBdanObjectsActionFunc)(struct BgBdanObjects*, PlayState*);
 
 typedef struct BgBdanObjects {
     /* 0x0000 */ DynaPolyActor dyna;
     /* 0x0164 */ BgBdanObjectsActionFunc actionFunc;
+    union {
     /* 0x0168 */ u8 switchFlag;
+    /* 0x0168 */ u8 camChangeTimer;
+    };
```
Dragorn421, line 24: I'm not opposed to using unions, on the contrary, but we avoided these constructs before so afaik there aren't many examples if any of it in the repo

the number one issue is it's not obvious with an anonymous union that the two fields share the same memory (which harms readability). Naming the union fixes that (e.g. `var` then `this->var.switchFlag`, `this->var.camChangeTimer`) but finding a name for it is *bruh* (and again no example of it afaik)

@fig02 ?

src/overlays/actors/ovl_Bg_Bdan_Objects/z_bg_bdan_objects.h
Dragorn421, line 13: can leave out the = 0, 1, etc

src/overlays/actors/ovl_Bg_Bdan_Objects/z_bg_bdan_objects.c
```diff
@@ -378,50 +398,50 @@ void func_8086C874(BgBdanObjects* this, PlayState* play) {
...
+void BgBdanObjects_WaitForSwitch(BgBdanObjects* this, PlayState* play) {
     if (Flags_GetSwitch(play, this->switchFlag)) {
         this->timer = 100;
-        this->actionFunc = func_8086C9F0;
+        this->actionFunc = BgBdanObjects_ChangeWaterboxLevel;
     }
 }
 
-void func_8086C9F0(BgBdanObjects* this, PlayState* play) {
+void BgBdanObjects_ChangeWaterboxLevel(BgBdanObjects* this, PlayState* play) {
```
Dragorn421, line 408: "water box" so
```suggestion
void BgBdanObjects_ChangeWaterBoxLevel(BgBdanObjects* this, PlayState* play) {
```
skawoUHHUH

src/overlays/actors/ovl_Bg_Bdan_Objects/z_bg_bdan_objects.c
```diff
@@ -278,63 +297,64 @@ void func_8086C3D8(BgBdanObjects* this, PlayState* play) {
...
     CollisionCheck_SetAT(play, &play->colChkCtx, &this->collider.base);
     if (Flags_GetClear(play, this->dyna.actor.room)) {
         Flags_SetSwitch(play, this->switchFlag);
         this->dyna.actor.home.rot.y = (s16)(this->dyna.actor.shape.rot.y + 0x2000) & 0xC000;
-        this->actionFunc = func_8086C6EC;
+        this->actionFunc = BgBdanObjects_SinkToFloorHeight;
     } else {
         this->dyna.actor.shape.rot.y += this->dyna.actor.world.rot.y;
+        // This plays a sound effect id (0x2063) that is not present in environmentbank_table?
         func_800F436C(&this->dyna.actor.projectedPos, 0x2063, ABS(this->dyna.actor.world.rot.y) / 512.0f);
```
Dragorn421, line 334: @engineer124 if you're interested (if not interesting/interested feel free to ignore)

(is this just `NA_SE_EV_ROLL_STAND - SFX_FLAG` ?)

Dragorn421 (commented):
Nice juicy actor here, good job
Partial review ahead

src/overlays/actors/ovl_Bg_Bdan_Objects/z_bg_bdan_objects.h
krm01: hmm I guess I should stick with `BDAN_OBJECTS` for now? seems more correct, maybe some eventual rename of the actor in the future to friendly common names will then go to JABU.

fig02: imo its fine to use friendly names for all enums in an actor, but think everyone has their own opinions/preferences on this

src/overlays/actors/ovl_Bg_Bdan_Objects/z_bg_bdan_objects.h
fig02: if we want to start doing the player approach everywhere, yeah, a named union would be better. I dont think we really ironed out that in discussions though. 
Id be fine with a named enum for all cases like this personally

src/overlays/actors/ovl_Bg_Bdan_Objects/z_bg_bdan_objects.c
krm01: I think you're right about this, considering 
1) `NA_SE_EV_ROLL_STAND` is otherwise unused
2) related sound effect `NA_SE_EV_ROLL_STAND_2` is always used with SFX_FLAG

going to change it in upcoming commit

src/overlays/actors/ovl_Bg_Bdan_Objects/z_bg_bdan_objects.h
krm01: changed to friendlies in aefde5c

src/overlays/actors/ovl_Bg_Bdan_Objects/z_bg_bdan_objects.h
krm01: named the union `var`, but going to leave this conversation unresolved in case there are other opinions, since we don't really have a strong consensus

src/overlays/actors/ovl_Bg_Bdan_Objects/z_bg_bdan_objects.h
```diff
@@ -1,17 +1,27 @@
-#ifndef Z_BG_BDAN_OBJECTS_H
-#define Z_BG_BDAN_OBJECTS_H
+#ifndef Z_BG_JABU_OBJECTS_H
+#define Z_BG_JABU_OBJECTS_H
```
fig02, line 3: header guards should match the file name, rest look ok though

src/overlays/actors/ovl_Bg_Bdan_Objects/z_bg_bdan_objects.h
```diff
@@ -1,17 +1,27 @@
...
+
 typedef void (*BgBdanObjectsActionFunc)(struct BgBdanObjects*, PlayState*);
 
 typedef struct BgBdanObjects {
     /* 0x0000 */ DynaPolyActor dyna;
     /* 0x0164 */ BgBdanObjectsActionFunc actionFunc;
+    union {
     /* 0x0168 */ u8 switchFlag;
+    /* 0x0168 */ u8 camChangeTimer;
+    } var;
```
fig02, line 24: indent the two members :D

krm01: oops. letting me down clang-format

PR #1581, opened 2023-11-14
Doc Actor: En_Brob (Jabu Jabu Wobble Flesh Block)

hiisuya:
First actor documentation attempt, feel free to let me know anything I've done wrong, should be changed, or that I missed. Suggestion welcome as well. 

Don't be afraid to tear into it, no better way to learn than through experience.

I was very unsure about the DL/Limb names, there were limbs inside of limbs that did nothing, so I checked another actor and they seemed to follow this pattern.

### DIFF
--- a/assets/xml/objects/object_brob.xml
+++ b/assets/xml/objects/object_brob.xml
@@ -6,7 +6,7 @@
         <DList Name="gBrobMiddleSegmentDL" Offset="0x9D0"/>
         <DList Name="gBrobUpperSegmentDL" Offset="0xAE8"/>
         <DList Name="gBrobTopSegmentDL" Offset="0xC00"/>
-        <Texture Name="object_brob_Tex_000D48" OutName="tex_00000D48" Format="rgba16" Width="32" Height="32" Offset="0xD48"/>
+        <Texture Name="gBrobFleshTex" OutName="flesh" Format="rgba16" Width="32" Height="32" Offset="0xD48"/>
         <Limb Name="gBrobRootLimb" LimbType="Standard" Offset="0x1548"/>
         <Limb Name="gBrobBaseSegmentRootLimb" LimbType="Standard" Offset="0x1554"/>
         <Limb Name="gBrobBaseSegmentLimb" LimbType="Standard" Offset="0x1560"/>
@@ -18,7 +18,7 @@
         <Limb Name="gBrobTopSegmentLimb" LimbType="Standard" Offset="0x15A8"/>
         <Skeleton Name="gBrobSkel" Type="Flex" LimbType="Standard" Offset="0x15D8"/>
         <Animation Name="gBrobShockAnim" Offset="0x1678"/>
-        <Animation Name="gBrobExtrudeAnim" Offset="0x1750"/>
+        <Animation Name="gBrobMoveUpAnim" Offset="0x1750"/>
         <Animation Name="gBrobWobbleAnim" Offset="0x1958"/>
         <Collision Name="gBrobCol" Offset="0x1A70"/>
     </File>
--- a/assets/xml/objects/object_brob.xml
+++ b/assets/xml/objects/object_brob.xml
@@ -1,21 +1,21 @@
 <Root>
     <File Name="object_brob" Segment="6">
         <Animation Name="gBrobStunnedAnim" Offset="0x290"/>
-        <DList Name="object_brob_DL_0007E0" Offset="0x7E0"/>
-        <DList Name="object_brob_DL_000898" Offset="0x898"/>
-        <DList Name="object_brob_DL_0009D0" Offset="0x9D0"/>
-        <DList Name="object_brob_DL_000AE8" Offset="0xAE8"/>
-        <DList Name="object_brob_DL_000C00" Offset="0xC00"/>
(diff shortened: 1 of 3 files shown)


src/overlays/actors/ovl_En_Brob/z_en_brob.c
```diff
@@ -322,7 +325,7 @@ void EnBrob_Draw(Actor* thisx, PlayState* play) {
     EnBrob* this = (EnBrob*)thisx;
 
     Gfx_SetupDL_25Opa(play->state.gfxCtx);
-    Matrix_Translate(0.0f, this->unk_1AE, 0.0f, MTXMODE_APPLY);
+    Matrix_Translate(0.0f, this->yOffset, 0.0f, MTXMODE_APPLY);
     SkelAnime_DrawFlexOpa(play, this->skelAnime.skeleton, this->skelAnime.jointTable, this->skelAnime.dListCount, NULL,
```
fig02, line 329: maybe `modelOffsetY`, to make clear its only the model/skeleton and not the whole actor? this is prob okay though

src/overlays/actors/ovl_En_Brob/z_en_brob.c
```diff
@@ -103,82 +104,82 @@ void EnBrob_Destroy(Actor* thisx, PlayState* play) {
...
+            EnBrob_SetupExtrude(this, play);
         }
     } else if (this->timer >= 81) {
         this->dyna.actor.colorFilterTimer = 80;
     }
 }
 
-void func_809CB114(EnBrob* this, PlayState* play) {
+void EnBrob_Extrude(EnBrob* this, PlayState* play) {
     f32 curFrame;
```
fig02, line 167: trying to think if theres a less technical term for this, like "Grow" or "MoveUp". But it is only the model moving up...
Extrude is probably fine, people who are using 3d modeling tools are already very familiar with the term lol, just wondering if there are other ideas

fig02 (commented):
Just a couple comments. Overall great job, you are very in-tune to our naming conventions already :D 

I did not really review the asset naming, if someone else wants to take a closer look at that

src/overlays/actors/ovl_En_Brob/z_en_brob.c
hiisuya: I did this based off a similar translation I noticed in Goma with `eggYOffset`, modified slightly as `brobYOffset` felt redundant. I can change it to `modelOffsetY` as that does align more with what is actually happening.

src/overlays/actors/ovl_En_Brob/z_en_brob.c
hiisuya: I did consider "Grow" and "Shrink" at first until I noticed it really was just a translation of it going up and down. Extrude might not be perfect as it might give off the impression of it being physically extended and not just moved.

I'm open to suggestions. `MoveUp` is fine, though I would probably change `Retract` to `MoveDown` then as well just to stay consistent

src/overlays/actors/ovl_En_Brob/z_en_brob.c
fig02: yeah id say specifying model is good to be clear its just an offset of the visuals and nothing else

src/overlays/actors/ovl_En_Brob/z_en_brob.c
fig02: alright, im fine with move up and move down, but if you want to wait for more opinions thats fair

fig02 (approved):

Dragorn421 (commented):
Good job :)

`EnBrob_Stunned` is a bit weird because it's not the stunned state itself, it's only the "goes to stunned" animation and the actual stunned state (including when you can walk on the blob) is part of `EnBrob_Idle`, but oh well

I only have one comment on the limb symbol names

assets/xml/objects/object_brob.xml
```diff
@@ -1,25 +1,25 @@
...
+        <DList Name="gBrobLowerSegmentDL" Offset="0x898"/>
+        <DList Name="gBrobMiddleSegmentDL" Offset="0x9D0"/>
+        <DList Name="gBrobUpperSegmentDL" Offset="0xAE8"/>
+        <DList Name="gBrobTopSegmentDL" Offset="0xC00"/>
+        <Texture Name="gBrobFleshTex" OutName="flesh" Format="rgba16" Width="32" Height="32" Offset="0xD48"/>
+        <Limb Name="gBrobRootLimb" LimbType="Standard" Offset="0x1548"/>
+        <Limb Name="gBrobBaseSegmentRootLimb" LimbType="Standard" Offset="0x1554"/>
+        <Limb Name="gBrobBaseSegmentLimb" LimbType="Standard" Offset="0x1560"/>
+        <Limb Name="gBrobLowerSegmentLimb" LimbType="Standard" Offset="0x156C"/>
+        <Limb Name="gBrobLowerSegmentRootLimb" LimbType="Standard" Offset="0x1578"/>
```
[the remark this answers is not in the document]
hiisuya: Oh thank you, good catch

Dragorn421 (approved):

PR #1584, opened 2023-11-19
Doc Actor: En_Ma1 (Child Malon)

hiisuya:
Here's a shot at doc'ing Malon. If this goes well, I'll use this as a basis and doc both `ma2` and `ma3` after this as well. But for now, I just wanted to start with 1 and see how it goes.

Some notable points to maybe look into:
- Appreciate help on the naming conventions for the `INFTABLE` and `EVENTCHKINF`. I think I did a fairly good job, but any help is welcome and appreciated
-~Change from `this->isNotSinging` to `this->isSinging` with the defines. I felt that the current setup was fairly confusing, especially with how `Audio_ToggleMalonSinging` uses its parameter~
- Didn't realize this wasn't matching so I changed it back oops

That is all please and thank you


src/overlays/actors/ovl_En_Ma1/z_en_ma1.c
```diff
@@ -71,7 +74,7 @@ static AnimationFrameCountInfo sAnimationInfo[] = {
     { &gMalonChildSingAnim, 1.0f, ANIMMODE_LOOP, -10.0f },
 };
 
-static Vec3f D_80AA16B8 = { 800.0f, 0.0f, 0.0f };
+static Vec3f sMultVec = { 800.0f, 0.0f, 0.0f };
```
Dragorn421, line 77: you can leave this unnamed, it's not super helpful

it's the position in limb space (more specifically, I assume the head limb) of the actor's focus position (i.e. the position link targets)
imo these should be named like `focus[Limb]LocalPos`, like `focusHeadLocalPos`, but that requires knowing what limb they use for the transform, so a lot of assets docs

anyway this data can be moved into the function if the mouth and eye textures arrays are also moved down. It'll probably be done eventually, or you can

src/overlays/actors/ovl_En_Ma1/z_en_ma1.c
```diff
@@ -402,15 +411,19 @@ void EnMa1_Update(Actor* thisx, PlayState* play) {
...
+
     if (this->actionFunc != EnMa1_DoNothing) {
         Npc_UpdateTalking(play, &this->actor, &this->interactInfo.talkState, (f32)this->collider.dim.radius + 30.0f,
                           EnMa1_GetTextId, EnMa1_UpdateTalkState);
     }
-    func_80AA0B74(this);
-    func_80AA0AF4(this, play);
+
+    EnMa1_HandleSinging(this);
+    EnMa1_HandleTracking(this, play);
```
Dragorn421, line 426: it's more in line with the overall names to use "Update" instead of "Handle" where appropriate, I think here Update works

Dragorn421 (commented):
Nice, there's a lot of undocumented NPCs, good to see that the recent-ish documentation of the `Npc_` system and talk system helps a lot

The names you used for the save flags (eventchkinf & co.) are fine to me (in style at least, didn't check all correctness yet)

src/overlays/actors/ovl_En_Ma1/z_en_ma1.c
```diff
@@ -242,19 +253,17 @@ void func_80AA0AF4(EnMa1* this, PlayState* play) {
...
 
-void func_80AA0B74(EnMa1* this) {
+void EnMa1_HandleSinging(EnMa1* this) {
     if (this->skelAnime.animation == &gMalonChildSingAnim) {
         if (this->interactInfo.talkState == NPC_TALK_STATE_IDLE) {
             if (this->isNotSinging) {
-                // Turn on singing
                 this->isNotSinging = false;
-                Audio_ToggleMalonSinging(false);
+                Audio_ToggleMalonSinging(TURN_ON_MALON_SINGING);
```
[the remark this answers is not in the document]
hiisuya: The issue I had when doc'ing this is that the double negative followed by a toggle where `false` is actually `on` felt completely unintuitive (this is why I had done a non-matching swap of logic initially before that was caught). While I do agree it does just take a minute to look the function over and figure it out, it's not very easy on the eyes to immediately determine it.

That's just my thoughts, however, if you think it's not necessary I can totally remove it that's not a problem

src/overlays/actors/ovl_En_Ma1/z_en_ma1.c
hiisuya: I see, I just noticed something similar was done in another actor so I just followed suit. I'll revert it 👍

fig02: the stuff he mentions about moving the variables would look like this.

Basically, data that gets referenced in only 1 function can be moved into that function. And we usually prefer this so you dont have to go looking around the file for the data. You just need to make sure to preserve the datas order in the file.
The order is still vec -> mouth textures -> eye textures.

If you dont want to mess with doing it this way thats fine, but I think itd be good to have if youre up for it
```c
void EnMa1_PostLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Vec3s* rot, void* thisx) {
    EnMa1* this = (EnMa1*)thisx;
    Vec3f vec = { 800.0f, 0.0f, 0.0f };

    if (limbIndex == 15) {
        Matrix_MultVec3f(&vec, &this->actor.focus.pos);
    }
}

void EnMa1_Draw(Actor* thisx, PlayState* play) {
    static void* sMouthTextures[] = {
        gMalonChildNeutralMouthTex,
        gMalonChildSmilingMouthTex,
        gMalonChildTalkingMouthTex,
    };
    static void* sEyeTextures[] = {
        gMalonChildEyeOpenTex,
        gMalonChildEyeHalfTex,
        gMalonChildEyeClosedTex,
    };
    EnMa1* this = (EnMa1*)thisx;
    Camera* activeCam;
    f32 distFromCamEye;
    s32 pad;

    OPEN_DISPS(play->state.gfxCtx, "../z_en_ma1.c", 1226);
}
```

src/overlays/actors/ovl_En_Ma1/z_en_ma1.c
fig02: +1 on this, UpdateSinging and UpdateTracking seem good, so it matches the fact that its called directly by the Update function

src/overlays/actors/ovl_En_Ma1/z_en_ma1.c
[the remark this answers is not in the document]
fig02: I agree its confusing but dont think the best solution is making a new define for true and false.

This seems like a problem that could possibly be handled with a better name for the audio function, but I dont have any ideas at the moment. The argument is at least named aptly, and if the user is confused they can hover over the function for the explanation of the fact that its backwards.

Also, I think `isNotSinging` could be `singingDisabled`. Its still a negative word, cause it kinda has to be, but "is not x" is clunky and makes understanding it harder I feel

include/z64save.h
```diff
@@ -662,9 +662,9 @@ typedef enum {
 #define INFTABLE_76 0x76
 #define INFTABLE_77 0x77
 #define INFTABLE_TALKED_TO_TALON_IN_RANCH_HOUSE 0x7E
-#define INFTABLE_84 0x84
-#define INFTABLE_85 0x85
-#define INFTABLE_8B 0x8B
+#define INFTABLE_TALKED_TO_MALON_FIRST_TIME 0x84
+#define INFTABLE_TOLD_EPONA_IS_SCARED 0x85
+#define INFTABLE_ENTERED_HYRULE_CASTLE_SCENE 0x8B
 #define INFTABLE_8C 0x8C
```
fig02, line 668: `INFTABLE_ENTERED_HYRULE_CASTLE_SCENE` should probably still be named centered around malon somehow. With the current name, youd expect some higher level code to be watching for link to enter hyrule castle *in general*. But really its "malon spawned in hyrule castle for the first time"

I know in practice malon spawning in hyrule castle is always going to coincide with link going there the first time. But I think its an important distinction still

include/z64save.h
```diff
@@ -388,13 +388,13 @@ typedef enum {
...
 #define EVENTCHKINF_11 0x11
-#define EVENTCHKINF_12 0x12
+#define EVENTCHKINF_RECEIVED_WEIRD_EGG 0x12
 #define EVENTCHKINF_TALON_WOKEN_IN_CASTLE 0x13
 #define EVENTCHKINF_TALON_RETURNED_FROM_CASTLE 0x14
-#define EVENTCHKINF_15 0x15
-#define EVENTCHKINF_16 0x16
+#define EVENTCHKINF_TALKED_TO_CHILD_MALON_AT_RANCH 0x15
+#define EVENTCHKINF_CAN_OBTAIN_EPONA_SONG 0x16
 #define EVENTCHKINF_EPONA_OBTAINED 0x18
```
fig02, line 398: EPONA_SONG -> EPONAS_SONG
and maybe instead of "can obtain", "can learn"? cause you have to pull your ocarina out and go through the whole song learning sequence, then you obtain it later

src/overlays/actors/ovl_En_Ma1/z_en_ma1.c
```diff
@@ -58,10 +58,10 @@ static ColliderCylinderInit sCylinderInit = {
 static CollisionCheckInfoInit2 sColChkInfoInit = { 0, 0, 0, 0, MASS_IMMOVABLE };
 
 typedef enum {
-    /* 0 */ ENMA1_ANIM_0,
-    /* 1 */ ENMA1_ANIM_1,
-    /* 2 */ ENMA1_ANIM_2,
-    /* 3 */ ENMA1_ANIM_3
+    /* 0 */ ENMA1_ANIM_UNUSED,
```
fig02, line 61: Even though its unused, its still nameable. IDLE_NOMORPH seems fine. if you want to comment on its unused-ness, `// unused` next to it is fine, but dont think its all that necessary personally

src/overlays/actors/ovl_En_Ma1/z_en_ma1.c
```diff
@@ -437,14 +433,25 @@ s32 EnMa1_OverrideLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Vec3f* p
...
 
 void EnMa1_Draw(Actor* thisx, PlayState* play) {
+    static void* sMouthTextures[] = {
+        gMalonChildNeutralMouthTex,
+        gMalonChildSmilingMouthTex,
+        gMalonChildTalkingMouthTex,
+    };
+
+    static void* sEyeTextures[] = {
+        gMalonChildEyeOpenTex,
```
fig02, line 451: tiny thing, but you can remove the line break between these two. 
We always keep the stack of variables at the top of a function bunched together, and then put a line break to signal the code is now starting

fig02 (approved):

src/overlays/actors/ovl_En_Ma1/z_en_ma1.c
```diff
@@ -58,10 +58,10 @@ static ColliderCylinderInit sCylinderInit = {
...
 
 typedef enum {
-    /* 0 */ ENMA1_ANIM_0,
-    /* 1 */ ENMA1_ANIM_1,
-    /* 2 */ ENMA1_ANIM_2,
-    /* 3 */ ENMA1_ANIM_3
+    /* 0 */ ENMA1_ANIM_NOMORPH,
+    /* 1 */ ENMA1_ANIM_IDLE,
+    /* 2 */ ENMA1_ANIM_SING_NOMORPH,
+    /* 3 */ ENMA1_ANIM_SING
```
Thar0, line 64: I think `ENMA1` -> `MALON_` is fine, considering it's local to the file it shouldn't be confused with the other actors

src/overlays/actors/ovl_En_Ma1/z_en_ma1.c
```diff
@@ -402,15 +394,19 @@ void EnMa1_Update(Actor* thisx, PlayState* play) {
...
     Collider_UpdateCylinder(&this->actor, &this->collider);
     CollisionCheck_SetOC(play, &play->colChkCtx, &this->collider.base);
+
     SkelAnime_Update(&this->skelAnime);
     EnMa1_UpdateEyes(this);
+
     this->actionFunc(this, play);
+
     if (this->actionFunc != EnMa1_DoNothing) {
         Npc_UpdateTalking(play, &this->actor, &this->interactInfo.talkState, (f32)this->collider.dim.radius + 30.0f,
```
Thar0, line 404: Is the cast on `this->collider.dim.radius` required for matching? I don't think it should be needed so if not it would be nice to have those removed from several places

src/overlays/actors/ovl_En_Ma1/z_en_ma1.c
```diff
@@ -437,14 +433,24 @@ s32 EnMa1_OverrideLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Vec3f* p
 
 void EnMa1_PostLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Vec3s* rot, void* thisx) {
     EnMa1* this = (EnMa1*)thisx;
-    Vec3f vec = D_80AA16B8;
+    Vec3f vec = { 800.0f, 0.0f, 0.0f };
```
Thar0, line 436: `offset` would be a slightly more descriptive name than `vec`, while you're here

src/overlays/actors/ovl_En_Ma1/z_en_ma1.c
```diff
@@ -437,14 +433,24 @@ s32 EnMa1_OverrideLimbDraw(PlayState* play, s32 limbIndex, Gfx** dList, Vec3f* p
...
 
     if (limbIndex == 15) {
```
Thar0, line 438: The limb index here and in OverrideLimbDraw should use an enum value from the skeleton. If you to go `object_ma1.xml` and add an `EnumName` attribute to the skeleton, e.g.
`<Skeleton Name="gMalonChildSkel" Type="Flex" LimbType="Standard" EnumName="ChildMalonLimb" Offset="0x8460"/>`
it should generate an enum if you re-run asset extraction (either `make setup` again or with `extract_assets.py -s objects/object_ma1`)
Then replace the raw values that `limbIndex` is being compared to with items from that enum.

Thar0 (changes requested):
Overall looks good, there's a couple of things that could be done in addition to make it more complete.

src/overlays/actors/ovl_En_Ma1/z_en_ma1.c
fig02: It is not needed. This issue happens in quite a few different files, just early mips2c artifacts https://github.com/zeldaret/oot/pull/1595

No action needed, I made a PR to handle all of them (at least the ones related to `this->collider`)

Thar0 (approved):

PR #1587, opened 2023-11-26
Document a bug in Actor_ChangeCategory where actors may not be updated correctly

cadmic:
If `Actor_ChangeCategory` is called during an actor update function, the loop in `Actor_UpdateAll` will continue from the next actor in the new category, rather than the next actor in the old category. This might cause some actors in the old category to be skipped over and not updated, or some actors in the new category to be updated twice.

I first noticed this with the pots in the spirit temple lobby:

https://github.com/zeldaret/oot/assets/146315916/fa1b4e39-cabc-4efa-ad72-916626670ce9

The two watches are the heights of the two pots, and here I activate the left pot before the right pot. When the right pot activates (and calls `Actor_ChangeCategory` to become an enemy), the left pot (which is already an enemy) gets updated twice, once when enemies are updated and once after the right pot is updated. So the left pot's y coordinate increases by 6 instead of 3 on this frame.

I think this is why this function was changed in MM: https://discord.com/channels/688807550715560050/688851976746041391/903481092508631091

### DIFF
--- a/src/code/z_actor.c
+++ b/src/code/z_actor.c
@@ -3397,10 +3397,12 @@ Actor* func_80033684(PlayState* play, Actor* explosiveActor) {
  * This is done by moving it to the corresponding category list and setting its category variable accordingly.
  */
 void Actor_ChangeCategory(PlayState* play, ActorContext* actorCtx, Actor* actor, u8 actorCategory) {
-    //! @bug If Actor_ChangeCategory is called during an actor update function, the loop in
+    //! @bug Calling this function immediately moves an actor from one category list to the other.
+    //! So, if Actor_ChangeCategory is called during an actor update function, the inner loop in
     //! Actor_UpdateAll will continue from the next actor in the new category, rather than the next
-    //! actor in the old category. This might cause some actors in the old category to be skipped
-    //! over and not updated, or some actors in the new category to be updated twice.
+    //! actor in the old category. This will cause any actors after this one in the old category to
+    //! be skipped over and not updated, and any actors in the new category to be updated more than
+    //! once.
     Actor_RemoveFromCategory(play, actorCtx, actor);
     Actor_AddToCategory(actorCtx, actor, actorCategory);
 }
--- a/src/code/z_actor.c
+++ b/src/code/z_actor.c
@@ -3397,6 +3397,10 @@ Actor* func_80033684(PlayState* play, Actor* explosiveActor) {
  * This is done by moving it to the corresponding category list and setting its category variable accordingly.
  */
 void Actor_ChangeCategory(PlayState* play, ActorContext* actorCtx, Actor* actor, u8 actorCategory) {
+    //! @bug If Actor_ChangeCategory is called during an actor update function, the loop in
+    //! Actor_UpdateAll will continue from the next actor in the new category, rather than the next
+    //! actor in the old category. This might cause some actors in the old category to be skipped
+    //! over and not updated, or some actors in the new category to be updated twice.
     Actor_RemoveFromCategory(play, actorCtx, actor);
     Actor_AddToCategory(actorCtx, actor, actorCategory);
 }
--- a/src/code/z_actor.c
+++ b/src/code/z_actor.c
@@ -3398,7 +3398,7 @@ Actor* func_80033684(PlayState* play, Actor* explosiveActor) {
  */
 void Actor_ChangeCategory(PlayState* play, ActorContext* actorCtx, Actor* actor, u8 actorCategory) {
     //! @bug Calling this function immediately moves an actor from one category list to the other.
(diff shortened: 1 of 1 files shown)


Dragorn421 (approved):
Nice catch, very interesting

They definitely noticed and fixed this in MM, the actual category lists changes were moved to after the "update all" main loop instead of being done directly in `Actor_ChangeCategory`

https://github.com/zeldaret/mm/blob/b25e64d1bd2ece30301f9fa5af78f963afd0cf76/src/code/z_actor.c#L2659-L2677
in OoT there's "nothing" there: https://github.com/zeldaret/oot/blob/f2c06ce441a38a032202b9024fdfa57bb36a67c2/src/code/z_actor.c#L2253

src/code/z_actor.c
```diff
@@ -3397,6 +3397,10 @@ Actor* func_80033684(PlayState* play, Actor* explosiveActor) {
  * This is done by moving it to the corresponding category list and setting its category variable accordingly.
  */
 void Actor_ChangeCategory(PlayState* play, ActorContext* actorCtx, Actor* actor, u8 actorCategory) {
+    //! @bug If Actor_ChangeCategory is called during an actor update function, the loop in
+    //! Actor_UpdateAll will continue from the next actor in the new category, rather than the next
+    //! actor in the old category. This might cause some actors in the old category to be skipped
+    //! over and not updated, or some actors in the new category to be updated twice.
```
fig02, line 3403: Just some slight rewording suggestions. 
The first sentence I think just helps the reader understand what is the cause of the problem. And I think its more encompassing to say "an actor that is currently updating" rather than "the update function", since it can be from any function called by the actors update function
```
Calling this function immediately moves an actor from one category list to the other. This can cause a problem if called from an actor that is currently updating. 
The loop in `Actor_UpdateAll` will continue from the next actor in the new category, rather than the next
actor in the old category. Depending on the category the actor who called this function was moved to, this can either cause some actors in the old category to skip updating on that frame, or some actors in the new category to be updated twice.
```

fig02 (commented):
Great catch. I agree with what dragorn said. Seems like a good opportunity to call `unk_08` in MMs actor entry as `categoryChanged` or something 😄 but that ofc should be discussed over there

src/code/z_actor.c
Dragorn421: Remove "either", it can (and typically would) be both

fig02: This can cause whole categories to be skipped as well right? might be worth rephrasing that whole part in general

mzxrules: it can't skip categories because the outer loop should iterate over every category; Actor_ChangeCategory just breaks the link for the category currently being processed.

Dragorn421: If I remember correctly the different relevant bits of code, an actor A moving from category B to category C means that

- any actor after A in B will not be updated (the recursive iterating on actors inside categories is from head to tail)
- and actors in C will be updated instead (actors added to categories are added to the head of the linked list so that would be all previous C actors). C actors will also be updated during the normal iteration for C (either before or after B)

If B is before C, A gets updated twice (first as part of B with the category switch, then as part of C once the iteration for C comes), otherwise A is only updated as part of B

-----

Personally I don't think this bug is major enough to warrant long explanations, just a short description would do and one can figure out the details.

mzxrules: also, in theory it's possible to execute actors more than twice per cycle. For example, let's say you have Actor A in category 1 moving to category 3, actor B in category 2 also moving into category 3, and Actor C in category 3

* Actor A will be updated twice
* Actor B will be updated once
* Actor C will be updated three times.

cadmic: Okay, I added the first sentence that Fig suggested, and made the wording about the effects a bit more precise.

(I disagree about needing to expand to "an actor that is currently updating", typically I count anything that happens during a function called from `f` as also happening during `f`, such as when asking "how long does this function take to run")

fig02: dunno, I think its better to describe the process of updating rather than referring to a specific function that happens to be the entry point for carrying it out

cadmic: meh I would understand them both the same so I think it's better to be more concise

AngheloAlf (approved):
Nice catch. The good old bug caused by modifying the list you are iterating lol

src/code/z_actor.c
fig02: The way I phrased it is not much longer and it leaves less room for confusion, it just seems better all around.. If someone else wants to weigh in real quick to tie break this, would be appreciated

cadmic: Alright, changed to "actor update"

fig02 (approved):

PR #1590, opened 2023-12-09
Delete old bootstrap tools

cadmic:
These don't really seem useful now without major changes. They won't be forgotten though, if we want to resurrect them we can always refer to this PR.


Dragorn421 (commented):
I approve of this but I would like to let the OGs weigh in more than me
It's hard to let go I know

ethteck:
Yeah, I do think indeed it is time to bid farewell these old companions.

> They won't be forgotten though

Aye 😿 

but in all seriousness, yee feel free to remove them :)

fig02 (approved):
I guess they will always be present in github history, but I feel like it would be nice if we preserved them somewhere more concrete for historical documentation purposes. that should not be your responsibility though and doesnt have to happen rn :)

playerskel:
Might I suggest to simply move them to a directory like `old/` in the root of the repository? That way they're still there and they're not "in the way" or confusing for people working on the project.

Dragorn421 (approved):
🫡

PR #1593, opened 2023-12-09
Player Docs: Name some high level update calls

fig02:
These are the low hanging fruit "Player_UpdateX" calls. The contents of the functions themselves are not documented in this PR and can happen at a later time.

`Player_SpawnMagicSpell` is also thrown in there because it gets called directly by update.


Dragorn421 (commented):
I think I wrote more than what you changed Despair

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -3473,7 +3482,15 @@ s32 func_80836AB8(Player* this, s32 arg1) {
     return var;
 }
 
-void func_80836BEC(Player* this, PlayState* play) {
+/**
+ * Update Z Targeting if the Z button is pressed.
+ * This includes both sub-types of Z Targeting:
+ *     - Lock On: Player locks onto an actor and a reticle appears on the target.
+ *     - Parallel: Player's Y rotation is locked and he will move in a parallel fashion.
```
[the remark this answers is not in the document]
fig02: What you described is what I mean by moving in a "parallel fashion" yeah. But I do agree its not great to leave that interpretation up to the reader, your suggestion seems fine

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -3473,7 +3482,15 @@ s32 func_80836AB8(Player* this, s32 arg1) {
     return var;
 }
 
-void func_80836BEC(Player* this, PlayState* play) {
+/**
+ * Update Z Targeting if the Z button is pressed.
+ * This includes both sub-types of Z Targeting:
+ *     - Lock On: Player locks onto an actor and a reticle appears on the target.
```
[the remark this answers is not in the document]
engineer124: It's also worth noting that this function is still important for updating part of the targeting system where `Z button` is not pressed. For example, talking to actors, or forced target (think redead scream), so `Update Z Targeting if the Z button is pressed.` doesn't feel general enough

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -10802,7 +10824,7 @@ void func_80848C74(PlayState* play, Player* this) {
     }
 }
 
-void func_80848EF8(Player* this) {
+void Player_UpdateStoneOfAgony(Player* this) {
```
engineer124, line 10827: Just curious, have you looked at the docs for this function in MM? It'd identical to the OoT function except it doesn't check `QUEST_STONE_OF_AGONY`. It's called `Player_DetectSecrets` in MM which is named on what this function is doing instead of tying it to the stone of agony.

fig02: I see. I dont really like the "secrets" terminology in general personally, if I read that for the first time I would not be able to deduce that its for stone of agony. But I do understand the desire for cross game sync

src/overlays/actors/ovl_player_actor/z_player.c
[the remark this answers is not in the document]
fig02: Where is talking involved in this function?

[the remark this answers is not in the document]
engineer124: `cond = func_8083224C(play);` is checking if player has `ACTOR_FLAG_TALK` set, which then leads to player stateFlags being set when "talking" sets `this->unk_664` to the actor that is talking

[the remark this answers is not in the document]
fig02: I see. This goes back to unk_664 being more than just `lockOnActor`.
I think it will probably hurt readability in like 90% of cases if the name isnt `lockOnActor`, but im still not really sure what to do :))

src/overlays/actors/ovl_player_actor/z_player.c
fig02: what is this doing in mm anyway? Does mm rumble for similar things as stone of agony, but you just dont need the in-game item for it to work?

fig02: If that is the case, id prefer "secrets" in all cases be "rumble secrets" at least, so that its obvious what its referring to

src/overlays/actors/ovl_player_actor/z_player.c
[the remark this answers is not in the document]
fig02: I unnamed this function and removed the comment for now. It should at least wait until the Target PR is in, I suppose. Which we should work on trying to get ready soon :D

src/overlays/actors/ovl_player_actor/z_player.c
hensldm: Fwiw, I would say `Player_DetectRumbleSecrets` would be fine for both games.

fig02: coming back to this-- I think I would like to keep the pattern of naming "update" type functions called directly by `_Update` as `UpdateX`
So, I think `Player_UpdateRumbleSecrets` could work. Its both updating the weird accumulator variable for the secrets and actually doing the rumbling

hashbrownomeleto:
esaelp pleH !xob eht raen dednal tsuj tub revo pmuj ot occuC a deirt I ?llimdniw eht raen ecnef eht revo owt eht teg t'nac I tub )dnuora gnimaor srehto dna xob ni eno( s'occuc lla tog I

src/overlays/actors/ovl_player_actor/z_player.c
fig02: After even more thought on this, Ill go with `Player_DetectRumbleSecrets`, I think its okay to not have "update" in this one

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -10185,12 +10200,12 @@ void func_808473D4(PlayState* play, Player* this) {
                            (this->getItemId < GI_MAX)) {
                     doAction = DO_ACTION_GRAB;
                 } else if (this->stateFlags2 & PLAYER_STATE2_11) {
-                    static u8 sDiveDoActions[] = { DO_ACTION_1, DO_ACTION_2, DO_ACTION_3, DO_ACTION_4,
-                                                   DO_ACTION_5, DO_ACTION_6, DO_ACTION_7, DO_ACTION_8 };
+                    static u8 sDiveNumberDoActions[] = { DO_ACTION_1, DO_ACTION_2, DO_ACTION_3, DO_ACTION_4,
+                                                         DO_ACTION_5, DO_ACTION_6, DO_ACTION_7, DO_ACTION_8 };
```
hensldm, line 10204: Wonder if a trailing comma format would look better?

hensldm (approved):
1 minor comment

src/overlays/actors/ovl_player_actor/z_player.c
fig02: There are a handful of them and they are just numbers that count up, so I think more compact is a plus in this case

PR #1594, opened 2023-12-10
Player Docs: Body Shock and Body Burn

fig02:
This in the same spirit as https://github.com/zeldaret/oot/pull/1593 , but it required some more in depth changes, so I made it a seperate PR.

I changed things like `shockTimer` and `isBurning` to `bodyShockTimer` and `bodyIsBurning`.
This is because when I went to name the functions that update these things, `Player_UpdateBurn` and `Player_UpdateShock` sounded a little funny. I added "body" to make it clear the shock and flames appear on his body. And it is in line with the fact that these functions concern his BodyPart array.

So, with the small terminology change, the functions are now `Player_UpdateBodyBurn` and `Player_UpdateBodyShock`.

### DIFF
--- a/include/z64player.h
+++ b/include/z64player.h
@@ -822,7 +822,7 @@ typedef struct Player {
     /* 0x088E */ u8 unk_88E;
     /* 0x088F */ u8 unk_88F;
     /* 0x0890 */ u8 unk_890;
-    /* 0x0891 */ u8 shockTimer;
+    /* 0x0891 */ u8 bodyShockTimer;
     /* 0x0892 */ u8 unk_892;
     /* 0x0893 */ u8 hoverBootsTimer;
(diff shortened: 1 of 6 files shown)


Dragorn421 (commented):
Should we take the opportunity to make `bodyIsBurning` and `bodyFlameTimers` use the same language? (e.g. `bodyIsBurning` and `bodyBurnTimers`, `bodyIsInFlames` and `bodyFlameTimers`, ...)

fig02:
Personally I think its okay if the language is different. Burning makes more sense for describing what its doing to the player as a whole, while the flames are the things doing the burning. We can try to add a comment on the member variables to make sure they are related though, if we feel its needed

Dragorn421 (approved):
:zap: :elf_man: :fire:

mzxrules (approved):

PR #1597, opened 2023-12-15
Create debug macros for common functions

cadmic:
Create macros for functions that are replaced by non-debug versions in retail ROMs, controlled by a `DEBUG` CPP define (except `osSyncPrintf`, which I'll leave for another PR).

Other changes:
* I also used `DEBUG` in assert.h instead of `NDEBUG`, since having both feels redundant and it seems like we're already okay modifying this header instead of strictly conforming to the C standard.
* I renamed `DmaMgr_SendRequest` -> `DmaMgr_RequestAsync` and `DmaMgr_RequestAsync` -> `DmaMgr_RequestAsyncDebug` to fit the naming scheme better.
* I created new functions `Matrix_ToMtxDebug` and `Matrix_NewMtxDebug` instead of modifying `Matrix_ToMtx` and `Matrix_NewMtx`, though I'm not sure if it was that was originally.


AngheloAlf:
> I also used `DEBUG` in assert.h instead of `NDEBUG`, since having both feels redundant and it seems like we're already okay modifying this header instead of strictly conforming to the C standard.

That's not true, we have been trying to adhere to the C standard whenever we can. Granted, there's lot of stuff to do, but that doesn't mean we should give up on it.

I think having separate defines for the standard `NDEBUG` and a specific oot `DEBUG` (maybe with a different name to avoid confusion) is fine

cadmic:
Alright, I've restored `NDEBUG`

fig02:
I have no attachment to NDEBUG personally and dont see when it would make more sense to use over DEBUG. The double negative-ness of using it with ifndef has always confused me 😄

Dragorn421:
I'm lukewarm about `DEBUG` as the macro name, it's quite overloaded. I would prefer if it was more unique, specific to oot

include/macros.h
```diff
@@ -156,6 +161,48 @@ extern struct GraphicsContext* __gfxCtx;
...
+#define MATRIX_NEW(gfxCtx, line, file) Matrix_NewMtxDebug(gfxCtx, line, file)
+#define MATRIX_CHECK_FLOATS(mtx, file, line) Matrix_CheckFloats(mtx, file, line)
+#define DMA_REQUEST_SYNC(ram, vrom, size, file, line) DmaMgr_RequestSyncDebug(ram, vrom, size, file, line)
+#define DMA_REQUEST_ASYNC(req, ram, vrom, size, unk5, queue, msg, file, line) DmaMgr_RequestAsyncDebug(req, ram, vrom, size, unk5, queue, msg, file, line)
+#define GAME_STATE_ALLOC(gameState, size, file, line) GameState_Alloc(gameState, size, file, line)
+#define SYSTEM_ARENA_MALLOC(size, file, line) SystemArena_MallocDebug(size, file, line)
+#define SYSTEM_ARENA_MALLOCR(size, file, line) SystemArena_MallocRDebug(size, file, line)
+#define SYSTEM_ARENA_FREE(size, file, line) SystemArena_FreeDebug(size, file, line)
+#define ZELDA_ARENA_MALLOC(size, file, line) ZeldaArena_MallocDebug(size, file, line)
+#define ZELDA_ARENA_MALLOCR(size, file, line) ZeldaArena_MallocRDebug(size, file, line)
```
fig02, line 175: Thinking for both of these it should be `MALLOC_R`, so that the "R" for reverse stands out more. 
It stands out in the function name because of the case difference, but doesnt really do so well in all caps

include/functions.h
```diff
@@ -1430,8 +1430,10 @@ void Matrix_RotateZYX(s16 x, s16 y, s16 z, u8 mode);
 void Matrix_TranslateRotateZYX(Vec3f* translation, Vec3s* rotation);
 void Matrix_SetTranslateRotateYXZ(f32 translateX, f32 translateY, f32 translateZ, Vec3s* rot);
 Mtx* Matrix_MtxFToMtx(MtxF* src, Mtx* dest);
-Mtx* Matrix_ToMtx(Mtx* dest, char* file, s32 line);
-Mtx* Matrix_NewMtx(GraphicsContext* gfxCtx, char* file, s32 line);
+Mtx* Matrix_ToMtx(Mtx* dest);
+Mtx* Matrix_ToMtxDebug(Mtx* dest, char* file, s32 line);
+Mtx* Matrix_NewMtx(GraphicsContext* gfxCtx);
+Mtx* Matrix_NewMtxDebug(GraphicsContext* gfxCtx, char* file, s32 line);
```
fig02, line 1436: Confused on your comment in the PR description about matrix functions. You made two whole new functions and you arent sure if they exist? Feel like we should confirm something like this before putting it in

cadmic: The debug ROM and the retail ROM both have two functions here. The debug ROM functions take additional `file` and `line` arguments while the retail versions don't. So I'm not sure whether to name both versions the same thing, or add a `Debug` suffix to the debug versions like for the other macros

cadmic:
I changed `DEBUG` -> `OOT_DEBUG` (here and in #1598)

include/z64dma.h
```diff
@@ -43,10 +43,12 @@ extern size_t gDmaMgrDmaBuffSize;
...
 // Standard DMA Requests
 
+s32 DmaMgr_RequestAsync(DmaRequest* req, void* ram, uintptr_t vrom, size_t size, u32 unk5, OSMesgQueue* queue,
+                        OSMesg msg);
 s32 DmaMgr_RequestSync(void* ram, uintptr_t vrom, size_t size);
+s32 DmaMgr_RequestAsyncDebug(DmaRequest* req, void* ram, uintptr_t vrom, size_t size, u32 unk5, OSMesgQueue* queue,
+                             OSMesg msg, const char* file, s32 line);
 s32 DmaMgr_RequestSyncDebug(void* ram, uintptr_t vrom, size_t size, const char* file, s32 line);
-s32 DmaMgr_RequestAsync(DmaRequest* req, void* ram, uintptr_t vrom, size_t size, u32 unk5, OSMesgQueue* queue,
-                        OSMesg msg, const char* file, s32 line);
```
Dragorn421, line 49: @Thar0 thoughts?

include/functions.h
fig02: Right now I kinda feel like they should be named the same cause of how they get swapped out. but more opinions needed

Dragorn421 (commented):
I'm not super convinced by the debug macros wrappers, it's not what I envisioned, trying to pin down why

One thing is even if they're macros I expected normal `Matrix_ToMtx` casing, not `UPPER_CASING`

Another thing is when `NON_MATCHING` is defined, file/line should be ignored from the for-ok-purposes arguments and `__FILE__`/`__LINE__` used instead (similar to how it's done in main for `ASSERT`)

for example:
```c
#ifdef OOT_DEBUG
#  ifndef NON_MATCHING
#    define Matrix_NewMtx(gfxCtx, file, line) Matrix_NewMtxDebug(gfxCtx, file, line)
#  else
#    define Matrix_NewMtx(gfxCtx, file, line) Matrix_NewMtxDebug(gfxCtx, __FILE__, __LINE__)
#  endif
#else
#  define Matrix_NewMtx(gfxCtx, file, line) Matrix_NewMtx(gfxCtx)
#endif
```

(
this is motivated to me by the approach I would assume the original code had (at least that's what I'd do), fully transparent to the programmer:
```c
#ifdef OOT_DEBUG
#  define Matrix_NewMtx(gfxCtx) Matrix_NewMtxDebug(gfxCtx, __FILE__, __LINE__)
#else
#  define Matrix_NewMtx(gfxCtx) Matrix_NewMtx(gfxCtx)
#endif
```
where we also need to pass in the OK file and line
)

fig02:
Personally dont mind the uppercase and appreciate the information that its a macro wrapping the debug differences

include/functions.h
Dragorn421: I would either not introduce the non-debug variants for now (using `Matrix_ToMtxDebug(..., "", 0)` for now to define the non-debug macro), or wait for a version to be worked on in the repo that has these non-debug variants

AngheloAlf: Those functions are swapped between debug and non debug they should probably be named the same.
Also I don't think it is a good idea to add new functions on a PR that was meant to create macros

cadmic: Okay, I've named the non-debug and debug functions the same

cadmic:
Yeah I think it would be too confusing if a function and a macro both had the same name. And I think it wouldn't work with the way the headers are set up.

I have no opinion about using the real `__FILE__` and `__LINE__` with `NON_MATCHING`. If you're not sure about `NON_MATCHING`, we can easily leave it til later: this PR does not make anything worse in that regard, and does not make adding it later any harder

include/z64dma.h
Thar0: Looks good to me :rocket:

fig02 (approved):

include/macros.h
```diff
@@ -156,6 +161,48 @@ extern struct GraphicsContext* __gfxCtx;
     }                                                   \
     (void)0
 
+#define GRAPH_ALLOC(gfxCtx, size) Graph_Alloc(gfxCtx, size)
+#define MTXF_TO_MTX(gfxCtx, line, file) Matrix_ToMtx(gfxCtx, line, file)
+#define MATRIX_NEW(gfxCtx, line, file) Matrix_NewMtx(gfxCtx, line, file)
+#define MATRIX_CHECK_FLOATS(mtx, file, line) Matrix_CheckFloats(mtx, file, line)
```
Dragorn421, line 167: `line, file` vs `file, line`

probably something's up

include/macros.h
```diff
@@ -156,6 +161,48 @@ extern struct GraphicsContext* __gfxCtx;
...
+    {                                       \
+        GraphicsContext* __gfxCtx = gfxCtx; \
+        s32 __dispPad
+
+#define CLOSE_DISPS(gfxCtx, file, line) \
+    (void)0;                            \
+    }                                   \
+    (void)0
+
+#define GRAPH_ALLOC(gfxCtx, size) ((void*)((gfxCtx)->polyOpa.d = (Gfx*)((u8*)(gfxCtx)->polyOpa.d - ALIGN16(size))))
```
Dragorn421, line 190: is this what is done in retail? kind of a wild change compared to debug (though it does seem equivalent)

include/ultra64/gbi.h
```diff
@@ -5134,6 +5136,25 @@ _DW({                                                   \
 #define gDPNoOpCloseDisp(pkt, file, line)   gDma1p(pkt, G_NOOP, file, line, 8)
 #define gDPNoOpTag3(pkt, type, data, n)     gDma1p(pkt, G_NOOP, data, n, type)
 
+#else
+
+#define gDPNoOp(pkt)
+#define gsDPNoOp()
```
Dragorn421, line 5142: not sure the static dlist commands `#define gsDPNoOp()` would work like this, for example

```c
Gfx dl[] = {
  gsDPNoOp(),
  gsSPEndDisplayList(),
};
```

would expand to invalid C

(pseudo-expansion:)

```c
Gfx dl[] = {
  ,
  { _SHIFTL(G_ENDDL, 24, 8), 0 },
};
```

not sure what to do about it (note in practice afaik there is no `gsDPNoOp*` anywhere)

include/macros.h
fig02: it is, yeah. can see MM for an example

include/macros.h
cadmic: Thanks for catching this

include/ultra64/gbi.h
cadmic: Good point, I've restored gsDPNoOp and friends, which I think makes things a bit more organized too

include/macros.h
```diff
@@ -156,6 +161,48 @@ extern struct GraphicsContext* __gfxCtx;
     }                                                   \
     (void)0
 
+#define GRAPH_ALLOC(gfxCtx, size) Graph_Alloc(gfxCtx, size)
+#define MTXF_TO_MTX(gfxCtx, file, line) Matrix_ToMtx(gfxCtx, file, line)
```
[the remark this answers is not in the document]
cadmic: Yeah I thought `MATRIX_TO_MTX` was confusing but I don't feel strongly about it, so I will change it. Do you have an opinion on `MATRIX_NEW`?

src/code/gamealloc.c
```diff
@@ -13,7 +13,7 @@ void GameAlloc_Log(GameAlloc* this) {
 }
 
 void* GameAlloc_MallocDebug(GameAlloc* this, u32 size, const char* file, s32 line) {
-    GameAllocEntry* ptr = SystemArena_MallocDebug(size + sizeof(GameAllocEntry), file, line);
+    GameAllocEntry* ptr = SYSTEM_ARENA_MALLOC(size + sizeof(GameAllocEntry), file, line);
```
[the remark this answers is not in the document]
cadmic: Good catch, thanks. I checked every function that took a `file` or `line` argument and this is the only instance I could find.

src/code/sys_matrix.c
```diff
@@ -603,16 +603,30 @@ Mtx* Matrix_MtxFToMtx(MtxF* src, Mtx* dest) {
...
+
+Mtx* Matrix_ToMtx(Mtx* dest) {
+    return Matrix_MtxFToMtx(sCurrentMatrix, dest);
 }
 
+Mtx* Matrix_NewMtx(GraphicsContext* gfxCtx) {
+    return Matrix_ToMtx(GRAPH_ALLOC(gfxCtx, sizeof(Mtx)));
+}
+
+#endif /* OOT_DEBUG */
```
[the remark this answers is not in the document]
cadmic: Sorry, "add new functions" was a poor choice of words on my part. I meant whether the we should name the debug and retail versions the same (as if they were "modified" for debug) or if they should have separate names (as if they were "new" functions for debug).

I don't think it really makes sense to split this change into another PR, because doing something like `Matrix_ToMtxDebug(..., "", 0)` for the retail version in the meantime just seems obviously broken, and it's not a large change to keep it here, and I've already made the change here.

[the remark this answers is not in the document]
Dragorn421: basically the way it is currently doesn't address my comment or anghelo's (?) from the previous thread

I personally feel very strongly about not adding code to the repo that doesn't represent the matching game

src/code/z_actor.c
```diff
@@ -2833,8 +2832,7 @@ Actor* Actor_Spawn(ActorContext* actorCtx, PlayState* play, s16 actorId, f32 pos
 
                 if (actorCtx->absoluteSpace == NULL) {
                     // "AMF: absolute magic field"
-                    actorCtx->absoluteSpace =
-                        ZeldaArena_MallocRDebug(ACTOROVL_ABSOLUTE_SPACE_SIZE, "AMF:絶対魔法領域", 0);
+                    actorCtx->absoluteSpace = ZELDA_ARENA_MALLOC_R(ACTOROVL_ABSOLUTE_SPACE_SIZE, "AMF:絶対魔法領域", 0);
```
Dragorn421, line 2835: og code didn't use `__FILE__`, `__LINE__` here so it's debatable if we want to reflect that using the debug function directly or not.

For non_matching debug purposes it seems more useful to me to indeed use the macro so it would use real file and line info

this also happens for other `ZeldaArena_Malloc*` calls in this function

src/code/sys_matrix.c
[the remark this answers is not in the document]
cadmic: Yeah sorry, I misunderstood before.

I disagree for exactly the same reason, the non-debug versions in this PR do match retail OOT versions, while `Matrix_ToMtxDebug(..., "", 0)` doesn't match anything at all

src/overlays/actors/ovl_Bg_Hidan_Rsekizou/z_bg_hidan_rsekizou.c
```diff
@@ -214,8 +214,8 @@ Gfx* BgHidanRsekizou_DrawFireball(PlayState* play, BgHidanRsekizou* this, s16 fr
     mf->zw = (tmpf7 * coss) + this->dyna.actor.world.pos.z;
 
     gSPMatrix(displayList++,
-              Matrix_MtxFToMtx(Matrix_CheckFloats(mf, "../z_bg_hidan_rsekizou.c", 543),
-                               Graph_Alloc(play->state.gfxCtx, sizeof(Mtx))),
+              Matrix_MtxFToMtx(MATRIX_CHECK_FLOATS(mf, "../z_bg_hidan_rsekizou.c", 543),
+                               GRAPH_ALLOC(play->state.gfxCtx, sizeof(Mtx))),
```
Dragorn421, line 218: This is interestingly an "inlined" version of `Matrix_NewMtx`

the same thing happens in `z_bg_hidan_sekizou.c` and `z_bg_hidan_sima.c` (looks like copypasting), and... `Play_Draw`

I wonder if it was a macro, like a retail-`GRAPH_ALLOC`-like equivalent to `Matrix_NewMtx`

Dragorn421 (commented):
Some weird things I noticed about the code, but non-blocking for the PR (the weirdness is inherited from "that's what the game does")

Also side note we should add `OOT_DEBUG` to `vscode.md` eventually

src/code/sys_matrix.c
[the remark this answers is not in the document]
Dragorn421: Oh I didn't know they match retail, I assumed you just winged it. Then okay. I have no easy way to check if they do match so I'll just trust you on it

Just one more thing, anghelo said

> Those functions are swapped between debug and non debug

I understood it as "they are in a different order", but in this PR they're in the same order?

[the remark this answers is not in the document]
cadmic: I think he meant that the debug versions completely replaced the non-debug versions in the ROM so it looks like they've been "swapped out". I don't think we can know what order they appeared in the original source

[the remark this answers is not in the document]
Dragorn421: as far as we know the order in the original source is the order they appear in the rom so I use "order" for both

ok so I misunderstood then, thanks all's good

Dragorn421:
This is good for me, just waiting on changing `MTXF_TO_MTX` to `MATRIX_TO_MTX`
(afaict we agree on the change https://github.com/zeldaret/oot/pull/1597#discussion_r1443832000 )

include/macros.h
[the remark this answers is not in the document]
cadmic: Okay, done

Dragorn421 (approved):
:space_invader: :dart: :chart_with_upwards_trend:

Thar0 (approved):

PR #1599, opened 2023-12-16
Change ?= to := for Makefile variables

cadmic:
@Dragorn421 suggested using we use `:=` instead of `?=` so that variables don't accidentally get overridden by environment variables (e.g. `COMPILER` or `VERSION`). You can still override them with `make VAR=value` though.

I kept the `?=` for `CFLAGS` and `CPPFLAGS` though, since it seems useful to use an environment variable there: `CFLAGS=-show make` will merely add `-show` to every compiler call, but `make CFLAGS=-show` will completely replace the flags with `-show` which may not be what you want.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,9 +15,6 @@ ORIG_COMPILER := 0
 # If COMPILER is "gcc", compile with GCC instead of IDO.
 COMPILER := ido
 
-CFLAGS ?=
-CPPFLAGS ?=
-
 # ORIG_COMPILER cannot be combined with a non-IDO compiler. Check for this case and error out if found.
 ifneq ($(COMPILER),ido)
   ifeq ($(ORIG_COMPILER),1)
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,9 @@ ORIG_COMPILER := 0
 # If COMPILER is "gcc", compile with GCC instead of IDO.
 COMPILER := ido
 
+CFLAGS ?=
+CPPFLAGS ?=
+
 # ORIG_COMPILER cannot be combined with a non-IDO compiler. Check for this case and error out if found.
 ifneq ($(COMPILER),ido)
   ifeq ($(ORIG_COMPILER),1)
--- a/Makefile
+++ b/Makefile
@@ -7,13 +7,13 @@ SHELL = /bin/bash
 # Build options can either be changed by modifying the makefile, or by building with 'make SETTING=value'
 
 # If COMPARE is 1, check the output md5sum after building
-COMPARE ?= 1
+COMPARE := 1
 # If NON_MATCHING is 1, define the NON_MATCHING C flag when building
-NON_MATCHING ?= 0
+NON_MATCHING := 0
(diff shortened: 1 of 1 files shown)


cadmic:
Actually, the `CFLAGS ?=` and `CPPFLAGS ?=` have no effect so I've removed them

Makefile
```diff
@@ -7,16 +7,13 @@ SHELL = /bin/bash
...
-NON_MATCHING ?= 0
+NON_MATCHING := 0
 # If ORIG_COMPILER is 1, compile with QEMU_IRIX and the original compiler
-ORIG_COMPILER ?= 0
+ORIG_COMPILER := 0
 # If COMPILER is "gcc", compile with GCC instead of IDO.
-COMPILER ?= ido
-
-CFLAGS ?=
-CPPFLAGS ?=
```
Dragorn421, line 19: There are `CFLAGS +=` later so I think it'd be cleaner (more obvious to the reader) to initialize them somewhere

Not initializing at all is also the same as `?=`

And finally CFLAGS and CPPFLAGS are standard variables in Make which may have default values or may be part of the environment per user preference

So overall it'd just be better to initialize them, at least for clarity

cadmic: Okay I've reverted that change. Although as I mentioned in the PR description, for environment vs command-line variables is not preference for CFLAGS and CPPFLAGS, it has a functional difference

Dragorn421: Sorry I don't understand what you mean

Why not have
```
CFLAGS :=
CPPFLAGS :=
```
?

cadmic: Because then I can't write, e.g.

```
CFLAGS=-show make build/src/code/z_collision_check.o
```

to prepend the flag `-show` to the C compiler command. If I instead write

```
make CFLAGS=-show build/src/code/z_collision_check.o
```

then the C flags that are carefully crafted by the Makefile are completely overridden, and it doesn't compile. For a single file perhaps I could copy all the CFLAGS into my command-line instead, but that doesn't work if I want to compile multiple files which might have different flags set by the Makefile.

(I've actually been doing this to debug bss ordering btw, it's not just a theoretical thing)

Dragorn421: ah I see, okay. I guess the spirit of make is clashing against our "we need specific flags to make OK builds" use case :sweat_smile: 
I guess it's fine to let these be then.

and to be clear afaik
```
SOMEVAR ?=
SOMEVAR += blabla
```
and
```
SOMEVAR += blabla
```
behave exactly the same in make

cadmic: Right, that's why I removed the 

```
CFLAGS ?=
CPPFLAGS ?=
```

lines earlier. Should I un-revert that, or is it clearer the way it is now?

Dragorn421 (approved):

Makefile
Dragorn421: I think it's better to keep the empty `?=`s, it's less confusing for reading the makefile

Sorry for the confusion lol

fig02 (approved):

PR #1600, opened 2023-12-16
Match fault and fault_drawer .bss

cadmic:
I was curious about this so I tried removing the `NON_MATCHING`s and it basically worked. There might be something I'm missing, why was this under `NON_MATCHING` before?


hensldm:
It's because fault and fault_drawer are compiled with `-g`, which has a much larger impact on bss reordering. Thus it was decided to just keep in NON_MATCHING so it didn't needed to be dealt with as global changes are made. Especially since fault is still reliant on `global.h`. Not sure if OoT would be interested in `prevent_bss_reordering.h` which has been very helpful in MM though not perfect.

fig02:
Yeah we just didnt want to deal with g3 bss while making large changes to things like headers that can heavily affect it.
Ofc when we actually start retail work, it will be more commonplace to use `prevent_bss_reordering.h`. But until multiversion is fully up and running, I think its fine to wait

cadmic:
FWIW I don't think `prevent_bss_reordering.h` is sufficient, we need a way to force a specific bss ordering per-version without being too messy. For example, z_collision_check requires a specific ordering even in the debug ROM

PR #1602, opened 2023-12-26
Removed unused imports and other minor improvements

adaliaramon:


jpburnett (approved):
Thanks for the PR! Some of the tooling probably once upon a time used the imports and when those files were updated the imports were probably not removed. As for the string formatting, I can see it either way, I'll defer to a python guru as to which they prefer as I don't know what is the most "pythonic" way to do things anymore...
The cleanup is appreciated!

diff.py
Dragorn421, line 1: This script is not the main source, it's only here in the repo for convenience.
The repo it comes from is here: https://github.com/simonlindholm/asm-differ

(so we don't change it besides updating it, any change should be done upstream)

Dragorn421 (commented):
Thanks for the cleanup!

Sorry I only realized now that I tagged you twice in the licensing issue, and that there's actually something to comment on here

tools/ZAPD/ZAPD/genbuildinfo.py
```diff
@@ -2,7 +2,6 @@
 
 import argparse
 from datetime import datetime
-import getpass
 import subprocess
```
fig02, line 5: similarly, this tool is located at https://github.com/NEstelami/ZAPD/ and should have changes made upstream instead of here

sym_info.py
```diff
@@ -17,7 +17,7 @@ def symInfoMain():
 
     args = parser.parse_args()
 
-    BUILTMAP = Path(f"build") / "gc-eu-mq-dbg" / f"z64.map"
+    BUILTMAP = Path(f"build") / "gc-eu-mq-dbg" / "z64.map"
```
Dragorn421, line 20:
```suggestion
    BUILTMAP = Path("build") / "gc-eu-mq-dbg" / "z64.map"
```

tools/asm_processor/asm_processor.py
```diff
@@ -2,7 +2,6 @@
 import argparse
 import tempfile
 import struct
-import copy
 import sys
```
fig02, line 5: this tool is also external https://github.com/simonlindholm/asm-processor
(I'm making changes for the PR author, but just commenting for documentation purposes, and incase the author feels like making PRs upstream)

fig02 (approved):

Dragorn421 (approved):

PR #1609, opened 2024-01-16
Move Scene_SetTransitionForNextEntrance from z_scene_table.c to z_scene.c

cadmic:
`Scene_SetTransitionForNextEntrance` is currently the first function in `z_scene_table.c`, but it can't be the first function in a compilation unit because it is not 0x10-aligned on gc-eu-mq. I believe moving it to `z_scene.c` is the correct file split because `Scene_SetTransitionForNextEntrance` has padding after it in gc-eu-mq, and the following function (`Scene_DrawConfigDefault` on GC or `Scene_Draw` on N64) is 0x10-aligned on all versions according to the [zelda64_compare spreadsheet](https://docs.google.com/spreadsheets/d/17yPD3DqqH5lZeR7c_QmJfgxWgVyYkGgTLZoKcvLTwtw/edit#gid=1299953822).

### DIFF
--- a/src/code/z_scene.c
+++ b/src/code/z_scene.c
@@ -483,6 +483,26 @@ void Scene_CommandMiscSettings(PlayState* play, SceneCmd* cmd) {
     }
 }
 
+void Scene_SetTransitionForNextEntrance(PlayState* play) {
+    s16 entranceIndex;
+
+    if (!IS_DAY) {
+        if (!LINK_IS_ADULT) {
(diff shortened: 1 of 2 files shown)


fig02 (approved):
This also just makes way more sense logically :)

AngheloAlf (approved):

fig02:
Has met merge requirements already but will wait a day to give other people a chance to see

mzxrules (approved):

Dragorn421 (approved):

PR #1611, opened 2024-01-16
Remove actor padding in non-debug versions

cadmic:
This causes diffs in basically all actor functions so I think it's important to change this early.

### DIFF
--- a/include/z64actor.h
+++ b/include/z64actor.h
@@ -305,7 +305,9 @@ typedef struct Actor {
     /* 0x130 */ ActorFunc update; // Update Routine. Called by `Actor_UpdateAll`
     /* 0x134 */ ActorFunc draw; // Draw Routine. Called by `Actor_Draw`
     /* 0x138 */ ActorOverlay* overlayEntry; // Pointer to the overlay table entry for this actor
+#ifdef OOT_DEBUG
     /* 0x13C */ char dbgPad[0x10]; // Padding that only exists in the debug rom
+#endif
 } Actor; // size = 0x14C
 
 typedef enum {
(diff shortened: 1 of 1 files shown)


fig02 (approved):

AngheloAlf (approved):

fig02:
Has met merge requirements already but will wait a day to give other people a chance to see

mzxrules:
the ambiguity of offset comments begins now

mzxrules (approved):

Dragorn421 (approved):

include/z64actor.h
```diff
@@ -305,7 +305,9 @@ typedef struct Actor {
     /* 0x130 */ ActorFunc update; // Update Routine. Called by `Actor_UpdateAll`
     /* 0x134 */ ActorFunc draw; // Draw Routine. Called by `Actor_Draw`
     /* 0x138 */ ActorOverlay* overlayEntry; // Pointer to the overlay table entry for this actor
+#ifdef OOT_DEBUG
     /* 0x13C */ char dbgPad[0x10]; // Padding that only exists in the debug rom
+#endif
```
fig02, line 310: we can maybe change "rom" to version, or just remove the comment altogether since now it will be obvious with the ifdef. Im ok with either

cadmic: Sure, is it because there is more than one debug version now? Or do you not like using the term "ROM" to refer to a specific version?

fig02: yeah I think the term rom is out of place now. it made sense when our focus was decompiling one specific rom, but that isnt the case anymore. 
removing the comment might be best, cause it gets the same information across now. we arent going around commenting "debug version only" on all debug specific things. the ifdef already tells you :D

PR #1613, opened 2024-01-17
Set up disassembly for gc-eu-mq

cadmic:
This sets up disassembly and diffing tools for decompiling `gc-eu-mq`. I've written some instructions in `docs/retail_versions.md`, but here's the TL;DR version:

* Copy target ROM to `baseroms/gc-eu-mq/baserom.z64` (md5sum: `1618403427e4344a57833043db5ce3c3`)
* Run `make setup -jN VERSION=gc-eu-mq`
* Run `make -jN VERSION=gc-eu-mq`
* Run `make disasm -jN VERSION=gc-eu-mq`
* Run e.g. `./diff.py -mwo3 -v gc-eu-mq Graph_TaskSet00`

Please try it out and let me know what you think.

Regarding the diff.py `-v` flag: I'm sorry for wanting to die on this hill, but I tried out several alternatives (including `--oot-version`, `--ov` and `-r` (for "ROM")), and they all felt pretty bad IMO. I expect people to use this a lot (and I myself have been using this a lot) and frankly I would feel embarrassed to make people suffer with something else. I don't think the conventions for `--verbose` or `--version` are nearly universal enough to override this.

Some problems / room for improvement:
* Data symbols could be improved for less diffs (for example, not exporting local data symbols in spimdisasm)
* `first_diff.py` does not work, despite ostensibly having an `--oot-version` flag, because it looks for a `.map` file in the `expected` dir. This probably needs more thought anyway on how to deal with compressed ROMs and we won't need it for a little while anyway.
* `functions.txt` and `variables.txt` will probably fall out of sync during documentation work. I don't think this is a huge deal, since it's pretty easy to update and we can delete it all once the decomp work is done.


tools/disasm/disasm.py
```diff
@@ -0,0 +1,138 @@
...
+    config_dir:Path,
+    dma_file: DmaFile,
+    f: BinaryIO,
+) -> spimdisasm.mips.FileSplits:
+    f.seek(dma_file.rom_start)
+    if dma_file.rom_end != 0:  # compressed
+        raw_data = f.read(dma_file.rom_end - dma_file.rom_start)
+        data = bytearray(libyaz0.decompress(raw_data))
+    else:
+        data = bytearray(f.read(dma_file.vrom_end - dma_file.vrom_start))
```
AngheloAlf, line 25: Decompressing each segment during disassembly doesn't sound like the best idea. Generating a decompressed rom would be better, specially since we'll need it for assets.

AngheloAlf (commented):
I'm not a huge fan of rolling yet-another-custom-disassembler when we have more standardized options like [splat](https://github.com/ethteck/splat), one of the issues being the specific csv format being used being terrible

tools/disasm/disasm.py
fig02: I would love to get a tool like z64compress up and running now, if we want to. We could just decompress the input rom at the start, right?
 (this purpose aside, itd be helpful for debug-rom modders to have access to compression being included in the build system now)

fig02: actually idr if z64compress does decompression :D it might not

AngheloAlf: I can setup the compression and decompression, should be easy to take the stuff from MM.
And no, z64compress doesn't do decompression lol

cadmic: I totally agree about compression and decompression but I was hoping to do those in separate PRs. (Although I'd be very happy if Anghelo beats me to it.) Compression at least is not necessary right now, and the decompression code here is 4 lines long so I didn't think it cost that much to add.

AngheloAlf: It isn't much code but it can be impactful on runtime, so separating decompression from disassembly would be a lot better

fig02: waiting is fine by me too personally, just wanted to express that I am a fan of compression infrastructure in general being introduced

cadmic: I'm not totally sure it would be faster to decompress separately, for the debug ROM at least the uncompressed source files are about 5MB, but decompressing the entire ROM is about 54MB:

```
$ wc -c baserom/{boot,code,ovl_*}
  5091104 total
$ wc -c baserom/*
 54252304 total
```

I actually did try to set up decompression briefly, but I thought it made the build system too complicated, and I wasn't sure if I was heading in the right direction. It might be better to combine that work with multi-version assets

cadmic:
Re splat: I'm somewhat concerned it's not flexible enough for this use case. For example, we want to output all sections for a compilation unit to the same assembly file (if we instead concat everything together afterwards, the assembler will complain about defining macros multiple times), and in general there doesn't seem to be a way to pass arbitrary options to spimdisasm.

It's true we can probably add anything we need to splat, but I'm not sure if it's worth rewriting all of this. This is not a full-blown disassembler like MM's, it's merely a wrapper around spimdisasm which is doing all the heavy lifting here. And my hope is that it will be deleted soon.

AngheloAlf:
> Re splat: I'm somewhat concerned it's not flexible enough for this use case. For example, we want to output all sections for a compilation unit to the same assembly file (if we instead concat everything together afterwards, the assembler will complain about defining macros multiple times), 

Why do we want to put all the sections in the same assembly file? Wouldn't it be better to just use a C file and like decomp the functions/data?

> and in general there doesn't seem to be a way to pass arbitrary options to spimdisasm.

What kind of flags do we want to pass to spimdisasm anyways besides the defaults? I really think the yaml format splat uses is a lot better than a bunch of csvs like spimdisasm does. If we really wanted to pass some custom flags to spimdisasm then we can just use splat as a library instead and add the flags there, we would get all the pros from using splat and the wrapper would like 25 lines or so

> And my hope is that it will be deleted soon.

We have tens of versions to match yet. Whatever we end up using will probably be around for some time

AngheloAlf:
> first_diff.py does not work, despite ostensibly having an --oot-version flag, because it looks for a .map file in the expected dir. This probably needs more thought anyway on how to deal with compressed ROMs and we won't need it for a little while anyway.

`first_diff.py` won't work because it is meant to find the first difference between your built rom and a matching expected rom.
This script doesn't need to worry about compressed roms either, because it will search the diff between uncompressed roms instead

cadmic:
> Why do we want to put all the sections in the same assembly file? Wouldn't it be better to just use a C file and like decomp the functions/data?

I mean so we can recompile the .o file from the disassembly. What do you mean by use a C file?

> What kind of flags do we want to pass to spimdisasm anyways besides the defaults?

Well, for example I was thinking about implementing a spimdisasm flag to not export local data symbols.

> We would get all the pros from using splat

What are the pros of splat besides the yaml format? It just doesn't seem worth it to me to add another layer right now. I agree the yaml format is better, but these CSVs are already written and I don't expect anyone to modify them, so we wouldn't really get any benefit from switching. I'm not ruling out switching to splat in the future though, especially if zelda64_compare has switched by then too since that's where this data comes from.

> We have tens of versions to match yet. Whatever we end up using will probably be around for some time

Yeah that's fair. I'm hoping at least some of them could be matched in a weekend and wouldn't need all this infrastructure though.

tools/disasm/disasm.py
Dragorn421: (de)compression will be handled in https://github.com/zeldaret/oot/pull/1614
as well as some baserom reorganization

cadmic:
I rebased this branch just now. For now, if the spec includes a ROM segment (like "$(BUILD_DIR)/baserom/Audiobank.o") then it's still taken from the Debug ROM. I plan to clean this up soon (by making extract_baserom.py multi-version). This was a bit cleaner before when `make setup` was not per-version, but alas.

cadmic:
I've cleaned up the progress script `./retail_progress.py` and added it to this branch. Let me know if I should split it out instead

fig02:
Im fine with progress script being included in here, and personally find it kinda necessary for people getting involved in actual code decomp.

Ill be pulling this branch and trying out the process soon.

sym_info.py
```diff
@@ -9,11 +9,12 @@
 def symInfoMain():
     parser = argparse.ArgumentParser(description="Display various information about a symbol or address.")
     parser.add_argument("symname", help="symbol name or VROM/VRAM address to lookup")
-    parser.add_argument("-e", "--expected", dest="use_expected", action="store_true", help="use the map file in expected/build/ instead of build/")
+    parser.add_argument("-v", "--oot-version", help="Which version should be processed", default="gc-eu-mq-dbg")
+    parser.add_argument("-e", "--expected", dest="use_expected", action="store_true", help="use the map file in expected/build/ instead of build/")    parser.add_argument("-v", "--oot-version", help="Which version should be processed", default="gc-eu-mq-dbg")
```
hensldm, line 13:
```suggestion
    parser.add_argument("-e", "--expected", dest="use_expected", action="store_true", help="use the map file in expected/build/ instead of build/")
```
Somehow this got duplicated.

hensldm (commented):
Does what it sets out to do. Process seems fine from my experience.

fig02 (approved):
My approval is from the perspective of usability. I do not have anything to add regarding the implementation of this system. 

I took this system for a test drive and matched 23 different functions in player. There are 20 more functions left in that file. The system is intuitive and easy to use.

I think this is a great first step to get started on retail .text decomp, but do think we can improve on this system, and even completely replace it soon, if we figure out a full matching rom setup.

There were some small issues to work out with the system, but my thorough test helped catch them, and they have since been fixed. Can check cadmics recent commit history for those changes.

docs/retail_versions.md
```diff
@@ -0,0 +1,166 @@
...
+Note that many tools will require activating the Python virtual environment
+in your terminal session. To do this, run:
+
+```sh
+source .venv/bin/activate
+```
+
+### retail_progress.py
+
+Running `./retail_progress.py file.c` will attempt to figure out which functions
```
Yanis42, line 132: not a big deal but I was expecting to just run ``./retail_progress.py graph.c`` and it would magically find the path, but yeah I guess it makes sense if several files have the same name

maybe you should add this to avoid any confusion
```suggestion
Running `./retail_progress.py path/to/file.c` will attempt to figure out which functions
```

Yanis42 (commented):
I have few comments here and there but nothing really important, I tried to mess with the tools (just by reading ``docs/retails_versions.md``) and I was able to match 3 functions from ``game.c`` very easily, and I'm someone with almost 0 MIPS knowledge lol

really nice system, definitely gonna try to match stuff for fun

Makefile
```diff
@@ -185,16 +198,23 @@ ASSET_FILES_OUT := $(foreach f,$(ASSET_FILES_XML:.xml=.c),$f) \
 
 UNDECOMPILED_DATA_DIRS := $(shell find data -type d)
 
+BASEROM_SEGMENTS_DIR := baseroms/gc-eu-mq-dbg/segments
```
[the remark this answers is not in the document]
Yanis42: now that I tried the thing I'm thinking this is probably intended, forgot to read again my comments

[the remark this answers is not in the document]
cadmic: Yeah, currently the build system doesn't know how to extract segments from anything but the debug ROM. This will change soon, but I added a comment to hopefully prevent confusion

retail_progress.py
```diff
@@ -0,0 +1,216 @@
...
+            encoding="utf-8"
+        ).stdout.splitlines()
+    except subprocess.CalledProcessError as e:
+        return []
+
+    result = []
+
+    func_name = None
+    i = 6  # skip preamble
+    while i < len(lines):
```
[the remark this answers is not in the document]
cadmic: Yeah, for this loop in particular `i` is always increasing so it'll always terminate

tools/disasm/disasm.py
```diff
@@ -0,0 +1,136 @@
...
+    context.addBannedSymbolRange(0xA0000000, 0xFFFFFFFF)
+
+    spimdisasm.mips.InstructionConfig.parseArgs(args)
+    spimdisasm.common.GlobalConfig.parseArgs(args)
+
+    spimdisasm.common.GlobalConfig.ASM_USE_PRELUDE = False
+    spimdisasm.common.GlobalConfig.PRODUCE_SYMBOLS_PLUS_OFFSET = True
+    spimdisasm.common.GlobalConfig.TRUST_USER_FUNCTIONS = True
+
+    dma_files = parse_file_addresses(args.config_dir / "file_addresses.csv")
```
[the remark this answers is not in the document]
cadmic: I kinda think this will work correctly and disassembly 0 files. Maybe that's confusing, but `file_addresses.csv` is committed here so I don't think it's a danger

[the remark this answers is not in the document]
Yanis42: yeah it should be fine, I was just thinking of the very unlikely situation where someone doesn't have that file for some reasons

Yanis42 (approved):

retail_progress.py
Dragorn421, line 1: can you move this file into tools/

also could you add copyright/license info to your new scripts. You can use SPDX identifiers, for example:

```py
# SPDX-FileCopyrightText: Copyright 2024 cadmic
# SPDX-License-Identifier: CC0-1.0
```

and lastly for diff noise reasons could you format the scripts using black https://github.com/psf/black
I can also PR formatting them to your branch if you want

Dragorn421 (commented):
Nice work

docs/retail_versions.md
```diff
@@ -0,0 +1,166 @@
...
+For register and stack allocation differences, often the code can be tweaked so
+that it matches both the retail ROM while continuing to match the Debug ROM (for
+example, by reordering assignments or moving a local variable declaration inside
+an `if` block). Since retail MM versions use the same compiler flags as retail
+OOT, checking MM decomp for similar code can help.
+
+We can disable code that was removed in retail builds by adding
+`#ifdef OOT_DEBUG` around these parts of the code. In order too keep the code
+readable, we should try to minimize the amount of `#ifdef` noise whenever
+possible. For example, instead of writing
```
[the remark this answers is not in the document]
cadmic: Sure, I have removed the example (it was rather noisy anyway)

retail_progress.py
cadmic: I don't think this should go in tools/, it's more similar to diff.py/first_diff.py/sym_info.py in that it's something that will be run often and run by humans.

I do not want to claim sole ownership of these files for copyright, so I added the "© 2024 ZeldaRET" which was used in the decompression files.

cadmic: (I reformatted the files too)

Dragorn421 (approved):

Dragorn421:
Still would like @AngheloAlf 's input since we initially wanted to use splat for disassembly.
To restate what I said on discord I think it's fine to go ahead since this works, most of the diff is the csv data (the scripts themselves aren't much), and we can work on a splat approach separately.

Makefile
```diff
@@ -185,16 +198,24 @@ ASSET_FILES_OUT := $(foreach f,$(ASSET_FILES_XML:.xml=.c),$f) \
 
 UNDECOMPILED_DATA_DIRS := $(shell find data -type d)
 
+# For now, ROM segments are still taken from the Debug ROM even when building other versions
```
hensldm, line 201: Very small nit picky, but could you add a `TODO`s for comments like this. I know they are meant to be short lived and you might even already be working on them, but it just makes them easier to find in the future.

hensldm (approved):
Again looks good. 1 small nit picky comment.

Makefile
cadmic: Sounds good, done

PR #1614, opened 2024-01-18
Setup rom decompression and compression

AngheloAlf:
- Setup decompressing a rom
  - Decompression is handled by the new `tools/decompress_baserom.py` script.
  - New script supersedes the old `fix_baserom.py` script.
  - The decompressed rom is named ``baserom_uncompressed.{VERSION}.z64``
- Builds a compressed ROM using `crunch64`
  - The compressed segments listed in the `spec` file are taken from `gc-eu-mq`
  - Makefile will still build uncompressed roms by default. To build a compressed one use `make compressed`.
- Renamed the baseroms: The old `baserom_original.z64` is now named `baserom.{VERSION}.z64`
- New python dependencies
  - The scripts to handle decompression and compression depend of a few python packages, so I made a `requirements.txt` file. Added install instructions on the readme.
  - I included instructions to setup a python virtual environment on the readme


README.md
```diff
@@ -29,7 +29,7 @@ The only build currently supported is Master Quest (Debug), but other versions a
 
 It builds the following ROM:
 
-* oot-gc-eu-mq-dbg.z64 `md5: f0b7f35375f9cc8ca1b2d59d78e35405`
+* oot_uncompressed-gc-eu-mq-dbg.z64 `md5: f0b7f35375f9cc8ca1b2d59d78e35405`
```
cadmic, line 32: It seems a bit strange to me that final output ROM, `oot-gc-eu-mq-dbg.z64`, does not match the debug ROM. What about a `COMPRESS` makefile variable, where if COMPRESS=0 then the uncompressed ROM is directly copied to the output ROM? It would be set to 1 for retail versions and 0 for debug, but modders could override it if they want their mod to be compressed.

Makefile
```diff
@@ -158,11 +158,19 @@ endif
...
+# rom compression flags
+COMPFLAGS := --threads $(N_THREADS)
+ifneq ($(NON_MATCHING),1)
+  COMPFLAGS += --matching
+endif
+
 #### Files ####
 
 # ROM image
 ROM := oot-$(VERSION).z64
```
cadmic, line 170:
```suggestion
```

Makefile
```diff
@@ -290,13 +304,18 @@ endif
 	$(N64_EMULATOR) $<
 
 
-.PHONY: all clean setup run distclean assetclean
+.PHONY: all uncompressed compressed clean setup run distclean assetclean
+.DEFAULT_GOAL := uncompressed
```
cadmic, line 308: I think `uncompressed` is already the default because it's the first goal. But why should the default be `uncompressed` instead of `all`?

tools/decompress_baserom.py
```diff
@@ -0,0 +1,258 @@
...
+import io
+import struct
+from pathlib import Path
+import argparse
+
+import crunch64
+import ipl3checksum
+import zlib
+
+def decompressZlib(data: bytes) -> bytes:
```
cadmic, line 18: Are any OOT roms zlib-compressed?

tools/decompress_baserom.py
cadmic, line 1: I find the mix of camelCase and snake_case jarring and I wouldn't know which one to use if I were to modify this file. Generally Python style is to always using snake_case for variables and functions

(ditto for z64compress_wrapper.py)

tools/decompress_baserom.py
```diff
@@ -0,0 +1,258 @@
+#!/usr/bin/env python3
+
+# SPDX-FileCopyrightText: © 2024 ZeldaRET
+# SPDX-License-Identifier: MIT
```
cadmic, line 4: Why does this file get a copyright text and license identifier? None of the other source files have it.

(ditto for z64compress_wrapper.py)

tools/decompress_baserom.py
```diff
@@ -0,0 +1,258 @@
...
+        output.extend(decomp.decompress(decomp.unconsumed_tail))
+    output.extend(decomp.flush())
+    return bytes(output)
+
+def decompress(data: bytes, is_zlib_compressed: bool) -> bytes:
+    if is_zlib_compressed:
+        return decompressZlib(data)
+    return crunch64.yaz0.decompress(bytes(data))
+
+FILE_TABLE_OFFSET = {
```
cadmic, line 32: Could we wait to add other versions for now? For example I'm not yet sure how to name NTSC-J and NTSC-U.

tools/decompress_baserom.py
```diff
@@ -0,0 +1,258 @@
...
+    "gc-us":            "903c64d4b9b58df256f3ee4c2714e3b0",
+    "gc-us-mq":         "ef50e687a31c49cdc49b5fce867c3c3c",
+    "gc-eu-mq-dbg":     "f0b7f35375f9cc8ca1b2d59d78e35405",
+    "gc-eu":            "71db1704e62966f2949653d4fe41e1c2",
+    "gc-eu-mq":         "1a438f4235f8038856971c14a798122a",
+    "ique-cn":          "17a9f30d722c29e6912bd4b66713d2b0",
+    "ique-zh":          "291d14928fbe5bc90642bd9cd9b2b8cd",
+}
+
+def round_up(n,shift):
```
cadmic, line 68: In general there's lots of formatting errors here and in z64compress_wrapper.py. Maybe run them through an autoformatter?

tools/spec.c
```diff
@@ -216,10 +217,13 @@ bool parse_segment_statement(struct Segment *currSeg, STMTId stmt, char* args, i
...
         currSeg->includes[currSeg->includesCount - 1].dataWithRodata = (stmt == STMT_include_data_with_rodata);
         break;
-        case STMT_increment:
+    case STMT_increment:
         if (!parse_number(args, &currSeg->increment))
             util_fatal_error("line %i: expected number after 'increment'", lineNum);
         break;
+    case STMT_compress:
+            currSeg->compress = true;
+            break;
```
cadmic, line 226:
```suggestion
        currSeg->compress = true;
        break;
```

Makefile
```diff
@@ -158,11 +158,19 @@ endif
...
+  COMPFLAGS += --matching
+endif
+
 #### Files ####
 
 # ROM image
 ROM := oot-$(VERSION).z64
-ELF := $(ROM:.z64=.elf)
+ROMC := oot-$(VERSION).z64
+ROM := oot_uncompressed-$(VERSION).z64
```
cadmic, line 172: I find the ROM naming a bit inconsistent. Right now there's

```
baserom.gc-eu-mq-dbg.z64
baserom_uncompressed.gc-eu-mq-dbg.z64
oot-gc-eu-mq-dbg.z64
oot_uncompressed-gc-eu-mq-dbg.z64
```

There's a mix of underscores and dashes, and the version is separated  by a period in `baserom.` but by a dash in `oot-`.

One option is to go all in on periods, e.g. something like

```
baserom.gc-eu-mq-dbg.z64
baserom.gc-eu-mq-dbg.uncompressed.z64
oot.gc-eu-mq-dbg.z64
oot.gc-eu-mq-dbg.uncompressed.z64
```

although I'm not totally sold on that

Makefile
```diff
@@ -279,7 +293,7 @@ distclean: clean assetclean
 
 setup:
 	$(MAKE) -C tools
-	python3 fixbaserom.py
+	python3 tools/decompress_baserom.py $(VERSION)
```
cadmic, line 296: This feels redundant right now because `extract_baserom.py` already knows how to decompress yaz0 files. Is the idea that multiversion asset extraction would work directly from the uncompressed rom instead of first extracting the rom files?

tools/z64compress_wrapper.py
```diff
@@ -0,0 +1,111 @@
...
+    parser.add_argument("spec", help="path to processed spec file", type=Path)
+    parser.add_argument("--cache", help="Cache directory")
+    parser.add_argument("--threads", help="Number of threads to run compression on, 0 disables multithreading", type=int, default=0)
+    parser.add_argument("--mb", help="compressed ROM size in MB, default is the smallest multiple of 8mb fitting the whole ROM")
+    parser.add_argument("--matching", help="matching compression, forfeits some useful optimizations", action="store_true")
+    parser.add_argument("--stderr", help="z64compress will write its output messages to stderr instead of stdout", action="store_true")
+
+    args = parser.parse_args()
+
+    dma_addr, dna_len = get_dmadata_start_len(args.elf)
```
cadmic, line 75: dna -> dma

tools/decompress_baserom.py
fig02: reading some discord history, ique is

Makefile
AngheloAlf: I think explicit instead of implicit is usually better.
Building a compressed rom is slow and just building an uncompressed one is usually enough to know the rom is matching or to test any changes

tools/decompress_baserom.py
AngheloAlf: Indeed, most iQue games use DEFLATE instead of yaz0.
We still don't have a way to match the specific version of DEFLATE used on iQue games yet, but iQue is usually low on everybody's priority list

tools/ZAPD/Jenkinsfile
```diff
@@ -38,7 +38,7 @@ pipeline {
                 stage('Setup OOT') {
                     steps {
                         dir('oot') {
-                            sh 'cp /usr/local/etc/roms/baserom_oot.z64 baserom_original.z64'
+                            sh 'cp /usr/local/etc/roms/baserom_oot.z64 baserom.gc-eu-mq-dbg.z64'
```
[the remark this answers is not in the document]
AngheloAlf: What do you mean by changed upstream?

README.md
AngheloAlf: Personally I think running `make compress` and getting an uncompressed rom on some versions but not on others is bad.
I would like to hear more people's opinions about this

tools/decompress_baserom.py
AngheloAlf: It is the result of mix and matching different scripts to produce this. I'll change them to use snake_case

tools/decompress_baserom.py
AngheloAlf: We have been adopting this on our scripts to make their license explicit. We have some tools in the repo that we don't even know where they came from and it have been the topic of discussion for some time.
As a reference we have been adopting the license identifier for scripts on other repos like MM and AF

tools/decompress_baserom.py
AngheloAlf: You would prefer this list to only have `gc-eu-mq-dbg` and `gc-eu-mq` then?

Makefile
AngheloAlf: Having a decompressed rom to begin with is nice since you can use it to compare your built uncompressed rom in case it doesn't match with tools like bdiff or vbindiff

tools/ZAPD/Jenkinsfile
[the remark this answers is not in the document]
cadmic: I'm in changed in the ZAPD repo, not this subrepo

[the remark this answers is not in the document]
AngheloAlf: oh frick, I didn't notice this was the ZAPD Jenkins file

tools/decompress_baserom.py
cadmic: Yeah (well technically only `gc-eu-mq-dbg` for the current state of the repo, but `gc-eu-mq` will be added soon anyway)

tools/decompress_baserom.py
cadmic: got it, thanks

README.md
fig02: Im not sure the best approach to solve this, but I do think itd be a bit weird if `oot-gc-eu-mq-dbg.z64` is not the name for the definitive matching debug rom

Makefile
fig02: personally not a fan of too many dots, my brain automatically associates things after dots as file extensions. but I dont have an alternative solution to offer at this time

cadmic: A dotless way could be something like

```
baserom-gc-eu-mq-dbg.z64
baserom-gc-eu-mq-dbg-uncompressed.z64
oot-gc-eu-mq-dbg.z64
oot-gc-eu-mq-dbg-uncompressed.z64
```

Makefile
cadmic: Alright, I'm on board

Makefile
Dragorn421: I would like baseroms to go in a subfolder tbh, since there are a lot of them and even variants like "pal mq debug with overdump", "pal mq debug stripped", "1.0 compressed", "1.0 uncompressed" ...

We can also move the md5 sums to that subfolder. And relevant scripts

And the current baserom/ folder to `baseroms/segments/{version}/` ? idk

fig02: I like just dashes personally

fig02: so with subfolders, youd rather just name them "baserom.z64" and let the folder handle the versioning?

README.md
fig02: I tbh am not sure where I stand on this. 
Theres two lenses to look at this through: OoT historical documentation/decomp mindset, and a general use (modding) codebase.

Logically it makes sense to label the uncompressed rom as uncompressed, and I do recognize that, but from an OoT "historian" viewpoint, it makes sense for the "not standard" format to have the "non standard" name.

Makefile
fig02: image
im fine with this approach mentioned on discord

tools/decompress_baserom.py
fig02: iirc the cc0 licensing effort explicitly states that tools can and should detail its own license if it wants to. I also am not sure on the details though, @Dragorn421 would know

README.md
fig02: is `COMPARE` at the top supposed to be `COMPRESS`?
and  "oot-$(VERSION)-uncompressed.z64 is the "compressed ROM"", I assume you mean is the uncompressed rom?

So if I am following, for gc-eu-mq-dbg, we are going to get both a "uncompressed rom" and an "output rom", and both are going to be the exact same in that specific case?

cadmic: Yes, apologies for the sloppiness.

Yeah, if you do not want compression, the output rom is an exact copy of the uncompressed rom

fig02: ok, I think this is the best sounding compromise so far

tools/decompress_baserom.py
Dragorn421: Not sure what to explain but it's good that tools are explicitly licensed

The cc0 agreement is a blanket cover for everything so we (and repo users in general) can do literally what we want without consulting all contributors (caveat: not all contributors agreed)

You may argue that since anghelo agreed as roughly "all my contributions are cc0" you may still take the MIT code he contributes as cc0, but that's not the spirit of it. Plus this MIT licensing is much more specific than the blanket cc0 statement so feels like it prevails
In the end it's the honor system anyway, what matters if you do have honor is to feel like you're abiding by the authors' wishes (which is easy to check when the wishes are expressed with licenses)

AngheloAlf: btw if we want to use a different license for those scripts I'm happy to change it.
I used MIT since that's the one I use as a default for almost everything, and I'm happy to gave up those scripts contributions to ZeldaRET instead of explicitly having my name on it

Makefile
AngheloAlf: I mean, the idea of `all` is to build `all`, and most of the time you won't want to build a compressed rom, so that's why it isn't the default.
At some point in MM we built the compressed rom by default, and it was painfully slow for getting work done, so we changed it to build the uncompressed rom by default

README.md
mzxrules: > Logically it makes sense to label the uncompressed rom as uncompressed

I don't agree. Generating the rom in it's decompressed form should be considered standard output for compilation, so it shouldn't need "uncompressed" tacked on to the end. On the other hand, compressed roms should be labeled as being compressed, since it makes it more clear that a transformation on the original data has occurred. 

Arguably, we may also want to specify the type of compression that has been applied to the rom, since not all versions use the same compression algorithm.

fig02: I keep flip flopping back and forth here, but I guess I do agree that "uncompressed" should probably be unlabeled, and compressed which has data altered should be labeled as such.
I do not agree on including the compression algorithm in the name tho, I dont find it necessary

.gitignore
```diff
@@ -10,12 +10,14 @@ __pycache__/
...
 cmake-build-debug
 venv/
+.venv/
 
 # Project-specific ignores
 build/
 expected/
 notes/
 baserom/
+baseroms/*/segments/
```
fig02, line 20: is this rule needed? doesnt the one above cover the whole folder?

.gitignore
```diff
@@ -10,12 +10,14 @@ __pycache__/
 CMakeLists.txt
 cmake-build-debug
 venv/
+.venv/
```
fig02, line 13: do we need two different venv ignores?

.gitignore
Dragorn421: baserom/
vs
baseroms/

That said I think this means we can remove baserom/ (moved to baseroms/VERSION/segments/)

.gitignore
Dragorn421: You can name your venv folder however you want, both are typically used so it doesn't hurt

Jenkinsfile
```diff
@@ -4,18 +4,31 @@ pipeline {
...
             }
         }
         stage('Build (qemu-irix)') {
             when {
                 branch 'main'
             }
             steps {
-                sh 'ORIG_COMPILER=1 make -j'
+                sh '''. .venv/bin/activate
+                make -j ORIG_COMPILER=1
```
Dragorn421, line 30: Cool I guess #1599 broke Jenkins building with irix-qemu ido :) nice catch

Makefile
```diff
@@ -321,14 +340,14 @@ $(BUILD_DIR)/ldscript.txt: $(BUILD_DIR)/$(SPEC)
 $(BUILD_DIR)/undefined_syms.txt: undefined_syms.txt
 	$(CPP) $(CPPFLAGS) $< > $@
 
-$(BUILD_DIR)/baserom/%.o: baserom/%
+$(BUILD_DIR)/baseroms/$(VERSION)/segments%.o: baseroms/$(VERSION)/segments%
```
Dragorn421, line 343: VERSION is already part of BUILD_DIR so ideally we'd have

```suggestion
$(BUILD_DIR)/baserom/%.o: baseroms/$(VERSION)/segments/%
```
But the relevant bit in setting O_FILES may become unreadable

This would matter a lot in the spec that iirc currently includes some baserom files, and in the near future many more if we ifdef-build-OK another version this way

Makefile
cadmic: Sorry, what was the resolution for the output rom names?

fig02: None so far, still being discussed in https://github.com/zeldaret/oot/pull/1614#discussion_r1456907385

tools/decompress_baserom.py
cadmic: (you can resolve this)

README.md
```diff
@@ -68,13 +68,15 @@ The build process has the following package requirements:
 * build-essential
 * binutils-mips-linux-gnu
 * python3
+* pip3
+* venv
```
cadmic, line 72: * python3-pip
* python3-venv

README.md
```diff
@@ -98,23 +100,47 @@ This will copy the GitHub repository contents into a new folder in the current d
...
+```bash
+. .venv/bin/activate
+```
+
+Take in mind for each new terminal you'll need to **active** the Python virtual environment again. There's no need to create the virtual environment again.
+
+Now you can install the Python dependencies, to do so run:
+
+```bash
+python3 -m pip install -U -r requirements.txt
```
cadmic, line 124: Why not `pip install -r requirements.txt`? That's what basically every other Python project uses so it's ingrained in my muscle memory

README.md
```diff
@@ -98,23 +100,47 @@ This will copy the GitHub repository contents into a new folder in the current d
...
+```
+
+#### 4. Prepare a base ROM
 
 Copy over your copy of the Master Quest (Debug) ROM inside the root of this new project directory.
-Rename the file to "baserom_original.z64", "baserom_original.n64" or "baserom_original.v64", depending on the original extension.
+Rename the file to "baserom.gc-eu-mq-dbg.z64", "baserom_original.n64" or "baserom_original.v64", depending on the original extension.
```
cadmic, line 130: Is this still accurate?

Makefile
cadmic: Oh I didn't realize you were trying to combine the discussions, sorry

fig02: They are the same conversation right? Just wanna make sure I'm not missing something

README.md
AngheloAlf: My way of seeing it is since the final result of the build process will produce a rom that matches the contents of the cartridge then that one should be favored and be unlabelled.
Another alternative would be to label both: `oot-compressed.z64` and `oot-uncompressed.z64`

.gitignore
AngheloAlf: I keep the original `baserom` because otherwise when you pull from main git will want to add all the old files from `baserom`. This way we prevent somebody from committing those files by mistake

.gitignore
AngheloAlf: I prefer using `.venv`  (with a dot) because that's a hidden folder on Linux. I'm not sure if explorer.exe would honor this when examining the WSL filesystem.
I didn't remove the other one in case somebody was already using it

Jenkinsfile
AngheloAlf: Yeah, that was totally my intention and not me just preferring to pass makefile variable at the right instead of the left. Totally

:sweat_smile:

Makefile
AngheloAlf: Oh right, I didn't notice this.
This will be fun to deal with...

README.md
AngheloAlf: Using `pip` instead of `python3 -m pip` can be problematic if you have more than one version of Python installed, which can happen if either you install it manually or your distro decides to push a newer version.
It is better practice to do this, but if anybody wants to just type `pip` then it should work in most cases.

As a safe-ish option you can slap `alias pip="python3 -m pip"` on your `~/.bashrc` file and forget about this issue.

README.md
fig02: Yeah I mentioned above that the 2 main ways of looking at it above is "oot rom historian" or "result from a generic build pipeline". And while I understand the historian viewpoint, I think just naming them after how it objectively is (the "game" and the "compressed game"), is much more consistent and better long term.

we can try to document somewhere which ones are originally not compressed and compressed

fig02: I do recognize the compromise option of labeling both, but do think mzx has a point here that labeling something as "uncompressed" seems a little redundant. It is just the output of make which builds the game. It get compressed after the fact to create a new variation of that original output. So I think "uncompressed" being unlabeled does make alot of sense here

README.md
Dragorn421: Yeah I also wanted to suggest `pip` over `python3 -m pip` but couldn't find solid reference on it. Everything I found even used `python3 -m pip`, including the "official" pip tutorial

Makefile
```diff
@@ -201,7 +210,7 @@ TEXTURE_FILES_OUT := $(foreach f,$(TEXTURE_FILES_PNG:.png=.inc.c),$(BUILD_DIR)/$
 					 $(foreach f,$(TEXTURE_FILES_JPG:.jpg=.jpg.inc.c),$(BUILD_DIR)/$f) \
 
 # create build directories
-$(shell mkdir -p $(BUILD_DIR)/baserom $(BUILD_DIR)/assets/text $(foreach dir,$(SRC_DIRS) $(UNDECOMPILED_DATA_DIRS) $(ASSET_BIN_DIRS),$(BUILD_DIR)/$(dir)))
+$(shell mkdir -p $(BUILD_DIR)/baserom/$(VERSION)/segments $(BUILD_DIR)/assets/text $(foreach dir,$(SRC_DIRS) $(UNDECOMPILED_DATA_DIRS) $(ASSET_BIN_DIRS),$(BUILD_DIR)/$(dir)))
```
Dragorn421, line 213:
```suggestion
$(shell mkdir -p $(BUILD_DIR)/baserom $(BUILD_DIR)/assets/text $(foreach dir,$(SRC_DIRS) $(UNDECOMPILED_DATA_DIRS) $(ASSET_BIN_DIRS),$(BUILD_DIR)/$(dir)))
```

README.md
cadmic: `pip` here refers to the command in the virtualenv, not the distro one

Dragorn421: > `pip` here refers to the command in the virtualenv, not the distro one

Yeah I haven't found reference on that

I do use it typically but since I can't find it documented I wouldn't advocate document using it

Makefile
cadmic: In my mind they were slightly different (ROM naming vs semantics of the build system) so I was a bit confused why you resolved this, but combining them is fine by me. I should have DMed you instead

README.md
cadmic: The official docs at https://docs.python.org/3/library/venv.html say

> Changed in version 3.4: Installs pip by default, added the `--without-pip` and `--copies` options

and later it suggests using `pip install -r requirements.txt` (though oddly, only in a warning)

README.md
```diff
@@ -98,23 +100,48 @@ This will copy the GitHub repository contents into a new folder in the current d
...
+```
+
+To start using the virtual environment on your current terminal run:
+
+```bash
+. .venv/bin/activate
+```
+
+Take in mind for each new terminal you'll need to **active** the Python virtual environment again. There's no need to create the virtual environment again.
+
```
fig02, line 120:
```suggestion
Keep in mind for each new terminal session you'll need to **activate** the Python virtual environment again. There's no need to create the virtual environment again.

```

Makefile
```diff
@@ -246,26 +255,32 @@ $(BUILD_DIR)/src/libultra/rmon/%.o: CC := $(CC_OLD)
...
 else
 $(BUILD_DIR)/src/libultra/libc/ll.o: OPTFLAGS := -Ofast
 $(BUILD_DIR)/src/%.o: CC := $(CC) -fexec-charset=euc-jp
 endif
 
 #### Main Targets ###
 
-all: $(ROM)
-ifeq ($(COMPARE),1)
+uncompressed: $(ROM)
```
hensldm, line 270: I'm wondering if we should actually just name this makefile target `rom` instead of `uncompressed` (i.e. we run `make rom`)? Since we decided to label compressed roms as they are the transformed data.

Makefile
cadmic: This is fine by me now that the "default" output rom is uncompressed

Makefile
fig02: I agree I think that makes the most sense

extract_baserom.py
```diff
@@ -1,12 +1,12 @@
...
-import sys
 import struct
 from multiprocessing import Pool, cpu_count
+from pathlib import Path
 
 
-ROM_FILE_NAME = 'baserom.z64'
+ROM_FILE_NAME = Path('baseroms/gc-eu-mq-dbg/baserom_uncompressed.z64')
+SEGMENTS_PATH = Path('baseroms/gc-eu-mq-dbg/segments/')
 FILE_TABLE_OFFSET = 0x12F70
```
fig02, line 10: slight semantics thing, but im wondering if `baserom_decompressed` makes more sense as a name. as in, the rom that was once compressed and then got de-compressed for extraction.

Either way the name doesnt really make that much sense for debug rom since its not compressed... but idk

fig02: no matter what name we go with, we might want to switch to using a dash instead of an underscore, to match the rest of the rom names we've been going with

Makefile
```diff
@@ -184,9 +192,10 @@ UNDECOMPILED_DATA_DIRS := $(shell find data -type d)
 # source files
 C_FILES       := $(filter-out %.inc.c,$(foreach dir,$(SRC_DIRS) $(ASSET_BIN_DIRS),$(wildcard $(dir)/*.c)))
 S_FILES       := $(foreach dir,$(SRC_DIRS) $(UNDECOMPILED_DATA_DIRS),$(wildcard $(dir)/*.s))
+BIN_FILES     := $(wildcard baseroms/$(VERSION)/segments/*)
```
Dragorn421, line 195:
```suggestion
BASEROM_BIN_FILES := $(wildcard baseroms/$(VERSION)/segments/*)
```
rip the alignment but I think this is too generic of a name , in particular `$(foreach f,$(BIN_FILES),$(BUILD_DIR)/baserom/$(notdir $f).o)` later simply assumes it to be the baserom segments binaries

README.md
```diff
@@ -98,23 +100,48 @@ This will copy the GitHub repository contents into a new folder in the current d
...
 
-#### 3. Prepare a base ROM
+#### 3. Install python dependencies
 
-Copy over your copy of the Master Quest (Debug) ROM inside the root of this new project directory.
-Rename the file to "baserom_original.z64", "baserom_original.n64" or "baserom_original.v64", depending on the original extension.
+The build process has a few python packages required that are located in `requirements.txt`.
 
-#### 4. Setup the ROM and build process
+It is recommend to setup a virtual environment for python to localize all dependencies. To create a virtual environment:
```
Dragorn421, line 107:
```suggestion
It is recommended to set up a virtual environment for python to contain all dependencies. To create a virtual environment:
```

README.md
```diff
@@ -98,23 +100,48 @@ This will copy the GitHub repository contents into a new folder in the current d
...
+
+```bash
+python3 -m venv .venv
+```
+
+To start using the virtual environment on your current terminal run:
```
Dragorn421, line 113:
```suggestion
To start using the virtual environment in your current terminal run:
```

README.md
```diff
@@ -98,23 +100,48 @@ This will copy the GitHub repository contents into a new folder in the current d
...
+
+```bash
+. .venv/bin/activate
+```
+
+Keep in mind for each new terminal session you'll need to **activate** the Python virtual environment again. There's no need to create the virtual environment again.
```
Dragorn421, line 119:
```suggestion
Keep in mind that for each new terminal session, you will need to **activate** the Python virtual environment again. That is, run the above `. .venv/bin/activate` command.
```

README.md
```diff
@@ -98,23 +100,48 @@ This will copy the GitHub repository contents into a new folder in the current d
...
+
+Now you can install the Python dependencies, to do so run:
+
+```bash
+python3 -m pip install -U -r requirements.txt
+```
+
+#### 4. Prepare a base ROM
+
+Copy over your copy of the Master Quest (Debug) ROM inside the `baseroms/gc-eu-mq-dbg/` path of this new project directory.
```
Dragorn421, line 129: while we're at it, may as well
```suggestion
Place a copy of the Master Quest (Debug) ROM inside the `baseroms/gc-eu-mq-dbg/` folder.
```

README.md
```diff
@@ -98,23 +100,48 @@ This will copy the GitHub repository contents into a new folder in the current d
...
+#### 5. Setup the ROM and build process
 
 Setup and extract everything from your ROM with the following command:
 
 ```bash
 make setup
 ```
 
-This will generate a new ROM called "baserom.z64" that will have the overdump removed and the header patched.
+This will generate a new ROM called "baseroms/gc-eu-mq-dbg/baserom-uncompressed.z64" that will have the overdump removed and the header patched.
```
Dragorn421, line 141:
```suggestion
This will generate a new ROM "baseroms/gc-eu-mq-dbg/baserom-uncompressed.z64" that will have the overdump removed and the header patched.
```

docs/tutorial/object_decomp.md
```diff
@@ -113,7 +113,7 @@ If in doubt, look at completed objects in the repo, and if still in doubt, ask.
 Just as when you decomp an actor you have to change the `spec` to tell it to use the new files, you have to do a similar thing for the object. Find the appropriate section for the object you have decompiled, and replace the line
 
 ```c
-include "$(BUILD_DIR)/baserom/object_name.o"
+include "$(BUILD_DIR)/baseroms/gc-eu-mq-dbg/segments/object_name.o"
```
Dragorn421, line 116:
```suggestion
include "$(BUILD_DIR)/baserom/object_name.o"
```
and elsewhere

docs/vscode.md
```diff
@@ -74,7 +74,7 @@ Add the following to (or create) the `.vscode/settings.json` file for VSCode to
     "search.useIgnoreFiles": false,
     "search.exclude": {
         "**/.git": true,
-        "baserom/**": true,
+        "baseroms/gc-eu-mq-dbg/segments/**": true,
```
Dragorn421, line 77: does this syntax exist/work:
```suggestion
        "baseroms/*/segments/**": true,
```

extract_baserom.py
```diff
@@ -1,12 +1,12 @@
...
 
-import os
-import sys
 import struct
 from multiprocessing import Pool, cpu_count
+from pathlib import Path
 
 
-ROM_FILE_NAME = 'baserom.z64'
+ROM_FILE_NAME = Path('baseroms/gc-eu-mq-dbg/baserom-uncompressed.z64')
```
Dragorn421, line 8:
```suggestion
ROM_FILE_PATH = Path('baseroms/gc-eu-mq-dbg/baserom-uncompressed.z64')
```

tools/overlayhelpers/colliderinit.py
```diff
@@ -387,7 +387,7 @@ def GetColliderInit(address, type, num, path):
     else:
         return("ItemInit type must specify number of elements")
 
-    ovlFile = open(path + "/baserom/" + fileResult.name, "rb")
+    ovlFile = open(path + "/baseroms/" + fileResult.name, "rb")
```
Dragorn421, line 390: should be
```suggestion
    ovlFile = open(path + "/baseroms/gc-eu-mq-dbg/segments/" + fileResult.name, "rb")
```
afaict, but it's not like this script is going to be used again anyway (?)

same with other `tools/overlayhelpers/*.py`

tools/z64compress_wrapper.py
Dragorn421, line 1: Is there a reason to use z64compress over crunch64?

AngheloAlf: Responded here: https://github.com/zeldaret/mm/pull/1542#issuecomment-1904957459

z64compress is already aware of how a zelda64 rom is structured while crunch64 only provides matching (de)compressor for yaz0 and others. z64compress also offers other features that can be interesting in modding contexts

docs/vscode.md
fig02: should probably just exclude the entire baseroms folder? at least I cant think of anything I would want from there to show up in my searches

fig02 (approved):

README.md
cadmic: bump

cadmic (approved):

README.md
hensldm: https://poonam-jangid01.medium.com/to-clarify-when-you-use-it-you-are-invoking-the-pip-module-as-a-script-within-the-python-13170aa3f59f seems to be a good write up about the differences between them both. Tbh I don't really have a preference either way and not sure it really matters which we document (though `python3 -m pip` seems the "safer" option.)

hensldm (approved):

README.md
cadmic: That doesn't seem to be a particularly high-quality source. What does "multiple virtual environments" even mean? If you could have multiple virtual environments, wouldn't `python` be just as ambiguous as `pip`?

hensldm: > What does "multiple virtual environments" even mean?

You can create as many virtual environments as you would like (different python versions, different packages based on projects), so it is very possible people have multiple virtual environments "installed" though only 1 can be active.
As stated in the article `python` will always point to the current activated python interpreter, so you can be assured which python is being used. On the other hand if you do use `pip` it will use the first pip it finds in the system path. When activating a virtual environment it puts that venv's bin folder first in the path so if it has pip, it would be fine since that pip should be found first. However if it doesn't, it would mess with whatever pip it does find on your system path which might cause issues.
New venv's seem to have pip installed by default now, so again either should work for our use case.

tools/z64compress_wrapper.py
```diff
@@ -0,0 +1,138 @@
...
+        directives = segment.split("endseg")[0].split("\n")
+
+        for directive in directives:
+            directive = directive.strip()
+            if directive.startswith("flags") and "NOLOAD" in directive:
+                shift += 1
+            elif directive.startswith("compress"):
+                compress_segments.append(i - shift)
+
+    return list(ranges(compress_segments))
```
Dragorn421, line 38: idk if this is the case in MM but in oot-gc-eu-mq-dbg the dmadata and the spec aren't ordered exactly the same
(`include/tables/dmadata_table_mqdbg.h` is used unless `NON_MATCHING`)

Tharo https://discord.com/channels/688807550715560050/688851317593997489/1181707645934121020
> map_i_static and map_48x85_static are reordered, and syotes is the first scene in the matching layout while in the spec it's several scenes down

so this code relying on spec order to tell which dma entries are compressed only works "by chance" kinda :sweat_smile:

AngheloAlf: arrrrggg

I guess a "proper" fix for this issue would be to gather all the relocations from the dmatable from the built elf to know what are the proper indices.

Would you prefer if we tried to fix this now or in the future?

Dragorn421: since this is only an "issue" for matching builds but that it works anyway, I think this is fine to not fix at all. but it should be documented

or I may fix it if we yeet z64compress like I would like to

tools/overlayhelpers/damage_table.py
```diff
@@ -132,7 +132,7 @@ def get_damage_bytes(address, repo):
 
     print(file_result)
 
-    with open(repo + os.sep + "baserom" + os.sep + file_result.name, "rb") as ovl_file:
+    with open(repo + os.sep + "baseroms" + os.sep + file_result.name, "rb") as ovl_file:
```
Dragorn421, line 135:
```suggestion
    with open(repo + os.sep + "baseroms/gc-eu-mq-dbg/segments" + os.sep + file_result.name, "rb") as ovl_file:
```

fig02 (approved):

Makefile
```diff
@@ -290,13 +300,20 @@ endif
...
+.DEFAULT_GOAL := rom
+all: rom compressed
 
 #### Various Recipes ####
 
 $(ROM): $(ELF)
 	$(ELF2ROM) -cic 6105 $< $@
 
+$(ROMC): $(ROM) $(ELF) $(BUILD_DIR)/compress_ranges.txt
+# note: $(BUILD_DIR)/compress_ranges.txt should only be used for nonmatching builds. it works by chance for matching builds too though
```
Dragorn421, line 313: Would like to point out this hack in my implementation
(it's not a new "issue", it was already present in the previous `z64compress_wrapper.py`-based implementation, cf https://github.com/zeldaret/oot/pull/1614#discussion_r1463455888 )

`$(BUILD_DIR)/compress_ranges.txt` is generated from the spec, but the OK dmadata order isn't necessarily the spec's segments order (in oot-gc-eu-mq-dbg at least, that's why we have `include/tables/dmadata_table_mqdbg.h`)

as previously I think it's fine to not address this, since the OK builds aren't going to change, then as long as it works, it works

Dragorn421 (approved):

fig02:
I got confirmation from people that approved before the latest changes that their approvals still stand-- merging

PR #1618, opened 2024-01-24
Suggest pip install -U -r requirements.txt

cadmic:
This is what's used for the vast majority of Python projects, so I think the more complicated command is just needlessly confusing to newcomers. Note that `pip` is not more unsafe than `python3 -m pip` because pip is always installed by virtualenv since Python 3.4

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ Keep in mind that for each new terminal session, you will need to **activate** t
 Now you can install the Python dependencies, to do so run:
 
 ```bash
-python3 -m pip install -U -r requirements.txt
+pip install -r requirements.txt
 ```
 
 #### 4. Prepare a base ROM
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ Keep in mind that for each new terminal session, you will need to **activate** t
 Now you can install the Python dependencies, to do so run:
 
 ```bash
-pip install -r requirements.txt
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -121,7 +121,7 @@ Keep in mind that for each new terminal session, you will need to **activate** t
 Now you can install the Python dependencies, to do so run:
 
 ```bash
-python3 -m pip install -U -r requirements.txt
+pip install -r requirements.txt
```
Dragorn421, line 124: I'd keep `-U` "upgrade already installed packages"
```suggestion
pip install -U -r requirements.txt
```

out of curiosity do you have a source for pip is always in venvs

cadmic: Copied from the other thread:

The official docs at https://docs.python.org/3/library/venv.html say

Changed in version 3.4: Installs pip by default, added the --without-pip and --copies options

and later it suggests using pip install -r requirements.txt (though oddly, only in a warning)

cadmic: Sure, we can keep `-U` although it's not necessary for an install. Ideally I'd like people to understand how to use pip and not blindly copy commands, and keeping the commands as simple as possible helps with that, but I suppose it's fine

Dragorn421 (approved):

fig02 (approved):

PR #1619, opened 2024-01-24
Use virtualenv in Makefile for python dependencies

cadmic:
`make setup` will create a virtualenv if it doesn't already exist, and all python commands in the Makefile will use the virtualenv without the user having to activate it. This is so the build system can use pip packages without the user having to learn about virtualenv, since that can be hard for beginners.

You will still have to activate the virtualenv if you want to run scripts like `diff.py` by hand, though. I've moved the docs for this out of the main README and into the docs for the tools.

I'm not sure if `distclean` should nuke the virtualenv too. I chose not to here, because I couldn't find a way to automatically deactivate it (`deactivate` it is a shell function, not a command in the virtualenv), and it hopefully it doesn't get too messed up.

(Thanks to kentonm for the idea)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -263,6 +263,8 @@ endif
 
 #### Main Targets ###
 
+all: rom compressed
+
 rom: $(ROM)
 ifneq ($(COMPARE),0)
 	@md5sum $(ROM)
@@ -288,9 +290,12 @@ distclean: clean assetclean
 	$(RM) -r baseroms/$(VERSION)/segments
 	$(MAKE) -C tools distclean
 
-install-python-deps: $(VENV)/.install-python-deps
+venv:
+	test -d $(VENV) || python3 -m venv $(VENV)
+	$(PYTHON) -m pip install -U pip
+	$(PYTHON) -m pip install -U -r requirements.txt
 
-setup: install-python-deps
+setup: venv
 	$(MAKE) -C tools
 	$(PYTHON) tools/decompress_baserom.py $(VERSION)
 	$(PYTHON) extract_baserom.py
@@ -303,9 +308,8 @@ endif
 	$(N64_EMULATOR) $<
 
 
-.PHONY: all rom compressed clean install-python-deps setup run distclean assetclean
+.PHONY: all rom compressed clean assetclean distclean venv setup run
 .DEFAULT_GOAL := rom
-all: rom compressed
 
 #### Various Recipes ####
 
@@ -319,16 +323,6 @@ $(ROMC): $(ROM) $(ELF) $(BUILD_DIR)/compress_ranges.txt
 
 $(ELF): $(TEXTURE_FILES_OUT) $(ASSET_FILES_OUT) $(O_FILES) $(OVL_RELOC_FILES) $(BUILD_DIR)/ldscript.txt $(BUILD_DIR)/undefined_syms.txt
 	$(LD) -T $(BUILD_DIR)/undefined_syms.txt -T $(BUILD_DIR)/ldscript.txt --no-check-sections --accept-unknown-input-arch --emit-relocs -Map $(BUILD_DIR)/z64.map -o $@
-
-# Python virtual environment
-
-$(VENV):
-	python3 -m venv $(VENV)
-
-$(VENV)/.install-python-deps: requirements.txt | $(VENV)
-	$(PYTHON) -m pip install -U pip
-	$(PYTHON) -m pip install -U -r requirements.txt
-	touch $@
 
 ## Order-only prerequisites
 # These ensure e.g. the O_FILES are built before the OVL_RELOC_FILES.
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,7 @@ endif
 
 PROJECT_DIR := $(dir $(realpath $(firstword $(MAKEFILE_LIST))))
 BUILD_DIR := build/$(VERSION)
+VENV := .venv
 
 MAKE = make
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -289,6 +290,9 @@ distclean: clean assetclean
 
 setup:
 	$(MAKE) -C tools
+	test -d $(VENV) || python3 -m venv $(VENV)
+	$(PYTHON) -m pip install -U pip
+	$(PYTHON) -m pip install -r requirements.txt
```
hensldm, line 295: Thoughts on maybe doing something like this instead
```make
$(VENV): requirements.txt
	python3 -m venv $(VENV)
	$(PYTHON) -m pip install -U pip
	$(PYTHON) -m pip install -U -r requirements.txt
	touch $(VENV)

setup: $(VENV)
	$(MAKE) -C tools
	$(PYTHON) tools/decompress_baserom.py $(VERSION)
	$(PYTHON) extract_baserom.py
	$(PYTHON) extract_assets.py -j$(N_THREADS)
```
this way it keeps all the venv logic together, and would only update if requirements.txt gets updated. This also gives a way to just build/update the venv without needing a full setup again.

hensldm (commented):
fwiw, I initially didn't like the idea at first (tbh probably just a personal taste of messing with virtual environments in the Makefile feeling wrong, but not really having any concrete examples why). But thinking/researching it a bit I've started to warm up to the idea.

Makefile
cadmic: I thought directories behave badly if you use them as makefile targets, but I do like the idea of splitting out the logic into a separate target, and depending on requirements.txt. Maybe we can have the target be `.venv/pyvenv.cfg` instead, or our own sentinel file?

cadmic: Reading up on this a bit more, I think `setup: | $(VENV)` would work, I'll try it out

hensldm: Personally I'm fine with a sentinel file (would prefer making our own though). Though in that case we may want to add a phony `venv` target that depends on it, just to make setting up/updating the venv without a full setup a little easier (though maybe not necessary).

cadmic: Okay, I created 2 separate targets, one for creating the venv (which we only want to do once ever) and one for installing python deps if `requirements.txt` changes. I'm also not sure what to call these things so please suggest better names

Makefile
```diff
@@ -317,6 +320,16 @@ $(ROMC): $(ROM) $(ELF) $(BUILD_DIR)/compress_ranges.txt
 $(ELF): $(TEXTURE_FILES_OUT) $(ASSET_FILES_OUT) $(O_FILES) $(OVL_RELOC_FILES) $(BUILD_DIR)/ldscript.txt $(BUILD_DIR)/undefined_syms.txt
 	$(LD) -T $(BUILD_DIR)/undefined_syms.txt -T $(BUILD_DIR)/ldscript.txt --no-check-sections --accept-unknown-input-arch --emit-relocs -Map $(BUILD_DIR)/z64.map -o $@
 
+# Python virtual environment
+
+$(VENV):
+	python3 -m venv $(VENV)
+
+$(VENV)/.install-python-deps: requirements.txt | $(VENV)
```
hensldm, line 328: tbh I don't really see the need to separate this out, since if the venv already exists `python3 -m venv $(VENV)` does nothing so is fine with the others.

cadmic: I think `python3 -m venv $(VENV)` will reinstall the virtualenv, at least I tried editing `.venv/pyvenv.cfg` and ran `python3 -m venv $(VENV)` and it was changed back. It also took about 2 seconds to do so which is pretty slow. I guess the 2 seconds isn't so bad since it will only run when requirements.txt changes, but maybe advanced users want to set up their virtualenv differently (using a different python3 perhaps) and I don't think we should clobber that

hensldm: Hmm, it does seem to clobber `.venv/pyvenv.cfg` but seems to keep pip packages installed.
Perhaps we just go back to using `test -d $(VENV) || python3 -m venv $(VENV)` to see if it already exists. Would kind of like to keep one target as much as possible.

hensldm: Though if we do end up keeping both targets might I suggest `venv` and `venv-deps` as the target names.

Makefile
```diff
@@ -262,6 +263,8 @@ endif
 
 #### Main Targets ###
 
+all: rom compressed
```
Dragorn421, line 266:
```suggestion
all: rom compress
```
(or move this back imo it's just diff/blame noise but whatever)

cadmic: the other PR won so I merged main into this one

README.md
```diff
@@ -141,7 +117,7 @@ make setup
 This will generate a new ROM "baseroms/gc-eu-mq-dbg/baserom-decompressed.z64" that will have the overdump removed and the header patched.
```
Dragorn421, line 117: I think we should mention the venv step. maybe
```suggestion
This downloads some dependencies (from pip), and compiles tools for the build process.
Then it generates a new ROM "baseroms/gc-eu-mq-dbg/baserom-decompressed.z64" that will have the overdump removed and the header patched.
```

Makefile
cadmic: Oh, I didn't realize I messed it up. ty

Dragorn421 (approved):

fig02 (approved):
will wait for anon's approval before merging

hensldm (approved):

PR #1620, opened 2024-01-24
Rename `make compressed` to `make compress`

cadmic:
Phony make targets are usually verbs e.g. `run` or `install`. Sorry for not thinking about this during review


fig02 (approved):

Dragorn421 (approved):
just `make all` 4Head

PR #1621, opened 2024-01-24
Add dependencies for tools to requirements.txt

cadmic:
Now that many people will have a virtualenv active, it would be nice to install everything they need here

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,5 +14,5 @@ pycparser
 toml
 
 # tools
-mapfile-parser>=2.3.0,<3.0.0
+mapfile-parser>=1.2.1,<2.0.0
 rabbitizer>=1.0.0,<2.0.0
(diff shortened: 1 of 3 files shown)


fig02 (approved):

requirements.txt
```diff
@@ -1,3 +1,18 @@
...
+cxxfilt
+python-Levenshtein
+watchdog
+
+# decomp-permuter
+pycparser
+toml
+
+# tools
+mapfile-parser>=2.3.0,<3.0.0
```
Dragorn421, line 17: the ImportError handlers say `mapfile-parser>=1.2.1,<2.0.0`, is mapfile-parser 2.x.x compatible?

cadmic: That is a great question. Unfortunately I don't remember why I wrote this, I'm slowly trying to add code from my old branch. Anecdotally it seems to work with 2.x.x but I put it back to `mapfile-parser>=1.2.1,<2.0.0` for now

Dragorn421 (approved):
colorama is back! time to make flashy python scripts

PR #1622, opened 2024-01-26
Delete venv if venv creation fails

cadmic:
At least one person has run into this. Maybe there's a cleaner way to go about this

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -291,6 +291,8 @@ distclean: clean assetclean
 	$(MAKE) -C tools distclean
 
 venv:
+# Create the virtual environment if it doesn't exist.
+# Delete the virtual environment directory if creation fails.
 	test -d $(VENV) || python3 -m venv $(VENV) || { rm -rf $(VENV); false; }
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -291,7 +291,7 @@ distclean: clean assetclean
 	$(MAKE) -C tools distclean
 
 venv:
-	test -d $(VENV) || python3 -m venv $(VENV)
+	test -d $(VENV) || python3 -m venv $(VENV) || { rm -rf $(VENV); false; }
```
Dragorn421, line 294: a comment is probably warranted since not everyone cares about shell syntax yet would like to know what this does

```mk
# Create the virtual environment if it doesn't exist.
# Delete the virtual environment folder if creation fails.
	test -d $(VENV) || python3 -m venv $(VENV) || { rm -rf $(VENV); false; }
```

(note: comment should be unindented otherwise make will execute it as a command (and it'll be a shell comment that shows in stdout))

Dragorn421 (approved):

fig02 (approved):

PR #1637, opened 2024-01-28
Colliders: change "touch, bump" to AT and AC (elems)

Dragorn421:
Another major step in cleaning up colliders!

Mostly replace-all changes


include/z64collision_check.h
```diff
@@ -76,23 +76,23 @@ typedef struct {
 } ColliderInitToActor; // size = 0x08
 
 typedef struct {
-    /* 0x00 */ u32 dmgFlags; // Toucher damage type flags.
+    /* 0x00 */ u32 dmgFlags; // Damage types dealt by this collider element as AT.
     /* 0x04 */ u8 effect; // Damage Effect (Knockback, Fire, etc.)
-    /* 0x05 */ u8 damage; // Damage or Stun Timer
-} ColliderElementTouch; // size = 0x08
+    /* 0x05 */ u8 damage; // Damage
+} ColliderElementDamageInfoAT; // size = 0x08
```
fig02, line 82: Seems like a potentially good idea to remove "damage" from these field names since that information is now contained within the wrapping struct

`dmgFlags` -> `typeFlags` possibly-- both the old and new comments refer to it as a "damage type". 
(and less importantly, thats also what it was originally called: `at_type`, see en_arrow)

`damage` -> maybe `amount`-- This sort of helps in the cases where this value isnt even a damage value. But it will still be implied by its wrapper struct having "damage" in it. But either way I think it removes some redundancy.

with these changes you would have:
`this->collider.elem.atDmgInfo.typeFlags`
`this->collider.elem.atDmgInfo.amount`

Im not 100% sold on this yet, just an idea. 
I have a bone to pick with "effect" as well, but will leave that for another time.

fig02: also this change doesn't necessarily need to happen right now, I think I agree with the overall change that this PR introduces. just some more changes that can be made on top of it if we like it

Dragorn421: I like your suggestion better than what it is currently
I'll try to think of alternatives

We could also name `atDmgInfo.dmgFlags` differently from `acDmgInfo.dmgFlags` to help readability (those two only differ by one letter in ac/at)
`atDmgInfo.dmgFlags` -> `atDmgInfo.dealtTypes`
`acDmgInfo.dmgFlags` -> `acDmgInfo.sensibleTypes`
(bad names but to give an idea)

Dragorn421: I like your suggestion better than what it is currently
I'll try to think of alternatives

We could also name `atDmgInfo.dmgFlags` differently from `acDmgInfo.dmgFlags` to help readability (those two only differ by one letter in ac/at)
`atDmgInfo.dmgFlags` -> `atDmgInfo.dealtTypes`
`acDmgInfo.dmgFlags` -> `acDmgInfo.sensibleTypes`
(bad names but to give an idea)

mzxrules: `atDmgInfo.attackTypes` `acDmgInfo.weakTypes` maybe? can't really think of anything super great.

Other ideas: 

* `attackFlags` and `immuneFlags` and just explain that 0s are immune, 1s are vulnerable.
* `attackFlags` and `vulnerabilityFlags`

Dragorn421: (brainstorming)

I like "vulnerabilities", could be shortened to `vuln`

I think Types is more informative than Flags but either may work, unsure

`atDmgInfo.dealTypes`
`acDmgInfo.vulnTypes`

`attack Types/Flags` would be expanding our assumed meaning for AT to attack, then AC could expand to check
`atDmgInfo.attackTypes`
`acDmgInfo.checkTypes`

include/z64collision_check.h
```diff
@@ -349,28 +349,28 @@ typedef struct {
...
-#define BUMP_HIT (1 << 1) // Had an AC collision
-#define BUMP_HOOKABLE (1 << 2) // Can be hooked if actor has hookability flags set.
-#define BUMP_NO_AT_INFO (1 << 3) // Does not give its info to the AT collider that hit it.
-#define BUMP_NO_DAMAGE (1 << 4) // Does not take damage.
-#define BUMP_NO_SWORD_SFX (1 << 5) // Does not have a sound effect when hit by player-attached AT colliders.
-#define BUMP_NO_HITMARK (1 << 6) // Skips hit effects.
-#define BUMP_DRAW_HITMARK (1 << 7) // Draw hitmark for AC collision this frame.
+#define ATELEM_NONE 0 // No flags set. Cannot have AT collisions
+#define ATELEM_ON (1 << 0) // Can have AT collisions
+#define ATELEM_HIT (1 << 1) // Had an AT collision
```
fig02, line 354: I personally dont like smashing words together for macro/define names, and would be okay with `AT_ELEM_X` for example, but this kind of thing is already preveleant all over the codebase already. so i guess its a matter of preference

fig02 (approved):
The comment I made about the "damage" member variable names can be expanded on and implemented in the future. I think as-is the pr changes are good

include/z64collision_check.h
Dragorn421: There already are `AT_` flags (and ac/oc), I think `ATELEM_` is more readable as less easy to confuse with an `AT_` flag than `AT_ELEM_` would be

Besides this point (which weighs more to me than other considerations, since collider data is complicated) I agree at and elem should be separate

cadmic (approved):
Could you update the function names in the disassembly?

Dragorn421:
> Could you update the function names in the disassembly?

done
according to map diff I got them all

PR #1639, opened 2024-01-28
z_en_bom.c OK

cadmic:
I have a feeling that this

```
#ifdef OOT_DEBUG
    if (1) {}
#endif
```

was originally this

```
    if (OOT_DEBUG) {}
```

but maybe we should wait for more evidence before trying to change this.


Dragorn421 (approved):

fig02:
Yea agree on waiting. In general if (1)'s can be removed completely by rewriting the code in certain ways (like by introducing or removing {} blocks with returns) so in general its hard to be sure if its actually related to a debug ifdef

fig02 (approved):

PR #1641, opened 2024-01-28
is_debug.c and boot_main.c OK

cadmic:


Dragorn421 (commented):
would be good to also ifdef out the prototype for `isPrintfInit` (in `functions.h`)

src/boot/is_debug.c
```diff
@@ -92,3 +102,5 @@ NORETURN void func_80002384(const char* exp, const char* file, u32 line) {
         ;
     }
 }
+
+#endif /* OOT_DEBUG */
```
fig02, line 106: Not sure if we should be putting these comments. At least I'm not sure i like the idea of only putting it sometimes. 

I understand the rationale is that this is a longer block that is contained within the ifdef than usual. But most editors will show you what ifdef and endif belongs to if you highlight it

cadmic: Well, mine doesn't :p (VS Code with mostly default configuration). And GitHub won't show it either and I view code here a lot.

I noticed this in other parts of the codebase, and I've been putting it whenever I can't see the corresponding #if on my screen. I know screen sizes vary but I think it's a good heuristic

fig02: I dont like the inconsistency it introduces. Some more opinions needed

Dragorn421: I don't think this is useful because it doesn't tell you what the matching `#if`/`#ifdef`/`#ifndef` is. The block may be under `OOT_DEBUG` as well as !`OOT_DEBUG`

The only spots I've seen this idiom useful is in deeply nested conditional directives (headers that shouldn't exist at all with how unreadable they are) which will never be our case

cadmic: Okay, I've removed it here and in the z_collision_check PR

fig02 (approved):

src/boot/is_debug.c
```diff
@@ -35,11 +41,15 @@ void rmonPrintf(const char* fmt, ...) {
...
+#ifdef OOT_DEBUG
     _Printf(is_proutSyncPrintf, NULL, fmt, args);
+#endif
 
     va_end(args);
 }
 
+#ifdef OOT_DEBUG
+
 void* is_proutSyncPrintf(void* arg, const char* str, size_t count) {
```
Dragorn421, line 53: Why empty line
I mean why not my point is more "agree arbitrarily or standardize yes/no"

cadmic: I originally had two types of ifdefs:

* "Small" blocks that had no space between them, and had no /* OOT_DEBUG */ comment on the endif
* "Large" blocks that had spaces around them and had an /* OOT_DEBUG */ comment on the endif

But this distinction is gone now so I removed the extra lines

Dragorn421 (approved):

PR #1642, opened 2024-01-28
z_collision_check.c OK

cadmic:
These functions were removed in retail:
```
Collider_DrawRedPoly
Collider_DrawPoly
Collider_Draw
CollisionCheck_DrawCollision
```

Some of these functions were used in other places, so I had to remove those calls. The z_bgcheck.c changes are probably correct, while the z_actor.c one will have to be cleaned up by whoever does that file.

### DIFF
--- a/src/code/z_collision_check.c
+++ b/src/code/z_collision_check.c
@@ -7,6 +7,8 @@ typedef void (*ColChkApplyFunc)(PlayState*, CollisionCheckContext*, Collider*);
 typedef void (*ColChkVsFunc)(PlayState*, CollisionCheckContext*, Collider*, Collider*);
 typedef s32 (*ColChkLineFunc)(PlayState*, CollisionCheckContext*, Collider*, Vec3f*, Vec3f*);
 
+#ifdef OOT_DEBUG
+
 /**
  * Draws a red triangle with vertices vA, vB, and vC.
  */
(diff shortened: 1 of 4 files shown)


src/code/z_collision_check.c
```diff
@@ -2169,6 +2174,8 @@ void CollisionCheck_ATTrisVsACCyl(PlayState* play, CollisionCheckContext* colChk
             }
 
             if (Math3D_CylTriVsIntersect(&acCyl->dim, &atTrisElem->dim, &hitPos) == true) {
+                Vec3f atPos;
+                Vec3f acPos;
                 atPos.x = (atTrisElem->dim.vtx[0].x + atTrisElem->dim.vtx[1].x + atTrisElem->dim.vtx[2].x) * (1.0f / 3);
```
fig02, line 2179: pretty small thing, but a style pattern we have kept pretty consistent for a long time is to always put a line break after variable declarations. its relevant in a few other places in this pr

src/code/z_collision_check.c
```diff
@@ -70,6 +71,8 @@ void Collider_DrawPoly(GraphicsContext* gfxCtx, Vec3f* vA, Vec3f* vB, Vec3f* vC,
     CLOSE_DISPS(gfxCtx, "../z_collision_check.c", 757);
 }
 
+#endif
```
fig02, line 74: can remove this newline

src/code/z_collision_check.c
cadmic: Good to know for the future, thanks

fig02 (approved):

Dragorn421 (approved):

PR #1643, opened 2024-01-28
idle.c OK

cadmic:
I admit I don't really understand this. It looks like there's no reference here to `osViModeFpalLan1` in NTSC 1.0 either, so this might actually be a debug vs retail change, but maybe this will be clearer once we decomp other retail versions

### DIFF
--- a/src/boot/idle.c
+++ b/src/boot/idle.c
@@ -56,6 +56,9 @@ void Idle_ThreadEntry(void* arg) {
     gViConfigYScale = 1.0f;
 
     switch (osTvType) {
+#ifndef OOT_DEBUG
+        case OS_TV_PAL:
+#endif
         case OS_TV_NTSC:
             gViConfigModeType = OS_VI_NTSC_LAN1;
             gViConfigMode = osViModeNtscLan1;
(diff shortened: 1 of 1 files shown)


src/boot/idle.c
```diff
@@ -84,7 +89,5 @@ void Idle_ThreadEntry(void* arg) {
     osStartThread(&gMainThread);
     osSetThreadPri(NULL, OS_PRIORITY_IDLE);
 
-    while (1) {
-        ;
-    }
+    for (;;) {}
```
Dragorn421, line 92: is this stylistic or actual codegen difference

cadmic: Actual codegen difference, there's a different number of generated `nop`s. This was taken from MM

Dragorn421 (approved):
well I am puzzled as well for the VI config stuff

@Thar0 may know/be interested

anyway this is what it is so

Thar0 (approved):
I'm curious what this looks like on a PAL N64 version. I'm thinking this might just be a gamecube-emulator-aware change, and maybe NTSC 1.0 just didn't have PAL support in yet?

cadmic:
Yeah you may be right, PAL 1.0 does have a reference to `osViModeFpalLan1` unlike NTSC 1.0. There are more changes than that though, and I don't think I'm motivated enough right now to decompile it

PR #1645, opened 2024-01-28
Match ``main.c`` - OK

Yanis42:
This one is weird, ``retail_progress.py`` says file is OK but ``diff.py`` says score is 100 and not 0 like I had on the other matches for the ``Main()`` function

### DIFF
--- a/src/code/main.c
+++ b/src/code/main.c
@@ -34,7 +34,6 @@ void Main_LogSystemHeap(void) {
 }
 #endif
 
-// Note: retail_progress says ok but diff score is 100
 void Main(void* arg) {
     IrqMgrClient irqClient;
     OSMesgQueue irqMgrMsgQueue;
--- a/src/code/main.c
+++ b/src/code/main.c
@@ -25,21 +25,26 @@ AudioMgr gAudioMgr;
 OSMesgQueue sSerialEventQueue;
 OSMesg sSerialMsgBuf[1];
 
+#ifdef OOT_DEBUG
 void Main_LogSystemHeap(void) {
     PRINTF(VT_FGCOL(GREEN));
     // "System heap size% 08x (% dKB) Start address% 08x"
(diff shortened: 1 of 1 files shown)


src/code/main.c
```diff
@@ -25,21 +25,26 @@ AudioMgr gAudioMgr;
...
+#ifdef OOT_DEBUG
 void Main_LogSystemHeap(void) {
     PRINTF(VT_FGCOL(GREEN));
     // "System heap size% 08x (% dKB) Start address% 08x"
     PRINTF("システムヒープサイズ %08x(%dKB) 開始アドレス %08x\n", gSystemHeapSize, gSystemHeapSize / 1024, gSystemHeap);
     PRINTF(VT_RST);
 }
+#endif
 
+// Note: retail_progress says ok but diff score is 100
```
fig02, line 37: in general i wouldnt really too much about diff scores. you can remove the comment

src/code/main.c
```diff
@@ -25,21 +25,25 @@ AudioMgr gAudioMgr;
...
 void Main(void* arg) {
     IrqMgrClient irqClient;
     OSMesgQueue irqMgrMsgQueue;
     OSMesg irqMgrMsgBuf[60];
     uintptr_t systemHeapStart;
     uintptr_t fb;
+#ifdef OOT_DEBUG
     void* debugHeapStart;
     u32 debugHeapSize;
+#endif
```
fig02, line 46: can these temps be moved down into the relevant block about the debug heap? would remove an extra set of ifdefs

fig02: you may need to make your own block by wrapping the relevant code with {}

fig02: had to also move another temp down into a scope. but this seems way more likely to be something that was written https://github.com/zeldaret/oot/compare/main...fig02:oot:main-improved?expand=1

I can PR this to you if you want

Yanis42: feel free 👍

fig02: https://github.com/Yanis42/oot/pull/1

fig02 (approved):

cadmic (approved):

PR #1646, opened 2024-01-28
Match ``z_obj_switch.c`` - OK

Yanis42:
This was a tough one


src/overlays/actors/ovl_Obj_Switch/z_obj_switch.c
```diff
@@ -192,7 +192,10 @@ void ObjSwitch_RotateY(Vec3f* dest, Vec3f* src, s16 rotY) {
 void ObjSwitch_InitDynaPoly(ObjSwitch* this, PlayState* play, CollisionHeader* collision, s32 moveFlag) {
     s32 pad;
     CollisionHeader* colHeader = NULL;
+
+#ifdef OOT_DEBUG
     s32 pad2;
+#endif
 
     DynaPolyActor_Init(&this->dyna, moveFlag);
```
fig02, line 200: this block of code is very common among actors, including this weird pad (which could probably actually be used for something). search InitDynaPoly for examples

this could probably be one large macro but we need more research and discussion

Yanis42:
(turns out this requires more research)

PR #1647, opened 2024-01-28
Tweak OPTFLAGS for some files for retail

cadmic:
I went through MM's Makefile for files that were apparently compiled with `-O2` (not `-O2 -g3`) in retail. This made many files match completely, and I checked the rest to make sure that it looks "more matched" in some sense.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,11 +50,13 @@ endif
 
 # Version-specific settings
 ifeq ($(VERSION),gc-eu-mq)
+  DEBUG := 0
   CFLAGS += -DNON_MATCHING -DNDEBUG
   CPPFLAGS += -DNON_MATCHING -DNDEBUG
   OPTFLAGS := -O2 -g3
   COMPARE := 0
 else ifeq ($(VERSION),gc-eu-mq-dbg)
+  DEBUG := 1
   CFLAGS += -DOOT_DEBUG
   CPPFLAGS += -DOOT_DEBUG
   OPTFLAGS := -O2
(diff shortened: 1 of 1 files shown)


Thar0 (approved):

Makefile
```diff
@@ -50,11 +50,13 @@ endif
 
 # Version-specific settings
 ifeq ($(VERSION),gc-eu-mq)
+  DEBUG := 0
   CFLAGS += -DNON_MATCHING -DNDEBUG
```
fig02, line 54: confused a bit on `DEBUG` and `OOT_DEBUG`. I get one is for game code changes, but what warrants this stuff getting the "unnamed" debug define?

cadmic: well, I figured none of the other Makefile variables are prefixed with OOT (e.g. COMPARE, CFLAGS, etc)

cadmic: there's also no chance of collision here, which was the rationale for prefixing the CPP define with OOT

Dragorn421: To recap,

These are makefile variables (and not macros):
- `DEBUG`, currently 1 for gc-eu-mq-dbg and 0 for gc-eu-mq, indicates if doing a debug build. Currently only used in this PR to conditionally set compile flags on some files

These are macros (and not makefile variables):
- `NDEBUG`, specified by the C standard as "if it is defined then `assert` expands to nothing" (cf the C standard or `man assert` https://www.man7.org/linux/man-pages/man3/assert.3.html ). It is also traditionally used in C programs to if(n)def around debug code
- `OOT_DEBUG`, our own macro because we aren't forced to use `NDEBUG` at all, and the "ifndef ndebug" double negative is confusing though you get used to it (speaking from experience). Plus we probably eventually will gate individual debug features behind individual defines for documentation (and modding while we're at it) purposes

All three are currently directly derived from the VERSION value, there's no hierarchy between them

Dragorn421 (approved):
I think the newly introduced DEBUG makefile variable should also be used to define or not the NDEBUG and OOT_DEBUG macros but it can be looked into at a later time

fig02 (approved):

PR #1652, opened 2024-01-29
Check formatting in CI

cadmic:
`check_format.sh` is lifted directly from MM. Hopefully this will prevent random unrelated style changes in PRs. The downside is that this adds 21s to every build, and requires all contributors to set up clang-format 14 (which may not be trivial).

Example output:

```
[Pipeline] echo
Checking formatting...
[Pipeline] sh
+ tools/check_format.sh
Formatting files with 4 jobs
Running clang-format...
Running clang-tidy...
Adding missing final new lines...
Done formatting files.

Misformatted files found. Run ./format.py and verify codegen is not impacted.

--- Old git status
+++ New git status
@@ -0,0 +1 @@
+ M src/code/main.c

Exiting.
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -65,11 +65,9 @@ pipeline {
         failure {
             sh 'cat tools/check_format.txt'
         }
-        always {
+        cleanup {
             echo "Finished, deleting directory."
             deleteDir()
-        }
-        cleanup {
             echo "Clean up in post."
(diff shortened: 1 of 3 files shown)


cadmic:
Okay, I think this is working now

hensldm (approved):

Dragorn421 (approved):
TODO: only format files touched by PR to save time and power

fig02 (approved):

PR #1653, opened 2024-01-29
Extract baserom segments for gc-eu-mq

cadmic:
`decompress_baserom.py` now writes individual ROM segments too, replacing some duplicated logic in `extract_baserom.py`. As a bonus this extraction is about 1s faster now too.

Since hardcoding the file list for every version in the script seemed unwieldy, I moved all the data to config files. This is a little awkward for the file table offset, but I think it's pretty nice that this script does not need to be changed if we add a new version.


baseroms/gc-eu-mq-dbg/file-table-offset.txt
```diff
@@ -0,0 +1 @@
+12F70
```
hensldm, line 1: I would say we should just add this to the segments.txt file at the top instead of having two separate txt files. I would also say prefix the number with `0x` just to be clear it is a hex number.

Dragorn421 (commented):
segments.txt may be better named dmadata_names.txt or something maybe, in OoT one segment (in the spec) = one dmadata entry (in the dmadata table in the rom) but apparently MM has NOLOAD segments that aren't put into the rom

---------

Note the `Found valid baserom - exiting early` path makes segment extraction also be skipped, which doesn't sound like a good idea ?

-------

\+ can you black-format

--------

and ftr I'm not a fan of "whole bunch of txts" for configs but it will do

baseroms/gc-eu-mq-dbg/file-table-offset.txt
cadmic: It's nice that each line of segments.txt corresponds to a ROM file and I think it would be weird to have an exception to that. We could use a more structured format like json or yaml but I think that's a bit too complicated right now (but maybe it'll be more worth it in the future).

(I added `0x`, Dragorn had the same comment)

hensldm: I don't know. The exception seems worth it just so we don't need to have two different config files (especially when 1 is a single line). Perhaps this is just something we will disagree on.

cadmic:
Could you make separate threads in the future?

> segments.txt may be better named dmadata_names.txt or something maybe, in OoT one segment (in the spec) = one dmadata entry (in the dmadata table in the rom) but apparently MM has NOLOAD segments that aren't put into the rom

Hmm, the directory is already called `segments/` though the same concern applies to that. Each line of `segments.txt` corresponds to one file in `segments/` so it's nice that those two are consistent

> Note the Found valid baserom - exiting early path makes segment extraction also be skipped, which doesn't sound like a good idea ?

Yeah, I changed this so that segments are written before the decompressed baserom so we can recover if anything goes wrong

> + can you black-format

Done

> and ftr I'm not a fan of "whole bunch of txts" for configs but it will do

Yeah, I think we'll need other per-version config (like build date) too, so I could see this evolving into something better

baseroms/gc-eu-mq-dbg/file-table-offset.txt
cadmic: Yeah, I think two config files whose meanings are clear is better than one config file with a strange format

baseroms/gc-eu-mq-dbg/segments.txt
Dragorn421, line 1: just realized this essentially duplicates `include/tables/dmadata_table_mqdbg.h` but yeah whatever, for now at least

tools/decompress_baserom.py
```diff
@@ -235,6 +251,12 @@ def main():
     file_content = per_version_fixes(file_content, version)
 
     dmadata = read_dmadata(file_content, file_table_offset)
+    if len(segment_names) != len(dmadata):
+        print(
+            f"Error: segment names and dmadata length mismatch ({len(segment_names)} != {len(dmadata)})"
```
hensldm, line 256:
```suggestion
            f"Error: segment names and dmadata length mismatch ({len(segment_names)} != {len(dmadata)})", file=sys.stderr
```
Along with including sys
For these error messages we should probably be writing to stderr.

Dragorn421 (commented):
I don't see how

> Note the Found valid baserom - exiting early path makes segment extraction also be skipped, which doesn't sound like a good idea ?

was addressed

cadmic:
> Note the Found valid baserom - exiting early path makes segment extraction also be skipped, which doesn't sound like a good idea ?

Sorry, what scenario are you concerned about? I thought it was if anything goes wrong then we could potentially write the baserom but not the segments, leading to confusion the next time `make setup` was run.

cadmic:
Oh, I just realized that `make distclean` deleted the segments dir but not the decompressed baserom, maybe that's what you were concerned about. I changed it to delete both

Dragorn421:
There is a check early in the script for a valid baserom-uncompressed.z64 to stop execution then, regardless of `baseroms/version/segments/...` state

Historically we didn't distclean the baserom-uncompressed.z64, personally I've used that to delete baserom.z64, as well as symlinking baserom-uncompressed.z64, saving 100MiB per oot repo clone (now to multiply by n_versions)

hensldm:
Perhaps I'm missing something, but does this actually work for originally compressed roms? We first read the dmadata from the compressed rom, but when we write the segments we use the data of the decompressed rom without reading the new dmadata table of the decompressed rom.

cadmic:
> Perhaps I'm missing something, but does this actually work for originally compressed roms? We first read the dmadata from the compressed rom, but when we write the segments we use the data of the decompressed rom without reading the new dmadata table of the decompressed rom.

Great catch, thank you. It works properly now, I compared some extracted asset segments with the debug rom

cadmic:
> Historically we didn't distclean the baserom-uncompressed.z64, personally I've used that to delete baserom.z64, as well as symlinking baserom-uncompressed.z64, saving 100MiB per oot repo clone (now to multiply by n_versions)

Thanks for explaining. What do you think the behavior should be?

baseroms/gc-eu-mq-dbg/file-table-offset.txt
fig02: I dont like 2 files for this, but if the plan is to use json/yaml/something else soon then I dont really care for now

Dragorn421:
I would prefer the baseroms not be distcleaned, the whole point of extract_baserom (or uncompress_baserom now) being to ensure the baseroms are OK, once they are they can just stay there. And if they're not ok the script catches that

I guess uncompress_baserom could check if the segments/ folder exists along with checking the uncompressed-baserom hash for "early exit". Then even if segments/ is messed up at least distclean which rms segments/ will have it be regen'd

Also I guess like we did for venv / like Make does for target files, we should have the script rm segments/ if writing anything to it goes wrong. (`try: writesegments; except: rmsegments; raise`)

tools/decompress_baserom.py
```diff
@@ -32,31 +33,20 @@ def decompress(data: bytes, is_zlib_compressed: bool) -> bytes:
...
-}
-
-
 def round_up(n, shift):
     mod = 1 << shift
     return (n + mod - 1) >> shift << shift
 
 
-def as_word_list(b) -> list[int]:
+def as_word_list(b) -> List[int]:
```
AngheloAlf, line 41: Why use `List` when we already can use plain `list`?

tools/decompress_baserom.py
```diff
@@ -122,6 +112,20 @@ def decompress_rom(
...
+    dmadata: List[List[int]],
+    segment_names: List[str],
+    segments_dir: Path,
+):
+    segments_dir.mkdir(parents=True, exist_ok=True)
+    for segment_name, (v_start, v_end, p_start, p_end) in zip(segment_names, dmadata):
+        if p_start == 0xFFFFFFFF and p_end == 0xFFFFFFFF:
+            continue
+        with open(segments_dir / segment_name, "wb") as f:
+            f.write(file_content[v_start:v_end])
```
AngheloAlf, line 126:
```suggestion
        (segments_dir / segment_name, "wb").write_bytes(file_content[v_start:v_end])
```

AngheloAlf:
You claimed that the script is much faster now, could you provide some numbers to know how much faster it is now? It is kinda weird that the new implementation is faster given that the old one uses multiprocessing while this one does not

tools/decompress_baserom.py
Dragorn421: Compatibility with python 3.8

AngheloAlf: That's why we have the `from __future__ import annotations` at the top of this file

Dragorn421:
For me it's <1s and I don't run it every minute so I kinda think performance of this script doesn't matter

Wouldn't be surprised if the previous script is slower because of multiprocessing, which needs to copy the data between processes (slow and can easily outweigh the gains)

hensldm:
> I would prefer the baseroms not be distcleaned, the whole point of extract_baserom (or uncompress_baserom now) being to ensure the baseroms are OK, once they are they can just stay there. And if they're not ok the script catches that
> 
> I guess uncompress_baserom could check if the segments/ folder exists along with checking the uncompressed-baserom hash for "early exit". Then even if segments/ is messed up at least distclean which rms segments/ will have it be regen'd
> 
> Also I guess like we did for venv / like Make does for target files, we should have the script rm segments/ if writing anything to it goes wrong. (`try: writesegments; except: rmsegments; raise`)

Could possibly just always have the segments written out by adding it before the early return like
```python
    if check_existing_rom(uncompressed_path, correct_str_hash):
        print("Found valid decompressed baserom -- proceeding to writing segments")

        file_content = bytearray(uncompressed_path.read_bytes())

        dmadata = read_dmadata(file_content, file_table_offset)
        if len(segment_names) != len(dmadata):
            print(
                f"Error: segment names and dmadata length mismatch ({len(segment_names)} != {len(dmadata)})",
                file=sys.stderr,
            )
            exit(1)

        # Write ROM segments
        segments_dir = baserom_dir / "segments"
        print(f"Writing ROM segments to {segments_dir}...")
        write_segments(file_content, dmadata, segment_names, segments_dir)

        return
```

Does duplicate a little code, but seems fine to me.

cadmic:
> You claimed that the script is much faster now, could you provide some numbers to know how much faster it is now? It is kinda weird that the new implementation is faster given that the old one uses multiprocessing while this one does not

"Much faster" is probably just wrong, sorry about that. On my machine it's about 1.4s faster.

Should I scrap this and change extract_baserom.py instead?

cadmic:
I'm also thinking about putting this off a little longer to think about the other concerns that people have brought up:
* Better config format (what other version-specific config do we need? e.g. build date, incbin offsets, maybe even BSS ordering info)
* Refactor some logic into a small library to read this config + dmadata, for various scripts to use
* Generating things like include/tables/dmadata_table_mqdbg.h from this config

Dragorn421:
The two steps are very similar in that they use dmadata and deal with segments. The way it is implemented in this PR currently isn't much different from two different scripts but it could be refactored to avoid redundant code

tools/decompress_baserom.py
cadmic: Oh, good to know. Maybe I'll revert all the other typing changes we made to support fig's ancient python version...

PR #1654, opened 2024-01-29
Move __osPfsInodeCache and __osPfsPifRam

cadmic:
In retail, `__osPfsInodeCache` is present (currently in `pfsreadwritefile.c` BSS) but none of the functions in `pfsreadwritefile.c` are present. This is strange because it seems like all of the other pfs* files are either entirely present or entirely removed. ~~I think it makes sense to put `__osPfsInodeCache` in `pfsgetstatus.c` instead, because `pfsgetstatus.c` already has `__osPfsPifRam` which is also used in many pfs-related files.~~ (edit: see @hensldm's comment below on where to put these)

### DIFF
--- a/src/libultra/io/contpfs.c
+++ b/src/libultra/io/contpfs.c
@@ -4,6 +4,8 @@
 s32 __osPfsInodeCacheChannel = -1;
 u8 __osPfsInodeCacheBank = 250;
 
+__OSInode __osPfsInodeCache;
+
 u16 __osSumcalc(u8* ptr, s32 length) {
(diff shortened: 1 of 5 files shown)


hensldm:
Hmm based on ultralib, `__osPfsInodeCache` should be in `contpfs.c` and `__osPfsPifRam` in `pfsisplug.c`.

cadmic:
Thanks, that makes sense too

Dragorn421 (approved):

fig02 (approved):

PR #1655, opened 2024-01-29
z_demo retail code matched

fig02:
I named a flag here because I was really curious how it was related to a debug feature. Turns out the devs didnt want to watch the long cutscene with sheik after you get the master sword :)


src/code/z_demo.c
```diff
@@ -2220,6 +2244,14 @@ void CutsceneHandler_RunScript(PlayState* play, CutsceneContext* csCtx) {
...
     }
 }
+#else
+void CutsceneHandler_RunScript(PlayState* play, CutsceneContext* csCtx) {
+    if (gSaveContext.save.cutsceneIndex >= 0xFFF0) {
+        csCtx->curFrame++;
+        Cutscene_ProcessScript(play, csCtx, play->csCtx.script);
+    }
+}
+#endif
```
fig02, line 2254: I made an entirely new copy of this function for retail, cause ifdeffing inside the function got messy really fast. If anyone has other ideas tho lmk

AngheloAlf: Personally I think it would be better to put the ifdef inside the function, but at the very top
```c
void CutsceneHandler_RunScript(PlayState* play, CutsceneContext* csCtx) {
#ifdef OOT_DEBUG
    // ...
#else
    // ...
#endif
}
```
This is nicer for the IDE in stuff like doing `ctrl-click` on the name of the function so it can lead you to the source, otherwise it would show you a list of possible functions, which is kinda annoying

fig02: Possibly.
An alternative to that I guess is two separate ifdefs, one for the debug print info, and another for the processscript. Still messy and bad imo.
More opinions needed
```c
void CutsceneHandler_RunScript(PlayState* play, CutsceneContext* csCtx) {
#ifdef OOT_DEBUG
    if (0) {} // Necessary to match

    if (gSaveContext.save.cutsceneIndex >= 0xFFF0) {
        if (0) {} // Also necessary to match

        if (BREG(0) != 0) {
            Gfx* displayList;
            Gfx* prevDisplayList;

            OPEN_DISPS(play->state.gfxCtx, "../z_demo.c", 4101);

            prevDisplayList = POLY_OPA_DISP;
            displayList = Graph_GfxPlusOne(POLY_OPA_DISP);
            gSPDisplayList(OVERLAY_DISP++, displayList);
            Cutscene_DrawDebugInfo(play, &displayList, csCtx);
            gSPEndDisplayList(displayList++);
            Graph_BranchDlist(prevDisplayList, displayList);
            POLY_OPA_DISP = displayList;

            CLOSE_DISPS(play->state.gfxCtx, "../z_demo.c", 4108);
        }
#endif
        csCtx->curFrame++;

#ifdef OOT_DEBUG
        if (R_USE_DEBUG_CUTSCENE) {
            Cutscene_ProcessScript(play, csCtx, gDebugCutsceneScript);
        } else {
            Cutscene_ProcessScript(play, csCtx, play->csCtx.script);
        }
#else
        Cutscene_ProcessScript(play, csCtx, play->csCtx.script);
#endif
    }
}
```

fig02: also the required if 0's hint more toward cadmics idea of `if (OOT_DEBUG)`

Dragorn421: > also the required if 0's hint more toward cadmics idea of `if (OOT_DEBUG)`

Could you match this function with that idea and removing the if0s

fig02: yes

of note, just using `OOT_DEBUG_DEF` in the condition at the end is enough to remove the if(0) required in debug. I thought it wouldnt, because typically this means more { } blocks are needed. but seemingly not in this case? or I just dont understand ido's optimiation, which is likely 
```c
#define OOT_DEBUG_DEF 0

void CutsceneHandler_RunScript(PlayState* play, CutsceneContext* csCtx) {
    if (gSaveContext.save.cutsceneIndex >= 0xFFF0) {
        if (OOT_DEBUG_DEF) {
            if (BREG(0) != 0) {
                Gfx* displayList;
                Gfx* prevDisplayList;

                OPEN_DISPS(play->state.gfxCtx, "../z_demo.c", 4101);

                prevDisplayList = POLY_OPA_DISP;
                displayList = Graph_GfxPlusOne(POLY_OPA_DISP);
                gSPDisplayList(OVERLAY_DISP++, displayList);
                Cutscene_DrawDebugInfo(play, &displayList, csCtx);
                gSPEndDisplayList(displayList++);
                Graph_BranchDlist(prevDisplayList, displayList);
                POLY_OPA_DISP = displayList;

                CLOSE_DISPS(play->state.gfxCtx, "../z_demo.c", 4108);
            }
        }
        csCtx->curFrame++;

        if (OOT_DEBUG_DEF && R_USE_DEBUG_CUTSCENE) {
            Cutscene_ProcessScript(play, csCtx, gDebugCutsceneScript);
        } else {
            Cutscene_ProcessScript(play, csCtx, play->csCtx.script);
        }
    }
}
```

src/code/z_demo.c
```diff
@@ -1779,10 +1801,13 @@ void Cutscene_ProcessScript(PlayState* play, CutsceneContext* csCtx, u8* script)
         return;
     }
 
+#ifdef OOT_DEBUG
+    // With debug features enabled, pressing D-pad Right on controller 1 will stop the cutscene early and skip it
```
Dragorn421, line 1805: At first I thought "with debug features enabled" refered to an additional check by the game like a previous combo or something, but I guess you mean "if debug build"
I think being inside a `#ifdef OOT_DEBUG` is enough for documenting it's a debug feature.
Or if you really want to hammer it in just using `// Debug: description...` would be enough

------

I wouldn't use "skip" here because it's more "interrupt the cutscene" right? rather than force running the destination which looks more like what you'd expect from a skip

------

Also a style thing I prefer writing debug features docs in the imperative, like:

```
// Press D-pad Right on controller 2 to stop playing the current cutscene.
```

Since it's for the programmer reading the source. It's also usually more concise.
But may be just-me

src/code/z_demo.c
fig02: you can actually move the debug define into the first conditional with && too

include/z64save.h
```diff
@@ -438,7 +438,7 @@ typedef enum {
 #define EVENTCHKINF_4C 0x4C
 #define EVENTCHKINF_4D 0x4D
 #define EVENTCHKINF_4E 0x4E
-#define EVENTCHKINF_4F 0x4F
+#define EVENTCHKINF_WATCHED_SHEIK_AFTER_MASTER_SWORD_CS 0x4F
```
Dragorn421, line 441: Could you add a comment here to clarify which cutscene
Maybe

```suggestion
#define EVENTCHKINF_WATCHED_SHEIK_AFTER_MASTER_SWORD_CS 0x4F // Cutscene in Temple of Time as adult after the first time pulling the Master Sword
```

src/code/z_demo.c
fig02: I wasnt too convinced with this theory yet, but this is a pretty huge example. any other way of writing this with ifdefs just sucks :D

include/z64save.h
fig02: I can, but im not sure what this can be confused with. there is only one cutscene with sheik after the master sword

src/code/z_demo.c
Dragorn421: Yeah
My main worry now is warnings spam about unreachable code
Idk about ido but I imagine gcc would complain

include/z64save.h
Dragorn421: Yeah if you know all the cutscenes in the game "sheik after master sword" is enough, but if you don't more context doesn't hurt

src/code/z_demo.c
Dragorn421: Gcc *currently* does not warn for unreachable code. It used to be a feature, it was removed 10+ years ago because it was bugged
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=46476

But ideally a compiler should warn about dead code. e.g. if someone ever gets clang working (and its mips support gets less bad), clang does currently warn about dead code

https://stackoverflow.com/questions/45598080/clang-gcc-disable-unused-code-warning-as-a-result-of-conditional-macro
a comment says (some versions of?) clang are smart enough to not warn about unreachable code in these cases (i.e. recognize the macro may expand to not-0)

Note this post also links to
https://stackoverflow.com/questions/1644868/define-macro-for-debug-printing-in-c
which strongly advocates for `if(debug)` over `#ifdef debug`

But overall I'm not finding much on the internet about this which leads me to believe the `#ifdef` way is how traditional C handles this (it's what I've typically encountered personally, whereas I've seen C++ work with constexpr in these cases)


So currently my take would be "use #ifdef unless if() is really really worth it" (like in this case where it avoids code duplication, and if0s as a bonus)

src/code/z_demo.c
fig02: Ill omit the comments for now, maybe in the future we can standardize and document the various debug features in the game

fig02:
review has been addressed and the better match with OOT_DEBUG is used

Dragorn421 (approved):

src/code/z_demo.c
```diff
@@ -561,9 +575,7 @@ void CutsceneCmd_Destination(PlayState* play, CutsceneContext* csCtx, CsCmdDesti
         titleDemoSkipped = true;
     }
 
-    if ((csCtx->curFrame == cmd->startFrame) || titleDemoSkipped ||
-        ((csCtx->curFrame > 20) && CHECK_BTN_ALL(play->state.input[0].press.button, BTN_START) &&
-         (gSaveContext.fileNum != 0xFEDC))) {
+    if ((csCtx->curFrame == cmd->startFrame) || titleDemoSkipped || DEBUG_SKIP_CS) {
```
cadmic, line 578: Why not inline `OOT_DEBUG && ...`, too unwieldy? I checked that it matches, and I don't think it looks terrible when formatted:

```
if ((csCtx->curFrame == cmd->startFrame) || titleDemoSkipped ||
        // In debug mode, cutscenes can be skipped by pressing the start button (or whatever this does)
        (OOT_DEBUG && ((csCtx->curFrame > 20) && CHECK_BTN_ALL(play->state.input[0].press.button, BTN_START) &&
                       (gSaveContext.fileNum != 0xFEDC)))) {
```

cadmic (approved):

src/code/z_demo.c
fig02: I never tried it, cause the define didnt exist at the time. I kinda like the complicated debug define being separate in this case, cause of how important the condition is. But I also dont like the condition being defined so far away from its usage...

Not a huge fan of comments in the middle of conditions. its not something we really do at all so far. it may end up looking okay in this case but I can also see it going really badly depending on the length of the different lines.

for now ill go with using the define, without a comment (I already removed them for now per another review above)

PR #1656, opened 2024-01-29
Match stackcheck.c and code_800FC620.c

cadmic:
Is it worth making a debug macro for __osMalloc? This is the only place it would be used


fig02 (approved):
seems fine to omit a macro in this case

hensldm (approved):

fig02:
this failed cause of isghj-3. can either just force merge or poke it for a different agent

cadmic:
Looks like it re-ran after the previous merge into main

PR #1658, opened 2024-01-29
Define OOT_DEBUG=0 in retail builds

cadmic:
Define OOT_DEBUG=0 and change all `#ifdef OOT_DEBUG` into `#if OOT_DEBUG`. This is so we can write `if (OOT_DEBUG)` or `if (OOT_DEBUG && ...)`, which in many places cleans up the code significantly, especially in files with many debug features. Example: https://github.com/zeldaret/oot/pull/1655#discussion_r1469769614

I don't think all ifs can be written this way, for example in z_parameter.c it does not match the debug version if we write
```
    if (OOT_DEBUG && pauseCtx->debugState == 3) {
        FlagSet_Update(play);
    }
```
instead of
```
#ifdef OOT_DEBUG
    if (pauseCtx->debugState == 3) {
        FlagSet_Update(play);
    }
#endif
```

Apologies in advance for all the conflicts.

### DIFF
--- a/src/code/game.c
+++ b/src/code/game.c
@@ -1,6 +1,12 @@
 #include "global.h"
 #include "terminal.h"
 
+#if OOT_DEBUG
+#define VI_MODE_EDITOR_INACTIVE (R_VI_MODE_EDIT_STATE == VI_MODE_EDIT_STATE_INACTIVE)
+#else
+#define VI_MODE_EDITOR_INACTIVE true
+#endif
+
 SpeedMeter D_801664D0;
 VisCvg sVisCvg;
 VisZBuf sVisZBuf;
@@ -436,7 +442,7 @@ void GameState_Init(GameState* gameState, GameStateFunc init, GraphicsContext* g
     VisCvg_Init(&sVisCvg);
     VisZBuf_Init(&sVisZBuf);
     VisMono_Init(&sVisMono);
-    if (!OOT_DEBUG || (R_VI_MODE_EDIT_STATE == VI_MODE_EDIT_STATE_INACTIVE)) {
+    if (VI_MODE_EDITOR_INACTIVE) {
         ViMode_Init(&sViMode);
     }
     SpeedMeter_Init(&D_801664D0);
@@ -467,7 +473,7 @@ void GameState_Destroy(GameState* gameState) {
     VisCvg_Destroy(&sVisCvg);
     VisZBuf_Destroy(&sVisZBuf);
     VisMono_Destroy(&sVisMono);
-    if (!OOT_DEBUG || (R_VI_MODE_EDIT_STATE == VI_MODE_EDIT_STATE_INACTIVE)) {
+    if (VI_MODE_EDITOR_INACTIVE) {
         ViMode_Destroy(&sViMode);
     }
     THA_Destroy(&gameState->tha);
--- a/src/code/game.c
+++ b/src/code/game.c
@@ -1,12 +1,6 @@
 #include "global.h"
 #include "terminal.h"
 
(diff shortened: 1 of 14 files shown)


src/overlays/actors/ovl_En_Bom/z_en_bom.c
```diff
@@ -359,9 +359,7 @@ void EnBom_Draw(Actor* thisx, PlayState* play) {
     s32 pad;
     EnBom* this = (EnBom*)thisx;
 
-#ifdef OOT_DEBUG
-    if (1) {}
-#endif
+    if (OOT_DEBUG) {}
```
fig02, line 362: I think we should only use the define when we are sure its being used for debug features. We can and do remove if (0)'s all the time through other means, like restructuring how if conditionals are written + early returns. So we should leave the empty ones unless you can find a way to make the debug define usage meaningful

fig02: actually I missed that in this case the if (1) is ifdeffed out in retail. this one makes sense. but my point still stands for other cases :D

cadmic: Yeah I agree in general. I'm not sure if there's a better way to match this case

Dragorn421:
Merge #1657 first

fig02 (approved):

Dragorn421 (commented):
Missed at least one `ifndef` (idle.c)

cadmic:
I fixed the #ifndef and merged game.c

src/overlays/actors/ovl_En_Bom/z_en_bom.c
Dragorn421, line 362: this doesn't match for me

src/code/game.c
```diff
@@ -442,7 +436,7 @@ void GameState_Init(GameState* gameState, GameStateFunc init, GraphicsContext* g
     VisCvg_Init(&sVisCvg);
     VisZBuf_Init(&sVisZBuf);
     VisMono_Init(&sVisMono);
-    if (VI_MODE_EDITOR_INACTIVE) {
+    if ((R_VI_MODE_EDIT_STATE == VI_MODE_EDIT_STATE_INACTIVE) || !OOT_DEBUG) {
```
Dragorn421, line 439: personally would prefer `OOT_DEBUG` checks to come first (here and below)
```suggestion
    if (!OOT_DEBUG || (R_VI_MODE_EDIT_STATE == VI_MODE_EDIT_STATE_INACTIVE)) {
```

src/overlays/actors/ovl_En_Bom/z_en_bom.c
cadmic: Uhh, it does for me...

src/code/game.c
cadmic: I agree, but putting `!OOT_DEBUG` first did not match debug :/

src/overlays/actors/ovl_En_Bom/z_en_bom.c
cadmic: You're right, I swear I had tested this so I'm not sure what I did wrong

src/code/game.c
Dragorn421: oh it matches in `GameState_Init` but not `GameState_Destroy` :pensive: 

it would match in `GameState_Destroy` with just one more use of `gameState` for the compiler to put it in s0...
anyway whatever for this PR

Dragorn421 (approved):

PR #1659, opened 2024-01-29
Match ``z_select.c``

Yanis42:
I added the ``GameState`` temp outside the macro since debug version of ``z_play.c`` isn't matching with it

### DIFF
--- a/src/overlays/gamestates/ovl_select/z_select.c
+++ b/src/overlays/gamestates/ovl_select/z_select.c
@@ -756,6 +756,7 @@ void MapSelect_Init(GameState* thisx) {
 
     {
         u32 size = (uintptr_t)_z_select_staticSegmentRomEnd - (uintptr_t)_z_select_staticSegmentRomStart;
+
         this->staticSegment = GAME_STATE_ALLOC(&this->state, size, "../z_select.c", 1114);
         DMA_REQUEST_SYNC(this->staticSegment, (uintptr_t)_z_select_staticSegmentRomStart, size, "../z_select.c", 1115);
     }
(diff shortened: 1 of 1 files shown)


src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -739,17 +746,20 @@ void MapSelect_Init(GameState* thisx) {
...
         this->pageDownIndex = dREG(82);
     }
+
     R_UPDATE_RATE = 1;
 
-    this->staticSegment = GAME_STATE_ALLOC(&this->state, size, "../z_select.c", 1114);
-    DMA_REQUEST_SYNC(this->staticSegment, (uintptr_t)_z_select_staticSegmentRomStart, size, "../z_select.c", 1115);
+    {
+        u32 size = (uintptr_t)_z_select_staticSegmentRomEnd - (uintptr_t)_z_select_staticSegmentRomStart;
+        this->staticSegment = GAME_STATE_ALLOC(&this->state, size, "../z_select.c", 1114);
```
fig02, line 759: whenever a new stack variable is declared, we prefer to include a new line before the rest of the code. so you can add a break after the size line

src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -11,7 +11,11 @@
 
 void MapSelect_LoadTitle(MapSelectState* this) {
     this->state.running = false;
-    SET_NEXT_GAMESTATE(&this->state, ConsoleLogo_Init, ConsoleLogoState);
+    {
+        GameState* state = &this->state;
+
+        SET_NEXT_GAMESTATE(state, ConsoleLogo_Init, ConsoleLogoState);
+    }
```
cadmic, line 18: Perhaps add some TODOs here and below to update this when SET_NEXT_GAMESTATE is fixed?

cadmic (approved):

src/overlays/gamestates/ovl_select/z_select.c
Yanis42: When this will be added to the macro this file won't match anymore so I thought it wasn't needed but I can add one if everyone's fine with it

cadmic: Yeah that's true. I don't mind leaving it as-is

fig02 (approved):

PR #1662, opened 2024-01-30
Overlay_Relocate and Overlay_Load retail OK

hensldm:
No particularly happy with the fake match with `relocSectionOffset` and `gOverlayLogSeverity` I had to do in `Overlay_Load` but couldn't find anything better. `relocSectionOffset = 0` is close, but has compiler managed stack issues.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -247,6 +247,7 @@ $(BUILD_DIR)/src/code/code_800FD970.o: OPTFLAGS := -O2
 $(BUILD_DIR)/src/code/gfxprint.o: OPTFLAGS := -O2
 $(BUILD_DIR)/src/code/jpegutils.o: OPTFLAGS := -O2
 $(BUILD_DIR)/src/code/jpegdecoder.o: OPTFLAGS := -O2
+$(BUILD_DIR)/src/code/load.o: OPTFLAGS := -O2
 $(BUILD_DIR)/src/code/loadfragment2.o: OPTFLAGS := -O2
 $(BUILD_DIR)/src/code/logutils.o: OPTFLAGS := -O2
 $(BUILD_DIR)/src/code/mtxuty-cvt.o: OPTFLAGS := -O2
--- a/include/ultra64/libc.h
+++ b/include/ultra64/libc.h
@@ -5,7 +5,7 @@
 
 void osSyncPrintf(const char* fmt, ...);
 
-void bzero(void* __s, size_t __n);
+void bzero(void* __s, int __n);
 int bcmp(const void* __sl, const void* __s2, size_t __n);
 void bcopy(const void* __src, void* __dest, size_t __n);
 
--- a/include/ultra64/libc.h
+++ b/include/ultra64/libc.h
@@ -6,7 +6,7 @@
 void osSyncPrintf(const char* fmt, ...);
 
 void bzero(void* __s, int __n);
-int bcmp(const void* __sl, const void* __s2, size_t __n);
-void bcopy(const void* __src, void* __dest, size_t __n);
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -247,6 +247,7 @@ $(BUILD_DIR)/src/code/code_800FD970.o: OPTFLAGS := -O2
 $(BUILD_DIR)/src/code/gfxprint.o: OPTFLAGS := -O2
 $(BUILD_DIR)/src/code/jpegutils.o: OPTFLAGS := -O2
 $(BUILD_DIR)/src/code/jpegdecoder.o: OPTFLAGS := -O2
+$(BUILD_DIR)/src/code/load.o: OPTFLAGS := -O2
```
cadmic, line 250: (Sorry for missing this in the earlier PR, hopefully it wasn't too much trouble)

include/ultra64/libc.h
```diff
@@ -5,7 +5,7 @@
 
 void osSyncPrintf(const char* fmt, ...);
 
-void bzero(void* __s, size_t __n);
+void bzero(void* __s, int __n);
 int bcmp(const void* __sl, const void* __s2, size_t __n);
 void bcopy(const void* __src, void* __dest, size_t __n);
```
cadmic, line 10: Shouldn't all these be `int` instead of only `bzero`?

src/code/load.c
```diff
@@ -1,13 +1,13 @@
...
+size_t Overlay_Load(uintptr_t vromStart, uintptr_t vromEnd, void* vramStart, void* vramEnd, void* allocatedRamAddr) {
     s32 pad[3];
     uintptr_t end;
     OverlayRelocationSection* ovlRelocs;
     u32 relocSectionOffset;
-    size_t size;
+    s32 size = vromEnd - vromStart;
 
-    size = vromEnd - vromStart;
+    relocSectionOffset = gOverlayLogSeverity;
```
cadmic, line 10: Maybe add a comment for confused readers?

include/ultra64/libc.h
hensldm: Yes, since it didn't directly apply to what I was doing I didn't touch them. However I'll just fix them now.

src/code/relocation.c
```diff
@@ -105,12 +100,13 @@ void Overlay_Relocate(void* allocatedRamAddress, OverlayRelocationSection* ovlRe
...
                 // Extract the address from the target field of the J-type MIPS instruction.
                 // Relocate the address and update the instruction.
-
-                unrelocatedAddress = PHYS_TO_K0(MIPS_JUMP_TARGET(*relocDataP));
-                relocOffset = unrelocatedAddress - (uintptr_t)vramStart;
-                relocatedValue = (*relocDataP & 0xFC000000) | (((allocu32 + relocOffset) & 0x0FFFFFFF) >> 2);
-                relocatedAddress = PHYS_TO_K0(MIPS_JUMP_TARGET(relocatedValue));
-                *relocDataP = relocatedValue;
+                if (1) {
+                    relocOffset = PHYS_TO_K0(MIPS_JUMP_TARGET(*relocDataP)) - (uintptr_t)vramStart;
```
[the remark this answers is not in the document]
hensldm: In MM jp1.0 (which is an older revision that kept the debugging statements with the actual relocation logic) there is an unused string that really only makes sense if this case had the exact same structure like the other ones i.e.
```c
if (condition) {
    relocation logic goes here
} else if (gOverlaySeverity >= 3) {
    PRINT(debug string)
}

cadmic (approved):

Dragorn421 (approved):

PR #1665, opened 2024-01-30
Match retail z_std_dma.c

cadmic:
Notably, calls to `DmaMgr_Error` were replaced with `Fault_AddHungupAndCrash`, which takes a file and line number.

### DIFF
--- a/src/boot/z_std_dma.c
+++ b/src/boot/z_std_dma.c
@@ -137,27 +137,27 @@ s32 DmaMgr_DmaRomToRam(uintptr_t rom, void* ram, size_t size) {
 
     if (1) { // Also necessary to match
         s32 pad[2];
+    }
 
-        ioMsg.hdr.pri = OS_MESG_PRI_NORMAL;
-        ioMsg.hdr.retQueue = &queue;
-        ioMsg.devAddr = rom;
-        ioMsg.dramAddr = ram;
-        ioMsg.size = size;
+    ioMsg.hdr.pri = OS_MESG_PRI_NORMAL;
+    ioMsg.hdr.retQueue = &queue;
+    ioMsg.devAddr = rom;
+    ioMsg.dramAddr = ram;
+    ioMsg.size = size;
 
-        if (gDmaMgrVerbose == 10) {
-            PRINTF("%10lld ノーマルＤＭＡ %08x %08x %08x (%d)\n", OS_CYCLES_TO_USEC(osGetTime()), ioMsg.dramAddr,
-                   ioMsg.devAddr, ioMsg.size, MQ_GET_COUNT(&gPiMgrCmdQueue));
-        }
+    if (gDmaMgrVerbose == 10) {
+        PRINTF("%10lld ノーマルＤＭＡ %08x %08x %08x (%d)\n", OS_CYCLES_TO_USEC(osGetTime()), ioMsg.dramAddr,
+               ioMsg.devAddr, ioMsg.size, MQ_GET_COUNT(&gPiMgrCmdQueue));
+    }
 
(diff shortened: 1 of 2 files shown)


src/boot/z_std_dma.c
```diff
@@ -229,7 +235,7 @@ void DmaMgr_DmaFromDriveRom(void* ram, uintptr_t rom, size_t size) {
  *
  * This function does not return.
  */
-NORETURN void DmaMgr_Error(DmaRequest* req, const char* file, const char* errorName, const char* errorDesc) {
+NORETURN void DmaMgr_Error(DmaRequest* req, const char* filename, const char* errorName, const char* errorDesc) {
     uintptr_t vrom = req->vromAddr;
```
fig02, line 239: we use `file` way more than `filename`, at least currently. We can discuss fixing the inconsistency but it should prob be cleaned up in one pass
fwiw file is prob used more often for symmetry with the `__FILE__` directive

cadmic: ty

fig02 (approved):

src/boot/z_std_dma.c
```diff
@@ -132,27 +135,29 @@ s32 DmaMgr_DmaRomToRam(uintptr_t rom, void* ram, size_t size) {
         ram = (u8*)ram + buffSize;
     }
 
-    if (1) {} // Also necessary to match
+    if (1) { // Also necessary to match
+        s32 pad[2];
```
Dragorn421, line 140:
```suggestion
    if (1) { // Also necessary to match
        s32 pad[2];
    }
```
would prefer to not indent the whole thing if we can avoid it

src/boot/z_std_dma.c
```diff
@@ -460,11 +481,15 @@ s32 DmaMgr_RequestAsync(DmaRequest* req, void* ram, uintptr_t vrom, size_t size,
...
-    if ((1 && (ram == NULL)) || (osMemSize < OS_K0_TO_PHYSICAL(ram) + size) || (vrom & 1) || (vrom > 0x4000000) ||
+#if OOT_DEBUG
+    if ((ram == NULL) || (osMemSize < OS_K0_TO_PHYSICAL(ram) + size) || (vrom & 1) || (vrom > 0x4000000) ||
         (size == 0) || (size & 1)) {
-        //! @bug `req` is passed to `DmaMgr_Error` without rom, ram and size being set
-        DmaMgr_Error(req, NULL, "ILLIGAL DMA-FUNCTION CALL", "パラメータ異常です");
+        // The line numbers for `DMA_ERROR` are only used in retail builds, but this usage was removed so
+        // its line number is unknown.
+        //! @bug `req` is passed to `DMA_ERROR` without rom, ram and size being set
+        DMA_ERROR(req, NULL, "ILLIGAL DMA-FUNCTION CALL", "パラメータ異常です", "../z_std_dma.c", 0);
```
Dragorn421, line 490: why not just use `DmaMgr_Error` here

Dragorn421 (commented):
I'm confused by `sDmaMgrIsRomCompressed` being set only in debug code, when it's also used in retail code (and it's also initialized to false in retail)

maybe it's a bug (doesn't really matter either way, towards the end of `DmaMgr_ProcessRequest` either the rom hangs up properly or most likely misbehaves and crashes)

src/boot/z_std_dma.c
cadmic: I thought that the padding was some optimized-out local variables for the following code, and maybe we could figure out what they were later, but I don't mind the unindent

src/boot/z_std_dma.c
cadmic: I feel that is less consistent with the rest of the code

Dragorn421 (approved):

PR #1666, opened 2024-01-30
Revert bad uses of `if (OOT_DEBUG)`

cadmic:
My apologies, I should have known when Dragorn caught the other error that I just wasn't building properly or something.


fig02 (approved):

Dragorn421 (approved):
Happy little okcidents

PR #1667, opened 2024-01-30
Reorder boot/code spec includes for retail

cadmic:
Unfortunately this makes things a bit of a mess. I'm not sure if the original devs' tooling did this automatically.

For functions that are no longer linked, I chose not to `#if` the declarations. The function definitions themselves are not `#if`'d out, so a linker error seems more appropriate if you use one of these functions without updating the spec.

I made some temporary changes to camera.c to avoid referencing unlinked functions. These will have to be cleaned up for real when we decomp those files.

Moved from `boot` to `code`:
```
ldiv
sprintf
string
vimodefpallan1
xldtob
xlitob
xprintf
```
Moved around within `code`:
```
contpfs
pfsgetstatus
pfsisplug
```
No longer linked:
```
assert
db_camera
debug_malloc
epiwrite
mempak
mtxl2f
mtxuty-cvt
pfsallocatefile
pfschecker
pfsdeletefile
pfsfilestate
pfsfindfile
pfsfreeblocks
pfsinitpak
pfsreadwritefile
ucode_disas
z_text
```

### DIFF
--- a/src/code/z_camera.c
+++ b/src/code/z_camera.c
@@ -7281,7 +7281,6 @@ void Camera_PrintSettings(Camera* camera) {
         sp48[i] = '\0';
 
         sp48[camera->play->activeCamId] = 'a';
-#if OOT_DEBUG
         DebugCamera_ScreenTextColored(3, 22, DEBUG_CAM_TEXT_WHITE, sp58);
         DebugCamera_ScreenTextColored(3, 22, DEBUG_CAM_TEXT_PEACH, sp48);
         DebugCamera_ScreenTextColored(3, 23, DEBUG_CAM_TEXT_WHITE, "S:");
@@ -7292,7 +7291,6 @@ void Camera_PrintSettings(Camera* camera) {
         DebugCamera_ScreenTextColored(
             5, 25, DEBUG_CAM_TEXT_GOLD,
             sCameraFunctionNames[sCameraSettings[camera->setting].cameraModes[camera->mode].funcIdx]);
-#endif
 
         i = 0;
         if (camera->bgCamIndex < 0) {
@@ -7316,10 +7314,8 @@ void Camera_PrintSettings(Camera* camera) {
         sp50[i++] = ' ';
         sp50[i++] = ' ';
         sp50[i] = '\0';
-#if OOT_DEBUG
         DebugCamera_ScreenTextColored(3, 26, DEBUG_CAM_TEXT_WHITE, "I:");
         DebugCamera_ScreenTextColored(5, 26, DEBUG_CAM_TEXT_GOLD, sp50);
-#endif
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/code/z_camera.c
```diff
@@ -7277,6 +7281,7 @@ void Camera_PrintSettings(Camera* camera) {
         sp48[i] = '\0';
 
         sp48[camera->play->activeCamId] = 'a';
+#if OOT_DEBUG
         DebugCamera_ScreenTextColored(3, 22, DEBUG_CAM_TEXT_WHITE, sp58);
```
Dragorn421, line 7285: surprisingly, `DebugCamera_ScreenTextColored` is in `z_debug.c`, and double surprisingly `z_debug.o` is linked into the retail rom, so you don't need to ifdef it out

Dragorn421 (commented):
> a linker error seems more appropriate if you use one of these functions without updating the spec.

agreed

------

I have additional changes locally but we'll see later

cadmic:
> I have additional changes locally but we'll see later

Nice, thanks for verifying. Could you push up your branch so we can incorporate the changes here?

src/code/z_camera.c
cadmic: Huh. Removed these ifdefs

cadmic:
After diffing dragorn's branch, I missed removing z_text, so I've done that now. I think this is good to go now though

Dragorn421 (approved):

fig02 (approved):

MrCheeze:
Is this expected after this change?

```bash
$ make VERSION=gc-eu-mq
mips-linux-gnu-ld -T build/gc-eu-mq/oot-gc-eu-mq.ld -T build/gc-eu-mq/undefined_syms.txt --no-check-sections --accept-unknown-input-arch --emit-relocs -Map build/gc-eu-mq/oot-gc-eu-mq.map -o build/gc-eu-mq/oot-gc-eu-mq.elf
mips-linux-gnu-ld: build/gc-eu-mq/src/code/z_moji.o: in function `Moji_DrawChar':
src/code/z_moji.c:62: undefined reference to `gMojiFontTLUTs'
mips-linux-gnu-ld: src/code/z_moji.c:62: undefined reference to `gMojiFontTLUTs'
mips-linux-gnu-ld: build/gc-eu-mq/src/code/z_moji.o: in function `Moji_DrawString':
src/code/z_moji.c:88: undefined reference to `gMojiFontTex'
mips-linux-gnu-ld: src/code/z_moji.c:88: undefined reference to `gMojiFontTex'
make: *** [Makefile:398: build/gc-eu-mq/oot-gc-eu-mq.elf] Error 1
```

Dragorn421:
No, z_moji.o should also be left out for retail, woops

PR #1668, opened 2024-01-30
Fix some retail `boot` data

cadmic:
I know we're not officially matching data yet, this is extra credit.

I'm really unsure about the `stackcheck.c`'s unreferenced `"(null)"` string literal, I can't think of a way to make it seem natural. It could be in another file, but this file is the only place where `"(null)"` appears in debug.


fig02:
feel like we should probably wait till the system supports it, especially so it can be verified that its correct?
maybe at that point we'll have a better idea for what to do with the weird case as well

cadmic:
The data sections are disassembled, so you can see the string in expected/build/gc-eu-mq/src/boot/stackcheck.s, and we can verify with objdump:
```
$ mips-linux-gnu-objdump -s -j .rodata expected/build/gc-eu-mq/src/boot/stackcheck.o

expected/build/gc-eu-mq/src/boot/stackcheck.o:     file format elf32-tradbigmips

Contents of section .rodata:
 0000 286e756c 6c290000 00000000 00000000  (null)..........

$ mips-linux-gnu-objdump -s -j .rodata build/gc-eu-mq/src/boot/stackcheck.o

build/gc-eu-mq/src/boot/stackcheck.o:     file format elf32-tradbigmips

Contents of section .rodata:
 0000 286e756c 6c290000 00000000 00000000  (null)..........
```

In fact inspecting rodata like this was useful when decomping retail z_std_dma, and that's actually why I was reminded of this case.

I agree this is low priority but I don't think we necessarily need to wait for anything. These are the only data issues besides BSS ordering in `boot`. So maybe cleaning up data is easier than we though, and it can be done as we go rather than having an official "phase" for it

cadmic:
Revisiting this now that https://github.com/zeldaret/oot/pull/1706 gives us a way to verify this change

After figuring out that `LOG` macro evaluated its argument, I have a feeling the `PRINTF` macro originally did too. At least, that's the only explanation I can think of for the `"(null)"` here. I don't think we have a good way to evaluate the arguments in our fake variadic macro though so maybe this is the best we can do

fig02 (approved):

hensldm (approved):

PR #1669, opened 2024-01-30
Fixup vscode settings

Dragorn421:
- add missing include
- explicitly define OOT_DEBUG to 1
- change configuration name to `N64 oot-gc-eu-mq-dbg` (maybe we have several eventually for the different versions idk. better than "Linux" anyway)

### DIFF
--- a/docs/vscode.md
+++ b/docs/vscode.md
@@ -51,11 +51,12 @@ You can create a `.vscode/c_cpp_properties.json` file with `C/C++: Edit Configur
                 "${workspaceFolder}/**",
                 "src",
                 "build/gc-eu-mq-dbg",
-                "include"
+                "include",
+                "include/libc"
             ],
             "defines": [
                 "_LANGUAGE_C", // For gbi.h
-                "OOT_DEBUG" // If targeting a debug version
+                "OOT_DEBUG=1" // If targeting a debug version
             ],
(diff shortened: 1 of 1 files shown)


docs/vscode.md
```diff
@@ -41,7 +41,7 @@ You can create a `.vscode/c_cpp_properties.json` file with `C/C++: Edit Configur
 {
     "configurations": [
         {
-            "name": "Linux",
+            "name": "OoT64 gc-eu-mq-dbg",
             "compilerPath": "${default}", // Needs to not be "" for -m32 to work
```
fig02, line 45: no need for OoT64 imo, but meh

fig02 (approved):

docs/vscode.md
Dragorn421: yeah idk
personally I locally have another configuration I switch to for native tools (like mkdmadata and co), which I'd recommend to anyone, so it's more to separate "cross compile for N64" from "compile native code"
I can change it to "N64 gc-eu-mq-dbg", "N64 oot-gc-eu-mq-dbg" ?

fig02: sure either of those are fine with me. i guess i slightly prefer without "oot" but doesnt matter

hensldm (approved):

PR #1671, opened 2024-01-31
`.clang-format`: break short enums, TypenameMacros

Dragorn421:
Needs discussion


.clang-format
```diff
@@ -7,6 +7,7 @@ BreakBeforeBraces: Attach
 SpaceAfterCStyleCast: false
 Cpp11BracedListStyle: false
 IndentCaseLabels: true
+IndentPPDirectives: AfterHash
```
Dragorn421, line 10: can also be BeforeHash

hensldm: I like it :) . Though prefer `BeforeHash` myself.

src/code/ucode_disas.c
```diff
@@ -69,7 +69,12 @@ F3dzexFlag sUCodeDisasMtxFlags[] = {
     F3DZEX_FLAG(G_MTX_PUSH, G_MTX_NOPUSH),
 };
 
-typedef enum { COMBINER_A = 1, COMBINER_B, COMBINER_C, COMBINER_D } CombinerArg;
+typedef enum {
+    COMBINER_A = 1,
+    COMBINER_B,
+    COMBINER_C,
+    COMBINER_D
+} CombinerArg;
```
fig02, line 77: I guess following our current style these technically should have offset comments lol. but I dont find them that useful for cases like this personally

.clang-format
fig02: if we do this I would want before hash

fig02 (commented):
this is probably a good idea

src/code/ucode_disas.c
Dragorn421: in many cases the offset comments don't really have any use and we have them more to force the one-value-per-line formatting than for usefulness

fig02: very true. maybe we can start removing them eventually 😄 
theres also ofc struct comments that are going to be out of date between versions... whole other conversation though

fig02 (approved):

.clang-format
Thar0: I'm an AfterHash enjoyer 😨

I'm not really sure if this is worthwhile, at least currently, considering most nested preprocessor directives are in headers and we don't format headers (yet?)

fig02: this came up because of a case in another pr: https://github.com/zeldaret/oot/pull/1670#discussion_r1472156367
This will only become more common as we are working on retail

Dragorn421: we have a bunch of nested directives coming in along the lines of `#if debug #define STMT mydbgstmt #else #define STMT false #endif`

fig02: I guess this is going to be all preference at the end of the day, but for me I am always used to seeing, for example, `#define` grouped together, seeing the word "define" far away from the hash makes it read not as a directive for me anymore

src/boot/z_std_dma.c
```diff
@@ -296,9 +296,9 @@ const char* DmaMgr_FindFileName(uintptr_t vrom) {
     }
     //! @bug Since there is no return, in case the file isn't found, the return value will be a pointer to the end
     // of gDmaDataTable
-#ifdef AVOID_UB
+    #ifdef AVOID_UB
     return "";
-#endif
+    #endif
```
fig02, line 301: this one in particular is not ideal id say, but the benefits still outweigh some rarer cases like this i think

Dragorn421: I agree it kinda looks weird but still more readable than #endif into an unrelated #else

hensldm (approved):

fig02:
should prob wait for a few more than 2 approvals on this

cadmic:
Personally I'm slightly negative on indenting directives, for reasons others have brought up:

* It looks very weird to me that `#` is not the first character on lines with directives
* It is inconsistent with the existing style in headers
* There are already a few examples here where the formatting seems weird, so I'm not sure it would be rare

I think the fundamental reason why this feels weird to me is that normal code is not indented inside `#if/#endif`, so if there's a mix of normal code and directives it looks inconsistent if only the directives are indented.

I'm fine with the enum change though.

fig02:
The headers can be changed to match the preferred style, since they dont get formatted. That shouldn't really be a point of conversation here, cause if this is decided to be the style, then they can be updated

Dragorn421:
fwiw with `IndentPPDirectives: AfterHash` https://github.com/zeldaret/oot/compare/main...Dragorn421:oot:fmtconfig_AfterHash

Dragorn421:
but yeah may not be ideal in any case but is it worse than no indentation?

cadmic:
No indentation is how normal code looks between `#if/#endif` though, and while maybe that's not ideal (C preprocessor in general is not ideal, really) I do think it's better to be consistent

AngheloAlf (approved):
Looks a lot nicer. I will look at using this for other projects.
We may want to make one formater run on headers to keep consistency

Dragorn421:
ftr here's a "format headers" experiment
https://github.com/Dragorn421/oot/compare/fmtconfig...fmt_headers
note the .clang-format changes
```sh
./format.py -j30 `find include src -name *.h`
```

cadmic:
IMO indenting directives looks great, until headers try to define structs etc (e.g. the _LANGUAGE_C in include/ultra64/message.h, or the structs in gbi.h) where I think it's too jarring for me to tolerate.

Other than that I would love for the headers to be autoformatted, I'm tired of aligning `\` by hand when making macros

AngheloAlf:
> IMO indenting directives looks great, until headers try to define structs etc (e.g. the _LANGUAGE_C in include/ultra64/message.h, or the structs in gbi.h) where I think it's too jarring for me to tolerate.

I wouldn't worry too much about formatting libultra headers, those will be a subrepo eventually anyways.
Have you seen anything on a header that looks bad that is not libultra?

> Other than that I would love for the headers to be autoformatted, I'm tired of aligning `\` by hand when making macros
Alternatively you can run the format script on a single file, so you don't have to manually align stuff

cadmic:
> I wouldn't worry too much about formatting libultra headers, those will be a subrepo eventually anyways.
Have you seen anything on a header that looks bad that is not libultra?

I didn't look thoroughly, but I don't think any look bad, I don't think we've tried to mix code and CPP directives inside `#if/#endif` in headers yet.

Maybe we could match the style of `assert.h`? i.e. AfterHash and 1 space indent. I think that would look better when CPP directives and code (where lines start with a space and have 4 space indent)

fig02:
maybe we can take out the directive change for now, and get this in-- I think the other changes are good and not controversial

Dragorn421:
yeeted `IndentPPDirectives` changes

cadmic (approved):

PR #1672, opened 2024-01-31
Remove double #endif in PRINTF macro definition

cadmic:


hensldm (approved):

Dragorn421 (approved):
I think it was fine but sure

PR #1673, opened 2024-01-31
Only check formatting on modified files

Dragorn421:


Dragorn421:
Reafy

fig02 (approved):

format.py
```diff
@@ -175,10 +185,18 @@ def main():
...
-        files = args.files
+    if args.files or args.files_list:
+        files = []
+        files.extend(args.files)
+        if args.files_list:
+            files.extend(Path(args.files_list).read_text().split())
+        files = list(files)
         extra_files = []
+        if args.verbose:
+            print("Formatting specific files:", len(files), files)
```
cadmic, line 196: Nit: I think something like `f"Formatting {len(files)} files: ' '.join(files)"` would be more friendly, but I doubt many people will actually read this

cadmic (approved):

format.py
Dragorn421: yeah this is meant for debugging when this implementation explodes, I don't really care how it looks

hensldm (approved):
Neat. Useful for MM as well :) .

PR #1675, opened 2024-01-31
Gamealloc, Graph, and Graphalloc retail OK

hensldm:
Note had to split the last 3 functions from graph into a graphalloc.


src/code/graphalloc.c
```diff
@@ -0,0 +1,25 @@
...
+
+Gfx* Graph_GfxPlusOne(Gfx* gfx) {
+    return gfx + 1;
+}
+
+Gfx* Graph_BranchDlist(Gfx* gfx, Gfx* dst) {
+    gSPBranchList(gfx, dst);
+    return dst;
+}
+
```
fig02, line 11: ooo interesting that this is its own file. thinking about it, theres nothing about this that needs to be tied to "Graph" or graph thread, is there? I guess its gfx related and "graph" refers to graphics...

Basically im wondering if a different system name makes sense

fig02: I see mm already has this file, is "graphalloc" made up or known from somewhere?

src/code/gamealloc.c
```diff
@@ -12,6 +12,7 @@ void GameAlloc_Log(GameAlloc* this) {
     }
 }
 
+#if OOT_DEBUG
```
cadmic, line 15: Could you ifdef-out or delete the declarations in `functions.h` for the ones that are disabled here?

src/code/graph.c
```diff
@@ -377,7 +402,10 @@ void Graph_Update(GraphicsContext* gfxCtx, GameState* gameState) {
 
     {
         OSTime timeNow = osGetTime();
-        s32 pad[4];
+        s32 pad;
+#if OOT_DEBUG
+        s32 pad2[3];
```
cadmic, line 407: Could these padding variables go somewhere in the `#if OOT_DEBUG` just below instead of here?

src/code/gamealloc.c
cadmic: Probably can delete `func_800C55D0` there too, that seems to be a missed rename of `GameAlloc_Log`

src/code/graph.c
hensldm: Yes, they actually are probably part of the SET_NEXT_GAMESTATE macro, so I did what was done in z_select.c

src/code/graphalloc.c
hensldm: My guess is made up and probably came from the fact these functions were in OoT's graph, but I'm not sure.
I wouldn't be opposed to a new system name, though I have ideas to rename the functions anyways.

include/functions.h
```diff
@@ -1318,16 +1318,14 @@ GameStateFunc GameState_GetInit(GameState* gameState);
 u32 GameState_IsRunning(GameState* gameState);
 #if OOT_DEBUG
 void* GameState_Alloc(GameState* gameState, size_t size, char* file, s32 line);
-#endif
-void func_800C55D0(GameAlloc* this);
+
```
cadmic, line 1321: nit: remove 3 extra newlines (seems the only place one is used here is between `boot` and `code`)

cadmic (approved):

src/code/graphalloc.c
fig02: did you meant to say you dont have any ideas?
I dont really either rn tbh, but something to consider at a later time

hensldm: Ah, looks like AF has these functions in `gfxalloc.c`

fig02: yeahhh I was thinking it should be gfx instead of graph at least. closer tied to Gfx than the graph thread itself.
Would love to see that change but doesnt have to be in this pr

cadmic: IMO it would be nice the change the filename now at least, but not necessarily the function names

hensldm: Went ahead and just made the changes now.

fig02 (approved):

fig02:
will wait a bit to merge incase anyone wants to comment on the new function names

src/code/graphalloc.c
AngheloAlf, line 1: I think there's an impostor

(duplicated of gfxalloc.c)

PR #1677, opened 2024-01-31
Retail z_scene matching

fig02:
The match for `Scene_CommandObjectList` is pretty cursed. it needs a s32 return type, but that also means changing it for every other scene command also, which does still match.

I think its feasible that the return on all of these functions was just wrong originally. but if anyone wants to try to find a better match: https://decomp.me/scratch/mLyVX

### DIFF
--- a/src/code/z_scene.c
+++ b/src/code/z_scene.c
@@ -499,8 +499,7 @@ void Scene_SetTransitionForNextEntrance(PlayState* play) {
     play->transitionType = ENTRANCE_INFO_START_TRANS_TYPE(gEntranceTable[entranceIndex].field);
 }
 
-BAD_RETURN(s32)
-(*gSceneCmdHandlers[SCENE_CMD_ID_MAX])(PlayState*, SceneCmd*) = {
+SceneCmdHandlerFunc gSceneCmdHandlers[SCENE_CMD_ID_MAX] = {
     Scene_CommandPlayerEntryList,          // SCENE_CMD_ID_SPAWN_LIST
     Scene_CommandActorEntryList,           // SCENE_CMD_ID_ACTOR_LIST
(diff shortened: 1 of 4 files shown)


src/code/z_scene.c
```diff
@@ -503,7 +499,8 @@ void Scene_SetTransitionForNextEntrance(PlayState* play) {
     play->transitionType = ENTRANCE_INFO_START_TRANS_TYPE(gEntranceTable[entranceIndex].field);
 }
 
-void (*gSceneCmdHandlers[SCENE_CMD_ID_MAX])(PlayState*, SceneCmd*) = {
+BAD_RETURN(s32)
+(*gSceneCmdHandlers[SCENE_CMD_ID_MAX])(PlayState*, SceneCmd*) = {
     Scene_CommandPlayerEntryList,          // SCENE_CMD_ID_SPAWN_LIST
```
fig02, line 504: yikes. the only way around this I see is a typedef to use for the array, which tbh might not be a bad idea regardless of the formatting concern

hensldm (approved):

src/code/z_scene.c
AngheloAlf: I think this should be a typedef even without the `BAD_RETURN`, function pointers are annoying to read.
This made me realize we don't have a typedef for MM when we should

cadmic (approved):

fig02:
alright, i made the typedef

src/code/z_scene.c
Dragorn421: clang-format has a setting for this unexpectedly
https://github.com/zeldaret/oot/pull/1671/commits/817eb770211deec3d3ddafb727aa64a147fc3f18

still yeah typedef +1

PR #1678, opened 2024-01-31
Irqmgr and Sched Retail OK

hensldm:


src/code/sched.c
```diff
@@ -52,27 +52,37 @@ OSTime sRSPAudioTimeStart;
...
 
+#if OOT_DEBUG
+#define SCHED_PRINTF \
+    if (sLogScheduler)  \
+    PRINTF
+#elif defined(__sgi)
+// See comment about PRINTF in macros.h
+#define SCHED_PRINTF(args) (void)0
+#else
+#define SCHED_PRINTF(format, ...) (void)0
```
cadmic, line 63: Maybe should be called `SCHED_DEBUG_PRINTF` like the actor one (and `sLogScheduler` -> `sSchedDebugPrintfEnabled` or something)

cadmic (approved):

src/code/sched.c
fig02: agree on both

fig02 (approved):

PR #1680, opened 2024-01-31
Check format: only check modified files, v2

Dragorn421:
#1673 made check_format try to format deleted files.
This is take 2, now featuring using python instead of bash


Dragorn421:
reafy (take two)

cadmic (approved):

fig02 (approved):

PR #1681, opened 2024-01-31
z_view.c gc-eu-mq OK

Thar0:


include/macros.h
```diff
@@ -102,7 +102,7 @@
 
 #if OOT_DEBUG
 #define PRINTF osSyncPrintf
-#elif defined(__sgi) /* IDO compiler */
+#elif defined(__sgi) && !defined(__GNUC__) && !defined(PERMUTER) && !defined(M2CTX) /* IDO compiler */
```
cadmic, line 105: Are all of these checks necessary or could it be simplified?

We (will) have at least 3 printf-style macros now, maybe we could package this up somehow?

include/regs.h
```diff
@@ -277,6 +277,12 @@ typedef enum {
 #define R_PLAY_UCODE_DISAS_LOG_LEVEL HREG(92) // log level to use when switching to ucode disas
 
 // HREG_MODE_PERSPECTIVE
+#if OOT_DEBUG
+#define PERSPECTIVE_EDITOR_ACTIVE (R_HREG_MODE == HREG_MODE_PERSPECTIVE)
+#else
+#define PERSPECTIVE_EDITOR_ACTIVE false
+#endif
```
cadmic, line 284: I think we've started to prefer inlining `OOT_DEBUG && R_HREG_MODE == HREG_MODE_PERSPECTIVE` to this now (with the assumption that this will later be replaced by a macro

include/macros.h
Thar0: The issue I was having is some tools define `__sgi` even when using a modern preprocessor which rightly errors on the IDO version of `PRINTF`. I can think of two ways to simplify/package it up:
- `defined(__sgi) && (__STDC_VERSION__ < 199901L)` detects if modern cpp is pretending to be IDO, but it's not so obvious why it's needed
- I believe `#define USE_IDO_PRINTF defined(__sgi) && !defined(__GNUC__) && !defined(PERMUTER) && !defined(M2CTX)` would work if we intend to use the same check for selection other macros in future

cadmic: I guess I was implicitly asking if you wanted to package it up in this PR. It's totally fine if not, one of the places we would use it (#1678) has not yet been merged

Thar0: I'm happy to implement something in this PR, but I'd like to hear what the preferred solution is (either of the options I mentioned or an alternative)

cadmic: Oh sorry. I do think `__STDC_VERSION__` is pretty weird, and the second method is probably better.

Does it work as `__sgi && !__GNUC__ && !PERMUTER && !M2CTX`? It's shorter but I'm not sure if it's confusing

Thar0: That does work, I don't think it's especially confusing so I've gone with that for now

cadmic (approved):

src/code/z_view.c
```diff
@@ -511,7 +523,10 @@ s32 View_ApplyPerspectiveToOverlay(View* view) {
         view->eye.z += 1.0f;
     }
 
+#if OOT_DEBUG
     View_ErrorCheckEyePosition(view->eye.x, view->eye.y, view->eye.z);
+#endif
```
fig02, line 528: with how often it comes up in the file, I kinda think `VIEW_ERROR_CHECK_EYE_POS(view)` is worth it to remove a bunch of #if 's. It can be a macro local to the file

fig02: oh i didnt realize there was an external use. neat

fig02 (approved):

tools/m2ctx.py
```diff
@@ -27,7 +27,7 @@ def get_c_file(directory):
 
 def import_c_file(in_file):
     in_file = os.path.relpath(in_file, root_dir)
-    cpp_command = ["gcc", "-E", "-P", "-Iinclude", "-Isrc", "-undef", "-D__sgi", "-D_LANGUAGE_C",
+    cpp_command = ["gcc", "-E", "-P", "-Iinclude", "-Isrc", "-undef", "-DM2CTX", "-D__sgi", "-D_LANGUAGE_C",
```
hensldm, line 30: If we are going to be touching m2ctx.py, perhaps we should just update it fully? (https://github.com/ethteck/m2ctx/)

fig02: + subrepo wouldnt be a bad idea

Thar0: I'm not really very interested in doing more in this PR, I changed just enough to get it to do what I wanted. A follow-up that properly updates it would be nice though if you're interested :eyes:

hensldm (approved):

PR #1683, opened 2024-02-01
Match retail system_malloc.c and z_malloc.c

cadmic:
I reordered the declarations in functions.h because the debug functions are interleaved with the non-debug ones so I thought it was too messy to disable them individually.


hensldm (approved):

src/code/system_malloc.c
```diff
@@ -21,53 +23,66 @@ void SystemArena_CheckPointer(void* ptr, u32 size, const char* name, const char*
     }
 }
 
+#define CHECK_POINTER(ptr, size, name, action) SystemArena_CheckPointer(ptr, size, name, action)
+#else
+#define CHECK_POINTER(ptr, size, name, action) (void)0
+#endif
```
fig02, line 29: I know these are defined locally, but I think I would still like if they had a system identifier, like `SYSTEM_ARENA_` and `ZELDA_ARENA_`. It just helps me as Im reading the code to know its doing something locally instead of some generic "check pointer" macro

fig02 (approved):

PR #1686, opened 2024-02-01
[Faster CI] Make CI not run CC_CHECK

Dragorn421:
CC_CHECK is a `gcc -fsyntax-only` call on the ido-compiled .c files to get better warnings/errors than ido's

it's very useful locally but mostly wastes time during CI builds, and a lot of it it turns out.

example times, from a done build:
`rm -rf build/gc-eu-mq-dbg/src/ && time make -j32` : 36s
`rm -rf build/gc-eu-mq-dbg/src/ && time make -j32 CC_CHECK='@#'` : 25s
-> 10s for cc_check over src/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -35,7 +35,8 @@ pipeline {
                 branch 'main'
             }
             steps {
-                sh 'make -j ORIG_COMPILER=1'
+                // Do not run CC_CHECK to save time. @ makes Make not print the command and # starts a shell comment.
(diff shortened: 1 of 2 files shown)


cadmic:
Is there any scenario where the CC_CHECK will fail but IDO will not?

Dragorn421:
> Is there any scenario where the CC_CHECK will fail but IDO will not?

Never seen it before, it wouldn't really make sense.
Still CC_CHECK will run on the main branch (like emulated ido does instead of recomp ido) for super extra safety

PR #1692, opened 2024-02-02
Match retail Title and Sample gamestates

fig02:
only gamestate left is file select which engineer has covered.

Not happy with yet another BAD_RETURN, but I think its reasonable to think that `View_Apply` could have returned some kind of status at some point in time? idk. MM doesnt run into this issue cause that game gets rid of the sample gamestate.
The permuter found this one and any other match was extremely fake.

### DIFF
--- a/src/code/z_view.c
+++ b/src/code/z_view.c
@@ -275,13 +275,13 @@ s32 View_StepDistortion(View* view, Mtx* projectionMtx) {
 /**
  * Apply view to POLY_OPA_DISP, POLY_XLU_DISP (and OVERLAY_DISP if ortho)
  */
-BAD_RETURN(s32) View_Apply(View* view, s32 mask) {
+s32 View_Apply(View* view, s32 mask) {
     mask = (view->flags & mask) | (mask >> 4);
 
     if (mask & VIEW_PROJECTION_ORTHO) {
-        View_ApplyOrtho(view);
+        return View_ApplyOrtho(view);
     } else {
-        View_ApplyPerspective(view);
+        return View_ApplyPerspective(view);
     }
 }
 
--- a/src/code/z_view.c
+++ b/src/code/z_view.c
@@ -275,7 +275,7 @@ s32 View_StepDistortion(View* view, Mtx* projectionMtx) {
 /**
  * Apply view to POLY_OPA_DISP, POLY_XLU_DISP (and OVERLAY_DISP if ortho)
  */
-void View_Apply(View* view, s32 mask) {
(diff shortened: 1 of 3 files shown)


src/overlays/gamestates/ovl_title/z_title.c
```diff
@@ -130,7 +132,6 @@ void ConsoleLogo_Main(GameState* thisx) {
 
     if (OOT_DEBUG && gIsCtrlr2Valid) {
         Gfx* gfx = POLY_OPA_DISP;
-        s32 pad;
 
         ConsoleLogo_PrintBuildInfo(&gfx);
```
cadmic, line 136: It's a little weird that this function doesn't exist in retail, yet it is called here and expected to be optimized out. Maybe this block makes more sense as `#if OOT_DEBUG` rather than `if (OOT_DEBUG && ...)`?

fig02: maybe. I didnt put that OOT_DEBUG there, i think you did when you added the define :D
but yeah i can change it. I think as-is it introduces a warning anyway cause there is no function declaration to reference

AngheloAlf: I was going to ask if cc_check were complaining about it.
Probably better to do a `#if OOT_DEBUG` instead

cadmic: oh, whoops

cadmic (approved):

src/code/z_view.c
```diff
@@ -275,7 +275,7 @@ s32 View_StepDistortion(View* view, Mtx* projectionMtx) {
 /**
  * Apply view to POLY_OPA_DISP, POLY_XLU_DISP (and OVERLAY_DISP if ortho)
  */
-void View_Apply(View* view, s32 mask) {
+BAD_RETURN(s32) View_Apply(View* view, s32 mask) {
```
AngheloAlf, line 278: Just checked, this matches both debug and retail:
```c
s32 View_Apply(View* view, s32 mask) {
    mask = (view->flags & mask) | (mask >> 4);

    if (mask & VIEW_PROJECTION_ORTHO) {
        return View_ApplyOrtho(view);
    } else {
        return View_ApplyPerspective(view);
    }
}
```

fig02: ooo those functions return things, ok thats good

fig02:
alright BAD_RETURN is gone

AngheloAlf (approved):

PR #1693, opened 2024-02-02
Match retail padmgr.c and sys_rumble.c

cadmic:
This includes the line 
```
LOG_NUM("++errcnt", ++sRumbleErrorCount, "../padmgr.c", 282);
```
where the `++sRumbleErrorCount` is preserved in retail. To handle this I changed the `LOG` macro to evaluate the value even in retail. This has the bonus of fixing `LOG_STRING` too, for example z_elf_message.c has the line
```
LOG_STRING("企画外 条件", "../z_elf_message.c", 156); // "Unplanned conditions"
```
where the string `"企画外 条件"` also exists in retail .rodata. I checked all the `LOG_STRING`s and it's the same for all of them.

### DIFF
--- a/include/macros.h
+++ b/include/macros.h
@@ -117,14 +117,13 @@
 #endif
 
 #if OOT_DEBUG
-
 #define LOG(exp, value, format, file, line)         \
     do {                                            \
         LogUtils_LogThreadId(file, line);           \
         osSyncPrintf(exp " = " format "\n", value); \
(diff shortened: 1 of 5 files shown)


cadmic:
I added sys_rumble.c here too since it was really easy

fig02 (approved):

hensldm (approved):

PR #1694, opened 2024-02-02
Match z_debug for retail

fig02:


src/code/z_debug.c
```diff
@@ -291,7 +293,7 @@ void Debug_DrawText(GraphicsContext* gfxCtx) {
         DebugCamera_DrawScreenText(&printer);
     }
 
-    if (gRegEditor->regPage != 0) {
+    if (OOT_DEBUG && gRegEditor->regPage != 0) {
```
cadmic, line 296: `#if` since `Regs_DrawEditor` was completely removed?

cadmic (commented):
Build has failed for real this time

src/code/z_debug.c
fig02: yeah, this is also what was causing a match issue in debug. forgot to check

cadmic (approved):

hensldm (approved):

PR #1701, opened 2024-02-02
Match ``z_fbdemo.c`` and ``z_fbdemo_fade.c``

Yanis42:
MM's version works for both except for ``z_fbdemo.c`` where I had to do some tricks to get debug matching, I searched a bit for something better for both but this is the best I could do to get both versions to match (double checked before and after formatting just in case)

No idea why but the ``\`` are required to match retail in ``z_fbdemo.c`` (I aligned them because it looks better imo), thankfully it doesn't affect debug version, also I tried using ``if(OOT_DEBUG)`` for the second ``#if`` but it didn't match unfortunately

### DIFF
--- a/src/code/z_fbdemo.c
+++ b/src/code/z_fbdemo.c
@@ -40,6 +40,12 @@ Gfx sTransTileSetupDL[] = {
     gsSPEndDisplayList(),
 };
 
+#if OOT_DEBUG
+#define LOOP_INCREMENT col++
+#else
+#define LOOP_INCREMENT col++, colTex += 0x20
+#endif
+
 void TransitionTile_InitGraphics(TransitionTile* this) {
     s32 frame;
     s32 col;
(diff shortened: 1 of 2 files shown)


src/code/z_fbdemo.c
```diff
@@ -40,6 +40,12 @@ Gfx sTransTileSetupDL[] = {
     gsSPEndDisplayList(),
 };
 
+#if OOT_DEBUG
+#define LOOP_INCREMENT col++
+#else
+#define LOOP_INCREMENT col++, colTex += 0x20
+#endif
```
fig02, line 47: this one is definitely pretty sus. can you make a scratch and post it in decomp-help? maybe others will have ideas

Yanis42: how can I do that for the debug version?

Yanis42: I wanted to do the reverse thing for some reasons but I guess giving the scratch from debug works too https://decomp.me/scratch/oe7W2 (gonna send it on discord too)

cadmic: Hm I tried for a bit and didn't make any progress. Feel free to revert this (i.e. leave this function non-matching for retail) if you want merge the rest of the changes

fig02: agree. its good information to know that this can make it match, but it surely is not the best solution. will just need some more people to take a crack at it

Dragorn421: there's probably better for `TransitionFade_Update` too so I'd suggest to close the PR for minimal bother

fig02: that match is what mm has as well, currently. so its been tried at a bunch more, theoretically

Yanis42: I'll close this then, I don't think this PR is worth if the only change is that fake match for fbdemo_fade

PR #1708, opened 2024-02-04
Syms DMA Entries Extraction Support

hensldm:
This adds support for handling YAZ0 archives when extracting the baserom. The actual archive is already handled properly, however `extract_baserom.py` needed a small change to allow proper handling of `SYMS` DMA entries (see `is_syms` comment in `dmadata.py` for a small explanation of `SYMS` entries). This also adds `SYMS` support to the spec.

Note: This is completely unused in OoT, but is being added for MM compatibility and makes it easier to keep the build tools in sync.


tools/decompress_baserom.py
```diff
@@ -171,23 +167,30 @@ def find_baserom(version: str) -> Path | None:
...
-    uncompressed_path = Path(f"baseroms/{version}/baserom-decompressed.z64")
+    baserom_dir = Path(f"baseroms/{version}")
+    if not baserom_dir.exists():
+        print(f"Error: Unknown version '{version}'.", file=sys.stderr)
+        exit(1)
+
+    uncompressed_path = baserom_dir / "baserom-decompressed.z64"
+
+    with open(baserom_dir / "checksum.md5", "r") as f:
+        correct_str_hash = f.read().split()[0]
```
Dragorn421, line 191:
```suggestion
    correct_str_hash = (baserom_dir / "checksum.md5").read_text().split()[0]
```

Dragorn421 (commented):
Could you elaborate on what "unset" means?
maybe as a comment on `DmaEntry.is_unset` too

tools/compress.py
```diff
@@ -171,8 +171,12 @@ def compress_rom(
             dmadata.DmaEntry(
                 segment.vrom_start,
                 segment.vrom_end,
-                segment_rom_start,
-                segment_rom_end if segment.is_compressed else 0,
+                segment_rom_start if not segment.is_unset else 0xFFFFFFFF,
+                (
+                    segment_rom_end
+                    if segment.is_compressed
+                    else 0 if not segment.is_unset else 0xFFFFFFFF
```
cadmic, line 178: this is getting a bit unwieldy IMO. How about something like:

```
if segment.is_unset:
  segment_rom_start = 0xFFFFFFFF
  segment_rom_end = 0xFFFFFFFF
elif segment.is_compressed:
    segment_rom_end = 0

dmadata.DmaEntry(...)
```

cadmic (commented):
Yeah I agree, as someone who doesn't know what an archive is in this context, it's strange that the word "archive" does not appear anywhere in the code. I think a PR description would help too

hensldm:
Drafting until #1709 is in

hensldm:
This should be ready for review again with 1709 in.

tools/spec.c
```diff
@@ -85,6 +85,8 @@ static bool parse_flags(char *str, unsigned int *flags)
             f |= FLAG_RAW;
         else if (strcmp(str, "NOLOAD") == 0)
             f |= FLAG_NOLOAD;
+        else if (strcmp(str, "SYMS") == 0)
+            f |= FLAG_NOLOAD;
```
cadmic, line 89:
```suggestion
            f |= FLAG_SYMS;
```

tools/extract_baserom.py
```diff
@@ -56,12 +56,17 @@ def main():
 
     args.output_dir.mkdir(parents=True, exist_ok=True)
     for dma_name, dma_entry in zip(dma_names, dma_entries):
-        if dma_entry.is_compressed():
+        if dma_entry.is_syms():
+            if (dma_entry.vrom_end - dma_entry.vrom_start) == 0:
+                continue
```
cadmic, line 61: why do we not want to extract an empty file in this case?

tools/extract_baserom.py
```diff
@@ -56,12 +56,17 @@ def main():
...
+            segment_rom_start = dma_entry.vrom_start
+        elif not dma_entry.is_compressed():
+            segment_rom_start = dma_entry.rom_start
+        else: # Segment compressed
```
cadmic, line 65: nit: flipping this conditional around so that `if dma_entry.is_compressed()` comes first would remove the need for this comment

tools/decompress_baserom.py
```diff
@@ -233,19 +234,21 @@ def main():
...
-                    "The provided baserom is a rom which has been edited with ZeldaEdit and is not suitable for use with decomp. Find a new one."
+                    "The provided baserom is a rom which has been edited with ZeldaEdit and is not suitable for use with decomp. Find a new one.",
+                    file=sys.stderr,
                 )
 
         exit(1)
 
     # Write out our new ROM
-    print(f"Writing new ROM {uncompressed_path}.")
+    print(f"Writing new ROM {uncompressed_path}...")
```
cadmic, line 251: Personally I would prefer all human-readable output go to stderr instead of splitting it up into both stdout and stderr. But I can't really imagine what else we'd use stdout for so I don't really care

tools/extract_baserom.py
hensldm: Might just be my own taste, but I like the error message being in the `else` branch. My eyes notice it better.

tools/extract_baserom.py
hensldm: For some reason I thought it was needed for matching in MM, but it is not. I can remove it.

tools/decompress_baserom.py
hensldm: I'd much prefer errors going to stderr, while natural program writing go to stdout. This way even if I redirect stdout to some file, I still get to see the errors (if there are any) front and center.

cadmic: Well if this tool did write something useful to stdout which I then redirect to a file, I would rather have "Writing new ROM" appear in my terminal instead of in my output file. Or if I redirect stderr to get a log file, I would rather have these log messages mixed in with the error messages there. But, it's really not important here

cadmic (approved):

tools/compress.py
```diff
@@ -167,17 +168,20 @@ def compress_rom(
...
+        elif not segment.is_compressed:
+            segment_rom_end = 0
+
         compressed_rom_dma_entries.append(
             dmadata.DmaEntry(
-                segment.vrom_start,
-                segment.vrom_end,
-                segment_rom_start,
-                segment_rom_end if segment.is_compressed else 0,
+                segment.vrom_start, segment.vrom_end, segment_rom_start, segment_rom_end
```
Dragorn421, line 181:
```suggestion
                segment.vrom_start, segment.vrom_end, segment_rom_start, segment_rom_end,
```
And black-format so they're on single lines
The rationale behind one on each line is exactly to minimize diff noise

tools/decompress_baserom.py
Dragorn421: Agreed it's not important here but I also think all report-thing messages should go to a single stream (either stdout or stderr, unless the program's purpose is writing to stdout in which case stderr)
If you want outout verbosity control it should be a command line argument

Dragorn421 (commented):
Is there a reason SYMS segments are not NOLOAD? (besides matching obviously, I mean for build/game function)

hensldm:
> Is there a reason SYMS segments are not NOLOAD? (besides matching obviously, I mean for build/game function)

AFAIK, no. I tried turning all the SYMS segments into NOLOAD and bombers notebook seem to work fine (one instance where the syms segment gets used)

Dragorn421 (approved):

PR #1709, opened 2024-02-04
Extract ROM segments from gc-eu-mq (take 2)

cadmic:
Take 2, after feedback on #1653.

The extraction logic is now still `extract_baserom.py` (although I moved it to `tools/` and it now uses our new `dmadata` python library). The DMA data start address and list of DMA names are still in flat .txt files, although I renamed them to `dmadata_start.txt` and `dmadata_names.txt` to try to be more consistent with the naming elsewhere. Having a file containing just an integer is still a bit ugly, but IMO a more complicated config format doesn't seem worth it until we know what other data we should put in there.


Makefile
```diff
@@ -366,9 +366,9 @@ venv:
 setup: venv
 	$(MAKE) -C tools
 	$(PYTHON) tools/decompress_baserom.py $(VERSION)
-# TODO: for now, we only extract ROM segments and assets from the Debug ROM
+	$(PYTHON) tools/extract_baserom.py $(BASEROM_DIR)/baserom-decompressed.z64 -o $(BASEROM_SEGMENTS_DIR) --dma-start `cat $(BASEROM_DIR)/dma_start.txt` --dma-names $(BASEROM_DIR)/dma_names.txt
```
Dragorn421, line 369: Can you change the file names to be `dmadata_` instead of `dma_`, not that there's a risk of confusion but [DMA](https://en.wikipedia.org/wiki/Direct_memory_access) != dmadata (what the og devs called the table and how we refer to it in the repo as well) and I like technical correctness

fwiw the `--dma-start` in compress.py is because it was `--dma-range` in my script, which I mimicked from z64compress' command line argument `--dma`
`--dmadata-start` would be better

I guess you can also apply this comment to `--dma-` args in extract_baserom.py

tools/extract_baserom.py
```diff
@@ -0,0 +1,70 @@
...
+import argparse
+from pathlib import Path
+import sys
+
+import dmadata
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Extract DMA segments from an uncompressed ROM."
```
Dragorn421, line 17:
```suggestion
        description="Extract segments from an uncompressed ROM, based on its dmadata."
```
or similar

tools/extract_baserom.py
```diff
@@ -0,0 +1,70 @@
...
+        required=True,
+        help=(
+            "The dmadata location in the rom, as a hexadecimal offset (e.g. 0x12f70)"
+        ),
+    )
+    parser.add_argument(
+        "--dma-names",
+        type=Path,
+        required=True,
+        help="Path to file containing names of DMA segments",
```
Dragorn421, line 41:
```suggestion
        help="Path to file containing segments names",
```

tools/extract_baserom.py
```diff
@@ -0,0 +1,70 @@
...
+    )
+
+    args = parser.parse_args()
+
+    rom_data = args.rom.read_bytes()
```
Dragorn421, line 46:
```suggestion
    rom_data = memoryview(args.rom.read_bytes())
```
so that slices won't make a copy, instead will create views of the underlying data
this is a free perf boost

Dragorn421 (commented):
some overlaps with #1708, I think this (#1709) is a better fit for going first

Makefile
hensldm, line 369: Curious as to the reasoning to pass everything by arguments, instead of just passing a version like `decompress_baserom.py`?

Dragorn421: I can live with the inconsistency but if I had to choose I'd like "file paths in the makefile" better than "hardcoded in a script"

cadmic: Okay, I tried to change everything in sight to use the new names too

cadmic: Using arguments makes it easier to run the tool outside of the build system, either for testing or because it's already useful on its own. For example, I've already used `extract_baserom.py` to extract segments from the non-matching built ROM to make it easier to study. It could be nice if `decompress_baserom.py` took arguments too, but it does a lot more stuff (e.g. checking hashes), and I can z64decompress instead of running it on its own, so I can live with that too.

tools/compress.py
```diff
@@ -207,8 +207,7 @@ def main():
         help="path of the compressed rom to write out",
     )
     parser.add_argument(
-        "--dma-start",
-        dest="dma_start",
+        "--dmadata-start",
         type=lambda s: int(s, 16),
```
Dragorn421, line 211: can you keep `dest=` for grepability

tools/extract_baserom.py
```diff
@@ -0,0 +1,70 @@
...
+            file=sys.stderr,
+        )
+        exit(1)
+
+    args.output_dir.mkdir(parents=True, exist_ok=True)
+    for dma_name, dma_entry in zip(dma_names, dma_entries):
+        if dma_entry.is_compressed():
+            print(f"Error: segment {dma_name} is compressed", file=sys.stderr)
+            exit(1)
+        segment_data = rom_data[dma_entry.vrom_start : dma_entry.vrom_end]
```
Dragorn421, line 62:
```py
        segment_rom_start = dma_entry.rom_start
        segment_rom_end = dma_entry.rom_start + (
            dma_entry.vrom_end - dma_entry.vrom_start
        )
```
```suggestion
        segment_data = rom_data[segment_rom_start : segment_rom_end]
```
yes typically that's the same but not necessarily

cadmic:
This shows that the only difference between the spec and dmadata_names.txt is `buffers`:

```
$ diff -U 3 <(grep 'name "' build/gc-eu-mq/spec | sed 's/.*"\(.*\)".*/\1/g') <(cat baseroms/gc-eu-mq/dmadata_names.txt)
--- /dev/fd/63	2024-02-04 17:08:09
+++ /dev/fd/62	2024-02-04 17:08:09
@@ -27,7 +27,6 @@
 map_48x85_static
 map_i_static
 code
-buffers
 ovl_title
 ovl_select
 ovl_opening
```

hensldm (approved):

PR #1710, opened 2024-02-04
PreRender.c retail 1 not matching

shawlucas:
Not sure if this should be submitted but I'll try anyway. Most of the other functions were figured out from using MM decomp but the DivotFilter function is complicated and above my knowledge, couldn't figure it out.


fig02 (approved):
partial files are fine. feel free to post a scratch in the discord for the function gving you a problem

cadmic (approved):

Dragorn421 (approved):

src/code/PreRender.c
```diff
@@ -545,7 +549,7 @@ void PreRender_AntiAliasFilter(PreRender* this, s32 x, s32 y) {
     s32 buffB[5 * 3];
     s32 xi;
     s32 yi;
-    s32 pad;
+    s32 temp;
```
Thar0, line 552:
```suggestion
    s32 invCvg;
```

Thar0 (commented):
Temp var name suggestion

src/code/PreRender.c
Dragorn421: inv is 1/ though I get what you mean

Thar0: It's an "additive inverse" which should be clear from the context

PR #1713, opened 2024-02-05
rcp retail OK

hensldm:
Not sure how we want to actually handle the HREGs, but the only way I could get it to match retail was having the switch statements take 0s.

Note we could also do ternaries, but I think I like the defines approach better.

### DIFF
--- a/src/code/z_rcp.c
+++ b/src/code/z_rcp.c
@@ -1482,7 +1482,7 @@ void Gfx_SetupFrame(GraphicsContext* gfxCtx, u8 r, u8 g, u8 b) {
     if ((R_PAUSE_BG_PRERENDER_STATE <= PAUSE_BG_PRERENDER_SETUP) && (gTransitionTileState <= TRANS_TILE_SETUP)) {
         s32 letterboxSize = Letterbox_GetSize();
 
-        #if OOT_DEBUG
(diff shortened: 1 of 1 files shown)


Dragorn421 (approved):
hm

I guess it will do for now

src/code/z_rcp.c
```diff
@@ -1246,23 +1246,33 @@ Gfx* Gfx_SetupDL_69NoCD(Gfx* gfx) {
...
+#define HREG_22 0
+#endif
+
 Gfx* func_800947AC(Gfx* gfx) {
     gSPDisplayList(gfx++, sSetupDL[SETUPDL_65]);
     gDPSetColorDither(gfx++, G_CD_DISABLE);
 
     // clang-format off
-    switch (HREG(21)) {
+    switch (HREG_21) {
```
cadmic, line 1262: I think I prefer the ternary actually, IMO it's a bit easier to read since my eyes don't have to jump somewhere else to see what's going on. But I don't feel strongly about it so feel free to ignore

cadmic (approved):

fig02 (approved):
I guess I dont really have a preference at the moment. I do still think there is a consistent pattern to how debug regs are handled, and whatever direction we go with that may also work in this case. not sure

PR #1718, opened 2024-02-05
Retail Player_Update/Noclip mode + some docs

fig02:
This matches retail things related to Noclip mode and does some documentation with it. Some notes:
- I moved the declaration down to where it is relevant 
- I used the goto method of matching to keep the ifdef stuff confined in one area. 
- The semicolon on the skip_update label is purely to make the formatter play nice
- I spell it as "Noclip" because that seems to be the common spelling in other game contexts. 
- The return on the noclip mode is unfortunately backwards with regards to "true" and "false", so the naming is a bit weird

### DIFF
--- a/src/overlays/actors/ovl_player_actor/z_player.c
+++ b/src/overlays/actors/ovl_player_actor/z_player.c
@@ -13634,7 +13634,7 @@ void Player_Action_8084FBF4(Player* this, PlayState* play) {
 
 #if OOT_DEBUG
 /**
- * Updates the "Noclip" debug feature, which allows the player to fly around anywhere 
+ * Updates the "Noclip" debug feature, which allows the player to fly around anywhere
  * in the world and clip through any collision.
  *
  * Noclip can be enabled with two different button combos:
--- a/src/overlays/actors/ovl_player_actor/z_player.c
+++ b/src/overlays/actors/ovl_player_actor/z_player.c
@@ -13656,7 +13656,7 @@ void Player_Action_8084FBF4(Player* this, PlayState* play) {
 s32 Player_UpdateNoclip(Player* this, PlayState* play) {
     sControlInput = &play->state.input[0];
 
-    if ((CHECK_BTN_ALL(sControlInput->cur.button, BTN_A | BTN_L | BTN_R) &&
+    if ((CHECK_BTN_ALL(sControlInput->cur.button, BTN_L | BTN_R | BTN_A) &&
          CHECK_BTN_ALL(sControlInput->press.button, BTN_B)) ||
         (CHECK_BTN_ALL(sControlInput->cur.button, BTN_L) && CHECK_BTN_ALL(sControlInput->press.button, BTN_DRIGHT))) {
 
--- a/src/overlays/actors/ovl_player_actor/z_player.c
+++ b/src/overlays/actors/ovl_player_actor/z_player.c
@@ -13634,6 +13634,19 @@ void Player_Action_8084FBF4(Player* this, PlayState* play) {
 /**
  * Updates the "Noclip" debug feature, which allows the player to
  * fly around anywhere in the world and clip through any collision.
+ * It can be enabled with two seperate button combos:
+ * A + L + R held down and then B
+ * or
+ * L held down and then D-pad right
+ *
+ * To control Noclip mode:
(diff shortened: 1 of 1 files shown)


src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -13618,21 +13632,41 @@ void Player_Action_8084FBF4(Player* this, PlayState* play) {
...
+ * To control Noclip mode:
+ * - Move horizontally with the 4 D-pad directions
+ * - Move up with B
+ * - Move down with A
+ * - Hold R to move faster
+ *
+ * With Noclip enabled, another button combination can be pressed to set all "temp clear" flags
+ * in the current room. To do so Hold L and press D-pad right.
+ *
+ * @return  true if Noclip is disabled, false if enabled
```
cadmic, line 13653: Maybe would be clearer to say it returns whether Link should update normally (since that's how it's used)

cadmic (approved):

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -13618,21 +13632,41 @@ void Player_Action_8084FBF4(Player* this, PlayState* play) {
...
+ * L held down and then D-pad right
+ *
+ * To control Noclip mode:
+ * - Move horizontally with the 4 D-pad directions
+ * - Move up with B
+ * - Move down with A
+ * - Hold R to move faster
+ *
+ * With Noclip enabled, another button combination can be pressed to set all "temp clear" flags
+ * in the current room. To do so Hold L and press D-pad right.
```
Yanis42, line 13651: isn't it dpad left for the temp clear feature?

src/overlays/actors/ovl_player_actor/z_player.c
fig02: It would make more sense that way yeah. But in a vacuum the function doesnt know what it indicates with regards to player's update, just the state of noclip. So it feels weird to be talking about the implication on players update in this function.
Would like to hear others thoughts as well

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -13618,21 +13632,42 @@ void Player_Action_8084FBF4(Player* this, PlayState* play) {
     func_8002F8F0(&this->actor, NA_SE_VO_LI_TAKEN_AWAY - SFX_FLAG + this->ageProperties->unk_92);
 }
 
-s32 func_8084FCAC(Player* this, PlayState* play) {
+#if OOT_DEBUG
+/**
+ * Updates the "Noclip" debug feature, which allows the player to fly around anywhere
+ * in the world and clip through any collision.
+ *
+ * Noclip can be enabled with two different button combos:
```
Dragorn421, line 13640:
```suggestion
 * Noclip can be toggled on and off with two different button combos:
```

src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -13618,21 +13632,42 @@ void Player_Action_8084FBF4(Player* this, PlayState* play) {
...
+ * Hold A + L + R and press B
```
Dragorn421, line 13641: the
`if (!CHECK_BTN_ALL(sControlInput->cur.button, BTN_L))`
check before checking for A/B
makes me believe their combo was holding L before A, so Link isn't moved by holding A
anyway even for just practical reasons I'd suggest
```suggestion
 * Hold L + R + A and press B
```
the order in `BTN_A | BTN_L | BTN_R` is completely arbitrary after all (in this case from most significant bit to least)

src/overlays/actors/ovl_player_actor/z_player.c
Dragorn421: as much as I hate the double negative I agree it's better to reflect what the function does than what it's used for

src/overlays/actors/ovl_player_actor/z_player.c
fig02: sure, done

fig02: I also changed the button check to reflect this

Dragorn421 (approved):

PR #1720, opened 2024-02-05
Match even more retail actors

cadmic:
EDIT: never mind, no macros here

~~I created a macro for `DebugDisplay_AddObject` since it's rather common and there are a few cases where `if (OOT_DEBUG && ...)` did not work. Unfortunately there are still a few special cases, and it doesn't work in `EnWonderItem_Update` due to same-line issues. Please suggest a better name for the macro, maybe it should really be called `DEBUG_DISPLAY_ADD_OBJECT_DEBUG` but that sounds a bit redundant.~~

~~It's also possible to go a bit farther with the macros, for example this block of code is used a few times:~~
```
    if (this->outOfRangeTimer != 0) {
        if (this->outOfRangeTimer % 2 == 0) {
            DebugDisplay_AddObject(this->actor.world.pos.x, this->actor.world.pos.y, this->actor.world.pos.z,
                                   this->actor.world.rot.x, this->actor.world.rot.y, this->actor.world.rot.z, 1.0f,
                                   1.0f, 1.0f, 120, 120, 120, 255, 4, play->state.gfxCtx);
        }
    } else {
        DebugDisplay_AddObject(this->actor.world.pos.x, this->actor.world.pos.y, this->actor.world.pos.z,
                               this->actor.world.rot.x, this->actor.world.rot.y, this->actor.world.rot.z, 1.0f,
                               1.0f, 1.0f, 255, 0, 255, 255, 4, play->state.gfxCtx);
    }
```

~~I'll leave that as future work though.~~

### DIFF
--- a/src/overlays/actors/ovl_En_Dnt_Demo/z_en_dnt_demo.c
+++ b/src/overlays/actors/ovl_En_Dnt_Demo/z_en_dnt_demo.c
@@ -326,8 +326,7 @@ void EnDntDemo_Update(Actor* thisx, PlayState* play) {
     }
     this->actionFunc(this, play);
 
-#if OOT_DEBUG
-    if (BREG(0) != 0) {
+    if (OOT_DEBUG && BREG(0) != 0) {
         if (this->debugArrowTimer != 0) {
             if (!(this->debugArrowTimer & 1)) {
                 DebugDisplay_AddObject(this->actor.world.pos.x, this->actor.world.pos.y, this->actor.world.pos.z,
@@ -340,5 +339,4 @@ void EnDntDemo_Update(Actor* thisx, PlayState* play) {
                                    1.0f, 1.0f, 255, 255, 255, 255, 4, play->state.gfxCtx);
         }
     }
-#endif
 }
--- a/src/overlays/actors/ovl_En_Dnt_Demo/z_en_dnt_demo.c
+++ b/src/overlays/actors/ovl_En_Dnt_Demo/z_en_dnt_demo.c
@@ -185,8 +185,7 @@ void EnDntDemo_Judge(EnDntDemo* this, PlayState* play) {
                 case PLAYER_MASK_ZORA:
                 case PLAYER_MASK_GERUDO:
                     rand9 = Rand_ZeroFloat(8.99f);
-                    maskIdx = Player_GetMask(play);
-                    maskIdx--;
+                    maskIdx = (s16)Player_GetMask(play) - 1;
                     if (rand9 == 8) {
                         ignore = true;
                         delay = 8;
@@ -265,6 +264,9 @@ void EnDntDemo_Judge(EnDntDemo* this, PlayState* play) {
 
 void EnDntDemo_Results(EnDntDemo* this, PlayState* play) {
     s32 i;
+    s16 offsetAngle;
+    Vec3f leaderPos;
+    f32 offsetDist;
 
     if (this->leaderSignal != DNT_SIGNAL_NONE) {
         for (i = 0; i < 9; i++) {
@@ -284,9 +286,8 @@ void EnDntDemo_Results(EnDntDemo* this, PlayState* play) {
         this->actionFunc = EnDntDemo_Prize;
     } else if (this->prize == DNT_PRIZE_STICK) {
         for (i = 0; i < 9; i++) {
-            s16 offsetAngle = -this->leader->actor.shape.rot.y;
-            Vec3f leaderPos = this->leader->actor.world.pos;
-            f32 offsetDist;
+            offsetAngle = -this->leader->actor.shape.rot.y;
+            leaderPos = this->leader->actor.world.pos;
 
             if (!(i & 1)) {
                 offsetAngle -= 0x59D8;
@@ -324,7 +325,9 @@ void EnDntDemo_Update(Actor* thisx, PlayState* play) {
         this->unkTimer1--;
     }
     this->actionFunc(this, play);
-    if (BREG(0)) {
+
+#if OOT_DEBUG
+    if (BREG(0) != 0) {
         if (this->debugArrowTimer != 0) {
             if (!(this->debugArrowTimer & 1)) {
                 DebugDisplay_AddObject(this->actor.world.pos.x, this->actor.world.pos.y, this->actor.world.pos.z,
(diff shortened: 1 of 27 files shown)


Dragorn421 (commented):
I'm not convinced by the `DEBUG_DISPLAY_ADD_OBJECT` macro, it's too inconsistent to be real

cadmic:
Alright, I've reverted `DEBUG_DISPLAY_ADD_OBJECT`. I still used `OOT_DEBUG &&` everywhere I could (it didn't seem to work in `EnSkj_OcarinaMinigameShortStumpUpdate`, `EnWonderItem_MultitagFree`, and `EnWonderItem_MultitagOrdered`)

src/overlays/actors/ovl_En_Encount1/z_en_encount1.c
```diff
@@ -51,6 +51,8 @@ void EnEncount1_Init(Actor* thisx, PlayState* play) {
     spawnRange = 120.0f + (40.0f * this->actor.world.rot.z);
     this->spawnRange = spawnRange;
 
+    if (1) {}
+
     PRINTF("\n\n");
```
Dragorn421, line 56: kinda curious if this matches as `if(OOT_DEBUG)` surrounding the printfs
(if it doesn not sure if we'd want that)

src/overlays/actors/ovl_En_Skj/z_en_skj.c
```diff
@@ -445,13 +444,17 @@ void EnSkj_Init(Actor* thisx, PlayState* play2) {
...
-            PRINTF("Player_X : %f\n", player->actor.world.pos.x);
-            PRINTF("Player_Z : %f\n", player->actor.world.pos.z);
-            PRINTF("World_X  : %f\n", this->actor.world.pos.x);
-            PRINTF("World_Z  : %f\n", this->actor.world.pos.z);
-            PRINTF("Center_X : %f\n", this->center.x);
-            PRINTF("Center_Z : %f\n\n", this->center.z);
+#if OOT_DEBUG
+            {
+                Player* player = GET_PLAYER(play);
+                PRINTF("Player_X : %f\n", player->actor.world.pos.x);
```
Dragorn421, line 450: newline police reporting for duty

src/overlays/actors/ovl_En_Skj/z_en_skj.c
```diff
@@ -1502,15 +1505,14 @@ void EnSkj_OfferNextRound(EnSkj* this, PlayState* play) {
...
     if (Message_GetState(&play->msgCtx) == TEXT_STATE_CHOICE && Message_ShouldAdvance(play)) {
         switch (play->msgCtx.choiceIndex) {
-            case 0: // yes
-                player = GET_PLAYER(play);
+            case 0: { // yes
+                Player* player = GET_PLAYER(play);
                 player->stateFlags3 |= PLAYER_STATE3_5; // makes player take ocarina out right away after closing box
                 this->actionFunc = EnSkj_SetupWaitForOcarina;
                 break;
+            }
```
Dragorn421, line 1515:
```suggestion
            } break;

```

src/overlays/actors/ovl_En_Encount1/z_en_encount1.c
cadmic: It does (for the record, `if (1)` surrounding the printfs matches too)

src/overlays/actors/ovl_En_G_Switch/z_en_g_switch.c
```diff
@@ -80,6 +80,8 @@ void EnGSwitch_Init(Actor* thisx, PlayState* play) {
     s32 pad;
     EnGSwitch* this = (EnGSwitch*)thisx;
 
+    if (play) {}
+
     this->type = (this->actor.params >> 0xC) & 0xF;
```
fig02, line 85: does this match with a play2 ( a fake match for recasting the input parameter)
For a while now we've known that they very likely originally passed GameState* to the 4 main actor function and made a temp to recast them in all cases. But we've yet to figure out if it works in ever single case.
If it does, I would prefer a `play2` for now over this

cadmic: It has regalloc issues unfortunately, there's probably a way to get it to work but I wasn't able to figure it out

fig02: thanks for checking

fig02 (approved):

src/overlays/actors/ovl_En_Dnt_Demo/z_en_dnt_demo.c
```diff
@@ -185,8 +185,7 @@ void EnDntDemo_Judge(EnDntDemo* this, PlayState* play) {
                 case PLAYER_MASK_ZORA:
                 case PLAYER_MASK_GERUDO:
                     rand9 = Rand_ZeroFloat(8.99f);
-                    maskIdx = Player_GetMask(play);
-                    maskIdx--;
+                    maskIdx = (s16)Player_GetMask(play) - 1;
```
Dragorn421, line 188: this looks fake to me (specifically the cast, I think it's more likely it was a short assignment like in main), can you add a `// fake` comment

Dragorn421 (approved):
after my latest comment is addresses this is good for me

src/overlays/actors/ovl_En_Dnt_Demo/z_en_dnt_demo.c
cadmic: How would you feel about this?
```
// clang-format off
maskIdx = Player_GetMask(play); maskIdx--;
// clang-format on
```

Dragorn421: looks more plausible but formatting oddities suck so not sure if it's overall better
I guess it is slightly?

fig02: either one can have a fake match comment for now and see if anyone can find something better during a fake match pass

cadmic: Alright, I left it as is and left the alternative in a comment

Dragorn421 (approved):

PR #1721, opened 2024-02-05
Hacks for building nes_font_static and nes_message_data_static with gc-eu-mq

cadmic:
Unfortunately the text for gc-eu-mq has changed, and 2 textures in `nes_font_static` are different. Since these changes are small I've hacked around it by adding some special cases in the Makefile. I figure that the assets system probably needs to be rethought anyway for the next retail version, but it would be a big milestone to get gc-eu-mq matching before that.


Makefile
```diff
@@ -438,6 +438,24 @@ $(BUILD_DIR)/assets/%.o: assets/%.c
 	$(CC) -c $(CFLAGS) $(MIPS_VERSION) $(OPTFLAGS) -o $@ $<
 	$(OBJCOPY) -O binary $@ $@.bin
 
+# Temporary hacks to handle asset differences
+ifeq ($(VERSION),gc-eu-mq)
+$(BUILD_DIR)/assets/text/message_data.enc.h:
+	$(PYTHON) tools/msgdis.py --oot-version $(VERSION) $(BUILD_DIR)/assets/text/message_data.h
```
Dragorn421, line 444: this would be less weird to change with part of #1491

cadmic: Maybe, although I think we'd still need the hack here to avoid clobbering the debug ROM assets

tools/msgdis.py
```diff
@@ -404,3 +407,21 @@ def extract_all_text(text_out, staff_text_out):
...
+    args = parser.parse_args()
+
+    if args.oot_version == "gc-eu-mq":
+        # Yikes
+        version = "gc-eu-mq"
+        nes_message_entry_table_addr = 0xB7E8F0
+        ger_message_entry_table_addr = 0xB82B18
+        fra_message_entry_table_addr = 0xB84C28
+        staff_message_entry_table_addr = 0xB86D38
+        staff_message_entry_table_addr_end = 0xB86EC0
```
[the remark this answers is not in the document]
cadmic: I guess I don't mind refactoring this script, it's easy enough to test. But I'd like to still hardcode the addresses for now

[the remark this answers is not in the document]
cadmic: On second thought, I don't think this is worth doing until we're more sure about what this tool should do in the future. I moved all the addresses to the same place at least so it's a bit less confusing

Dragorn421:
Alternative proposal for msgdis handling
https://github.com/zeldaret/oot/compare/main...Dragorn421:oot:meta_version_conf

Dragorn421:
Moving forward on extracting message data for multiversion with #1730

fig02:
do changes need to happen here because of the new way of handling text?

cadmic:
I think at this point it might be easier to do things properly than hack it, so I'll hold of on rebasing for now

cadmic:
Actually after thinking about it a bit more, there might still a long way to go for multiversion assets, so I've updated this branch and re-tested it. The nes_font_static is still quite hacky, but now nes_message_data_static is a bit cleaner though

Makefile
```diff
@@ -439,6 +439,20 @@ $(BUILD_DIR)/assets/%.o: assets/%.c
 	$(CC) -c $(CFLAGS) $(MIPS_VERSION) $(OPTFLAGS) -o $@ $<
 	$(OBJCOPY) -O binary $@ $@.bin
 
+# Temporary hack to handle asset differences
+ifeq ($(VERSION),gc-eu-mq)
+$(BUILD_DIR)/assets/textures/nes_font_static/nes_font_static.o:
```
Dragorn421, line 444: Is including the baserom file in the spec not an option? (I don't know if this file provides symbols the source relies on. Even if it does though they could just go into undefined_syms.txt)

Also to be clear this is needed changes wrt moving towards OK right, it's not technically required to build the rom at all?

cadmic: I thought it would it be nice to quarantine to the Makefile since we'll have to update that to support multiversion assets. But I didn't think about including it in the spec and it is much simpler there, so I've changed to that

cadmic: > Also to be clear this is needed changes wrt moving towards OK right, it's not technically required to build the rom at all?

Right (although this did inadvertently fix the build for gc-eu-mq after the extracted/ change)

Dragorn421: yeah you want to basically revert #1911's Makefile changes lol, this is addressing it properly

fig02 (approved):

Dragorn421 (approved):
unrelated binutils infodump ahead (nothing holding the PR)

<details>

I checked this builds matching message data with
`vbindiff build/gc-eu-mq/assets/text/nes_message_data_static.o.bin baseroms/gc-eu-mq/segments/nes_message_data_static`
and vbindiff showed
image
this is the only diff

turns out this must be an objcopy problem / misuse because the section data is correct:
```
$ mips-linux-gnu-objdump -s -j.rodata build/gc-eu-mq/assets/text/nes_message_data_static.o | head

build/gc-eu-mq/assets/text/nes_message_data_static.o:     file format elf32-tradbigmips

Contents of section .rodata:
 00000 1a132d08 596f7520 626f7272 6f776564  ..-.You borrowed
 00010 20612005 41506f63 6b657420 45676705   a .APocket Egg.
 00020 40210901 4120506f 636b6574 20437563  @!..A Pocket Cuc
 00030 636f2077 696c6c20 68617463 68206672  co will hatch fr
 00040 6f6d0169 74206f76 65726e69 6768742e  om.it overnight.
 00050 20426520 73757265 20746f20 67697665   Be sure to give
```

and yeah objcopy -O binary is not behaving as expected,
with -j.rodata to only dump that section:
```
$ mips-linux-gnu-objcopy -j.rodata build/gc-eu-mq/assets/text/nes_message_data_static.o -O binary a.bin && { xxd a.bin | head -n3 ; }
00000000: 1a13 2d08 596f 7520 626f 7272 6f77 6564  ..-.You borrowed
00000010: 2061 2005 4150 6f63 6b65 7420 4567 6705   a .APocket Egg.
00000020: 4021 0901 4120 506f 636b 6574 2043 7563  @!..A Pocket Cuc
```
without:
```
$ mips-linux-gnu-objcopy build/gc-eu-mq/assets/text/nes_message_data_static.o -O binary a.bin && { xxd a.bin | head -n3 ; }
00000000: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000010: 0000 0000 0000 7ff0 6b65 7420 4567 6705  ........ket Egg.
00000020: 4021 0901 4120 506f 636b 6574 2043 7563  @!..A Pocket Cuc
```

I think this is `.reginfo` being copied at LMA 0
```
$ mips-linux-gnu-objdump -h build/gc-eu-mq/assets/text/nes_message_data_static.o

build/gc-eu-mq/assets/text/nes_message_data_static.o:     file format elf32-tradbigmips

Sections:
Idx Name          Size      VMA       LMA       File off  Algn
  0 .rodata       00038080  00000000  00000000  00012a30  2**4
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  1 .options      00000040  00000000  00000000  0004aab0  2**3
                  CONTENTS, READONLY
  2 .reginfo      00000018  00000000  00000000  0004aaf0  2**2
                  CONTENTS, ALLOC, LOAD, READONLY, DATA, LINK_ONCE_SAME_SIZE
  3 .mdebug       00023658  00000000  00000000  0004ab08  2**2
                  CONTENTS, READONLY, DEBUGGING
```

TLDR it may be objcopying relocatable elf files is useless due to sections overlapping from not having gone through linking

</details>

PR #1723, opened 2024-02-06
Split msgdis from extract_assets

Dragorn421:


tools/msgdis.py
```diff
@@ -404,3 +404,23 @@ def extract_all_text(text_out, staff_text_out):
 
         with open(staff_text_out, "w", encoding="utf8") as outfile:
             outfile.write(out.strip() + "\n")
+
+
+def main():
+    from argparse import ArgumentParser
```
cadmic, line 410: why here and not with the other imports?

cadmic (approved):

tools/msgdis.py
Dragorn421: why not, it's not needed elsewhere

cadmic: It doesn't save any runtime, especially now that msgdis.py is only run by itself and not as a library (although even then argparse was already imported). And generally Python programs import everything globally, not per-function, unless there's a strong reason to break from that

fig02 (approved):

PR #1726, opened 2024-02-06
Rename func_80853148 to Player_StartTalking

mckinlee:
Simply what the title says. Over at Shipwright I'm working on a PR where I renamed this for clarity.

### DIFF
--- a/src/overlays/actors/ovl_player_actor/z_player.c
+++ b/src/overlays/actors/ovl_player_actor/z_player.c
@@ -261,7 +261,7 @@ s32 func_80852F38(PlayState* play, Player* this);
 s32 Player_TryCsAction(PlayState* play, Actor* actor, s32 csAction);
 void func_80853080(Player* this, PlayState* play);
 s32 Player_InflictDamage(PlayState* play, s32 damage);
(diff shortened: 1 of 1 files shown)


fig02:
While I agree with this name change, Talking as a whole is a larger system in player that should all be docuemented together in a more cohesive manner. Ive already started the work to do this and will PR it when its ready

Dragorn421:
Hi, thanks for the PR!

Yeah as fig said it's typically better to document systems at once, you have a lot less chance to miss out on some subtleties. And on the odd chance that something was named a bit too confidently and turns out inaccurate it can be pretty hard to spot and demonstrate (especially if later passes built upon that flawed assumption)
Especially with z_player which is extra complicated we want to be extra careful

Since fig plans broader docs passes on player I think we'll just close this, but feel free to contribute more in the future :)

fig02:
Closing per the explanations above.

As mentioned feel free to get involved with more documentation efforts. It may help if you join the discord and mention what you plan to work on, so it can be discussed with others who may have already planned to work on something already. Just helps to avoid stepping on each others toes 😄 
Thank you

PR #1727, opened 2024-02-06
Match retail z_fbdemo.c and z_fbdemo_fade.c

cadmic:
I wasn't able to find anything better than MM's fake match for z_fbdemo_fade.c, so I think we should go with it for now.

### DIFF
--- a/src/code/z_fbdemo.c
+++ b/src/code/z_fbdemo.c
@@ -40,6 +40,22 @@ Gfx sTransTileSetupDL[] = {
     gsSPEndDisplayList(),
 };
 
+#define SET_VERTEX(vtx, ob0, ob1, ob2, f, tc0, tc1, n0, n1, n2, alpha) \
+    {                                                                  \
+        Vtx_tn* vtxn = &(vtx)->n;                                      \
+        vtxn->ob[0] = (ob0);                                           \
+        vtxn->ob[1] = (ob1);                                           \
+        vtxn->ob[2] = (ob2);                                           \
+        vtxn->flag = (f);                                              \
+        vtxn->tc[0] = (tc0);                                           \
+        vtxn->tc[1] = (tc1);                                           \
+        vtxn->n[0] = (n0);                                             \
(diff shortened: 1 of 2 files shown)


src/code/z_fbdemo.c
```diff
@@ -40,6 +40,22 @@ Gfx sTransTileSetupDL[] = {
     gsSPEndDisplayList(),
 };
 
+#define SET_VERTEX(vtx, ob0, ob1, ob2, f, tc0, tc1, n0, n1, n2, alpha) \
+    {                                                                  \
+        Vtx_tn* vtxn = &(vtx)->n;                                      \
+        vtxn->ob[0] = (ob0);                                           \
```
fig02, line 46: does this make sense to use in more places?
Believe the answer is yes but it doesnt match in all cases, idr

fig02 (approved):

src/code/z_fbdemo.c
Dragorn421: I'm really interested in the answer being yes but I've seen enough subtle differences to have my doubts

I hope, though

cadmic: I think this is the only place it would be used (specifically for `vtx->n`, not `vtx->v`). I'm not sure about the `v` counterpart, there's a lot of scary code in e.g. z_parameter.c which I think I don't want to mess with right now

src/code/z_fbdemo.c
```diff
@@ -40,6 +40,22 @@ Gfx sTransTileSetupDL[] = {
     gsSPEndDisplayList(),
 };
 
+#define SET_VERTEX(vtx, ob0, ob1, ob2, f, tc0, tc1, n0, n1, n2, alpha) \
```
Dragorn421, line 43:
```suggestion
#define SET_VERTEX(vtx, x, y, z, s, t, nx, ny, nz, alpha) \
```
for the f / flag argument, just put 0 inside the macro, it's padding

Dragorn421 (approved):

PR #1730, opened 2024-02-07
Introduce extracted/VERSION, with text extracted there

Dragorn421:
The plan :tm: for assets,
(at least it's what I suggested over a year ago and there are no concerns nor other suggestions)
is to extract from the rom into `extracted/`, and commit `assets/**/*.c` files `#include`ing from there.

`extracted/` is fully gitignored.

With multiversion I extended this to extract into `extracted/VERSION/`. `extracted/VERSION` is added to the include path so that files can include from there using `#include "..."`

This PR also moves extracting text to `extracted/VERSION/text`. There are also some changes to the message extraction/build logic I had to make to accommodate for that change.

The extracted text `extracted/gc-eu-mq-dbg/text/message_data.h`, `extracted/gc-eu-mq-dbg/text/message_data_staff.h` is included by now-committed (one line include each) `assets/text/message_data.h` and `assets/text/message_data_staff.h` (respectively).
Note that this means in a modding context messages can be added into git-tracked `assets/text/message_data.h`, and the vanilla messages can be removed simply by removing the include.

Discussion/suggestions welcome!


Yanis42:
I don't understand what's the point of having the ``version_assets`` folder when you can extract assets to ``_extracted/VERSION``, seems like you include this path in the makefile so I don't see any reason why this wouldn't work

am I missing something? maybe some edge case I'm not seeing here? or some future change that will make this necessary? 🤔

Dragorn421:
I think it's more obvious where an include is pulling from as
`#include "version_assets/text/message_data.h"`
rather than
`#include "text/message_data.h"`

Currently I'm not even sure the latter would work given the include path also has `-Iassets` and `assets/text/message_data.h` also exists
(Note I think we want to no longer `-Iassets` eventually, so that could be addressed)

cadmic:
I agree with this in principle (extracting assets for different versions into different directories) but I would like to bikeshed the path a bit also.

First, why the underscore in `_extracted`? I understand the desire to mark this dir as "generated", but we don't do this for any of our other versioned directories (`expected`, `build` and baserom `segments`).

Second, `assets/_extracted/$(VERSION)/version_assets` is too long for my taste. I'd most like it to be extracted into `assets/$(VERSION)`, since that has a nice symmetry with `build/$(VERSION)` and `baseroms/$(VERSION)`. The `version_assets` part should not be necessary if the include paths are in the right order, which might be confusing temporarily but not in the long run if we plan to get rid of `-Iassets`. For the non-gitignored stuff currently in `assets/`: `assets/text` probably doesn't need to exist (it could be generated by the text extractor, especially to deal with NTSC vs PAL), and `assets/xml` could be moved somewhere else (e.g. a `zapd` dir).

I'd also be okay with a different toplevel directory name other than `assets/`, although I couldn't think of a good one. But I would like it if the extracted asset dir was `topleveldir/$(VERSION)`.

fig02:
I have not read this PR in depth yet, but would like to say I agree that I dont think a leading underscore is needed

tools/msgenc.py
```diff
@@ -51,21 +52,27 @@ def cvt_str(m):
 def main():
     parser = argparse.ArgumentParser(description="Encode message_data_static text headers")
     parser.add_argument("charmap", help="path to charmap file specifying custom encoding elements")
-    parser.add_argument("input", help="path to file to be encoded")
-    parser.add_argument("output", help="encoded file")
+    parser.add_argument("--input", help="path to file to be encoded, or - for stdin", default="-")
```
cadmic, line 55: FWIW I'd rather this be positional and charmap to be --charmap than the other way around

Also `-o` for `--output` might be nice, but not important since I don't think too many people will run this by hand

assets/text/message_data.h
```diff
@@ -0,0 +1 @@
+#include "version_assets/text/message_data.h"
```
cadmic, line 1: Oh I see, this was the purpose behind `version_assets`. Does this file need to exist at all if the include paths are set up correctly?

Dragorn421: Having an indirection like this is very useful for modding, it means you can add your own messages directly in git-tracked assets/text/message_data.h, and can remove the vanilla messages just by not including them. Or using messages from a specific version with `#include "assets/.../gc-eu-mq/.../message_data.h"`

This may seem a bit contrived in the case of text, but this approach is generalizing what I want to do for all assets, that is committing .c files that #include extracted assets, which has the same benefits as listed above.

as to version_assets/ yeah I think it is clearer with it where it's pulling from, even if not technically required. but I'm open to changing it (I do value a lot ability to understand what's going on)

cadmic: I think it would be pretty understandable if the assets were only built out of `assets/$(VERSION)`, the extra indirection is more complex.

Could we revisit the modding issue later? With no indirection, this PR seems compatible with the current modding workflow, which IIUC is tracking custom assets in a separate folder and cp-ing them into `assets/` during the build. Perhaps later if we deem a `version_assets` path component is necessary we can change it then

Dragorn421: > Could we revisit the modding issue later?

no, that's the main thing this build system change addresses

cadmic: Oh, I thought the primary reason was to support multiversion assets for building matching ROMs. Could we separate the multiversion assets vs modding issues then, and make a separate PR for versioning the assets dir?

Dragorn421: I don't see the point given both involve directory structure layout, may as well resolve both at once rather than artificially one after the other

cadmic: Because I agree with the multiversion assets approach (split asset directories by version) but not yet with the modding approach (layer of indirection with #include). So splitting them is a way for us to make progress without stalling everything

Dragorn421:
For the purpose of clarification

So here is the layout proposed in this PR, with text only
```
assets/
  _extracted/  # everything besides assets/_extracted/** is committed
    gc-eu-mq/
      version_assets/
        text/
          message_data.h
          message_data_staff.h
    gc-eu-mq-dbg/
      version_assets/
        text/
          message_data.h
          message_data_staff.h
  text/
    *.c -> assets/text/message_data(.enc).h, assets/text/message_data_staff(.enc).h
    message_data.h -> version_assets/text/message_data.h
    message_data_staff.h -> version_assets/text/message_data_staff.h
```

it would generalize to other assets, using the `gameplay_keep` object as an example, as
```
assets/
  _extracted/  # everything besides assets/_extracted/** is committed
    gc-eu-mq/
      version_assets/
        objects/
          gameplay_keep/
            hilite_1.rgba16.png
            gLinkPauseChildJointTable.inc.c
        text/
          message_data.h
          message_data_staff.h
    gc-eu-mq-dbg/
      version_assets/
        text/
          message_data.h
          message_data_staff.h
  objects/
    gameplay_keep/
      gameplay_keep.c -> version_assets/objects/gameplay_keep/hilite_1.rgba16.inc.c, version_assets/objects/gameplay_keep/gLinkPauseChildJointTable.inc.c
      gameplay_keep.h
  text/
    *.c -> assets/text/message_data(.enc).h, assets/text/message_data_staff(.enc).h
    message_data.h -> version_assets/text/message_data.h
    message_data_staff.h -> version_assets/text/message_data_staff.h
```

assets/text/message_data.h
Dragorn421: > not yet with the modding approach (layer of indirection with #include)

I know you weren't involved then but I made that suggestion over a year ago
https://discord.com/channels/688807550715560050/688851451014938642/1051488155099987998
https://github.com/Dragorn421/decomp_assets_and_multiversion_demo
and no other option has emerged, this option has even gained some implicit approvals

Dragorn421:
Updated to "future layout v3"

------

Productive discussion on discord happened
from https://discord.com/channels/688807550715560050/688851317593997489/1207388581639880745
to https://discord.com/channels/688807550715560050/688851317593997489/1208006804127088692 + possibly ongoing

The general result is a new top level folder `extracted/` for extracted things, including assets, text and baserom segments

Include path would be like `-Ibuild/VERSION -Iextracted/VERSION`

in the following structure layouts:
`[first/part/]second/part` paths are where the file does `#include "second/part"` and CPP resolves it to `first/part/second/part`

I compiled the changes to "future layout v2" (**outdated**):

<details>

```
# v2
assets/  # note: everything under assets/ committed
  objects/
    gameplay_keep/
      gameplay_keep.c -> [extracted/VERSION/]objects/gameplay_keep/gameplay_keepVtx_00C0A0.vtx.inc , [extracted/VERSION/]objects/gameplay_keep/deku_stick.i8.inc.c
      gameplay_keep.h
  text/
    nes_message_data_static.c -> [extracted/VERSION/]text/message_data.enc.h
  xml/
    objects/
      gameplay_keep.xml
extracted/  # .gitignore'd
  VERSION/
    baserom/
      gameplay_keep
    objects/
      gameplay_keep/
        deku_stick.i8.png  # written on make setup for reference, but otherwise unused
        deku_stick.i8.inc.c  # written on make setup (does *not* depend on the png)
        gameplay_keepVtx_00C0A0.vtx.inc
    text/
      message_data.h
      message_data.enc.h  # processed from ./message_data.enc.h
build/
  VERSION/
    baserom/
      gameplay_keep.o  # objcopy'd from extracted/VERSION/baserom/gameplay_keep
```

</details>

but realized while updating this PR I oversimplified message_data.h handling. Here's the "future layout v3" that fixes that:

```
# v3
assets/  # note: everything under assets/ committed
  objects/
    gameplay_keep/
      gameplay_keep.c -> [extracted/VERSION/]objects/gameplay_keep/gameplay_keepVtx_00C0A0.vtx.inc , [extracted/VERSION/]objects/gameplay_keep/deku_stick.i8.inc.c
      gameplay_keep.h
  text/
    nes_message_data_static.c -> [build/VERSION/]assets/text/message_data.enc.h
    message_data.h -> [extracted/VERSION/]text/message_data.h
  xml/
    objects/
      gameplay_keep.xml
extracted/  # .gitignore'd
  VERSION/
    baserom/
      gameplay_keep
    objects/
      gameplay_keep/
        deku_stick.i8.png  # written on make setup for reference, but otherwise unused
        deku_stick.i8.inc.c  # written on make setup (does *not* depend on the png)
        gameplay_keepVtx_00C0A0.vtx.inc
    text/
      message_data.h
      message_data.enc.h  # processed from ./message_data.enc.h
build/
  VERSION/
    baserom/
      gameplay_keep.o  # objcopy'd from extracted/VERSION/baserom/gameplay_keep
    assets/
      text/
        message_data.enc.h  # processed from assets/text/message_data.h
```

diff from v2:

<details>

```diff
--- dirlayout_future_v2.txt     2024-02-16 12:40:29.688695211 +0100
+++ dirlayout_future_v3.txt     2024-02-16 12:40:35.436757360 +0100
@@ -1,27 +1,31 @@
-# v2
+# v3
 assets/  # note: everything under assets/ committed
   objects/
     gameplay_keep/
       gameplay_keep.c -> [extracted/VERSION/]objects/gameplay_keep/gameplay_keepVtx_00C0A0.vtx.inc , [extracted/VERSION/]objects/gameplay_keep/deku_stick.i8.inc.c
       gameplay_keep.h
   text/
-    nes_message_data_static.c -> [extracted/VERSION/]text/message_data.enc.h
+    nes_message_data_static.c -> [build/VERSION/]assets/text/message_data.enc.h
+    message_data.h -> [extracted/VERSION/]text/message_data.h
   xml/
     objects/
       gameplay_keep.xml
 extracted/  # .gitignore'd
   VERSION/
     baserom/
       gameplay_keep
     objects/
       gameplay_keep/
         deku_stick.i8.png  # written on make setup for reference, but otherwise unused
         deku_stick.i8.inc.c  # written on make setup (does *not* depend on the png)
         gameplay_keepVtx_00C0A0.vtx.inc
     text/
       message_data.h
       message_data.enc.h  # processed from ./message_data.enc.h
 build/
   VERSION/
     baserom/
       gameplay_keep.o  # objcopy'd from extracted/VERSION/baserom/gameplay_keep
+    assets/
+      text/
+        message_data.enc.h  # processed from assets/text/message_data.h
```

</details>

For reference, here is "current zeldaret/main layout v1":

<details>

`-Ibuild/gc-eu-mq-dbg`

```
assets/
  objects/  # .gitignore'd
    gameplay_keep/
      gameplay_keep.c -> ./gameplay_keepVtx_00C0A0.vtx.inc , [build/gc-eu-mq-dbg/]assets/objects/gameplay_keep/deku_stick.i8.inc.c
      gameplay_keep.h
      deku_stick.i8.png
      gameplay_keepVtx_00C0A0.vtx.inc
  text/
    message_data.h  # .gitignore'd
    nes_message_data_static.c -> [build/gc-eu-mq-dbg/]assets/text/message_data.enc.h
  xml/
    objects/
      gameplay_keep.xml
baseroms/
  gc-eu-mq-dbg/
    segments/
      gameplay_keep  # written on make setup
build/
  gc-eu-mq-dbg/
    assets/
      objects/
        gameplay_keep/
          deku_stick.i8.inc.c  # processed from assets/objects/gameplay_keep/deku_stick.i8.png on make (depends on the png)
      text/
        message_data.enc.h  # processed from assets/text/message_data.h
    baserom/
      gameplay_keep.o  # objcopy'd from baseroms/gc-eu-mq-dbg/segments/gameplay_keep on make
```

</details>

tools/msgenc.py
Dragorn421: lmk if this is better

Makefile
```diff
@@ -361,7 +362,8 @@ setup: venv
 # TODO: for now, we only extract assets from the Debug ROM
 ifeq ($(VERSION),gc-eu-mq-dbg)
 	$(PYTHON) extract_assets.py -j$(N_THREADS)
-	$(PYTHON) tools/msgdis.py --text-out assets/text/message_data.h --staff-text-out assets/text/message_data_staff.h
+	mkdir -p $(EXTRACTED_DIR)/text
```
cadmic, line 365: should this go with the `create build directories` step above instead of here? Ditto with the one for `$(BUILD_DIR)` just below

Makefile
```diff
@@ -417,8 +419,9 @@ $(BUILD_DIR)/baserom/%.o: $(BASEROM_SEGMENTS_DIR)/%
 $(BUILD_DIR)/data/%.o: data/%.s
 	$(AS) $(ASFLAGS) $< -o $@
 
-$(BUILD_DIR)/assets/text/%.enc.h: assets/text/%.h assets/text/charmap.txt
-	$(PYTHON) tools/msgenc.py assets/text/charmap.txt $< $@
+$(BUILD_DIR)/assets/text/%.enc.h: assets/text/%.h $(EXTRACTED_DIR)/text/%.h assets/text/charmap.txt
```
cadmic, line 422: The lack of symmetry between `$(BUILD_DIR)/assets/text/%.enc.h` and `$(EXTRACTED_DIR)/assets/text/%.enc.h` bugs me so I started a discord discussion about it, but I guess we don't have to do anything for now

assets/.gitignore
```diff
@@ -2,6 +2,7 @@
 *.c
 !text/*.c
 *.h
+!text/*.h
```
cadmic, line 5:
```suggestion
*.h
*.c
!text/*.h
!text/*.c
```
this is really splitting hairs here

Makefile
```diff
@@ -130,7 +131,7 @@ NM      := $(MIPS_BINUTILS_PREFIX)nm
 
 N64_EMULATOR ?= 
 
-INC := -Iinclude -Iinclude/libc -Isrc -I$(BUILD_DIR) -I.
+INC := -Iinclude -Iinclude/libc -Isrc -I$(BUILD_DIR) -I. -I$(EXTRACTED_DIR)
```
cadmic, line 134: I haven't thought deeply about it, but why this order? i.e. `.` between `BUILD_DIR` and `EXTRACTED_DIR`

Dragorn421: idk about BUILD_DIR and ., there may be a reason to be found in git blame but I don't remember

Why EXTRACTED at the end, afaict it currently doesn't matter and could be elsewhere. But it ties into the other discussion of EXTRACTED/assets
https://discord.com/channels/688807550715560050/688851317593997489/1208148581945708626

Makefile
Dragorn421: May be personal preference (out of what I believe to be good practice) but I prefer running commands as part of rules (as intended) rather than as part of the Makefile being processed and parsed

Dragorn421: I'm not dying on this hill though I can also merge it into the global mkdir

cadmic: Yeah I agree running `mkdir` when eval-ing the makefile is weird and running it as part of a rule is better. But I think it's even better to be consistent now and change them all at once (e.g. https://www.twoistoomany.com/blog/2020/07/23/make-order-only-prerequisites/ claims that order-only prerequisites would be even better than a mkdir in each rule, but I haven't looked into it)

Dragorn421:
I found a small issue specific to text processing due to the encoding step done by msgenc.py,
errors in `message_data.h` are reported as being in `message_data.enc.h` at random lines,
due to message_data.enc.h being processed from message_data.h without linemarkers (with -P cpp flag)

e.g. given `assets/text/message_data.h` :
```
#include "text/message_data.h"

DEFINE_MESSAGE(0x1111, TEXTBOX_TYPE_BLUE, TEXTBOX_POS_BOTTOM,
hi
,
""
,
""
)
```

the following error happens on make:
```
cfe: Error: build/gc-eu-mq-dbg/assets/text/message_data.enc.h, line 38939: Syntax Error
 const char _message_0x1111_nes[sizeof( hi )] = {  hi  "\x02"     }; 
 -----------------------------------------------------------------^
The token read was unexpected.
cfe: Warning 625: build/gc-eu-mq-dbg/assets/text/message_data.enc.h, line 38939: Empty declaration
 const char _message_0x1111_nes[sizeof( hi )] = {  hi  "\x02"     }; 
 ------------------------------------------------------------------^
Empty declarations are invalid in standard C.
Failed to compile file assets/text/nes_message_data_static.c.
```

Note it says `build/gc-eu-mq-dbg/assets/text/message_data.enc.h, line 38939` instead of a more helpful hypothetical `assets/text/message_data.h at line 4`

I fixed (worked around) this with https://github.com/Dragorn421/oot/commit/42e4d4ac4aff763626f2d7dd655aa5d725cb245e (not in this PR) but it's not super great. Especially hardcoding the ido path because CC is asm-processor-wrapped. And though it works, msgenc.py wasn't written with handling cpp'd input in mind (e.g. take care not to touch linemarker filename strings)

Note there is a similar issue in main anyway, errors also report `message_data.enc.h` and msgenc.py may shift lines due to replacing comments / joining continuation lines

I think I'd probably want to rewrite msgenc.py as well.

----

Handle this in this PR or just live with errors on text stuff compile being reported in message_data.enc.h for now?

fig02:
Re: message_enc stuff above
If the problem already exists in main and is not a product of the changes in this PR, seems best to handle later, especially if you are just going to rewrite the script

fig02 (approved):
My approval is from the standpoint of "I agree with the idea of these changes" and "works on my machine"

Do encourage build system-savy people to continue to review the internals of this, like is happening above

assets/text/nes_message_data_static.c
```diff
@@ -5,8 +5,15 @@
 #define DEFINE_MESSAGE(textId, type, yPos, nesMessage, gerMessage, fraMessage) \
     const char _message_##textId##_nes[sizeof(nesMessage)] = { nesMessage END };
 
-#define DEFINE_MESSAGE_FFFC
-
 #include "assets/text/message_data.enc.h"
 
-#undef DEFINE_MESSAGE_FFFC
+DEFINE_MESSAGE(0xFFFC, 0, 0,
```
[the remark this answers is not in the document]
Dragorn421: I agree, but

the reason for this change is technical: this PR cpp's `assets/message_data.h` before .enc-oding it, because otherwise files included by `assets/message_data.h` are not cpp'd

Note having `assets/message_data.h` include a .enc.h version of the extracted assets is not viable, as I want to support modding text by placing text in `assets/message_data.h` or by including files from there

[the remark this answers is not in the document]
Dragorn421: I missed a step in the explanation. The fact cpp runs "early" means the `#ifdef` are resolved early as well, and the defines around .enc includes have no effect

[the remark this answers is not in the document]
Dragorn421: Should now be addressed by essentially reverting the changes, but using a `DEFINE_MESSAGE_NES` macro instead of ifdefs (you suggested `DEFINE_MESSAGE_FFFC` but I figured NES looked better and is correct)

assets/text/message_data.h
```diff
@@ -0,0 +1 @@
+#include "text/message_data.h"
```
Thar0, line 1: In view of adding includes below this one for modding, the vanilla text debugger won't recognize them anymore since it stops search at message `FFFD`. I suggest adding messages 0xFFFC and 0xFFFD to this header instead of in the extracted header

hensldm: If we still want to extract them, could possibly just add them to a separate `message_data_nes.h` that gets included below `message_data.h`. Then custom text data could go in between.

Dragorn421: Yeah I was considering a footer include as well as an option, extracted or not

Not sure yet what to do, anyone feel free to chip in
I won't get to this immediately

fig02: I think including them in the committed header works fine

cadmic (approved):
I don't have an opinion on message 0xFFFC, but the build system and structure look good to me

assets/text/message_data.h
Dragorn421: d75522b243d1f43171fc5f2b574cc976b85c7f43 Move messages 0xFFFC, 0xFFFD to committed message_data.h

and explanation comment

fig02 (approved):

Thar0 (approved):

hensldm (approved):

PR #1738, opened 2024-02-08
Match OnePointCutscene_Init by inlining static data

cadmic:
OnePointCutscene_Init had some local regalloc issues that could be fixed by moving static data to the scope where it's used, so I did this for all the data and it matched with no further changes. After this there didn't seem to be much point to z_onepointdemo_data.inc.c, so I inlined the rest of the file too.

Note that `} break;` did not match in some places, so I used `break; }` everywhere.


fig02 (approved):
I hate it, but if we gotta do it then we gotta do it

hensldm (approved):

PR #1740, opened 2024-02-09
Match retail PreRender_DivotFilter

cadmic:
This was tricky because the compiler was saving s0 and s1 but not actually using them for anything. According to the [-O2 decompilation guide](https://hackmd.io/vPmcgdaFSlq4R2mfkq4bJg#if----nothing--), this can happen if variables are mentioned in the conditions of empty if statements that are later optimized away. So, I tried adding more temps in the condition.

This makes me think that the one true REG debug macro was something like this:
```c
#if OOT_DEBUG
#define HREG_DEBUG(i) HREG(i)
#else
#define HREG_DEBUG(i) ((void)i, 0)
#endif 
```

### DIFF
--- a/src/code/PreRender.c
+++ b/src/code/PreRender.c
@@ -681,6 +681,14 @@ void PreRender_AntiAliasFilter(PreRender* this, s32 x, s32 y) {
     (((a2) >= (a1)) ? (((a3) >= (a2)) ? (a2) : (((a1) >= (a3)) ? (a1) : (a3))) \
                     : (((a2) >= (a3)) ? (a2) : (((a3) >= (a1)) ? (a1) : (a3))))
 
+#if OOT_DEBUG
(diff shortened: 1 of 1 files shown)


cadmic:
Well, I tried `((void)0, default_value)` for the OREG stuff in z_camera.c and it didn't work as well, so maybe not the one true macro just yet

fig02 (approved):
Ive also been thinking that something along these lines will be needed for a debug reg macro. I believe something like it will be needed for a function in player that hasnt been matched yet. 

We do know that regs themselves are very developer-centric (OREG being for ogawa the camera developer, for example. HREG for hayakawa the systems dev.. etc) So, its unfortunately possible that individual devs had there own way of macroing out their debug regs. But hopefully we can find something consistent

Dragorn421 (approved):
Susge

PR #1741, opened 2024-02-09
Match retail code data sizes

cadmic:
Notable changes:

* Realigned padding after sFaultDrawer so that it fits in NON_MATCHING mode
* Disabled terminal colors in debug builds (for __osMalloc.c rodata)
* Removed actor names from the actor table
* For src/audio/data.c, either `D_801333EC` or `D_801333EC` could be removed so I arbitrarily chose one

### DIFF
--- a/data/fault_drawer.bss.s
+++ b/data/fault_drawer.bss.s
@@ -12,7 +12,7 @@
 .balign 16
 
 glabel sFaultDrawer
-    .space 0x3C
+    .space 0x40
 
-glabel D_8016B6BC
-    .space 0x24
+glabel D_8016B6C0
+    .space 0x20
--- a/data/fault_drawer.bss.s
+++ b/data/fault_drawer.bss.s
@@ -12,7 +12,8 @@
 .balign 16
 
 glabel sFaultDrawer
-    .space 0x40
+    .space 0x3C
+    .space 0x04 # padding
 
 glabel D_8016B6C0
     .space 0x20
--- a/src/code/z_actor_dlftbls.c
+++ b/src/code/z_actor_dlftbls.c
@@ -41,19 +41,18 @@
 
 #else
 
-#define DEFINE_ACTOR(name, _1, allocType, nameString) \
-    { (uintptr_t)_ovl_##name##SegmentRomStart,        \
-      (uintptr_t)_ovl_##name##SegmentRomEnd,          \
-      _ovl_##name##SegmentStart,                      \
-      _ovl_##name##SegmentEnd,                        \
-      NULL,                                           \
-      &name##_InitVars,                               \
-      NULL,                                           \
-      allocType,                                      \
+#define DEFINE_ACTOR(name, _1, allocType, _3)  \
+    { (uintptr_t)_ovl_##name##SegmentRomStart, \
+      (uintptr_t)_ovl_##name##SegmentRomEnd,   \
+      _ovl_##name##SegmentStart,               \
+      _ovl_##name##SegmentEnd,                 \
+      NULL,                                    \
+      &name##_InitVars,                        \
+      NULL,                                    \
+      allocType,                               \
       0 },
 
-#define DEFINE_ACTOR_INTERNAL(name, _1, allocType, nameString) \
-    { 0, 0, NULL, NULL, NULL, &name##_InitVars, NULL, allocType, 0 },
+#define DEFINE_ACTOR_INTERNAL(name, _1, allocType, _3) { 0, 0, NULL, NULL, NULL, &name##_InitVars, NULL, allocType, 0 },
 
 #endif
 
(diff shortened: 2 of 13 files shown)


src/code/z_play.c
```diff
@@ -451,14 +458,16 @@ void Play_Init(GameState* thisx) {
     AnimationContext_Update(this, &this->animationCtx);
     gSaveContext.respawnFlag = 0;
 
-    if (OOT_DEBUG && R_USE_DEBUG_CUTSCENE) {
+#if OOT_DEBUG
+    if (R_USE_DEBUG_CUTSCENE) {
         gDebugCutsceneScript = sDebugCutsceneScriptBuf;
         PRINTF("\nkawauso_data=[%x]", gDebugCutsceneScript);
 
         // This hardcoded ROM address extends past the end of the ROM file.
```
fig02, line 466: can `sDebugCutsceneScriptBuf` be brought down as in function static to remove an ifdef? Did not check closely if the data ordering the file works out for it

fig02: er wait it gets used somewhere else also, nevermind

fig02 (approved):

src/code/z_actor_dlftbls.c
```diff
@@ -37,6 +39,24 @@
 #define DEFINE_ACTOR_INTERNAL(name, _1, allocType, nameString) \
     { 0, 0, NULL, NULL, NULL, &name##_InitVars, nameString, allocType, 0 },
 
+#else
+
+#define DEFINE_ACTOR(name, _1, allocType, nameString) \
+    { (uintptr_t)_ovl_##name##SegmentRomStart,        \
```
fig02, line 45: to follow the convention i suppose nameString should become _3 because its now an unused arg

src/code/z_play.c
cadmic: It's used only in this block I think, I tried it and it worked

fig02: oh, i somehow misread that when going to check myself lmao. cool.

fig02 (approved):

src/code/z_actor_dlftbls.c
```diff
@@ -37,6 +39,23 @@
 #define DEFINE_ACTOR_INTERNAL(name, _1, allocType, nameString) \
     { 0, 0, NULL, NULL, NULL, &name##_InitVars, nameString, allocType, 0 },
 
+#else
+
+#define DEFINE_ACTOR(name, _1, allocType, _3)  \
```
Dragorn421, line 44: could you add a comment like
`// non-debug sets ActorOverlay.name to NULL`
or similar

Dragorn421 (commented):
can't say I expected terminal.h of all things to have version differences

can you elaborate on what the fault_drawer bss changes do? could it (technically) just be
```c
#if OOT_DEBUG
char D_8016B6BC[0x24];
#else
static char sUnknownPadding[0x20];
#endif
```
, would that make more sense

cadmic:
Right now the fault_drawer padding doesn't match the debug ROM either. If we used the `NON_MATCHING` version there this is what the map file looks like:

```
build/gc-eu-mq-dbg/src/code/fault_drawer.o(.bss)
 .bss           0x000000008016b680       0x70 build/gc-eu-mq-dbg/src/code/fault_drawer.o
                0x000000008016b680                sFaultDrawer
                0x000000008016b6c0                D_8016B6BC
                0x000000008016b6f0                . = ALIGN (0x10)
```

So D_8016B6BC should really be D_8016B6C0 since that's its address in the debug version, and the size can't be 0x24 because then fault_drawer BSS is too big.

data/fault_drawer.bss.s
```diff
@@ -12,7 +12,7 @@
 .balign 16
 
 glabel sFaultDrawer
-    .space 0x3C
+    .space 0x40
 
-glabel D_8016B6BC
-    .space 0x24
+glabel D_8016B6C0
+    .space 0x20
```
Dragorn421, line 18: one of the things I don't get is `sFaultDrawer` is 0x3C bytes so why change to 0x40

the changes seem to go from bss size = 0x3C+0x24=0x60 to 0x40+0x20=0x60 ? am I missing the assembler aligning the symbols somehow?

cadmic: I changed this to be in sync with the code, which now reads:

```
FaultDrawer sFaultDrawer;
char D_8016B6C0[0x20];
```

where the distance between `sFaultDrawer` and `D_8016B6C0` is 0x40 bytes (0x3C bytes for the struct and 0x4 bytes of padding). What about something like this?

```
glabel sFaultDrawer
    .space 0x3C
    .space 0x4  # padding

glabel D_8016B6C0
    .space 0x20
```

Dragorn421: oh yeah that would be clearer
not sure why there's padding there :thinking:

cadmic: yeah, me neither

dannygraves82:
Trying to fight the long red monster in the belly and have to use the boomerang, but i cant seem to aim it. help....

hensldm (approved):

PR #1742, opened 2024-02-13
Use Romfile in place of vromStart and vromEnd in structs

hensldm:
I had casually brought this up in mm-decomp. I liked the way it turned out so thought I would PR it to see what other people's thoughts are.

### DIFF
--- a/include/z64dma.h
+++ b/include/z64dma.h
@@ -32,7 +32,7 @@ typedef struct {
 #define ROM_FILE_EMPTY(name) \
     { (uintptr_t)_##name##SegmentRomStart, (uintptr_t)_##name##SegmentRomStart }
 #define ROM_FILE_UNSET \
-    { 0 }
+    { 0, 0 }
 
 extern DmaEntry gDmaDataTable[];
 
--- a/include/z64dma.h
+++ b/include/z64dma.h
@@ -5,6 +5,18 @@
 #include "alignment.h"
 
 typedef struct {
+    /* 0x00 */ uintptr_t vromStart;
+    /* 0x04 */ uintptr_t vromEnd;
+} RomFile; // size = 0x8
+
+#define ROM_FILE(name) \
+    { (uintptr_t)_##name##SegmentRomStart, (uintptr_t)_##name##SegmentRomEnd }
+#define ROM_FILE_EMPTY(name) \
+    { (uintptr_t)_##name##SegmentRomStart, (uintptr_t)_##name##SegmentRomStart }
+#define ROM_FILE_UNSET \
+    { 0, 0 }
+
+typedef struct {
     /* 0x00 */ uintptr_t    vromAddr; // VROM address (source)
     /* 0x04 */ void*        dramAddr; // DRAM address (destination)
     /* 0x08 */ size_t       size;     // File Transfer size
@@ -16,23 +28,10 @@ typedef struct {
 } DmaRequest; // size = 0x20
 
 typedef struct {
-    /* 0x00 */ uintptr_t vromStart;
-    /* 0x04 */ uintptr_t vromEnd;
+    /* 0x00 */ RomFile file;
     /* 0x08 */ uintptr_t romStart;
     /* 0x0C */ uintptr_t romEnd;
 } DmaEntry;
-
-typedef struct {
-    /* 0x00 */ uintptr_t vromStart;
-    /* 0x04 */ uintptr_t vromEnd;
-} RomFile; // size = 0x8
-
-#define ROM_FILE(name) \
-    { (uintptr_t)_##name##SegmentRomStart, (uintptr_t)_##name##SegmentRomEnd }
-#define ROM_FILE_EMPTY(name) \
-    { (uintptr_t)_##name##SegmentRomStart, (uintptr_t)_##name##SegmentRomStart }
-#define ROM_FILE_UNSET \
-    { 0, 0 }
 
 extern DmaEntry gDmaDataTable[];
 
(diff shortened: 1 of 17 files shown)


include/z64dma.h
```diff
@@ -4,6 +4,18 @@
 #include "ultra64.h"
 #include "alignment.h"
 
+typedef struct {
+    /* 0x00 */ uintptr_t vromStart;
+    /* 0x04 */ uintptr_t vromEnd;
+} RomFile; // size = 0x8
```
fig02, line 10: Do wonder if in all cases if we should call the variable `rom` and then change the variable names to just `start` and `end`
So in context it would look like `thing->rom.start`

I guess the way it is right now is fine, especially if you really value the specificity of "vrom", but it does feel redundant to say in the variable name now. More thoughts needed

cadmic: FWIW `RomFile rom` doesn't seem right to me, I don't think I would call a start and end address a "rom" in conversation, it's more like a section of a rom

cadmic (approved):

include/z64dma.h
fig02: the type is still RomFile which tells you what it is, but `rom.start` reads exactly as you would use it in conversation "the rom start address"
but yeah idk

include/z64dma.h
```diff
@@ -16,24 +28,11 @@ typedef struct {
 } DmaRequest; // size = 0x20
 
 typedef struct {
-    /* 0x00 */ uintptr_t vromStart;
-    /* 0x04 */ uintptr_t vromEnd;
+    /* 0x00 */ RomFile file;
     /* 0x08 */ uintptr_t romStart;
     /* 0x0C */ uintptr_t romEnd;
 } DmaEntry;
```
Dragorn421, line 34: interesting, the conservative in me doesn't like touching such a fundamental struct but I appreciate that it links together the RomFile (vrom) references and the physical rom

src/code/z_actor_dlftbls.c
```diff
@@ -24,20 +24,26 @@
...
+    },
+
+#define DEFINE_ACTOR_INTERNAL(name, _1, allocType, nameString)                        \
+    {                                                                                 \
+        ROM_FILE_UNSET, NULL, NULL, NULL, &name##_InitVars, nameString, allocType, 0, \
+    },
+
+#define DEFINE_ACTOR_UNSET(_0)                              \
+    {                                                       \
+        ROM_FILE_UNSET, NULL, NULL, NULL, NULL, '\0', 0, 0, \
```
Dragorn421, line 45:
```suggestion
        ROM_FILE_UNSET, NULL, NULL, NULL, NULL, NULL, 0, 0, \
```
but I'd just keep the existing 0 initializer `{0}` tbh

include/z64dma.h
Dragorn421: I don't find there's anything superfluous in `file.vromStart`/End, and I think it's better to "abstract" that a file is located by vrom by naming the fields after that rather than the file member

Dragorn421, line 10: Given this struct is used a lot outside of z64dma.h context I think it would be a good time to also move RomFile stuff to its own header like `romfile.h`

hensldm: I had thought about doing something like `RomFile vrom` with `vrom.start` and `vrom.end`. Biggest hesitation for me was the few things that just use `RomFile` directly. For example the ObjectTable would have accesses like `gObjectTable[i].end - gObjectTable[i].start`, which doesn't seem right. I would want to wrap it in another struct to have `vrom` as the struct member. Not sure if that was worth it.

hensldm: Is it? Don't most use it to make a Dma Request?
(Though I guess most go through wrapper functions)

fig02: yeah does not seem worth it. its fine as is

fig02: I think the close relationship with dma makes sense, i dunno

Dragorn421: Yeah anything using RomFile will probably also use other z64dma.h stuff
but not everything seeing the RomFile type (for the purpose of struct layout) will use z64dma.h stuff

this is with header splitting in mind, ideally changing a header causes minimal rebuilding stuff

AngheloAlf:
Given that `RomFile` doesn't really describe rom addresses of files but it instead it describes rom addresses of segments I think we should consider renaming this struct to `RomSegment` instead

hensldm:
> Given that `RomFile` doesn't really describe rom addresses of files but it instead it describes rom addresses of segments I think we should consider renaming this struct to `RomSegment` instead

I would be fine with that. What are other people's thoughts?

Dragorn421:
I agree `RomSegment` > `RomFile`

fig02:
I get why its technically correct, but I like it less. its nicer to think of them as files, which for example the DMA system still refers to them as

include/z64dma.h
mzxrules: I think what I don't like about it is that when it's used here, it's obvious that the RomFile use case is too narrow. `vromStart`/`vromEnd` should be the same type of thing as `romStart`/`romEnd`

mzxrules:
Personally I'd prefer it if they were called rom segments, a lot of different parts of decomp already refers to them as segments this and it would make z_std_dma a lot clearer if we switched to segment name, source file name.

fig02:
I guess id be fine with it if its made consistent, sure. Seems to me like itd be best to go with this as-is and then file -> segment can happen in another pr

Dragorn421 (approved):

PR #1743, opened 2024-02-13
Match the rest of retail BG actors

cadmic:
Most interesting is BgMizuBwall_Init, which is another place where the `break;` must be inside the curly braces rather than outside it


fig02 (approved):

src/overlays/actors/ovl_Bg_Mizu_Bwall/z_bg_mizu_bwall.c
```diff
@@ -217,7 +220,15 @@ void BgMizuBwall_Init(Actor* thisx, PlayState* play) {
                 }
             }
             break;
-        case MIZUBWALL_RUTO_ROOM:
+        }
+        case MIZUBWALL_RUTO_ROOM: {
```
Dragorn421, line 224:
```c
} break;

case ...
```

cadmic: See PR description, break outside of the block didn't work in a couple places in this function so I changed them all for consistency

Dragorn421: Oh interesting
Sorry didn't pay attention to the description

Dragorn421 (approved):

PR #1744, opened 2024-02-14
Update `mapfile_parser`

AngheloAlf:
Updates `mapfile_parser` to version 2.X.

The most note worthy thing this version has is the mapfile parsing was rewritten in Rust, making it a lot faster than the old pure-python version.

Even if this is a new major version of the lib, none of interface directly used by the scripts changed, so this is a drop-in replacement for this repo.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,6 +14,6 @@ pycparser
 toml
 
 # tools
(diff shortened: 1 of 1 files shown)


fig02 (approved):

Dragorn421 (approved):
ez

PR #1746, opened 2024-02-17
Fixup permuter_settings

Dragorn421:
- Fix: make `IDO_PRINTF_WORKAROUND` not consider PERMUTER since the permuter preserves the PRINTF macros for IDO to process
- Misc fixup of permuter_settings.toml


cadmic (approved):

zedzevon:
I have 1 heart piece left but I can't get it. It's the roof guy in kakario village. I never did anything before but he won't give me it. Why is that?

PR #1747, opened 2024-02-17
Suggest explorer.exe in README.md

cadmic:


README.md
```diff
@@ -102,7 +102,8 @@ cd oot
 
 #### 3. Prepare a base ROM
 
-Place a copy of the Master Quest (Debug) ROM inside the `baseroms/gc-eu-mq-dbg/` folder.
+Place a copy of the Master Quest (Debug) ROM inside the `baseroms/gc-eu-mq-dbg/` folder. In you are under WSL, you can
```
Dragorn421, line 105:
```suggestion
Place a copy of the Master Quest (Debug) ROM inside the `baseroms/gc-eu-mq-dbg/` folder. If you are under WSL, you can
```
would also put the sentence on the next line for diff reasons
(it's still merged into the same paragraph)

Dragorn421 (approved):

Issue #1719, opened 2024-02-05
Speed up builds: bypass asm-processor


Dragorn421:
I've been looking into how to speed up build times. I found that asm-processor adds non negligible overhead (due to python startup and imports, the asm-processor code itself runs in negligible time)

-----------

I wrote a C program "crwasmp" https://github.com/Dragorn421/DragoStuff/tree/crwasmp that performs what we currently use asm-processor for, that is

- replace floats by the IEEE-754 binary representation ("in hex") in CutsceneData
- convert encoding from utf-8 to euc-jp
- run ido on the processed source

Note: oot is not using `GLOBAL_ASM`, the main feature of asm-processor, since gc-eu-mq-dbg was matched

it saves ~25% of time for building src/
(from built, `rm -rf build/gc-eu-mq-dbg/src && time make`)
in details, on my machine:
`make -j1`: 71s -> 54s (-24%)
`make -j4`: 27s -> 20s (-26%)
`make -j8`: 22s -> 17s (-23%)

This shows we have a lot of time to gain from not using asm-processor in the build process

---------

An option "option-crwasmp" is to introduce above program "crwasmp" into the repo, replacing `$(PYTHON) tools/asm_processor/build.py` in the Makefile with it.

Another option "option-justusehex" would be to have ZAPD write CutsceneData floats directly as IEEE-754 (possibly with the float value as a comment like `0x42142142 /* 421.0f */`).
Then all that's left to do before calling ido is the utf-8 to euc-jp encoding conversion, which is easily done with the `iconv` program. As a bonus this allows removing `#pragma asmproc recurse`.

Either way we would still keep asm-processor in the repo, which could selectively be used on specific files (cherry-pick in the makefile), if we do end up reintroducing `GLOBAL_ASM`s in the repo.

The two options are outlined in more details below

option-crwasmp has been tested, it's what I have right now.
There's no reason to believe option-justusehex would be slower (should be (marginally) faster, even)

## option-crwasmp

- Put crwasmp in the repo https://github.com/Dragorn421/DragoStuff/tree/crwasmp
  - Copy source into zeldaret/oot
  - Add dependencies to readme (apt)
  - Build crwasmp as part of building tools
- Modify the makefile to use crwasmp in place of asm-processor: https://github.com/Dragorn421/oot/commit/6b315426d1ce970f3b4db449f9da10efa63a74d1

## option-justusehex

[...]

PR #1824, opened 2024-02-19
No longer use asm-processor

Dragorn421:
Speeds up builds noticeably. `time make RUN_CC_CHECK=0` is 20-25% faster for me

asm-processor is still there ready to be used if we ever have more `GLOBAL_ASM`s to use

Addresses #1719

Changes:
- update ZAPD (for https://github.com/NEstelami/ZAPD/pull/317 )
- ZAPD is passed `--cs-float both` (`extract_assets.py`)
- cutscene floats are now extracted as `CS_FLOAT(0x42700000, 60.0f)`
- `z64cutscene_commands.h` :
```c
#ifdef __GNUC__
#define CS_FLOAT(ieee754bin, f) (f)
#define CMD_F(a) {.f = (a)}
#else
#define CS_FLOAT(ieee754bin, f) (ieee754bin)
#define CMD_F(a) {(a)}
#endif
```
- update csdis.py
- add csdis_re.py to reextract committed CutsceneData (using csdis.py)
- add reencode.sh to handle writing EUC-JP source for ido to process (could be done in the makefile directly but would look quite ugly. it is unfortunate ido cannot read source from stdin)
- add bss padding in z_collision_check.c (doesn't match without)


tools/reencode.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -e
+
+srcfile="${@: -1}"
+
+tempfile=`mktemp --tmpdir oot_XXXXXXXX.c`
```
cadmic, line 7: Does this litter /tmp with files? It should probably remove the tempfile with a `trap` on exit

include/z64cutscene_commands.h
```diff
@@ -22,8 +22,10 @@
  */
 
 #ifdef __GNUC__
+#define CS_FLOAT(ieee754bin, f) (f)
 #define CMD_F(a) {.f = (a)}
```
fig02, line 26: Not the biggest fan of this macro taking two numbers, its pretty confusing to see two numbers being passed into a macro and only one to be used depending on the context. I tbh would prefer if hex was used in all cases with comments to describe the float representation

I guess this is a solution to the "portability" problem that was mentioned before?

Dragorn421: Yeah we have the option to just write the hex as well rather than the macro

Personally I don't mind either

Makefile
```diff
@@ -308,11 +308,11 @@ $(BUILD_DIR)/src/libultra/rmon/%.o: CC := $(CC_OLD)
 $(BUILD_DIR)/src/code/jpegutils.o: CC := $(CC_OLD)
 $(BUILD_DIR)/src/code/jpegdecoder.o: CC := $(CC_OLD)
 
-$(BUILD_DIR)/src/boot/%.o: CC := $(PYTHON) tools/asm_processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
-$(BUILD_DIR)/src/code/%.o: CC := $(PYTHON) tools/asm_processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
-$(BUILD_DIR)/src/overlays/%.o: CC := $(PYTHON) tools/asm_processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
+# For using asm_processor on some files:
+#$(BUILD_DIR)/.../%.o: CC := $(PYTHON) tools/asm_processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
+
```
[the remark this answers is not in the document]
Dragorn421: I didn't touch `DEP_FILES` which is part of asmproc Makefile logic, I don't want to completely pluck it out so I'd rather document explicitly asm_processor can be used

tools/csdis_re.py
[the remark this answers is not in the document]
Dragorn421: we would definitely use it again if (when) constant names used in cutscene data change

include/z64cutscene_commands.h
Dragorn421: changed to "hex"

Makefile
```diff
@@ -308,11 +308,11 @@ $(BUILD_DIR)/src/libultra/rmon/%.o: CC := $(CC_OLD)
...
+$(BUILD_DIR)/%.o: CC := tools/reencode.sh $(CC)
```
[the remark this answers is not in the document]
Dragorn421: good remark

It does break, I think because `reencode.sh` expects the input file to be the last argument,
but the permuter uses `[compile command (including reencode.sh)] input -o output`
instead of I-think-would-work `[compile command (including reencode.sh)] -o output input`

<details>

```
$ ../decomp-permuter/import.py src/code/z_play.c expected/build/gc-eu-mq/functions/src/code/z_play/Play_Update.s VERSION=gc-eu-mq
Compiler type: ido
Function name: Play_Update
Compiler: tools/reencode.sh tools/ido_recomp/linux/7.1/cc -c -DNON_MATCHING -DNDEBUG -DOOT_DEBUG=0 -G 0 -non_shared -fullwarn -verbose -Xcpluscomm -Iinclude -Iinclude/libc -Isrc -Ibuild/gc-eu-mq -I. -Wab,-r4300_mul -woff 516,609,649,838,712 -mips2 -O2 -g3 {input} -o {output}
Assembler: mips-linux-gnu-as -march=vr4300 -mabi=32 {input} -o {output}
Preserving macros: gSPDisplayList, gSPSegment, gSPEndDisplayList, SQ, PRINTF, LOG, SET_FULLSCREEN_VIEWPORT, POLY_OPA_DISP, POLY_XLU_DISP, OVERLAY_DISP, OPEN_DISPS, CLOSE_DISPS, GRAPH_ALLOC. Use --preserve-macros='<regex>' to override.
/tmp/permuteru46fgu8v.c:
/tmp/oot_YZxuMAhC.c:

Done. Imported into nonmatchings/Play_Update
```

```
$ ../decomp-permuter/permuter.py nonmatchings/Play_Update/
Loading...
nonmatchings/Play_Update/base.c (Play_Update)
No perm macros found. Defaulting to randomization.
Will try 1 different base sources.
/tmp/permuterl7z0lk2a.c:
/tmp/oot_Ss8beWU8.c:
Traceback (most recent call last):
  File "/home/dragorn421/Documents/oot/../decomp-permuter/permuter.py", line 5, in <module>
    main()
  File "/home/dragorn421/Documents/decomp-permuter/src/main.py", line 771, in main
    run(options)
  File "/home/dragorn421/Documents/decomp-permuter/src/main.py", line 283, in run
    return run_inner(options, heartbeat)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dragorn421/Documents/decomp-permuter/src/main.py", line 360, in run_inner
    permuter = Permuter(
               ^^^^^^^^^
  File "/home/dragorn421/Documents/decomp-permuter/src/permuter.py", line 134, in __init__
    ) = self._create_and_score_base()
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dragorn421/Documents/decomp-permuter/src/permuter.py", line 157, in _create_and_score_base
    base_result = base_cand.score(self.scorer, o_file)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dragorn421/Documents/decomp-permuter/src/candidate.py", line 97, in score
    self.score_value, self.score_hash = scorer.score(o_file)
                                        ^^^^^^^^^^^^^^^^^^^^
  File "/home/dragorn421/Documents/decomp-permuter/src/scorer.py", line 46, in score
    objdump_output, cand_seq = self._objdump(cand_o)
                               ^^^^^^^^^^^^^^^^^^^^^
  File "/home/dragorn421/Documents/decomp-permuter/src/scorer.py", line 39, in _objdump
    lines = objdump(o_file, self.arch, stack_differences=self.stack_differences)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dragorn421/Documents/decomp-permuter/src/objdump.py", line 443, in objdump
    output = subprocess.check_output([executable] + arch.arguments + [o_filename])
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 466, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['mips-linux-gnu-objdump', '-drz', '-m', 'mips:4300', '/tmp/permuter4xrrvpby.o']' returned non-zero exit status 1.

$ ls /tmp/permuter4xrrvpby.o
ls: cannot access '/tmp/permuter4xrrvpby.o': No such file or directory
```

</details>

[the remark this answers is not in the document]
Dragorn421: actually no asm-processor also expects the last argument to be input so idk what's going on
https://github.com/simonlindholm/asm-processor/blob/fed1e3ddb9967bd427f4599b1b8ededa296bd9f9/build.py#L31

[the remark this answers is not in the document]
Dragorn421: Fixed by e9b52b493

Dragorn421:
Ready for more review

Makefile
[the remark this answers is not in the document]
cadmic: Sorry, why don't you want to pluck out the asmproc.d stuff too? Are you expecting people to re-enable asm_processor locally only for dependency tracking?

I'm not sure I like the idea of only "half-heartedly" removing asm-processor, ideally we would delete it from the repo entirely to try to make things simpler. I personally don't think we'll need `GLOBAL_ASM` ever again, but if that's still an issue, maybe it's better to wait on this change than making the build system more complicated.

tools/reencode.sh
cadmic: the tempfile should be removed on compile errors too IMO. Something like `trap "rm -f $tempfile" EXIT` instead of only removing it if we reach the end of the script

src/code/z_collision_check.c
```diff
@@ -1408,6 +1408,7 @@ void CollisionCheck_NoBlood(PlayState* play, Collider* collider, Vec3f* v) {
  * Used by collider types HIT0 and HIT8.
  */
 void CollisionCheck_BlueBlood(PlayState* play, Collider* collider, Vec3f* v) {
+    static s32 pad;
```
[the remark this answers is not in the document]
cadmic: Why `sBssDummyNeg1` instead of renumbering? Keeping it consistent helps with debugging bss ordering, it's surprising that `sBssDummy0` is not in fact the first dummy variable

include/z64cutscene_commands.h
fig02: do the zapd cutscene files have the float representation as a comment? I see the non-zapd ones dont. feel like that is necessary at least for the the cam commands where the floats need to be intelligible

Dragorn421: They don't
I can probably PR another "hex-comment" option to ZAPD, sounds simple enough for me to do

Makefile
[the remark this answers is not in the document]
Dragorn421: This change is a noticeable performance boost for IDO builds so I don't want to post-pone it

I do think we could see `GLOBAL_ASM`s again in the repo, for example if/when I figure out how to build matching in a non-intrusive way maybe a single function in a file would resist matching and would be `global_asm`'d

maybe I refactor now which files to run asm-processor on? i.e. `ASM_PROCESSOR_O_FILES :=`

src/code/z_collision_check.c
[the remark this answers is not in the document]
Dragorn421: because lazy and preserve diff and blame (especially for something as sketchy as this)

there may even be a better solution to this bss stuff now that it seemingly changed (no clue), but it's complete wizardry to me

I'll renumber them from 0

include/z64cutscene_commands.h
fig02: alright
I feel like we may have not had enough input on this-- I like just hex, but the macro with dual input might be a good compromise for those who feel strongly about them being in float form

Dragorn421: FTR I don't mind changing back

It's not particularly difficult either:
- the macro definitions
- the zapd argument in `extract_assets.py`
- the similar hardcoded argument in `csdis.py`
- run `csdis_re.py`

tools/reencode.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# The last argument, the input source file to be compiled
+srcfile="${@: -1}"
+
+# Create a temporary file, and remove it on script exit
+tempfile=`mktemp --tmpdir oot_XXXXXXXX.c`
+trap "rm -f $tempfile" EXIT
+
+# Re-encode from UTF-8 to EUC-JP
+iconv -f UTF-8 -t EUC-JP -o "$tempfile" "$srcfile"
```
cadmic, line 18: sorry, this only just occurred to me when thinking about the block numbers script. Should this script add a line directive the file too for better error messages?

Dragorn421: Yeah I also only now realized that. working on it right now

Dragorn421: done but it broke matching gc-eu-mq-dbg. Reverting the bss changes in zcolchk seems to have fixed zcolchk but there's more diffs bruh. I pushed what I have rn

include/z64cutscene_commands.h
Dragorn421: changed back to "both" (`CS_FLOAT(hex, f)` macro) after having too much trouble trying to have a tool encode floats on build like asm-processor did

fig02 (approved):

cadmic (approved):

Thar0 (approved):

PR #1825, opened 2024-02-20
Match retail En actors V-Z

cadmic:
`s32 pad;`

### DIFF
--- a/src/overlays/actors/ovl_En_Xc/z_en_xc.c
+++ b/src/overlays/actors/ovl_En_Xc/z_en_xc.c
@@ -349,7 +349,6 @@ s32 EnXc_BoleroCS(EnXc* this, PlayState* play) {
 }
 
 void EnXc_SetupSerenadeAction(EnXc* this, PlayState* play) {
-    // Player is adult and has not learned Serenade and, if in debug mode, does not have iron boots
     if (!(CHECK_OWNED_EQUIP(EQUIP_TYPE_BOOTS, EQUIP_INV_BOOTS_IRON) && OOT_DEBUG) && !GET_EVENTCHKINF(EVENTCHKINF_52) &&
         LINK_IS_ADULT) {
         s32 pad;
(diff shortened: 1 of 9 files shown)


src/overlays/actors/ovl_En_Xc/z_en_xc.c
```diff
@@ -226,6 +226,7 @@ void func_80B3C7D4(EnXc* this, s32 action1, s32 action2, s32 action3) {
     }
 }
 
+#if OOT_DEBUG
 s32 EnXc_NoCutscenePlaying(PlayState* play) {
     if (play->csCtx.state == 0) {
         return true;
```
fig02, line 232: this should be using the cs state enum. yet another thing to cleanup at some point.
as usual, dont have to do this now if you dont want

src/overlays/actors/ovl_En_Xc/z_en_xc.c
```diff
@@ -280,19 +282,26 @@ void func_80B3C9EC(EnXc* this) {
...
         this->action = SHEIK_ACTION_INIT;
     } else {
         Actor_Kill(&this->actor);
     }
 }
 
 s32 EnXc_MinuetCS(EnXc* this, PlayState* play) {
+    Player* player;
+    f32 z;
+
```
fig02, line 296: I would be a little bit more descriptive like `playerPosZ`. I understand its used right after its assignment but still think its best

src/overlays/actors/ovl_En_Xc/z_en_xc.c
```diff
@@ -336,9 +349,11 @@ s32 EnXc_BoleroCS(EnXc* this, PlayState* play) {
 }
 
 void EnXc_SetupSerenadeAction(EnXc* this, PlayState* play) {
-    // Player is adult and does not have iron boots and has not learned Serenade
-    if (!CHECK_OWNED_EQUIP(EQUIP_TYPE_BOOTS, EQUIP_INV_BOOTS_IRON) && !GET_EVENTCHKINF(EVENTCHKINF_52) &&
+    // Player is adult and has not learned Serenade and, if in debug mode, does not have iron boots
```
fig02, line 352: comment can prob be removed, this was from a time before `CHECK_OWNED_EQUIP` and `GET_EVENTCHKINF` even existed (the flag will be named at some point)

hensldm (approved):

PR #1889, opened 2024-02-23
Match En actors D-G

cadmic:
Highlights:
* Ugly `#if` around `if` in z_en_elf.c
* Speed logic in z_en_goroiwa.c is pretty different in retail


fig02 (approved):

src/overlays/actors/ovl_En_Goroiwa/z_en_goroiwa.c
```diff
@@ -84,7 +84,15 @@ static ColliderJntSphInit sJntSphInit = {
...
 
-static f32 sUnused[] = { 10.0f, 9.2f };
+static f32 sSpeeds[] = { 10.0f, 9.2f };
+
+#if OOT_DEBUG
+#define GET_EN_GOROIWA_SPEED(this) (R_EN_GOROIWA_SPEED * 0.01f)
+#define SET_EN_GOROIWA_SPEED(speed) R_EN_GOROIWA_SPEED = speed
+#else
+#define GET_EN_GOROIWA_SPEED(this) sSpeeds[(this)->isInKokiri]
+#define SET_EN_GOROIWA_SPEED(speed)
```
Dragorn421, line 94: Not a huge fan of the set macro between expanding to an assignment, or expanding to nothing
I'd rather see the assignment ifdefed out in code

Also the GET_ macro looks too much like the macros for getting data from actor params, but I don't have a suggestion

src/overlays/actors/ovl_En_Elf/z_en_elf.c
```diff
@@ -1379,9 +1379,15 @@ void func_80A053F0(Actor* thisx, PlayState* play) {
 
     if (player->naviTextId == 0) {
         if (player->unk_664 == NULL) {
+#if OOT_DEBUG
             if (((gSaveContext.save.info.playerData.naviTimer >= 600) &&
                  (gSaveContext.save.info.playerData.naviTimer <= 3000)) ||
-                (nREG(89) != 0)) {
+                (nREG(89) != 0))
```
Dragorn421, line 1385: Idk if this has been discussed before but is

```c
if (things
#if OOT_DEBUG
|| debugCheck
#endif
) {
```

not an option

cadmic: It is an option and I'm not sure it's been discussed before, would you prefer that?

src/overlays/actors/ovl_En_Goroiwa/z_en_goroiwa.c
cadmic: Okay I remove the SET_ macro and left the GET_ macro alone for now

fig02: fwiw I agree about the speed macro looking the a params macro, and think it would be fine as just "SPEED" without GET. If this isnt changed now, I assume it would be at some point in the future when the actor+params are documented. So either now or later is fine with me

cadmic: changed to `EN_GOROIWA_SPEED(this)`

fig02 (approved):
the discussion in discord about the condition is whatever to me, cause we still plan to tinker with these debug regs at a larger scale at a later time

Dragorn421 (approved):
I have an opinion but it can be redone later

I also don't expect the topic to be relevant too often (hence can revisit later)

hensldm (approved):

PR #1890, opened 2024-02-23
Match En actors A-C, and a few more actors

cadmic:
Highlights:
* Fake `if (mtx) {}` in z_en_ba.c for regalloc


Dragorn421 (approved):
`if (1) {} CLOSE_DISPS` 📈 invest now

PR #1891, opened 2024-02-23
Filename & line number args cleanup

Thar0:
* Use int for line number args over s32/u32
* Add missing const qualifiers from filename args
* Fix gcc warning in game.c

Line numbers are printed with the format `%d` so they shouldn't really be of long type. It could be argued that a lot of s32/u32 are also printed with `%d` and we can't do a lot about this in general, but line numbers are a nice subset of args where we can do something about it.

### DIFF
--- a/src/code/game.c
+++ b/src/code/game.c
@@ -408,7 +408,7 @@ void GameState_Init(GameState* gameState, GameStateFunc init, GraphicsContext* g
     gameState->destroy = NULL;
     gameState->running = 1;
     startTime = osGetTime();
-    gameState->size = gameState->init = 0;
+    gameState->size = (u32)(gameState->init = NULL);
 
     {
         s32 requiredScopeTemp;
@@ -494,7 +494,7 @@ u32 GameState_IsRunning(GameState* gameState) {
 }
 
 #if OOT_DEBUG
-void* GameState_Alloc(GameState* gameState, size_t size, char* file, s32 line) {
+void* GameState_Alloc(GameState* gameState, size_t size, const char* file, int line) {
     void* ret;
 
     if (THA_IsCrash(&gameState->tha)) {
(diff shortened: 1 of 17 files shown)


src/code/game.c
```diff
@@ -408,7 +408,7 @@ void GameState_Init(GameState* gameState, GameStateFunc init, GraphicsContext* g
     gameState->destroy = NULL;
     gameState->running = 1;
     startTime = osGetTime();
-    gameState->size = gameState->init = 0;
+    gameState->size = (u32)(gameState->init = NULL);
```
Dragorn421, line 411: That is an interesting change, is it fixing a warning?
Tbh I'd rather have the warning, there's only so much we can do about them anyway in a matching context

Dragorn421 (approved):
I sure wish there was any semblance of proper use of printf format strings in the codebase

In general I prefer the standard C types over libultra fixed width types so I can't not like these changes

src/code/game.c
Thar0: This fixes a "makes integer from pointer without a cast" warning. My view is that whenever we can fix a warning we should, and I think making this line a little uglier is even a good thing since it makes the weirdness more apparent than if it was left looking cleaner.

fig02: if we go with this id like a comment saying that this is effectively setting both to 0. that is not obvious at all when written this way

fig02: something like `// sets 0 and NULL respectively` or something is fine with me

Dragorn421: Alternatively in similar chained-assignments-we-wish-were-not we have just the list of values like `0, NULL`

fig02: possibly, but this one is espcially weird with the casting and parenthesis that make it not even look like a chained assign at first glance. it warrants a comment more than other cases imo

Dragorn421: Sure

Thar0: Added a comment

Dragorn421 (approved):

fig02 (approved):

PR #1893, opened 2024-02-24
Match more of player for retail

cadmic:
After this there are three functions left. `Player_Action_8084411C` and `Player_Action_8084EAC0` can be matched with fake labels as a last resort, but `Play_DrawGameplay` is proving to be tricky ([decomp.me scratch](https://decomp.me/scratch/wXhmy)).


fig02 (approved):
spent a long time staring at `func_80833350` and cant believe i didnt try moving the temps

Dragorn421 (approved):

PR #1894, opened 2024-02-26
Mostly match retail bosses, part 2

cadmic:
Part 1 coming soon. I skipped BossMo_DrawCore (https://decomp.me/scratch/sLMBY)

Highlights:
* There's already some `if (1) { ... }` used some scoping, and I added some more but maybe that's not fake?
* `goto` match in `BossMo_IntroCs`, but I noticed a similar pattern is already used in `BossMo_Tentacle` and other bosses as well
* To squash some fake diffs, I banned pointers in disassembly in the range 0x0000F000 to 0x00010100 (not for any principled reason, but because all the pointers in that range currently are fake)

### DIFF
--- a/src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
+++ b/src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
@@ -407,51 +407,14 @@ void BossMo_SetupTentacle(BossMo* this, PlayState* play) {
 }
 
 void BossMo_Tentacle(BossMo* this, PlayState* play) {
-    s16 tentXrot;
-    s16 sp1B4 = 0;
-    s32 buttons;
-    Player* player = GET_PLAYER(play);
-    s16 indS0;
     s16 indS1;
-    Camera* mainCam1;
-    Camera* mainCam2;
-    BossMo* otherTent = (BossMo*)this->otherTent;
-    f32 maxSwingRateX;
-    f32 maxSwingLagX;
-    f32 maxSwingSizeX;
-    f32 maxSwingRateZ;
-    f32 maxSwingLagZ;
-    f32 maxSwingSizeZ;
-    f32 swingRateAccel;
-    f32 swingSizeAccel;
-    s16 rippleCount;
-    s16 indT5;
-    Vec3f ripplePos;
-    f32 randAngle;
-    f32 randFloat;
+    s16 sp1B4 = 0;
+    Player* player = GET_PLAYER(play);
     f32 tempf1;
     f32 tempf2;
     f32 sin;
     f32 cos;
-    f32 temp;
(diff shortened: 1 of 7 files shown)


src/overlays/actors/ovl_Boss_Tw/z_boss_tw.c
```diff
@@ -4654,8 +4651,13 @@ void BossTw_UpdateEffects(PlayState* play) {
                 off.z = sTwinrovaPtr->actor.world.pos.z - eff->pos.z;
 
                 if (sTwinrovaPtr->actionFunc != BossTw_TwinrovaStun) {
+                    Vec3f spF4;
+                    Vec3f spE8;
+                    Vec3f spDC;
+
                     if ((SQ(off.x) + SQ(off.y) + SQ(off.z)) < SQ(60.0f)) {
                         for (j = 0; j < 50; j++) {
+
```
fig02, line 4660: can remove

fig02 (approved):

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
```diff
@@ -570,25 +542,33 @@ void BossMo_Tentacle(BossMo* this, PlayState* play) {
...
-            }
-            for (indS1 = 0; indS1 < rippleCount; indS1++) {
-                randFloat = Rand_ZeroFloat(50.0f);
-                randAngle = Rand_ZeroFloat(0x10000);
-                ripplePos = this->actor.world.pos;
-                ripplePos.x += sinf(randAngle) * randFloat;
-                ripplePos.z += cosf(randAngle) * randFloat;
-                ripplePos.y = MO_WATER_LEVEL(play);
-                BossMo_SpawnRipple(play->specialEffects, &ripplePos, 40.0f, 110.0f, 80, 290, MO_FX_SMALL_RIPPLE);
+            if (1) {
```
hensldm, line 545: More so out of my own curiosity, is the `if (1)` needed? Do we have a preference of having an `if (1) {` to start a new block for variable declaration, or just starting with a `{`?

hensldm (approved):
1 curiosity question

src/overlays/actors/ovl_Boss_Mo/z_boss_mo.c
fig02: we've been doing it when its needed and also when scoped temps are needed, since it knocks two birds out with one stone. whether or not its needed in this case im not sure

cadmic: I've been preferring a plain `{`, but in this case `if (1) {` is needed. I'm not sure why, or if there's another way to write it

PR #1896, opened 2024-02-26
Clean up custom LoadTextureBlock in z_map_mark.c and z_lmap_mark.c

mzxrules:
the definitions of gdpLoadTextureBlock and similar use token replacement to construct the constants passed into macro's internal display list calls, meaning the macro cannot be used to pass in raw IM_SIZ values during runtime. To solve this, the original uses a series of six lookup tables to fetch the correct constant values. 

To get the original gdpLoadTextureBlock to reference these arrays, decomp currently creates 6 specially named defines for every uniquely named `siz` argument that needs to be passed into `gdpLoadTextureBlock`. 

There's a few things I find wrong with this: 

* In general, I find it poor design to have to create 6 defines for every different possible argument you'd pass into a macro, and I fear that somebody who is too afraid to look into the `gdpLoadTextureBlock` internals is going to blindly copypaste the pattern because it just works
* If a custom mod updates `gdpLoadTextureBlock` to allow raw IM_SIZ values, the changes won't automatically eliminate the array access since the decomp code still passes in the result of the sBaseImageSizes lookup into the macro.
* Technically, z_map_mark.c and z_lmap_mark.c is coupled to `gdpLoadTextureBlock`'s internals more than it probably should be.


include/gDPLoadTextureBlock_Runtime.inc.c
```diff
@@ -0,0 +1,67 @@
...
+    G_IM_SIZ_16b,
+    G_IM_SIZ_32b,
+};
+static ZCONST s32 sLoadTextureBlock_siz_LOAD_BLOCK[] = {
+    G_IM_SIZ_4b_LOAD_BLOCK,
+    G_IM_SIZ_8b_LOAD_BLOCK,
+    G_IM_SIZ_16b_LOAD_BLOCK,
+    G_IM_SIZ_32b_LOAD_BLOCK,
+};
+static ZCONST u32 sLoadTextureBlock_siz_INCR[] = {
```
hensldm, line 27: Can this be `s32` like the others, or perhaps they could all be `u32`?

hensldm (commented):
Fwiw, I don't mind this solution. Kind of annoying dealing with data that is both in data and in rodata in separate places, but this doesn't seem too bad.

include/gDPLoadTextureBlock_Runtime.inc.c
mzxrules: MM doesn't like all `s32`, but is fine with all `u32`.

hensldm (approved):

include/gDPLoadTextureBlock_Runtime.inc.c
```diff
@@ -0,0 +1,67 @@
...
+#include "global.h"
+#include "gfx.h"
+
+// The Following arrays must be defined as const in z_lmap_mark.c to appear in rodata
+
+#ifndef GDP_LOADTEXTUREBLOCK_RUNTIME_CONST
+#define ZCONST
+#else
+#define ZCONST const
+#endif
```
fig02, line 13: what is the Z here supposed to mean?

mzxrules: Nothing really, It's just to avoid having an identifier named `CONST`

fig02: well it can probably be more descriptive then. COND_CONST or something to make it clear that its conditional based on something else being defined, or something

cadmic (approved):

Thar0 (approved):

Dragorn421 (changes requested):
licensing

mzxrules:
CC0

PR #1899, opened 2024-02-27
Restore terminal colors in retail, except in __osMalloc.c

cadmic:
In #1741 I attempted to match retail __osMalloc.c .rodata by disabling terminal colors in retail. Turns out this was probably not the right move, because they still show up in fault.c and fault_drawer.c. So, I've reverted that and ifdef'd out the string in __osMalloc.c instead.


fig02 (approved):

Dragorn421 (approved):
:thinking:

PR #1900, opened 2024-02-27
Add proper bug comment for gSaveContext.save.dayTime getting cleared as s32 when it is u16

Thar0:


AngheloAlf (approved):

fig02 (approved):

Dragorn421 (approved):
dang

petition to change
`bzero(&gSaveContext.save.linkAge, sizeof(s32));`
to
`bzero(&gSaveContext.save.linkAge, sizeof(gSaveContext.save.linkAge));`
and similar

PR #1902, opened 2024-02-27
AVOID_UB for out-of-bounds access in AudioLoad_Init

Thar0:
This is (potentially) no longer harmless if using a compiler that is allowed to reorder global variables

### DIFF
--- a/src/audio/lib/load.c
+++ b/src/audio/lib/load.c
@@ -1133,9 +1133,19 @@ void AudioLoad_Init(void* heap, u32 heapSize) {
         s32 i;
         u8* audioContextPtr = (u8*)&gAudioCtx;
 
+#ifndef AVOID_UB
+        //! @bug This clearing loop sets one extra byte to 0 following gAudioCtx.
+        //! In practice this is harmless as it would set the most significant byte in gAudioCustomUpdateFunction to 0,
+        //! which was just reset to NULL above.
         for (i = sizeof(gAudioCtx); i >= 0; i--) {
             *audioContextPtr++ = 0;
         }
+#else
+        // Avoid out-of-bounds variable access
+        for (i = sizeof(gAudioCtx); i > 0; i--) {
+            *audioContextPtr++ = 0;
+        }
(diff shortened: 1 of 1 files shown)


src/audio/lib/load.c
```diff
@@ -1133,9 +1133,19 @@ void AudioLoad_Init(void* heap, u32 heapSize) {
...
+        //! In practice this is harmless as it would set the most significant byte in gAudioCustomUpdateFunction to 0,
+        //! which was just reset to NULL above.
         for (i = sizeof(gAudioCtx); i >= 0; i--) {
             *audioContextPtr++ = 0;
         }
+#else
+        // Avoid out-of-bounds variable access
+        for (i = sizeof(gAudioCtx); i > 0; i--) {
+            *audioContextPtr++ = 0;
+        }
```
AngheloAlf, line 1147: Why not `bzero` instead?
```suggestion
        bzero(&gAudioCtx, sizeof(gAudioCtx));
```

Thar0: I wanted to keep the behavioral differences minimal between the two. If more people are fine with larger departures from the original in AVOID_UB blocks I can change it.

Dragorn421:
This is a bugfix, not avoid_ub

avoid_ub would be zeroing gAudioCtx and `(u8)&gAudioCustomUpdateFunction`

Thar0:
It's already guaranteed to be 0, there is no need to 0 it again

Dragorn421 (approved):
that is quite true

cadmic (approved):

src/audio/lib/load.c
Dragorn421: I also prefer minimal difference. calling a function or fundamentally different code wouldn't be in the spirit of avoid_ub "replicate the undefined behavior as it behaves in vanilla but without undefined behavior"

fig02 (approved):

AngheloAlf (approved):

PR #1903, opened 2024-02-27
Some GCC flags improvements

Thar0:
Thanks to the prior alignment/UB changes these changes are now stable

* Move `-fno-toplevel-reorder -fno-zero-initialized-in-bss` to assets only
* Add `-fno-PIC` since some gcc versions use `-fPIC` as default (notably newer versions of mips-linux-gnu-)
* Enable builtin functions on gcc, with appropriate changes to missing_gcc_functions.c and ultra64/libc.h
* Move `-fno-merge-constants -mno-explicit-relocs -mno-split-addresses` to overlays only as they are only needed there for reloc reasons

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,7 +154,7 @@ endif
 ASFLAGS := -march=vr4300 -32 -no-pad-sections -Iinclude
 
 ifeq ($(COMPILER),gcc)
-  CFLAGS += -G 0 -nostdinc $(INC) -march=vr4300 -mfix4300 -mabi=32 -mno-abicalls -mdivide-breaks -fno-zero-initialized-in-bss -fno-toplevel-reorder -ffreestanding -fno-common -fno-merge-constants -mno-explicit-relocs -mno-split-addresses $(CHECK_WARNINGS) -funsigned-char
+  CFLAGS += -G 0 -nostdinc $(INC) -march=vr4300 -mfix4300 -mabi=32 -mno-abicalls -mdivide-breaks -fno-PIC -fno-common -ffreestanding -fbuiltin -fno-builtin-sinf -fno-builtin-cosf $(CHECK_WARNINGS) -funsigned-char
   MIPS_VERSION := -mips3
 else
   # Suppress warnings for wrong number of macro arguments (to fake variadic
@@ -314,8 +314,11 @@ $(BUILD_DIR)/src/overlays/%.o: CC := $(PYTHON) tools/asm_processor/build.py $(CC
 
 $(BUILD_DIR)/assets/%.o: CC := $(PYTHON) tools/asm_processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
 else
+# The / is intentionally left out to capture other directories starting with "assets"
+$(BUILD_DIR)/assets%.o: CFLAGS += -fno-zero-initialized-in-bss -fno-toplevel-reorder
+$(BUILD_DIR)/src/%.o: CFLAGS += -fexec-charset=euc-jp
 $(BUILD_DIR)/src/libultra/libc/ll.o: OPTFLAGS := -Ofast
-$(BUILD_DIR)/src/%.o: CC := $(CC) -fexec-charset=euc-jp
+$(BUILD_DIR)/src/overlays/%.o: CFLAGS += -fno-merge-constants -mno-explicit-relocs -mno-split-addresses
 endif
 
 #### Main Targets ###
--- a/Makefile
+++ b/Makefile
@@ -314,8 +314,8 @@ $(BUILD_DIR)/src/overlays/%.o: CC := $(PYTHON) tools/asm_processor/build.py $(CC
 
 $(BUILD_DIR)/assets/%.o: CC := $(PYTHON) tools/asm_processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
 else
-# The / is intentionally left out to capture other directories starting with "assets"
-$(BUILD_DIR)/assets%.o: CFLAGS += -fno-zero-initialized-in-bss -fno-toplevel-reorder
+# Note that if adding additional assets directories for modding reasons these flags must also be used there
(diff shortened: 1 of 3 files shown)


include/ultra64/libc.h
```diff
@@ -5,8 +5,14 @@
 
 void osSyncPrintf(const char* fmt, ...);
 
+#ifdef __GNUC__
+void bzero(void* __s, unsigned int __n);
+int bcmp(const void* __sl, const void* __s2, unsigned int __n);
+void bcopy(const void* __src, void* __dest, unsigned int __n);
+#else
```
AngheloAlf, line 12: Why?

Thar0: gcc recognizes these as builtins and warns because they have the wrong prototype, and since this is in a header the warning repeats for all files 💀
this fixes that warning (and avoids the spam)

fig02 (approved):

include/ultra64/libc.h
Dragorn421: are the implementations still the same using gcc? does it not error or at least warn that the prototypes are different in the definition

src/gcc_fix/missing_gcc_functions.c
```diff
@@ -10,10 +10,10 @@
 // Self-hosted libc memory functions, gcc assumes these exist even in a freestanding
 // environment and there is no way to tell it otherwise.
 
-int memcmp(void* s1, const void* s2, size_t n) {
-    u8* m1 = (u8*)s1;
-    u8* m2 = (u8*)s2;
-    u32 i;
+int memcmp(const void* s1, const void* s2, size_t n) {
+    const u8* m1 = (const u8*)s1;
+    const u8* m2 = (const u8*)s2;
```
Dragorn421, line 15:
```suggestion
    const u8* m1 = s1;
    const u8* m2 = s2;
```

src/gcc_fix/missing_gcc_functions.c
```diff
@@ -26,17 +26,38 @@ int memcmp(void* s1, const void* s2, size_t n) {
     return 0;
 }
 
-void* memset(void* str, s32 c, size_t n) {
-    u8* m1 = (u8*)str;
-    u32 i;
+void* memset(void* str, int c, size_t n) {
+    u8* m = (u8*)str;
```
Dragorn421, line 30:
```suggestion
    u8* m = str;
```

include/ultra64/libc.h
Thar0: I'm not sure what you're asking exactly; the implementations for these are assembly and are still called by gcc, but gcc has prototypes for these registered internally and doesn't agree with the libultra prototypes creating a warning

Dragorn421: oh the implementations are assembly, I forgot that point. okay then nvm

Makefile
```diff
@@ -154,7 +154,7 @@ endif
 ASFLAGS := -march=vr4300 -32 -no-pad-sections -Iinclude
 
 ifeq ($(COMPILER),gcc)
-  CFLAGS += -G 0 -nostdinc $(INC) -march=vr4300 -mfix4300 -mabi=32 -mno-abicalls -mdivide-breaks -fno-zero-initialized-in-bss -fno-toplevel-reorder -ffreestanding -fno-common -fno-merge-constants -mno-explicit-relocs -mno-split-addresses $(CHECK_WARNINGS) -funsigned-char
+  CFLAGS += -G 0 -nostdinc $(INC) -march=vr4300 -mfix4300 -mabi=32 -mno-abicalls -mdivide-breaks -fno-PIC -fno-common -ffreestanding -fbuiltin -fno-builtin-sinf -fno-builtin-cosf $(CHECK_WARNINGS) -funsigned-char
```
Dragorn421, line 157: asking for the sake of asking, why `-fno-PIC` rather than `-fno-pic`

I will look into `-fbuiltin -fno-builtin-sinf -fno-builtin-cosf` but if you feel like giving a short explanation I'd appreciate it

Makefile
```diff
@@ -314,8 +314,11 @@ $(BUILD_DIR)/src/overlays/%.o: CC := $(PYTHON) tools/asm_processor/build.py $(CC
 
 $(BUILD_DIR)/assets/%.o: CC := $(PYTHON) tools/asm_processor/build.py $(CC) -- $(AS) $(ASFLAGS) --
 else
+# The / is intentionally left out to capture other directories starting with "assets"
+$(BUILD_DIR)/assets%.o: CFLAGS += -fno-zero-initialized-in-bss -fno-toplevel-reorder
```
Dragorn421, line 318: still not a fan of this
not too sure what's the point either given the build rule
`$(BUILD_DIR)/assets/%.o: assets/%.c`
still is with-`/`

Makefile
Thar0: - iirc PIC vs pic is just whether to use a big GOT or not, afaik disabling just one is sufficient but maybe disabling both is best..
- `-fbuiltin` followed by `-fno-builtin-X` enables all builtins except X, so here I enable builtins in general but disable the trig ones

Makefile
Thar0: The idea is to cover for people "less in the know", as it were, who may make a second asset dir for mods and then wonder why gcc is crashing on them. I think the flags override is less obvious than the makefile rules themselves but maybe that's not true, idk

fig02: I get the idea but I do think its a bit messy to not be explicit. With the direction our assets are headed it should be sooner than later that custom asset dirs wouldnt be needed/encouraged

Thar0: I've added the `/` but left a comment which can hopefully be removed when assets are further along

fig02 (approved):

Dragorn421 (approved):

fig02:
@AngheloAlf would you like to approve since you reviewed above

AngheloAlf (approved):

PR #1905, opened 2024-03-02
Match retail z_fishing.c

cadmic:
Notes:
* Much of this code was not changed in MM, so that helped a bunch (although it took me a while to think about that). But MM's `Fishing_UpdateLure` is probably fake since it doesn't use `spE4` in the same places as the debug version. Instead, the permuter suggested changing a `4.0f` to a `4`.
* I added `KREG_DEBUG(i)` to change `KREG(i)` to `0` in retail, which helped in many places but not all
* I tried hard to avoid ifdef-ing out the `if (play) {}` in `Fishing_HandleOwnerDialog` but I couldn't find a better match for debug
* I added some more disassembly hacks for an especially bad fake diff where spimdisasm was finding floats in the .text section, which was confusing the diff script since it thought these floats were new functions

### DIFF
--- a/src/overlays/actors/ovl_Fishing/z_fishing.c
+++ b/src/overlays/actors/ovl_Fishing/z_fishing.c
@@ -14,6 +14,12 @@
 
 #define WATER_SURFACE_Y(play) play->colCtx.colHeader->waterBoxes->ySurface
 
+#if OOT_DEBUG
+#define KREG_DEBUG(i) KREG(i)
+#else
+#define KREG_DEBUG(i) 0
+#endif
+
 void Fishing_Init(Actor* thisx, PlayState* play2);
 void Fishing_Destroy(Actor* thisx, PlayState* play2);
 void Fishing_UpdateFish(Actor* thisx, PlayState* play2);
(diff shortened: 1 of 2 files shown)


fig02 (approved):
the missing f on many floats is a common pattern for this dev in particular so that one checks out

src/overlays/actors/ovl_Fishing/z_fishing.c
```diff
@@ -3385,8 +3415,14 @@ void Fishing_UpdateFish(Actor* thisx, PlayState* play2) {
...
 
+#if OOT_DEBUG
             if (((this->timerArray[0] == 1) || (Rand_ZeroOne() < chance)) &&
-                ((Rand_ZeroOne() < (this->perception * multiplier)) || ((this->isLoach + 1) == KREG(69)))) {
+                ((Rand_ZeroOne() < (this->perception * multiplier)) || ((this->isLoach + 1) == KREG(69))))
+#else
+            if (((this->timerArray[0] == 1) || (Rand_ZeroOne() < chance)) &&
+                (Rand_ZeroOne() < (this->perception * multiplier)))
+#endif
+            {
```
Dragorn421, line 3425: I forgot what we made of this last time / in which PR it was
```suggestion
            if (((this->timerArray[0] == 1) || (Rand_ZeroOne() < chance)) &&
                ((Rand_ZeroOne() < (this->perception * multiplier))
#if OOT_DEBUG
                 || ((this->isLoach + 1) == KREG(69))
#endif
                     )) {
```

fig02: id say just do whatever you think is best for now and we will decide on something and consistently use it after retail is matched

Dragorn421 (commented):
not convinced by `KREG_DEBUG` given it's only 1/3rd of KREG uses, but whatever for now

Dragorn421 (approved):

fig02:
KREG is **K**azuaki Morita-- this much is clear by the files/usages
- debug in z_actor to spawn arwing
- debug in collision check for the boss-only `CollisionCheck_GetSwordDamage` function
- effectss sibuki, bubbles used by gohma
- sda the unused shadow
- ganondorfs cape in en_viewer
- fishing

its possible individual devs had their own way of ifdeffing out their own debug regs, or there is a global way to do it for all. but we've yet to figure something out, so we are just doing what works for now

PR #1910, opened 2024-03-02
Remove 0xFF padding from uncompressed ROMs

cadmic:
Recall that the Debug ROM dump looks like this:
* The ROM begins at offset 0x00000000
* The last DMA segment file in the ROM ends at offset 0x035CE040
* The dump is filled with 0x00 up to offset 0x035CF000
* Then, the dump is filled 0xFF up to offset 0x03600000 (54 MiB)
* Then, the dump contains Pokemon Stadium 2 data(?) that has been partially overwritten.

Handling these 0xFFs makes the build process slightly more complicated so it's nice to get rid of it. In fact, right now decompress_baserom.py pads with 0xFF up to 0x4000 while elf2rom pads up to 0x100000, which causes problems for gc-eu-mq. I suspect that the 0xFFs are not actually part of the padded ROM, but are erased data from the development cartridge, since Flash or EEPROM memory has a default state of all 1s after being erased.

After this change, both decompress_baserom.py and elf2rom will pad ROMs with 0s up to the next 0x1000-byte boundary with no further padding (elf2rom via `_RomSize` in the linker script). Compressed ROMs are still the same as before, where they are filled with the 0x00,0x01,...,0xFE,0xFF pattern up to a multiple of 8 MiB.

### DIFF
--- a/baseroms/gc-eu-mq/checksum.md5
+++ b/baseroms/gc-eu-mq/checksum.md5
@@ -1,1 +1,1 @@
-1a438f4235f8038856971c14a798122a  build/gc-eu-mq/oot-gc-eu-mq.z64
+4920520254b9aab86de57b42ab477dbb  build/gc-eu-mq/oot-gc-eu-mq.z64
--- a/tools/decompress_baserom.py
+++ b/tools/decompress_baserom.py
@@ -128,7 +128,8 @@ def byte_swap(file_content: bytearray) -> bytearray:
 
 def per_version_fixes(file_content: bytearray, version: str) -> bytearray:
     if version == "gc-eu-mq-dbg":
-        # Strip the overdump
+        # Strip the overdump, which consists of an area of 0xFF bytes that may
+        # be erased memory and ROM data from an unrelated game
         print("Stripping overdump...")
         file_content = file_content[0:0x035CF000]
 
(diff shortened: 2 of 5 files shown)


tools/decompress_baserom.py
```diff
@@ -129,23 +130,14 @@ def per_version_fixes(file_content: bytearray, version: str) -> bytearray:
     if version == "gc-eu-mq-dbg":
         # Strip the overdump
         print("Stripping overdump...")
```
Dragorn421, line 132:
```suggestion
         # Strip the overdump
         # That is, an area filled with 0xFF that may be from erased EEPROM data, followed by rom data from another unrelated game
         print("Stripping overdump...")
```
Or something (may need a line break too)

Dragorn421 (commented):
> Uncompressed ROMs are still the same as before, ...

You meant "compressed"?

fig02 (approved):

baseroms/gc-eu-mq/checksum.md5
Dragorn421, line 1: ftr I checked correctness of all hashes in this PR (I have a script which sorts oot roms by variants as well as extrapolating variants)
for this artificial rom I had to go out of my way,
```
>>> from pathlib import Path
>>> import hashlib
>>> hashlib.md5(Path("baseroms/gc-eu-mq/baserom-decompressed.z64").read_bytes()[:0x34C7000]).hexdiges
t()
'4920520254b9aab86de57b42ab477dbb'
```

Dragorn421 (approved):

Issue #1904, opened 2024-02-28
New assets system (outdated)


Dragorn421:
Just need a home for the directory layout we agreed upon, so I can find it again

---------

Productive discussion on discord happened
from https://discord.com/channels/688807550715560050/688851317593997489/1207388581639880745
to https://discord.com/channels/688807550715560050/688851317593997489/1208159004694413312

The general result is a new top level folder `extracted/` for extracted things, including assets, text and baserom segments

I compiled the changes to "future layout", see below

----------

in the following directory structure layouts:
`[first/part/]second/part` paths are where the file does `#include "second/part"`, and CPP resolves it to `first/part/second/part` (due to `first/part` being in the include path)

## future layout

"future layout v4":

<details>

Include path would be like `-Ibuild/VERSION -Iextracted/VERSION`

```
# v4
assets/  # note: everything under assets/ committed
  objects/
    gameplay_keep/
      gameplay_keep.c -> [extracted/VERSION/]objects/gameplay_keep/gameplay_keepVtx_00C0A0.vtx.inc , [extracted/VERSION/]objects/gameplay_keep/deku_stick.i8.inc.c
      gameplay_keep.h
  text/
    nes_message_data_static.c -> [build/VERSION/]assets/text/message_data.enc.h
    message_data.h -> [extracted/VERSION/]text/message_data.h
  xml/
    objects/
      gameplay_keep.xml
extracted/  # .gitignore'd
  VERSION/
    baserom/
      gameplay_keep
    objects/
      gameplay_keep/
        deku_stick.i8.png  # written on make setup for reference, but otherwise unused
        deku_stick.i8.inc.c  # written on make setup (does *not* depend on the png)
        gameplay_keepVtx_00C0A0.vtx.inc
    text/
      message_data.h
build/
  VERSION/
    baserom/
      gameplay_keep.o  # objcopy'd from extracted/VERSION/baserom/gameplay_keep
    assets/
      text/
        message_data.enc.h  # processed from assets/text/message_data.h
```

</details>

history:

<details>

v3->v4: removed extracted/VERSION/text/message_data.enc.h (no clue why it was there)
v2->v3: fix oversimplified text handling. cf https://github.com/zeldaret/oot/pull/1730#issuecomment-1948247178
v1->v2: cf discord discussion linked above
v1: `assets/_extracted/`, cf https://github.com/zeldaret/oot/pull/1730#issuecomment-1932832284

</details>

## current zeldaret/main layout

For reference.

[...]

PR #1916, opened 2024-03-04
Move baserom segments to extracted/

cadmic:
Sorry for breaking ZAPD CI yet again

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ build/
 expected/
 notes/
 baserom/
+baseroms/*/segments/
 docs/doxygen/
 *.elf
 *.sra
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,6 @@ build/
 expected/
 notes/
 baserom/
-baseroms/*/segments/
 docs/doxygen/
 *.elf
 *.sra
--- a/Makefile
+++ b/Makefile
@@ -208,8 +208,7 @@ ASSET_FILES_OUT := $(foreach f,$(ASSET_FILES_XML:.xml=.c),$f) \
 
 UNDECOMPILED_DATA_DIRS := $(shell find data -type d)
 
-BASEROM_SEGMENTS_DIR := $(EXTRACTED_DIR)/segments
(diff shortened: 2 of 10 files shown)


.gitignore
```diff
@@ -17,7 +17,6 @@ build/
 expected/
 notes/
 baserom/
-baseroms/*/segments/
```
fig02, line 20: we could possibly make a "legacy" section of the gitignore and keep both of these baserom paths there? 

the idea behind a legacy section is that it would eventually be deleted, but it gives people time to stop working with old branches that use the old folder structure

fig02 (approved):

fig02:
I was wrongly under the impression that this change would allow use to remove the `baseroms/` folder entirely, but I guess that doesn't make sense. cause we still need a place to put it for the extraction step to use.

Is `extracted/{VERSION}/segments` clear enough? do we want to indicate that they are baserom/binary somehow, or does the term "segments" already convey that well enough?

cadmic:
Oh yeah, it's actually called `extracted/$VERSION/baserom` in the latest version in #1904 (similar to how the .o versions are in `build/$VERSION/baserom`). I'll make that change

Dragorn421:
`extracted/` conveys that it's from the baserom

Dragorn421:
> Oh yeah, it's actually called `extracted/$VERSION/baserom` in the latest version in #1904 (similar to how the .o versions are in `build/$VERSION/baserom`). I'll make that change

Yeah may as well

fig02:
it sounds like itll be more clear for the user, but it will be a bit silly that `baserom/` only has 1 item which is `segments/`, right?

cadmic:
I think the idea is that there will be no `segments` path component, it will be e.g. `extracted/gc-eu-mq-dbg/baserom/nes_message_data_static` (similar to `build/gc-eu-mq-dbg/baserom/nes_message_data_static.o`)

fig02 (approved):

Dragorn421 (approved):

PR #1917, opened 2024-03-05
Match retail camera

cadmic:
I tried to split this into nice commits for easier review, since some of the matches are a bit sketchy. ~~The only function left is `Camera_Battle1`~~ (EDIT: now hackily matched). Thanks to @engineer124 for getting this started.

Most of the `OREG` are replaced by their values in `sOREGInit`, so I created debug macros for all of them. Except, the OREGs that are scaled by 0.01f didn't work as-is, because e.g. `20 * 0.01f != 0.2f` (0x3e4ccccc vs 0x3e4ccccd), so I duplicated the list so I could write the floats directly. And I had to write some floats in a different way (e.g. no leading 0). Is there a better way?

`Camera_Demo5` had a lot of scoping problems and I don't think there's a way to manufacture enough scoped temps to cover them. So I moved all the static data inline. It's consistent with `z_onepointdemo.c` now at least.


include/regs.h
```diff
@@ -67,29 +67,6 @@
 #define R_FB_FILTER_ENV_COLOR(c)                 SREG(85 + (c))
 #define R_ENABLE_FB_FILTER                       SREG(88)
 #define R_PAUSE_BG_PRERENDER_STATE               SREG(94) // `PauseBgPreRenderState`
-#define R_CAM_XZ_OFFSET_UPDATE_RATE              OREG(2)
-#define R_CAM_Y_OFFSET_UPDATE_RATE               OREG(3)
-#define R_CAM_FOV_UPDATE_RATE                    OREG(4)
-#define R_CAM_MAX_PITCH                          OREG(5)
-#define R_CAM_R_UPDATE_RATE_INV                  OREG(6)
```
fig02, line 74: Not a fan of losing all the oreg names from this file, but I dont have any better ideas. I do agree that keeping the camera-specific data defined in here would be even weirder so this seems like the lesser of two evils

src/code/z_camera.c
```diff
@@ -706,12 +788,12 @@ Vec3f Camera_CalcUpFromPitchYawRoll(s16 pitch, s16 yaw, s16 roll) {
 f32 Camera_ClampLERPScale(Camera* camera, f32 maxLERPScale) {
     f32 ret;
 
-    if (camera->atLERPStepScale < CAM_DATA_SCALED(R_CAM_AT_LERP_STEP_SCALE_MIN)) {
-        ret = CAM_DATA_SCALED(R_CAM_AT_LERP_STEP_SCALE_MIN);
+    if (camera->atLERPStepScale < R_CAM_AT_LERP_STEP_SCALE_MIN) {
```
Dragorn421, line 791: I don't like `CAM_DATA_SCALED` being inside `R_CAM_AT_LERP_STEP_SCALE_MIN` and similar, `R_` macros should read s16s (e.g. as set/shown in the reg editor), perhaps (?) expanding to constant defaults in NDEBUG

The options I see are remove the `R_` prefix or don't have `CAM_DATA_SCALED` in the macro
the former doesn't solve not knowing the type (and more importantly the unit/range) of the value
the latter sounds annoying for default values (may even not match)

src/code/z_camera.c
```diff
@@ -1443,10 +1533,9 @@ s32 Camera_Normal1(Camera* camera) {
     f32 rate = 0.1f;
 
     playerHeight = Player_GetHeight(camera->player);
-    if (RELOAD_PARAMS(camera) || R_RELOAD_CAM_PARAMS) {
+    if (RELOAD_PARAMS(camera) || CAM_DEBUG_RELOAD_PARAMS) {
```
Dragorn421, line 1536: I am surprised `CAM_DEBUG_RELOAD_PARAMS` is true in non-debug, so the condition always passes

Is `RELOAD_PARAMS` entirely optimized out then?
Is it usually required for matching? If not maybe there's further refactoring of this conditional that could be done in some way

src/code/z_camera.c
Dragorn421: to bounce back on https://github.com/zeldaret/oot/pull/1917#discussion_r1513403713 ,
we could have the R_ regs in regs.h
and not-`R_`-prefixed macros with the default and `CAM_DATA_SCALED` in z_camera.c maybe
and eventually find a convention for indicating what is scaled or s16

include/regs.h
mzxrules: You can always leave the symbol names in for documentation purposes

src/code/z_camera.c
cadmic: I tried replacing this with `if (1)` everywhere. One benefit is that Camera_KeepOn4 currently has weird match of removing a PRINTF, which would not be necessary if we used `if (1)` instead of `if (RELOAD_PARAMS(camera) || R_RELOAD_CAM_PARAMS)`. However, Camera_Jump3 writes `if (RELOAD_PARAMS(camera) || modeSwitch || R_RELOAD_CAM_PARAMS)` instead of `if (RELOAD_PARAMS(camera) || R_RELOAD_CAM_PARAMS)`, and replacing that with `if (1)` didn't work (neither did rearranging the condition so we could write `modeSwitch || DEBUG_RELOAD_PARAMS(camera)` or something instead). So I went with the `|| CAM_DEBUG_RELOAD_PARAMS` so that case could be consistent too, but I'm not sure if it's worth the weird match in Camera_KeepOn4

src/code/z_camera.c
```diff
@@ -25,6 +42,128 @@ s32 Camera_UpdateWater(Camera* camera);
 // Load the next value and scale down from camera read-only data stored in CameraModeValue
 #define GET_NEXT_SCALED_RO_DATA(values) CAM_DATA_SCALED(GET_NEXT_RO_DATA(values))
 
+#if OOT_DEBUG
+#define OREG_DEBUG(i, default) OREG(i)
+#define OREG_DEBUG_SCALED(i, default) CAM_DATA_SCALED(OREG(i))
+#else
+#define OREG_DEBUG(i, default) (default)
+#define OREG_DEBUG_SCALED(i, default) (default)
+#endif
```
fig02, line 51: i may be misundersranding something, but cant these be removed now?

src/code/z_camera.c
```diff
@@ -6,13 +6,30 @@
...
 
+#if OOT_DEBUG
+s32 Camera_QRegInit(void);
+#endif
+
+#if OOT_DEBUG
+#define CAM_DEBUG_RELOAD_PREG(camera) \
+    if (R_RELOAD_CAM_PARAMS)          \
+    Camera_CopyPREGToModeValues(camera)
+#else
```
fig02, line 19: tiny thing but we can probably just include braces in the definition. its not a huge deal since its hidden by the macro, but if we can continue to follow the repos style we might as well imo

src/code/z_camera.c
cadmic: Sorry about that

src/code/z_camera.c
cadmic: Alright, I've added braces and a `(void)0`

fig02 (approved):

cadmic:
I pushed a small but hacky match for `Camera_Battle1`, because I've given up hope on trying to match it properly.

cadmic:
Thanks to @inspectredc `Camera_Battle1` is matched properly now

fig02 (approved):

Yanis42 (commented):
too bad there's lots of define at the beginning of ``z_camera.c`` (which became bigger because of the static data), though I don't see any alternative except moving them to their own file

anyway, just a little thought I had but otherwise seems good to me

src/code/z_camera.c
```diff
@@ -2693,12 +2802,12 @@ s32 Camera_Jump3(Camera* camera) {
...
-        eyeDiffGeo.pitch = R_CAM_MAX_PITCH;
+    if (eyeDiffGeo.pitch > CAM_MAX_PITCH) {
+        eyeDiffGeo.pitch = CAM_MAX_PITCH;
     }
 
-    if (eyeDiffGeo.pitch < R_CAM_MIN_PITCH_1) {
-        eyeDiffGeo.pitch = R_CAM_MIN_PITCH_1;
+    if (eyeDiffGeo.pitch < CAM_MIN_PITCH_1) {
+        eyeDiffGeo.pitch = CAM_MIN_PITCH_1;
     }
```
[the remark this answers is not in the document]
fig02: they can potentially use clamp even. but its probably out of scope for this PR, since its just trying to match things for retail. this seems more of a general improvement that can be researched separately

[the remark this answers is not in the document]
Yanis42: yeah true, I was concerned by the size of the file but it's fine for now

Yanis42 (approved):

Dragorn421:
Are we waiting on anyone specific to review?

Dragorn421 (approved):

fig02:
was hoping for a possible engineer review, however hes been busy and did not get back to my request. will merge soon

PR #1918, opened 2024-03-05
Line up code segment for gc-eu-mq

cadmic:
Fix the following gc-eu-mq padding/alignment problems, with help from the [zelda64_compare spreadsheet](https://docs.google.com/spreadsheets/d/17yPD3DqqH5lZeR7c_QmJfgxWgVyYkGgTLZoKcvLTwtw/edit#gid=204902945):
* The VRAM address of `code` was wrong; it should be 80010f00 but it was assigned address 80010ef0. This doesn't seem to be due to missing padding because the `boot` and `dmadata` segments have the right size. From the spreadsheet it looks like the `code` segment is 0x20-aligned on all versions, so I changed the implementation of the `align` directive to align the entire segment instead of individual sections and used that.
* There's 16 bytes of padding between the .text for audio/general.o and audio/sfx.o. It doesn't look like this is due to alignment and only exists in some versions, so I can't explain this. I added a `pad_text` in all retail builds for now and added a comment for later (same with the other `pad_text` before audio/lib/data.o).

I tested this with `tools/bss_reordering.py` (after hackily matching Camera_Battle1), since that seems to pretty reliably detect when things are shifted. For the curious, here are the non-overlay files with BSS ordering problems:
```
boot/idle.o
code/z_actor.o
code/z_camera.o
code/z_collision_check.o
code/z_common_data.o
code/z_demo.o
code/z_kankyo.o
code/z_kaleido_scope_call.o
code/z_play.o
code/main.o
code/speed_meter.o
code/sys_math3d.o
code/fault.o
```

### DIFF
--- a/tools/mkldscript.c
+++ b/tools/mkldscript.c
@@ -51,11 +51,13 @@ static void write_ld_script(FILE *fout)
                   "    ..%s ", seg->name, seg->name, seg->name, seg->name);
 
         if (seg->fields & (1 << STMT_after))
-            fprintf(fout, "_%sSegmentEnd ", seg->after);
+            fprintf(fout, "(_%sSegmentEnd + %i) & ~ %i ", seg->after, seg->align - 1, seg->align - 1);
         else if (seg->fields & (1 << STMT_number))
             fprintf(fout, "0x%02X000000 ", seg->number);
         else if (seg->fields & (1 << STMT_address))
             fprintf(fout, "0x%08X ", seg->address);
+        else
+            fprintf(fout, "ALIGN(0x%X) ", seg->align);
 
         // (AT(_RomSize) isn't necessary, but adds useful "load address" lines to the map file)
         fprintf(fout, ": AT(_RomSize)\n    {\n"
@@ -63,9 +65,6 @@ static void write_ld_script(FILE *fout)
                   "        . = ALIGN(0x10);\n"
                   "        _%sSegmentTextStart = .;\n",
                   seg->name, seg->name);
-
-        if (seg->fields & (1 << STMT_align))
-            fprintf(fout, "        . = ALIGN(0x%X);\n", seg->align);
 
         for (j = 0; j < seg->includesCount; j++)
         {
@@ -168,9 +167,6 @@ static void write_ld_script(FILE *fout)
                       "        . = ALIGN(0x10);\n"
                       "        _%sSegmentBssStart = .;\n",
                       seg->name, seg->name, seg->name, seg->name);
-
-        if (seg->fields & (1 << STMT_align))
-            fprintf(fout, "        . = ALIGN(0x%X);\n", seg->align);
 
         for (j = 0; j < seg->includesCount; j++)
             fprintf(fout, "            %s (.sbss)\n"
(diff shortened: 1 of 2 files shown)


tools/mkldscript.c
```diff
@@ -51,11 +51,13 @@ static void write_ld_script(FILE *fout)
                   "    ..%s ", seg->name, seg->name, seg->name, seg->name);
 
         if (seg->fields & (1 << STMT_after))
-            fprintf(fout, "_%sSegmentEnd ", seg->after);
+            fprintf(fout, "(_%sSegmentEnd + %i) & ~ %i ", seg->after, seg->align - 1, seg->align - 1);
```
Dragorn421, line 54: `ALIGN(segmentEnd, align)`
Also only if align is set

tools/mkldscript.c
```diff
@@ -51,11 +51,13 @@ static void write_ld_script(FILE *fout)
...
         else if (seg->fields & (1 << STMT_number))
             fprintf(fout, "0x%02X000000 ", seg->number);
         else if (seg->fields & (1 << STMT_address))
             fprintf(fout, "0x%08X ", seg->address);
+        else
+            fprintf(fout, "ALIGN(0x%X) ", seg->align);
```
Dragorn421, line 60: Probably shouldn't write this at all when align is unset (align(0) doesn't hurt but idk if the align field is initialized to 0. Anyway just check stmt_align)

Dragorn421 (commented):
I also had to mess with this stuff when researching
FTR I will rewrite mkldscript eventually TM

The 0x20 alignment reminds me of how some text sections produced by ido are to be 0x20-aligned (per elf section header) rather than the usual 0x10-alignment. (e.g. is_debug.o , https://discord.com/channels/688807550715560050/688851451014938642/1207252894903046194 )
Maybe underlying is an implementation difference between the og linker and the gnu linker (as in, I guess the gnu linker doesn't align the output section using the largest alignment of input sections inside, and maybe the og linker did? Or maybe the gnu linker does behave like that and in nondebug no .o text is 0x20-aligned)

tools/mkldscript.c
cadmic: This didn't seem to work:
`mips-linux-gnu-ld:build/gc-eu-mq-dbg/oot-gc-eu-mq-dbg.ld:3924: non constant or forward reference address expression for section ..code`

tools/mkldscript.c
cadmic: `seg->align` defaults to 16. Is this a bug?

I did try checking if `STMT_align` was set but I thought it made the logic here too complicated. I think writing `ALIGN(0x1)` or whatever in all cases is fine

cadmic: I mean, I think checking for `STMT_align` together with `STMT_address` was too complicated, so I didn't check here to be consistent

Dragorn421 (approved):
The mkldscript changes feel weird (changing `align` behavior which is semi historical afaik (at least from... sm64 tooling maybe?))
but whatever, I'll probably rewrite mkldscript and keep campaigning against the spec

fig02 (approved):

PR #1920, opened 2024-03-06
Make BSS ordering script check ordering for all sections

cadmic:
I renamed bss_reordering.py to check_ordering.py and had it check all sections. Turns out this was more useful than I thought, as it caught several data issues in gc-eu-mq.


tools/check_ordering.py
```diff
@@ -178,10 +182,9 @@ def main():
     pointers.sort(key=lambda p: p.base_value)
 
     # Go through BSS sections and report differences
```
Dragorn421, line 184:
```suggestion
    # Go through sections and report differences
```

tools/check_ordering.py
```diff
@@ -178,10 +182,9 @@ def main():
...
-    i = 0
-    for mapfile_segment in mapfile:
+    for mapfile_segment in mapfile_segments:
         for file in mapfile_segment:
-            if not file.sectionType == ".bss":
+            if file.vram is None:
                 continue
```
Dragorn421, line 188:
```suggestion
```
vram is never None, (from a linker perspective) VMAs are always defined

tools/check_ordering.py
```diff
@@ -118,30 +119,33 @@ def main():
     mapfile = mapfile_parser.mapfile.MapFile()
     mapfile.readMapFile(f"build/{version}/oot-{version}.map")
 
-    reloc_mapfile_segments = []
+    mapfile_segments = []
```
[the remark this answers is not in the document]
cadmic: I changed to "source code" because only "code" makes me think of .text and "data-only" makes me think of .data/.rodata. What do you think?

Dragorn421 (approved):

fig02 (approved):

